PR #105, opened 2019-08-15
Authentication with Auth0

sergiogcx:
To run:

1. Add a user account to https://manage.auth0.com/dashboard/us/atd-datatech/users

2. Once created your user (with a secure password), click on the user to edit it. On the first page, scroll down to the Metadata section, and add this json script to the `app_metadata` textfield:

```
{
  "roles": [
    "editor"
  ]
}
```

3. Save, and test away!


atd-vze/.env
```diff
@@ -0,0 +1,5 @@
+# IMPORTANT: YOU NEED TO INSERT *YOUR* DOMAIN AND CLIENT_ID BELOW
+REACT_APP_AUTH0_DOMAIN=atd-datatech.auth0.com
+REACT_APP_AUTH0_CLIENT_ID=Y5Q304ADgUOYq8199aSgSxLEc6Xzdkb4
+REACT_APP_AUTH0_CALLBACK_URL="https://localhost:3000/callback"
+REACT_APP_AUTH0_LOGOUT_URL=https://localhost:3000
```
sergiogcx: It should be safe to share these openly, we only need the id_token which is given back to us via Auth0.

PR #116, opened 2019-08-17
Crash count widgets + description lookups for units and injury severity

mateoclarke:
_Closes #113_
_Closes #114_

### Adds fatality/injury widgets to the top row

Screen Shot 2019-08-17 at 2 49 14 PM

### Add counts to related records tables

Screen Shot 2019-08-17 at 3 18 12 PM

### Adds injury severity description (instead of lookup IDs) to People table

Screen Shot 2019-08-17 at 3 18 17 PM

### Adds unit type, body style, make and model descriptions (instead of lookup IDs) to Units table

Screen Shot 2019-08-17 at 3 18 23 PM

### DIFF
--- a/atd-vze/src/views/Crashes/Crash.js
+++ b/atd-vze/src/views/Crashes/Crash.js
@@ -14,15 +14,15 @@ const calculateYearsLifeLost = people => {
   // Find the differance between person.prsn_age & 75
   // Sum over the list of ppl with .reduce
   return people.reduce((accumulator, person) => {
-    let yearsLifeLost = 0;
+    let years = 0;
     if (person.injury_severity.injry_sev_desc === "KILLED") {
       let yearsLifeLost = 75 - Number(person.prsn_age);
       // What if the person is older than 75?
       // For now, so we don't have negative numbers,
       // Assume years of life lost is 0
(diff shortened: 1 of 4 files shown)


atd-vze/src/views/Crashes/Crash.js
```diff
@@ -2,116 +2,29 @@ import React from "react";
...
+
+const calculateYearsLifeLost = people => {
+  // Assume 75 year life expectancy,
+  // Find the differance between person.prsn_age & 75
+  // Sum over the list of ppl with .reduce
+  return people.reduce((accumulator, person) => {
+    let years = 0;
+    if (person.injury_severity.injry_sev_desc === "KILLED") {
+      let yearsLifeLost = 75 - Number(person.prsn_age);
+      // What if the person is older than 75?
```
mddilley: Does it make sense to round ages over 75 down to 75? That way any years up to 75 are counted in years of life lost (just not the years above 75)? Not sure what is best for this metric.

mateoclarke: Yeah, thats basically whats happening in line 23 below in this reduce function. If we calculate that the years of life lost is negative, we just round it to zero.

PR #159, opened 2019-08-29
Changes to sort/search/pagination from PR feedback

### LINKED ISSUE #144: VZE: Changes to sort/search/pagination from PR feedback
These are outstanding changes from PR review of #131 

- [x] Abstract hex codes to constants file for elements that fall outside reactstrap
- [x] Change default of Field option for search from "Field" (app crashes if not set)
- [x] Handle case change of search field values - Need to consider that DB strings are all uppercase, but the QA process may introduce DB edits that are not uppercase - uppercase edit strings before insertion?
- [x] Change pagination buttons (Prev, Page #, Next) into one `<ButtonGroup>` to clean up the UI
- [ ] ~Modify CSV export to include more than what is displayed in view~  Moved to #161

mddilley:
Closes #144.

### DIFF
--- a/atd-vze/src/Components/TableSearchBar.js
+++ b/atd-vze/src/Components/TableSearchBar.js
@@ -38,7 +38,8 @@ const TableSearchBar = ({ setSearchFilter, clearFilters, fieldsToSearch }) => {
 
   const handleSearchSubmission = e => {
     e.preventDefault();
-    setSearchValue(searchFieldValue);
+    const uppercaseSearchValue = searchFieldValue.toUpperCase();
+    fieldToSearch !== "" && setSearchValue(uppercaseSearchValue);
   };
 
   const handleClearSearchResults = () => {
--- a/atd-vze/src/Components/TableSearchBar.js
+++ b/atd-vze/src/Components/TableSearchBar.js
@@ -32,7 +32,7 @@ const TableSearchBar = ({ setSearchFilter, clearFilters, fieldsToSearch }) => {
     };
     const queryStringArray = searchQuery();
     searchValue !== "" &&
-      searchFieldValue !== "" &&
+      fieldToSearch !== "" &&
       setSearchFilter(queryStringArray);
   }, [searchValue, setSearchFilter, fieldToSearch, searchFieldValue]);
 
--- a/atd-vze/src/Components/TableSearchBar.js
+++ b/atd-vze/src/Components/TableSearchBar.js
@@ -14,7 +14,6 @@ import {
   DropdownMenu,
   Alert,
 } from "reactstrap";
-import { isField } from "apollo-utilities";
 
 // TODO add query operators to each field to better fit data types (_eq, etc.)?
 const TableSearchBar = ({ setSearchFilter, clearFilters, fieldsToSearch }) => {
--- a/atd-vze/src/Components/TableSearchBar.js
+++ b/atd-vze/src/Components/TableSearchBar.js
@@ -12,7 +12,9 @@ import {
   DropdownToggle,
   DropdownItem,
   DropdownMenu,
+  Alert,
 } from "reactstrap";
+import { isField } from "apollo-utilities";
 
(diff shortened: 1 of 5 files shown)


mddilley:
This PR fixes some issues following the addition of sort/pagination/and search. This branch starts a pattern of abstracting Bootstrap colors to a colors file to keep the theme consistent when adding non-reactstrap elements. I also wrapped the pagination buttons in the same ButtonGroup to make one pill shaped group.

It also fixes a bug where the app would crash without setting a field to search. I added logic to do nothing when the search button is clicked without setting a search field in the dropdown and added a warning banner that appears once typing starts in the search field and then goes away once the search field is chosen.

Last, I added toUpperCase() to all values typed in the search field before querying the DB. @mateoclarke @sergiogcx I was thinking that once edits are implemented we could uppercase strings on the frontend before DB insertion (to keep case consistent). What do y'all think?

### Better pagination buttons
Screen Shot 2019-08-30 at 11 34 25 AM

### Prompt for search field
Screen Shot 2019-08-30 at 11 46 29 AM

atd-vze/src/Components/TableSortHeader.js
```diff
@@ -1,9 +1,10 @@
 import React, { useState, useEffect } from "react";
 import { withApollo } from "react-apollo";
 import styled from "styled-components";
+import { colors } from "../styles/colors";
 
 const StyledGreyArrow = styled.i`
-  color: #c8ced3;
+  color: ${colors.light};
```
mateoclarke: ❤️

atd-vze/src/styles/colors.js
```diff
@@ -0,0 +1,5 @@
+// Hex values of colors from Bootstrap
+export const colors = {
+  // Bootstrap color: hex value
+  light: "#c8ced3",
```
mateoclarke: thanks for doing this, starts small but sets a pattern that will pay dividends for our productivity and consistency.

atd-vze/src/Components/TableSearchBar.js
```diff
@@ -32,27 +34,32 @@ const TableSearchBar = ({ setSearchFilter, clearFilters, fieldsToSearch }) => {
...
-      searchFieldValue !== "" &&
+      fieldToSearch !== "" &&
       setSearchFilter(queryStringArray);
   }, [searchValue, setSearchFilter, fieldToSearch, searchFieldValue]);
 
   const handleSearchSubmission = e => {
-    e.preventDefault();
-    setSearchValue(searchFieldValue);
+    if (isFieldSelected) {
+      e.preventDefault();
```
mateoclarke: Something weird happens when I use "Return" to submit instead of clicking search. The whole page refreshes. Is it possible that we need to move line 43 up before the `if` block?

atd-vze/src/Components/TableSearchBar.js
```diff
@@ -20,6 +21,7 @@ const TableSearchBar = ({ setSearchFilter, clearFilters, fieldsToSearch }) => {
   const [searchValue, setSearchValue] = useState("");
   const [isDropdownOpen, setIsDropdownOpen] = useState(false);
   const [fieldToSearch, setFieldToSearch] = useState("");
```
mateoclarke: I think we need a default value for the `fieldToSearch` dropdown.

atd-vze/src/Components/TableSearchBar.js
mddilley: Good call! Even just hitting enter in the field refreshes the page. I moved 43 outside the block, and all is better now.

atd-vze/src/styles/colors.js
mddilley: 👍

atd-vze/src/Components/TableSearchBar.js
mddilley: I modified the code in the PR around `""` being the default value for the field. The form will never submit with this value set in `fieldToSearch` and the popup warning depends on this value being set in state to prevent the app from crashing.

PR #162, opened 2019-09-06
Crash Index table date range search

### LINKED ISSUE #153: VZE: Crash Index table date range search
- consensus from `Sprint Review #3` that we should have date range searching. 

Maybe a calendar range plugin input separate from the main search bar?

mddilley:
Closes #153


mddilley:
This PR adds a date range picker and a date range to the Crashes table query. A bug was discovered after adding this feature in which the `loading` variable instantiated by `useQuery` in `<TableWithFilters />` never returns to false after retrieving data from the DB in some cases (date range queries with long time spans). A workaround (disabling cache of the query data) was implemented and the issue below is being followed for updates on a better fix.

https://github.com/apollographql/react-apollo/issues/3361

**Date range picker**
Screen Shot 2019-09-10 at 2 03 09 PM

atd-vze/src/Components/TableDateRange.js
```diff
@@ -0,0 +1,99 @@
...
+    line-height: 1.5;
+    color: ${colors.grey700};
+    background-color: ${colors.white};
+    background-clip: padding-box;
+    border: 1px solid ${colors.grey200};
+    border-radius: 0.25rem;
+  }
+
+  .react-datepicker__day--selecting-range-start {
+    background-color: ${colors.primary} !important;
```
mateoclarke: YAAAS!

atd-vze/src/Components/TableDateRange.js
```diff
@@ -0,0 +1,99 @@
...
+    background-color: ${colors.secondary};
+  }
+
+  .react-datepicker__header {
+    background-color: ${colors.light};
+  }
+`;
+
+const TableDateRange = ({ setDateRangeFilter, databaseDateColumnName }) => {
+  const minDate = new Date("2010/01/01"); // TODO add programatic way to insert earliest crash record in DB
```
mateoclarke: Instead of having the programmatic solution get the oldest record in the DB, perhaps we just provide a useful default date range. Like YTD, getting just the last year of data as the starting place. 

```js
let d = new Date();
d.setFullYear(d.getFullYear() - 1);
```

or with moment.js since you import it already:
```
moment().subtract(1, 'year').format();
```

atd-vze/src/Components/TableWithFilters.js
```diff
@@ -98,12 +129,17 @@ const TableWithFilters = ({
...
   ]);
 
   const { loading, error, data } = useQuery(
     gql`
       ${tableQuery}
-    `
+    `,
+    // Disables cache as a workaround for loading never returning to false
+    // after queries with long time spans
+    // https://github.com/apollographql/react-apollo/issues/3361
```
mateoclarke: 👀

atd-vze/src/Components/TableWithFilters.js
```diff
@@ -79,14 +83,41 @@ const TableWithFilters = ({
             );
           });
         }
-        if (searchFilter !== "") {
+        if (dateRangeFilter !== "" && searchFilter !== "") {
+          // Combine both 'where' queries into one argument and replace in queryWithFilters
```
mateoclarke: This block is a little heady. Looking below how the other two `if` blocks work makes it a little easier to follow. This kind of makes me wonder if theres a library that provides tools and helpers for building graphql query strings.

atd-vze/src/Components/TableDateRange.js
mddilley: 🎉

atd-vze/src/Components/TableDateRange.js
mddilley: That's a great idea. I like defaulting to a small range of data too. I went with the moment option.

atd-vze/src/Components/TableWithFilters.js
mddilley: I added some comments about an alternate workaround that Sergio came up with so we can keep that in the equation down the line as well.

atd-vze/src/Components/TableWithFilters.js
mddilley: I'm in agreement on this. Adding this part of the feature was not as nice and tidy as I imagined. I'll take another look for a library along those lines. All of these if blocks are begging to be refactored as well. Lots of repetition.

PR #163, opened 2019-09-06
Create GIS QA Editing Workflow

### LINKED ISSUE #133: VZE: Create GIS QA Editing Workflow
User needs to add/modify lat/lon of a crash via map interface.

Idea: User clicks and edit button, drops/drags pin to the crash location. Clicks save.

Workaround is to retrieve lat/lon from another tool and use the normal Edit Crash interface to update lat/lon.

mddilley:
Closes #133


mddilley:
This PR adds the first iteration of the GIS QA workflow for crash records without latitude/longitude. This feature adds to the existing warning banner rendered when no lat/lon exists and allows the user to drag and drop a map pin to submit a location for the crash record.

**Map in Crash Details view**
Screen Shot 2019-09-09 at 11 37 54 AM

atd-vze/src/styles/mapPinStyles.js
```diff
@@ -0,0 +1,10 @@
+// Styles for pins in Mapbox maps
+export const pinStyles = {
+  stroke: "black",
+  strokeWidth: 2,
+};
+
+export const PIN_ICON = `M50,93.5c0,0,31.1-30.5,31.1-56C81.1,20.4,67.2,6.5,50,6.5c-17.2,0-31.1,13.9-31.1,31.1C18.9,63.1,50,93.5,50,93.5z   M35.2,37.6c0-8.2,6.6-14.8,14.8-14.8c8.2,0,14.8,6.6,14.8,14.8c0,8.2-6.6,14.8-14.8,14.8C41.8,52.4,35.2,45.8,35.2,37.6z`;
```
mateoclarke: 🔥

atd-vze/src/views/Crashes/Crash.js
```diff
@@ -29,17 +30,36 @@ const calculateYearsLifeLost = people => {
...
-  const { loading, error, data } = useQuery(GET_CRASH, {
+  const { loading, error, data, refetch } = useQuery(GET_CRASH, {
     variables: { crashId },
   });
 
   if (loading) return "Loading...";
   if (error) return `Error! ${error.message}`;
 
+  const createGeocoderAddressString = data => {
+    const geocoderAddressFields = [
```
mateoclarke: Just a heads up that this might need updating once we start referencing a pre-concatenated address field from the DB.

atd-vze/src/styles/mapPinStyles.js
mddilley: Yesss, this was fun to write.

atd-vze/src/views/Crashes/Crash.js
mddilley: 👍 and this also may need to be removed if it is confusing to the user.

Issue #169, opened 2019-09-12
VZE: Fix Edge Cases Bugs


sergiogcx:
There are important things that need fixing in this review:
https://github.com/cityofaustin/atd-vz-data/pull/160

- [x] Pagination Issue when clearing search results.
- [x] Text upper case search bug
- [x] Get rid of the console.log entries

sergiogcx:
Ready for review please:
https://github.com/cityofaustin/atd-vz-data/pull/182

sergiogcx:
Closing, merged to master.

PR #182, opened 2019-09-16
Fix Location Bugs

sergiogcx:
In this PR you should be able to see the following fixes:

- [ ] Case-insensitive search via `_ilike` operator.
- [ ] Fix pagination bug where if click through and paginate a few pages, then do a search with limited results, you end up in a state where you're on page 3 out 1. This was fixed by resetting the page number to 1 every time there is a new search. It should also default to 1 page when there are zero results.
- [ ] Got rid of `console.log` statements.

This is in relation to issue #169

### DIFF
--- a/atd-vze/src/App.js
+++ b/atd-vze/src/App.js
@@ -88,11 +88,7 @@ class App extends Component {
             "x-hasura-role": this.state.role,
           },
         });
-      } else {
-        alert("Client not yet initialized, no role available.");
       }
-    } else {
-      alert("Client not initialized, not authenticated.");
     }
   }
 
--- a/atd-vze/src/App.js
+++ b/atd-vze/src/App.js
@@ -68,11 +68,10 @@ class App extends Component {
               "x-hasura-allowed-roles"
             ][0];
           localStorage.setItem("hasura_user_role", role);
-          console.log("Role initialized: " + role);
           this.setState({ role: role });
           this.initializeClient();
         } catch (error) {
-          console.log(error);
+          alert("Error: " + error);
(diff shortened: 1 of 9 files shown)


atd-vze/src/App.js
```diff
@@ -68,11 +68,10 @@ class App extends Component {
               "x-hasura-allowed-roles"
             ][0];
           localStorage.setItem("hasura_user_role", role);
-          console.log("Role initialized: " + role);
           this.setState({ role: role });
           this.initializeClient();
         } catch (error) {
-          console.log(error);
+          alert("Error: " + error);
```
mateoclarke: console.logs are fine on errors imo, but good call, alert is better.

atd-vze/src/Components/GridTable.js
```diff
@@ -81,6 +80,10 @@ const GridTable = ({ title, query }) => {
     changePage(page - 1);
   };
 
+  const resetPageOnSearch = () => {
+    changePage(1);
```
mateoclarke: 💅

atd-vze/src/Components/GridTable.js
```diff
@@ -108,13 +111,12 @@ const GridTable = ({ title, query }) => {
   if (searchParameters["column"] && searchParameters["value"]) {
     query.setWhere(
       searchParameters["column"],
-      `_like: "%${searchParameters["value"]}%"`
+      `_ilike: "%${searchParameters["value"]}%"`
```
mateoclarke: 💅

PR #207, opened 2019-09-26
VZE: Map drags when dragging cursor in the CrashQAMap geocoder address bar

### LINKED ISSUE #168: VZE: Map drags when dragging cursor in the CrashQAMap geocoder address bar
When hovering over the address bar of the geocoder in the CrashQAMap component, clicking and dragging with the cursor moves the map below. The issue is possibly related to the z-index of the geocoder input or event bubbling. This bug is also present in the example from the creator of the [library](https://github.com/SamSamskies/react-map-gl-geocoder) (https://codesandbox.io/s/l7p179qr6m).

64653067-0a97b300-d3eb-11e9-9da0-12a680029318

mddilley:
Closes #168

### DIFF
--- a/atd-vze/src/views/Crashes/Maps/customGeocoderMapController.js
+++ b/atd-vze/src/views/Crashes/Maps/customGeocoderMapController.js
@@ -25,6 +25,18 @@ export class CustomGeocoderMapController extends MapController {
     super._onDoubleTap(event);
   }
 
+  _onPanStart(event) {
+    // ignore double taps on geocoder input
+    if (
+      this._isGeocoderInputNode(event.target) ||
+      this._isMapStyleInputNode(event.target)
+    ) {
+      return;
+    }
+
+    super._onPanStart(event);
+  }
+
   _isGeocoderInputNode(node) {
     return node.classList.contains("mapboxgl-ctrl-geocoder--input");
   }
--- a/atd-vze/src/views/Crashes/Maps/customGeocoderMapController.js
+++ b/atd-vze/src/views/Crashes/Maps/customGeocoderMapController.js
@@ -1,5 +1,9 @@
 import { MapController } from "react-map-gl";
 
+// This custom controller fixes the issue where mouse events were propagating through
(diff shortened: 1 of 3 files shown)


mddilley:
This PR fixes an issue where double-clicking or dragging over the geocoder or map style controls propagated events to the `<CrashQAMap />` below. Invaluable advice from @SamSamskies (https://github.com/SamSamskies/react-map-gl-geocoder/issues/53) led to a quick fix of adding a custom `MapController` that ignores these events for these specific control elements (https://github.com/uber/react-map-gl/blob/master/docs/advanced/custom-map-controller.md). 

The string in the geocoder control can now be selected by clicking and dragging without causing changes in the map below.

atd-vze/src/views/Crashes/Maps/customGeocoderMapController.js
```diff
@@ -0,0 +1,46 @@
+import { MapController } from "react-map-gl";
+
+export class CustomGeocoderMapController extends MapController {
+  _onPan(event) {
```
mateoclarke: so these methods overwrite the same `MapController` methods, I'm assuming?

mateoclarke: now digging into your links, I'm seeing that's exactly what it does. Can you add a comment to this file that gives a brief explanation and links to [the examples](https://github.com/uber/react-map-gl/blob/master/docs/advanced/custom-map-controller.md)?

mateoclarke:
It's super cool that you were able to reach out to the library maintainer and get a helpful pointer. We should send them some Tiff's Treats 🍪

atd-vze/src/views/Crashes/Maps/customGeocoderMapController.js
mddilley: Great idea - I added in some documentation so it is easier to see what is going on. I'm sure that we will see this come up again with this library too.

PR #209, opened 2019-09-27
Research drawing tool for Location boundaries

### LINKED ISSUE #198: VZE: Research drawing tool for Location boundaries
Location definitions are fluid, so the editor will need an interface to edit the polygon for each location. 

Corresponding database work  in #195 will handle version control and street name standardization.

mddilley:
Closes #198


mddilley:
This PR is a first pass on the Location Polygon Drawing/Editing tool. The tool loads the existing polygon on the map centered on the lat/lon from the DB. The user can edit the existing polygon (drag/drop existing points or create new points on the polygon), delete the existing polygon and draw a new one, and also reset the changes and start over. The edits can then be submitted to the DB.

### View Polygon
Screen Shot 2019-10-02 at 4 14 03 PM

### Edit Polygon
Screen Shot 2019-10-02 at 4 14 47 PM

atd-vze/src/views/Locations/LocationEditMap.js
```diff
@@ -0,0 +1,224 @@
...
+
+class LocationEditMap extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      // map
+      viewport: {
+        width: 800,
+        height: 600,
+        longitude: this.props.data.atd_txdot_locations[0].longitude,
```
mateoclarke: potential code cleanliness suggestion. Go ahead and make `this.props.data.atd_txdot_locations[0]` a variable since you're using it a bunch.

```js
const location = this.props.data.atd_txdot_locations[0]
```

mddilley: Agreed, I made this change and also added a isSubmitting boolean to change the text in the Submit button as a first step to a better loading indicator.

Issue #86, opened 2019-08-12
VZE: Edit Crash `/crash/:id/edit`


mateoclarke:
- [ ] form fields
- [ ] hopefully a subset of all
- [ ] audit trail, modified by, modified at
- [x] edit lat/lon, map component to drag and drop a pin

PR #214, opened 2019-09-30
Edit Crash details: `investigator_narrative` & `fhe_collsn_id` fields

mateoclarke:
This is be the first PR of many. Trying to set a pattern here with how we configure our fields so that we define each of their human readable labels, whether it is editable, what UI input type we use for editing (ex: free text vs dropdown select), and what lookup table the value/ID references so we can display the human-readable description instead of an ID.

This PR only effects the `investigator_narrative` & `fhe_collsn_id` fields. More to come...

I recommend testing on Crash ID = 123
https://localhost:3000/#/crashes/123

Sep-30-2019 12-38-04

Related to #86


atd-vze/src/queries/crashes.js
```diff
@@ -165,3 +167,105 @@ export const UPDATE_COORDS = gql`
     }
   }
 `;
+
+export const UPDATE_CRASH = gql`
+  mutation update_atd_txdot_crashes(
```
mateoclarke: @sergiogcx here's the example of the mutation for editing the fields...

atd-vze/src/views/Crashes/Crash.js
```diff
@@ -55,6 +78,35 @@ function Crash(props) {
...
+  const handleCancelClick = e => {
+    e.preventDefault();
+
+    setEditField("");
+  };
+
+  const handleFieldUpdate = e => {
+    e.preventDefault();
+
+    props.client
```
mateoclarke: @sergiogcx, this is an important part too

atd-vze/src/views/Crashes/Crash.js
```diff
@@ -1,19 +1,33 @@
...
+  Col,
+  Row,
+  Table,
+  Alert,
+  Input,
+} from "reactstrap";
 import { withApollo } from "react-apollo";
 import { useQuery } from "@apollo/react-hooks";
-import crashDataMap from "./crashDataMap";
+import { crashDataMap, geoFields } from "./crashDataMap";
```
sergiogcx: It appears neither setState or geoFields is being used, once removed this should be good to ship.

mddilley:
I'm still looking through the changes, but, when I tested locally, I encountered this error that happens now and again. This is an awesome feature that looks and works great so this bug could probably be captured in another issue. I can reproduce by looking at any crash record details view and then refreshing the page.

Screen Shot 2019-10-02 at 11 15 36 AM

sergiogcx:
> I'm still looking through the changes, but, when I tested locally, I encountered this error that happens now and again. This is an awesome feature that looks and works great so this bug could probably be captured in another issue. I can reproduce by looking at any crash record details view and then refreshing the page.
> 
> Screen Shot 2019-10-02 at 11 15 36 AM

I'm unable to replicate the issue even after following the same steps, tested on both firefox and chrome.

atd-vze/src/views/Crashes/crashDataMap.js
```diff
@@ -1,104 +1,328 @@
...
+        editable: false,
+      },
+      day_of_week: {
+        label: "Day of Week",
+        editable: false,
+      },
+      fhe_collsn_id: {
+        label: "Manner of Collision ID",
+        editable: true,
+        uiType: "select",
```
mddilley: I like this pattern a lot - especially setting the uiType here and then handling it in the component. 🙌

atd-vze/src/views/Crashes/Crash.js
mateoclarke: fixed in another branch

PR #220, opened 2019-10-01
Crash Page Fix

sergiogcx:
It appears the Crash page is not able to handle objects, in this particular case it is polygon data that tries to be converted into a string. This is a temporary fix that checks whether the current field is an object, if so it tries to print via JSON.stringify


sergiogcx:
Tested by Mike, merging to master.

mateoclarke:
If this a truly a temporary fix, we need to create an issue to come back and resolve it. On that note, looks fine to me.

PR #221, opened 2019-10-01
VZE: Location details page add map

### LINKED ISSUE #177: VZE: Location details page add map
LL requested a map be added to the location details page with an aerial from Near map tiles and timestamped to know the date of the capture.

mddilley:
Closes #177


mddilley:
This PR adds an aerial map (tiles from NearMap) to the Location Details view with a timestamp retrieved from the NearMap API.

@mateoclarke - the last question on this PR is handling the API key. I was sure to keep it out of my commits, and I'd like to discuss since I was able to generate a key with my CoA credentials but I can't see any stats on its usage.

### Aerial Map w/ timestamp
Screen Shot 2019-10-02 at 2 11 01 PM

atd-vze/src/views/Locations/LocationMap.js
```diff
@@ -0,0 +1,132 @@
+import React, { Component } from "react";
+import MapGL, { NavigationControl, FullscreenControl } from "react-map-gl";
+import axios from "axios";
+import moment from "moment";
+
+import styled from "styled-components";
+import { Button } from "reactstrap";
+
+const TOKEN = process.env.REACT_APP_MAPBOX_TOKEN;
```
mateoclarke: we could add the `NEARMAP_TOKEN` here like we did for Mapbox

mddilley: Cool, I'll make that change before merge (minus the actual token).

Issue #217, opened 2019-10-01
VZD: Rename UniqueID columns in postgres


sergiogcx:
There are a number of columns called unique_id which is not the best pattern to implement, especially as we scale up our code base there will be need for more consistent naming.

- [x] Find all instances of the unique_id column in postgres tables.
- [x] Find all instances of unique_id in triggers.
- [x] Find all instances of unique_id in the code base.
- [x] Find any instances of unique_id in Hasura (rename, re-adjust permissions)

sergiogcx:
The code base changes are minimal in comparison to the database:
https://github.com/cityofaustin/atd-vz-data/pull/222/files

sergiogcx:
This is done and merged to master, closing.

PR #222, opened 2019-10-01
Refactor unique_id

sergiogcx:
This PR refactors the code in compliance with issue #217, and the purpose of this PR is to find all instances of unique_id in the code base that need refactoring (luckily there are only four files that needed editing)

NOTE:
- The changes were actually made to the database for the primary key, it was renamed from unique_id to location_id; however, a the column was cloned and named unique_id so that the old code would not crash.

Once this PR is merged to master, we should be able to remove the helper column unique_id for good.

### DIFF
--- a/atd-vze/src/queries/Locations.js
+++ b/atd-vze/src/queries/Locations.js
@@ -2,8 +2,8 @@ import { gql } from "apollo-boost";
 
 export const GET_LOCATION = gql`
   query GetLocation($id: String) {
-    atd_txdot_locations(where: { unique_id: { _eq: $id } }) {
-      unique_id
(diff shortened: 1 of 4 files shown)


atd-vze/src/queries/Locations.js
```diff
@@ -2,11 +2,10 @@ import { gql } from "apollo-boost";
...
   query GetLocation($id: String) {
-    atd_txdot_locations(where: { unique_id: { _eq: $id } }) {
-      unique_id
+    atd_txdot_locations(where: { location_id: { _eq: $id } }) {
+      location_id
       address
       description
-      geometry
       metadata
       last_update
```
sergiogcx: We no longer need the geometry column we now use shape.

PR #255, opened 2019-10-07
Bug: Fix deaths & injuries filter to be OR not AND

### LINKED ISSUE #240: Bug: Fix deaths & injuries filter to be OR not AND
When both the "Deaths" and "Serious Injuries" options are active, we'd expect to see a list that stacks all the records with deaths or serious injuries. Currently, we only see records that have both deaths and serious injuries (just 9 records for the default date range).

Screen Shot 2019-10-07 at 12 25 30 PM

mddilley:
Closes #240


mddilley:
This PR changes the Death & Injuries advanced filters on the Crashes QA view and the Location Details view to stack using OR instead of AND logic. This required extending the GraphQL query abstract to handle the `_or` keyword within the `where` filter (https://docs.hasura.io/1.0/graphql/manual/queries/query-filters.html#using-multiple-filters-in-the-same-query-and-or). I added to Sergio's pattern of setting the filters for a table as props to `<GridTable />` and by adding an `_or` key to parameters of the filter (see example below). I tied some of the logic for handling `_or` and `where` filters together in the abstract since the `_or` filters need to be nested in the `where` filters.

Before:
```
        {
          id: "dni_deaths",
          label: "Deaths",
          filter: {
            where: [{ death_cnt: "_gt: 0" }],
          },
        },
```

After:
```
        {
          id: "dni_deaths",
          label: "Deaths",
          filter: {
            where: [{ _or: { death_cnt: "_gt: 0" } }],
          },
        },
```
### With single filter
Screen Shot 2019-10-08 at 1 54 33 PM

### With two filters
Screen Shot 2019-10-08 at 1 54 38 PM

Screen Shot 2019-10-08 at 1 55 01 PM

sergiogcx:
It looks good, but I have a question: should all filters fall under the _or operator? At the moment it seems just the first three (Deaths, Serious Injuries, and Non-Fatal Injuries) are the only ones being placed in the _or section.

atd-vze/src/queries/gqlAbstract.js
```diff
@@ -122,6 +133,15 @@ gqlAbstractTableAggregateName (
     delete this.config["where"][key];
   }
 
+  /**
+   * Removes a column from the or condition
+   * @param {string} key - The name of the column
```
mateoclarke: maybe i'm confused but I think the documentation/comments here should say that the param is called `orObject`

mateoclarke: and I'm thinking its probably and `{object}` not a string?

atd-vze/src/queries/gqlAbstract.js
```diff
@@ -224,10 +244,20 @@ gqlAbstractTableAggregateName (
 
     if (this.config["where"] !== null) {
       let where = [];
+      let _or = [];
       for (let [key, value] of this.getEntries("where")) {
         where.push(`${key}: {${value}}`);
       }
-      output.push(`where: {${where.join(", ")}}`);
+      if (this.config["or"] && this.config["or"] !== null) {
```
mateoclarke: would the double-bang trick work here?  
`!!this.config["or"]` instead of the two conditionals? 

https://medium.com/better-programming/javascript-bang-bang-i-shot-you-down-use-of-double-bangs-in-javascript-7c9d94446054

atd-vze/src/queries/gqlAbstract.js
```diff
@@ -224,10 +244,20 @@ gqlAbstractTableAggregateName (
 
     if (this.config["where"] !== null) {
       let where = [];
+      let _or = [];
```
mateoclarke: I don't have a strong opinion about `_` prefix for indicating private/class variables and methods in JS. But it seems like we should at least be consistent with the existing `where` variable naming. Either both underscore or neither. Unless I'm misunderstanding.

mateoclarke: ohhh, wait, it has to do with the naming of the `_or` query param in the graphql query object, right? my bad.

atd-vze/src/queries/gqlAbstract.js
mddilley: Good catch! That is my copypasta. Fixed this for the future.

atd-vze/src/queries/gqlAbstract.js
mddilley: Yep, it is following the GraphQL syntax but it feels a bit out of place here. I changed it everywhere except for where the gql query is built since that makes two of us. Tested and it works.

PR #256, opened 2019-10-08
collision type

### LINKED ISSUE #248: VZE: Add new infographic card to location details page for manner of collisions
It should show a ranked list of the most frequent manner of collisions type. Top three types at least?

tcwcodes:
Fixes #248 

Had an issue with the labels being way too long and disappearing off the side of the widget, so I used a workaround to split the labels at about half their length onto more than one line by converting them to arrays. Not very elegant, and looks weird at certain sizes, but at least you can read them. There may be a better workaround. I added numbers to each label to make them easier to read and will also will likely reformat them to be in sentence case.

Screen Shot 2019-10-07 at 8 46 59 PM

https://giphy.com/gifs/jTwLaLNAXok0yzYrL8

### DIFF
--- a/atd-vze/src/views/Locations/Location.js
+++ b/atd-vze/src/views/Locations/Location.js
@@ -116,12 +116,6 @@ function Location(props) {
     ],
   };
 
-  const test = data.atd_txdot_locations[0].crashes_by_manner_collision.map(
-    a => a.collsn_desc
-  );
-  const testB = test.map(a => formatLabel(a));
-  console.log(test, testB);
-
   const horizontalBar = {
     labels: data.atd_txdot_locations[0].crashes_by_manner_collision
       .map((a, index) => `${index + 1}. ${a.collsn_desc}`)
--- a/atd-vze/src/views/Locations/Location.js
+++ b/atd-vze/src/views/Locations/Location.js
@@ -23,7 +23,7 @@ import LocationCrashes from "./LocationCrashes";
 
 import { GET_LOCATION } from "../../queries/Locations";
 import Widget02 from "../Widgets/Widget02";
-import { Doughnut } from "react-chartjs-2";
+import { Doughnut, HorizontalBar } from "react-chartjs-2";
 
 function Location(props) {
   const locationId = props.match.params.id;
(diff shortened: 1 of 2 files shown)


tcwcodes:
> It needed lint in only two lines, I figured I would fix them and merge.

Thanks, Sergio!

atd-vze/src/views/Locations/Location.js
```diff
@@ -41,12 +41,57 @@ function Location(props) {
...
+
+      if (item.length < maxwidth) {
+        temp = item;
+      } else {
+        sections.push(item);
+      }
+    });
+
+    return sections;
+  };
```
mateoclarke: @tcwcodes, this is a pretty complex function that took me a full 30 minutes to dig through and understand. If I may make a suggestion for a more simple method, how would you feel about this... (it assumes we only need to split the string in half because I noticed the ones that wrapped to 3 lines looked a little crammed)

```js
  const formatLabel = str => {
    let sections = [];

    // Get the approx midpoint of string
    let splitPoint = Math.floor(str.length / 2);

    // If the midpoint is not a space,
    // find the closest " " to the left of the midpoint and split there instead
    if (str.charAt(splitPoint) !== " ") {
      splitPoint = str.substring(0, splitPoint).lastIndexOf(" ");
    }

    sections.push(str.substring(0, splitPoint));
    sections.push(str.substring(splitPoint));

    return sections;
  };
```

Result:

Screen Shot 2019-10-09 at 4 10 05 PM

atd-vze/src/views/Locations/Location.js
```diff
@@ -130,6 +194,66 @@ function Location(props) {
       )}
       <Row>
         <Col lg={6}>
+          <Card>
+            <CardHeader>Types of Vehicles - Count Distribution</CardHeader>
+            <CardBody>
+              {data.atd_txdot_crashes_aggregate.aggregate.count === 0 && (
```
mateoclarke: we use `data.atd_txdot_crashes_aggregate.aggregate.count` at least 4 times here. Before the `return (` on 138, why don't we do

```js
const { count: crashCount } = data.atd_txdot_crashes_aggregate.aggregate
```

so you can use a more human friendly variable `crashCount` instead.

mateoclarke:
@tcwcodes, if you don't mind making a new PR addressing my feedback once you're back on Monday that would be great. Since @sergiogcx has already merged this to master, just didn't want my feedback to get lost and never implemented.

tcwcodes:
> @tcwcodes, if you don't mind making a new PR addressing my feedback once you're back on Monday that would be great. Since @sergiogcx has already merged this to master, just didn't want my feedback to get lost and never implemented.

Totally understand - will take in your edits and submit a new PR. Thanks for the feedback!

atd-vze/src/views/Locations/Location.js
tcwcodes: This is much better/cleaner! I went ahead with this function.

atd-vze/src/views/Locations/Location.js
```diff
@@ -71,6 +116,25 @@ function Location(props) {
...
   };
 
+  const horizontalBar = {
+    labels: data.atd_txdot_locations[0].crashes_by_manner_collision
+      .map((a, index) => `${index + 1}. ${a.collsn_desc}`)
+      .map(a => formatLabel(a)),
+    datasets: [
+      {
+        label: "Number of Collisions",
+        backgroundColor: "rgba(255,99,132,0.2)",
```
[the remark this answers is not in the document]
tcwcodes: Imported the colors.js file and used the color associated with the "total crashes" widget so that they match. I made the hover color a darker green, which I also added to colors.js.

PR #267, opened 2019-10-09
Add summary counts

### LINKED ISSUE #244: VZE: Add summary counts as table columns to Locations Index page
- total crashes
- total fatal
- total serious
- total comprehensive cost

(all should be sortable)

### LINKED ISSUE #266: VZE Bug: Crashes table sorting breaks when switching between columns sorted by
In the Crashes table (using `<GridTable />` and the gql abstract), the sorting breaks after switching from one column to another. For example, if you sort by Crash ID first and then try to sort by Case Number, the query looks like `order_by: {crash_id: asc, case_id: desc}` so sorting breaks for everything except the Crash ID.

The issue is that each new sort selection adds to the GraphQL query but does not remove the previous sort.

Screen Shot 2019-10-09 at 11 22 42 AM

sergiogcx:
Closes #244  Closes #266 

This branch will show:
- [x] total crashes
- [x] total fatal
- [x] total serious
- [x] total comprehensive cost

Currently Broken:
- [x] (all should be sortable)

The abstract library will need changes in order to fix the sort problem for composite fields.


atd-vze/src/queries/gqlAbstract.js
```diff
@@ -37,6 +37,43 @@ gqlAbstractTableAggregateName (
...
+   * @param {string} exp - The nested key (usually a graphql expression)
+   * @returns {string}
+   */
+  sortifyNestedKey = (exp, val) =>
+    this.isNestedKey(exp)
+      ? `${this.getExpKey(exp)}: { ${this.sortifyNestedKey(
+          this.getExpValue(exp),
+          val
+        )} }`
+      : `${exp}: ${val}`;
```
sergiogcx: This is a recursive expression, basically transforms this:

`location { location_crashes_aggregate { count }}: asc`

into this (example):

`location: { location_crashes_aggregate: {count: asc}}`

atd-vze/src/queries/gqlAbstract.js
```diff
@@ -181,10 +239,50 @@ gqlAbstractTableAggregateName (
...
+  getDefault(columnName) {
+    return `${this.config["columns"][columnName]["default"] || "-"}`;
+  }
+
+  /**
+   * Attempts to format value based on configuration specification `format`
+   * @param {string} columnName - The column to read the configuration from
+   * @param {object} value - The actual value to be presented to the component
+   */
+  getFormattedValue(columnName, value) {
```
mddilley: This is going to be very handy!

PR #270, opened 2019-10-09
VZE: Modify CSV export to include more than what is displayed in view

### LINKED ISSUE #161: VZE: Modify CSV export to include more than what is displayed in view
Need more feedback from the Vision Zero Team on use cases of exporting CSVs. Moved from #144 to close that out.

Need to think about whether it makes sense to preserve existing function (export only rows and columns displayed and select # of records by setting "Rows per page") or to extend functionality to allow choosing columns and number of records before export.

mddilley:
Closes #161


mddilley:
This PR refactors the export feature as `<GridExportData />` and adds `columnsToExport` as a prop to `<GridTable />` which contains a string of columns for the .csv file. The export options now open as a modal where the limit can be defined. The component uses the `useLazyQuery` hook to execute the query whenever the modal is opened or when a change in the limit is made (through a number input or an "All" checkbox).

This PR also addresses the issue of objects exporting as references (obscuring the nested data) in the Location Crashes table. The export now takes all nested data and brings it to the top level as export columns. Any duplicate keys, like units and unit descriptions, are concatenated.

### Export modal
Screen Shot 2019-10-10 at 3 40 11 PM

### Previous export
Screen Shot 2019-10-12 at 5 23 59 PM

### New export
Screen Shot 2019-10-12 at 5 22 08 PM

atd-vze/src/Components/GridExportData.js
```diff
@@ -0,0 +1,160 @@
...
+    // 2. Keys with values that are an array of objects
+    const newData = data.map(column => {
+      // Look through each column for nested object and move all to top level
+      Object.entries(column).forEach(([key, value]) => {
+        // Remove __typename from export (contains table name which is already in filename)
+        if (key === "__typename") {
+          delete column["__typename"];
+        } else if (Array.isArray(value)) {
+          // If value is array, recursive call and handle objects in array
+          value = formatExportData(value);
```
sergiogcx: Brilliant! 👍

atd-vze/src/Components/GridExportData.js
```diff
@@ -0,0 +1,166 @@
...
+   * @returns {array}
+   */
+  const formatExportData = data => {
+    // Moves nested data to top level object (CSVLink uses each top level key as a column header)
+    // Handles:
+    // 1. Nested objects
+    // 2. Nested arrays of objects
+    const newData = data.map(row => {
+      // Look through each row for nested objects or arrays and move all to top level
+      Object.entries(row).forEach(([key, value]) => {
```
mateoclarke: 💅

atd-vze/src/Components/GridExportData.js
```diff
@@ -0,0 +1,166 @@
...
+        <i
+          className="fa fa-save fa-2x ml-2 mt-1"
+          onClick={toggleModalAndExport}
+        />
+      </StyledSaveLink>
+      <Modal isOpen={isModalOpen} toggle={toggleModal} className={"modal-sm "}>
+        <ModalHeader toggle={toggleModal}>
+          Export to .csv ({query.limit} rows)
+        </ModalHeader>
+        <ModalBody>
```
mateoclarke: Thinking here we add a conditional `<Alert />` component if the `totalRecords` > 4k records. "For larger downloads, please expect a delay while the CSV file is generated. This may take multiple minutes."

atd-vze/src/Components/GridExportData.js
```diff
@@ -0,0 +1,166 @@
...
+              className=""
+              data={formatExportData(data[query.table])}
+              filename={query.table + moment(Date.now()).format()}
+            >
+              <Button color="primary" onClick={toggleModal}>
+                Save
+              </Button>
+            </CSVLink>
+          ) : (
+            <Spinner className="mt-2" color="primary" />
```
mateoclarke: I'm wondering if it would be at all possible to user a progress bar here instead of a spinner. I guess you would need the `<CSVLink>` component to be signaling how far along it is in processing the data which might not be possible, but just a thought....

https://localhost:3000/#/dev/base/progress-bar

mateoclarke: I guess we could also fake it by creating an estimate based on the number of records and filling up the progress bar based on that estimate... Not sure how most people do this tbh.

atd-vze/src/Components/GridExportData.js
mddilley: This is a great idea because it is a really long wait for the larger exports. Sergio brought this up as well, and I could see the spinner going for minutes on end leading to some confusion. I'll add this before merge.

atd-vze/src/Components/GridExportData.js
```diff
@@ -0,0 +1,166 @@
...
+                id="csv-number-input"
+                type="number"
+                placeholder="Number of rows"
+                min={0}
+                max={totalRecords}
+                onChange={setExportLimit}
+              />
+            </Col>
+            <Col sm="4">
+              <Input
```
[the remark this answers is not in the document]
mddilley: Yep, the tiny checkbox isn't the best look. I'll switch it out for that nice slider.

atd-vze/src/Components/GridExportData.js
mddilley: Yeah! I'll look into this, create an issue, and estimate the effort it would take. That would be a really cool improvement over staring at the spinner.

PR #277, opened 2019-10-10
Fix Null Values in Sort

### LINKED ISSUE #276: VZE: Location cost sort issues
The VZE locations table seems to show both null and 0 as the same. This can be extremely confusing to the user when sorting, since null values take priority and 0 values go last; so we need a way to separate the null values from the 0 values.

sergiogcx:
Closes #276 

I noticed grid table would show null values and 0's  both as null or empty. This can be confusing when sorting, since null values will go first, and zeroes go last, but both seem the same. So we needed to implement a way to differentiate nulls from zeroes.

### DIFF
--- a/atd-vze/src/views/Locations/Locations.js
+++ b/atd-vze/src/views/Locations/Locations.js
@@ -30,7 +30,7 @@ let queryConf = {
       sortable: true,
       label_search: null,
       label_table: "Total Crashes",
-      default: "-",
+      default: 0,
       type: "Integer",
     },
     "crashes_count_cost_summary { total_deaths }": {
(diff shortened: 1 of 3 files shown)


sergiogcx:
Before:
Screen Shot 2019-10-10 at 3 26 46 PM

After:
Screen Shot 2019-10-10 at 3 27 34 PM

atd-vze/src/views/Locations/Locations.js
```diff
@@ -30,31 +30,31 @@ let queryConf = {
...
+      default: 0,
       type: "Integer",
     },
     "crashes_count_cost_summary { est_comp_cost }": {
       searchable: false,
       sortable: true,
       label_search: null,
       label_table: "Comp. Cost",
-      default: "-",
+      default: 0,
```
mateoclarke: nice, I like it.

PR #280, opened 2019-10-11
VZE: Persist table filters within a user session (for QA)

### LINKED ISSUE #243: VZE: Persist table filters within a user session (for QA)
The workflow to have in mind here is that if a user is going through QA, they've gone through and set up the Crash table with the appropriate filters and search and they go into each records one by one down the list in the table of results. The will be going back and forth between Crash details pages and their filtered search results. 

How could we persist this search for this user going back and forth? local storage on the browser?

mddilley:
Closes #243

### DIFF
--- a/atd-vze/src/Components/GridDateRange.js
+++ b/atd-vze/src/Components/GridDateRange.js
@@ -74,9 +74,9 @@ const GridDateRange = ({ setDateRangeFilter, initStartDate, initEndDate }) => {
           .format()
       );
 
-  const maxDate = initEndDate ? parseDate(initEndDate) : new Date();
+  const maxDate = new Date();
   const [startDate, setStartDate] = useState(minDate);
-  const [endDate, setEndDate] = useState(maxDate);
+  const [endDate, setEndDate] = useState(parseDate(initEndDate));
 
   useEffect(() => {
     setDateRangeFilter({
--- a/atd-vze/src/Components/GridDateRange.js
+++ b/atd-vze/src/Components/GridDateRange.js
@@ -5,7 +5,6 @@ import moment from "moment";
 import DatePicker from "react-datepicker";
 import "react-datepicker/dist/react-datepicker.css";
 import { colors } from "../styles/colors";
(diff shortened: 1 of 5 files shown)


mddilley:
This PR adds the ability to persist table filters in `<GridTable />`. The state that holds the filter settings is saved to `localStorage` and is then loaded in `<GridTable />` from `localStorage` if it exists. I made some adjustments to the children of `<GridTable />` so that they would display the saved filters (like the advanced filter sliders, search and search columns fields, date picker, etc.).

I noticed that setting a default sort column breaks the new changes so I removed it from the Locations table for now since we don't have one on the Crashes table. The clear button also may need to be extended to clear the date range and the limit (and maybe page number?) since the defaults only kick in on the initial load of the table now.

Please try to break this - there are lots of combinations of filters, and I did my best to test them.

### Screengrab
saveFilters

atd-vze/src/Components/GridDateRange.js
```diff
@@ -103,7 +108,7 @@ const GridDateRange = ({ setDateRangeFilter, initStartDate, initEndDate }) => {
           startDate={startDate}
           endDate={endDate}
           minDate={startDate}
-          maxDate={maxDate}
+          maxDate={new Date()}
```
sergiogcx: So the max date is not persisted?

sergiogcx: It looks like it is persisted when I run it locally 👍

mddilley: I changed this because I noticed that the selectable date range window would narrow each time I selected a date earlier than today. This keeps the max selectable date as the current day.

mddilley:
That is an awesome find, @mateoclarke . 🙌  I decided to fix that bug in this PR by passing `resetPageOnSearch()` from GridTable to the GridFilters component that contains the logic for advanced filters. Then, it is called when an advanced filter is added or removed so that the page resets to 1 whenever an advanced filter change occurs.

PR #294, opened 2019-10-14
VZE: Clear date range and limit in data tables (GridTable component)

### LINKED ISSUE #290: VZE: Clear date range and limit in data tables (GridTable component)
Since table filters are now persisting, need to clear all filters upon clicking the clear button.

#280
> The clear button also may need to be extended to clear the date range and the limit (and maybe page number?) since the defaults only kick in on the initial load of the table now.

mddilley:
Closes #290


mddilley:
This PR extends the clear button within `<GridTable />` to reset the result limit, page number, and date range. The date range logic was refactored to make sure the range updates uniformly in the GraphQL query and the Datepicker components when the range is loaded from local storage, default, or user input and also when the table is cleared.

atd-vze/src/Components/GridDateRange.js
```diff
@@ -63,52 +63,62 @@ const GridDateRange = ({ setDateRangeFilter, initStartDate, initEndDate }) => {
...
-    : new Date(
-        moment()
-          .subtract(1, "year")
-          .format()
-      );
+  const minDate = new Date(
+    moment()
+      .subtract(1, "year")
+      .format()
+  );
```
sergiogcx: The prettier log shows this variable as unused, should we remove it if not needed?

```
./src/Components/GridDateRange.js
  Line 69:  'minDate' is assigned a value but never used  no-unused-vars
```

mddilley: Thanks for catching this - I removed it before merge.

PR #295, opened 2019-10-14
Display comp cost per location

### LINKED ISSUE #250: VZE: Display sum of comprehensive costs for all crashes at a location on Location Details page
- [ ] aggregated statistic to the location details page for all crashes by location

tcwcodes:
Fixes #250 

This totally works, but there may be a better/more programmatic method for getting the nested value (comprehensive cost in this case). Happy to rework if necessary.

Screen Shot 2019-10-14 at 4 23 50 PM


mateoclarke:
@tcwcodes hope you don't mind but I'm going to go ahead and merge this for you so we can include it in our demo to VZ tomorrow...

PR #303, opened 2019-10-15
Clear query offset before export

### LINKED ISSUE #302: Bug: VZE CSV downloads affected by page number
It appears the CSV download contains all the data only if you are on the 1st page, probably the offset is not cleared whenever the CSV query is generated.

mddilley:
Closes #302 

In this PR, I added one line to the gqlAbstract class method `queryCSV()` that sets offset to 0 before returning the GraphQL query used for export.

### DIFF
--- a/atd-vze/src/queries/gqlAbstract.js
+++ b/atd-vze/src/queries/gqlAbstract.js
@@ -344,9 +344,9 @@ gqlAbstractTableAggregateName (
       let or = [];
       for (let [key, value] of this.getEntries("where")) {
         // If we have a nested expression for a key, then append to 'or'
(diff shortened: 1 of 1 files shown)


atd-vze/src/queries/gqlAbstract.js
```diff
@@ -344,9 +344,9 @@ gqlAbstractTableAggregateName (
       let or = [];
       for (let [key, value] of this.getEntries("where")) {
         // If we have a nested expression for a key, then append to 'or'
-        if(this.isNestedKey(key)) {
+        if (this.isNestedKey(key)) {
```
mddilley: Prettier changes

PR #325, opened 2019-10-17
VZE: Drive Location Details stats with table filters

### LINKED ISSUE #246: VZE: Location details page summary statistics (charts and counts) should drive table filtering
- [x] When a search filter updates on the crashes table within the Location Details page, we should update the data viz components to use only the filtered data.

- [x] Moved to #352 ~Date range of current data view displayed next to `<h2>` at top to signal to the user that the numbers/charts are not a total count but a subset when its in that state.~

mddilley:
Closes #246


mddilley:
This PR updates the widgets and charts in the Location view to be driven by the data filtered in the Location Crashes table. The widgets, doughnut chart, and horizontal chart were refactored as optional children components of GridTable. The charts were also refactored to query data from the Crashes table in Hasura instead of the Locations table to allow them to mirror changes from the GridTable filters (which are applied to a Crashes query).

Configurations for the widgets and charts are now defined in `LocationsCrashes.js` and passed down to `<GridTable />`.

I updated the layout so that I could see the changes in the charts and widgets when I was applying the filters, but we can change that since I was focusing on function with that change.

### Charts and widgets with default filters
Screen Shot 2019-10-28 at 9 57 03 AM

### Charts and widgets with serious injuries advanced filters and different date range
Screen Shot 2019-10-28 at 10 16 11 AM

atd-vze/src/Components/gridTableHelpers.js
```diff
@@ -0,0 +1,38 @@
...
+
+export const renderAlert = alert => <Alert color="warning">{alert}</Alert>;
+
+export const getChartData = (chartData, config) => {
+  // Look for records that match labels and increment total
+  const data = config.labels.map(type => {
+    let typeTotal = 0;
+    chartData[config.table].forEach(record => {
+      // If record is a string, increment total if match
+      if (config.isSingleRecord) {
```
mateoclarke: you could get rid of one level of nested `if`s if line 13 were just: 
`if(config.isSingleRecord && record[config.nestedKey][config.nestedPath] === type) {`

not sure that makes the code that much better 🤷‍♂

PR #337, opened 2019-10-21
Init VZV and create dashboard components

### LINKED ISSUE #335: VZV: Implement static VZ Dashboard
Using the paper copy that VZ team circulates as a first draft/MVP of VZV 

- query against Socrata
- create new react app in vzv directory of atd-vz-data
- 4 different graphs/counts types
- at least 7 different queries

mddilley:
Closes #335 

Initialize VZV with CRA and add VZ Dashboard components outline


mddilley:
This PR outlines the VZV Dashboard starting point and sticks to the libraries we used in VZE. I haven't added CoreUI yet, but the biggest takeaway is that we need data about vehicle mode in the open dataset for the pie charts.

PR #358, opened 2019-10-28
Update Location Details map Nearmap API key, whitelist VZE, and add polygon

### LINKED ISSUE #228: VZE: Update Location Details map Nearmap API key, whitelist VZE, and add polygon
This finishes out the aerial map of the Location Details view and consists of the following

- [x] Add Mapbox satellite layer until key is obtained
- [x] Contact CTM Help Desk
- [x] Create new Nearmap app and API key
- [x] Whitelist VZE production and staging
- [x] Add location polygon to aerial map
- [x] Add comments to README.md and add comment in codebase where API key is used to document how key was obtained and who to go to if changes need to be made

PR #221:
> As we discussed IRL, let's get our Nearmap token provisioned from a separate app by escalating to CTM GIS support if needed. And from there, we can also whitelist our production and staging instances. With those measures in place, we can store the API key as a env var. We can assume CTM GIS will track usage and let us know if there is unsustainable usage coming from external use somehow.

mddilley:
Closes #228


mddilley:
This PR adds a new API key that has staging and production URLs for VZE whitelisted. This also overlays the polygon on the aerial map. I commented out the Mapbox satellite layer so we can use it for testing or as a fallback. Also, `react-map-gl` version was updated to fix an import issue with `<Source />` and `<Layer />` ([link](https://github.com/uber/react-map-gl/issues/902)).

As I was adding the polygon, I started thinking that adding a slider or some way to control the opacity of the polygon could improve this map. I could create another issue for that if y'all think it would be worth the time.

### Aerial map with polygon overlay (using Mapbox satellite layer locally)
Screen Shot 2019-10-28 at 2 44 40 PM

mateoclarke:
I hope you don't mind me adding two commits to your branch. I thought it would be easier to play around in the code than to leave my ideas as comments.

1. I wanted to try to make the fallback to Mapbox more programmit so I'm checking the `window.location` to see if we're on localhost.
2. Because you mentioned opacity, I thought a line paint might look better than the fill and I think I do like it more. What do you think?

Screen Shot 2019-10-28 at 6 24 21 PM

atd-vze/src/views/Locations/LocationMap.js
```diff
@@ -1,12 +1,19 @@
...
+} from "react-map-gl";
 import axios from "axios";
 import moment from "moment";
 
 import styled from "styled-components";
+import { colors } from "../../styles/colors";
 import { Button } from "reactstrap";
 
 const TOKEN = process.env.REACT_APP_MAPBOX_TOKEN;
+// This API key is managed by CTM. Contact help desk for maintenance and troubleshooting.
```
mateoclarke: 👍

atd-vze/src/views/Locations/LocationMap.js
```diff
@@ -134,15 +134,17 @@ export default class LocationMap extends Component {
 
   render() {
     const { viewport } = this.state;
+    const isDev = window.location.hostname === "localhost";
```
mddilley: 🙌

PR #367, opened 2019-10-28
years_life_lost

### LINKED ISSUE #355: Replace pie charts with horizontal bar charts on VZV
Replace the following pie charts with bar charts

### LINKED ISSUE #359: VZV: Number of years of life lost viz
monthly or yearly chart for years of life lost

tcwcodes:
Fixes #359 

Very simple component displaying current year's years of life lost YTD and previous year's years of life lost. Created a new dataset in Socrata that this component queries. So far I've only added yearly data. I can turn this into a line graph of some kind to show year over year trends, or add monthly data and show that for the current year. OR I could create a line graph with two lines showing monthly trends for the current year vs. the last year similar to the existing Serious Injury + Fatal Crashes by Month component.

Fixes #355 in that it converts the Serious Injury + Fatal Crashes by Mode pie chart into a horizontal bar chart. It should be noted that the chart is still using dummy data.

Screen Shot 2019-10-28 at 5 18 27 PM


tcwcodes:
That all makes sense! Sorry, I hadn't considered the YTD issue and the fact that it's not an apples-to-apples comparison. I'll work on making it more fine-grained by adding monthly figures and creating a file with the GraphQL queries. I suppose in that case it will be an apples to apples comparison, but as you said it will always be through the previous month, so up to a month behind. If there is a way to automate the process to do a daily GraphiQL query on the Hasura database and add a new daily record to the Socrata dataset, perhaps we can get a comparison up to the previous day? This would be the closest I think we could get without just posting all the data to Socrata every day (minus sensitive fields) and querying Socrata directly. Sorry if I'm not using the right terminology, but I hope that makes sense! Just brainstorming ...

tcwcodes:
Hey @mateoclarke, I just resubmitted this PR to populate years of life lost YTD by directly querying the new Vision Zero demographics Socrata dataset I published last week, rather than relying on aggregates as before. The widget now compares the current year and previous year up to the current date of each year, so when we get the ETL up and running people should be getting really up-to-date information. The figure returned for the current year matches exactly with what's in the VZE dashboard, which was populated from the aggregate tables in Hasura, so it appears we're getting the same outputs! I also created a queries folder and added the Hasura query that I used to populate the Vision Zero demographics Socrata dataset and saved it into a js file using the gql syntax for future reference.

I think this was a good first step and a good way to test the new dataset. I'm going to work on other widgets using the new data that's available (looking at other demographics, like gender) and ways to make this years of life lost one more robust (like showing more years/trendlines outside of just current vs. previous year). In the meantime, can you merge this code in? Thanks!

Screen Shot 2019-11-14 at 7 05 59 PM

tcwcodes:
Haha, thanks! Yup, I saved the script there just for reference until we get
the ETL up and running. I'll talk to Mike and make sure to move it in an
upcoming PR.

PR #373, opened 2019-10-29
Location Details: Replace Type of Vehicle pie chart with horizontal bar chart of crashes by mode

### LINKED ISSUE #372: VZE: Location Details: Replace Type of Vehicle pie chart with horizontal bar chart of crashes by mode
Replace the chart pictured in the attached screenshot with a chart of crashes by mode.

mddilley:
Closes #372

### DIFF
--- a/atd-vze/src/Components/GridTableWidgets.js
+++ b/atd-vze/src/Components/GridTableWidgets.js
@@ -29,7 +29,7 @@ const GridTableWidgets = ({ aggData, widgetsConfig }) => {
     } else if (!!data) {
       return data.toString();
     } else {
-      return "0";
+      return "--";
     }
   };
 
--- a/atd-vze/src/Components/GridTableWidgets.js
+++ b/atd-vze/src/Components/GridTableWidgets.js
@@ -23,10 +23,13 @@ const GridTableWidgets = ({ aggData, widgetsConfig }) => {
   };
 
   const formatWidgetData = (data, format) => {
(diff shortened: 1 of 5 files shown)


atd-vze/src/Components/GridTableWidgets.js
```diff
@@ -23,10 +23,13 @@ const GridTableWidgets = ({ aggData, widgetsConfig }) => {
   };
 
   const formatWidgetData = (data, format) => {
-    if (format === "dollars") {
+    // Handles null return values and format
+    if (!!data && format === "dollars") {
       return formatCostToDollars(data);
-    } else {
+    } else if (!!data) {
```
mateoclarke: 👍

atd-vze/src/Components/GridTableWidgets.js
```diff
@@ -23,10 +23,13 @@ const GridTableWidgets = ({ aggData, widgetsConfig }) => {
...
       return data.toString();
+    } else {
+      return "0";
```
mateoclarke: just wondering if "0" is always going to be the right value to return. Maybe "--" or something that indicates "no data" might be a better one-size-fits-all solution?

mddilley: This is a good idea because "0" could be misleading. I'll add "--" for now and we could update with something else if we think of a good icon or if it produces any confusion to test users.

PR #374, opened 2019-10-30
Add download CR3 button to Crash Details page

### LINKED ISSUE #330: ETL: Capybara for CR3s to help initial set of QA
from either Brazos or CRIS Search we need to gather CR3 forms for: 

First we'll need to identify the Case ID & Crash ID numbers for crashes with
- no lat/lon
- fatalities 
- OR serious injuries 

As a secondary set, we could also gather IDs for crashes 
- that have been geocoded by Here
- AND have fatalities 

With these case/crash IDs, we should be able to write a Capybara program to go through interface and one by one download CR3s and store the PDF files securely S3.

mateoclarke:
Closes #330 

## Empty State
Screen Shot 2019-10-30 at 3 55 41 PM

## CR3 Available
Screen Shot 2019-10-30 at 3 55 28 PM

### DIFF
--- a/atd-vze/src/views/Crashes/CR3Record.js
+++ b/atd-vze/src/views/Crashes/CR3Record.js
@@ -1,0 +1,38 @@
+import React, { Component, useState } from "react";
+import { Card, CardHeader, CardBody, Button } from "reactstrap";
+import axios from "axios";
+
+function CR3Record(props) {
+  const requestCR3 = () => {
+    console.log(props);
+    let requestUrl =
+      "https://atd-vz-api-staging.austinmobility.io/cr3/download/11156151";
+    let token = window.localStorage.getItem("id_token");
+
+    axios
+      .request(requestUrl, {
+        headers: {
+          Authorization: `Bearer ${token}`,
+        },
+      })
+      .then(res => {
+        console.log("res", res);
+        var win = window.open(res, "_blank");
+        console.log(" should open a new window");
+        win.focus();
+      });
+  };
+
+  return (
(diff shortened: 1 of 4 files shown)


atd-vze/src/views/Crashes/CR3Record.js
```diff
@@ -0,0 +1,44 @@
+import React from "react";
+import { Card, CardHeader, CardBody, Button, Alert } from "reactstrap";
+import axios from "axios";
+
+function CR3Record(props) {
+  const requestCR3 = () => {
+    console.log(props);
```
mddilley: There might be a good reason for this but just in case.

mateoclarke: nope

johnclary:
@mateoclarke i love this and can't believe y'all have got this implemented so quickly. imho it's better UX to put the cr3 card above the related records. let me know if you want a separate issue for it.

atd-vze/src/views/Crashes/Crash.js
```diff
@@ -224,8 +224,8 @@ function Crash(props) {
           </div>
         </Col>
         <Col>
-          <CrashCollapses data={data} />
           <CR3Record crashId={crashId} isCr3Stored={cr3StoredFlag === "Y"} />
+          <CrashCollapses data={data} />
```
johnclary: 😵

mateoclarke: 😎

PR #375, opened 2019-10-30
VZE: Add badges to GridTable component to indicate filters enabled

### LINKED ISSUE #352: VZE: Add badges to GridTable component to indicate filters enabled
This issue comes out of #246. Need a way to indicate to the user which filters are applied to the `<GridTable />` results and children visualizations  (`<GridTableWidgets />`, `<GridTableHorizontalBar />`, and `<GridTableDoughnut />`).

@sergio suggested badges for this feature, and we think that is a great fit.

From #246:
>Date range of current data view displayed next to `<h2>` at top to signal to the user that the numbers/charts are not a total count but a subset when it's in that state.

mddilley:
Closes #352

### DIFF
--- a/atd-vze/src/Components/GridTableFilterBadges.js
+++ b/atd-vze/src/Components/GridTableFilterBadges.js
@@ -18,6 +18,7 @@ const GridTableFilterBadges = ({
     const searchText = searchParams.value
       ? `Search: ${searchParams.value}`
       : null;
+
     setSearchBadgeText(searchText);
   }, [searchParams]);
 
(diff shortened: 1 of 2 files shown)


mddilley:
This PR adds badges within the card body of `<GridTable />` to indicate which filters are active on the data feeding the table and visualizations. I created `<GridTableFilterBadges />` that uses state passed down from `<GridTable />` and the `customFilters` object that defines the advanced filters to add badges for active filters as they are applied.

I'd appreciate any feedback on where the badges are placed and their color and size. I tested making them a bit bigger but it looked odd with the badges having larger text than the table header.

### Badges in table body
Screen Shot 2019-10-30 at 3 10 15 PM

johnclary:
@mddilley awesome! i think the sizing is good. maybe make the badge color the same dark color as the "advanced filters" button? how about a funnel icon instead of the `Filters applied:` label?

perhaps it makes sense to stick the badges row between the advanced filters row and the pagination row? so when the advanced filters are collapsed, the badges are visible right there to indicate what's active? i don't feel strongly about any of this

atd-vze/src/Components/GridTableFilterBadges.js
```diff
@@ -0,0 +1,91 @@
...
+    const humanReadableBadgeText = updateFiltersWithConfigLabels(
+      advancedFilterText
+    );
+    setAdvancedFilterBadgeText(humanReadableBadgeText);
+  }, [advancedFilterParams]);
+
+  return (
+    <Row className="mb-2">
+      <Col>
+        <span className="mr-1">Filters applied:</span>
```
mateoclarke: Looks great overall. Could we make it bigger though? Maybe Wring the `<span>` and the Badge components in an `<h3>` ?

mddilley: Yep, I agree with this since a date range will always be shown with the default of one year. I incorporated these changes along with @johnclary's feedback about the icons instead of text.

Screen Shot 2019-10-31 at 2 04 33 PM

PR #382, opened 2019-10-31
Fatality flag future edits did not appear to be saved

### LINKED ISSUE #313: Crash Details: After editing the fatality flag future edits did not appear to be saved
After editing the fatality flag, future edits did not appear to be saved (at least, they were not reflected in the UI)

Try editing the fatality flag in this crash:
https://vze-staging.austintexas.io/#/crashes/16871421

mddilley:
Closes #313 

This PR makes fatality flag non-editable since its value will change on the DB level as the value of death count changes ( if death count > 0 then fatality flag is "Y" and if death count = 0 then fatality flag is "N"). It also changes the order of the death count and fatality flag in the UI.

### Death count is 0
Screen Shot 2019-10-31 at 1 38 33 PM

### Death count is greater than 0
Screen Shot 2019-10-31 at 1 38 47 PM

### DIFF
--- a/atd-vze/src/views/Crashes/crashDataMap.js
+++ b/atd-vze/src/views/Crashes/crashDataMap.js
@@ -92,17 +92,17 @@ export const crashDataMap = [
   {
     title: "Fatalities/Injuries",
     fields: {
-      apd_confirmed_fatality: {
-        label: "Fatality Flag",
-        editable: true,
-        uiType: "select",
-        lookupOptions: "atd_txdot__y_n_lkp",
-        lookupPrefix: "y_n",
-      },
       apd_confirmed_death_count: {
         label: "Death Count",
         editable: true,
         uiType: "text",
+      },
+      apd_confirmed_fatality: {
+        label: "Fatality Flag",
+        editable: false,
+        uiType: "select",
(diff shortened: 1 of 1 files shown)


atd-vze/src/views/Crashes/crashDataMap.js
```diff
@@ -92,18 +92,15 @@ export const crashDataMap = [
   {
     title: "Fatalities/Injuries",
     fields: {
-      apd_confirmed_fatality: {
-        label: "Fatality Flag",
-        editable: true,
-        uiType: "select",
-        lookupOptions: "atd_txdot__y_n_lkp",
-        lookupPrefix: "y_n",
```
mateoclarke: you can leave the `lookupOptions` and the `lookupPrefix`. This just enables the field to display as `YES` or `NO` instead of `Y` or `N`. And keeping those would be consistent with the rest of the fields on the "Flags" card section.

mddilley: Oops - good catch! I replaced those options and now we have `YES` or `NO` again.

Screen Shot 2019-10-31 at 2 20 07 PM

PR #390, opened 2019-10-31
VZE: Map pin drop cancel bug

### LINKED ISSUE #377: VZE: Map pin drop cancel bug
> The mapping tool with the pin drop – I’m noticing as I mess around with it that it saves a new location even when I hit “cancel.”  Is that just because it’s not live yet?

mddilley:
Closes #377


mddilley:
This PR fixes a bug where the `CrashEditCoordsMap` component was not calling the correct function when the “cancel” button was clicked due to a type in the function name. The form wrapper in the JSX also had an onSubmit handler that was removed to keep all click handlers in the `Button` components within. This also stops query strings from being added to the url which I missed previously.

I also added the `componentDidUpdate` lifecycle method  to `CrashMap` to catch updates to the `latitude_primary` and `longitude_primary`. `CrashMap` was centering on outdated lat/lon after editing or updating the location, and it now centers on the updated point.

PR #399, opened 2019-11-04
VZE: Enable filters on Locations Crashes GridTable too

### LINKED ISSUE #333: VZE: Enable filters on Locations Crashes GridTable too
Same as the new new advanced filters added for Crash GridTable on mode (aka. Unit Desc)

mddilley:
Closes #333


mddilley:
This PR unifies the columns and advanced filters between the Crashes table and the Location Crashes table. It also extends `gqlAbstract.queryAggregate()` to handle nested keys in the same manner that they are handled in `gqlAbstract.generateFilters()`. This allows the aggregate queries to filter on units involved in the crashes as well.

### Unit filters in Location Crashes table
Screen Shot 2019-11-04 at 3 10 58 PM

PR #405, opened 2019-11-05
VZE: Bug with new records: "null & null" location and "null" fatalities

### LINKED ISSUE #400: VZE: Bug with new records: "null & null" location and "null" fatalities
It appears that whenever there is a new record that has not been assigned to a location the crash details page, this could be a common thing with new records coming in, ie: 

https://visionzero.austin.gov/editor/#/crashes/17333080

mddilley:
Closes #400

### DIFF
--- a/atd-vze/src/views/Crashes/Crash.js
+++ b/atd-vze/src/views/Crashes/Crash.js
@@ -117,13 +117,17 @@ function Crash(props) {
     <div className="animated fadeIn">
       <Row>
         <Col>
-          <h2 className="h2 mb-3">{`${primaryAddress} & ${secondaryAddress}`}</h2>
+          {primaryAddress && secondaryAddress ? (
+            <h2 className="h2 mb-3">{`${primaryAddress} & ${secondaryAddress}`}</h2>
+          ) : (
+            <h2 className="h2 mb-3">{`Location Not Yet Assigned`}</h2>
+          )}
         </Col>
       </Row>
       <Row>
         <Col xs="12" sm="6" md="4">
           <Widget02
-            header={`${deathCount}`}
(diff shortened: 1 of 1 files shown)


mddilley:
This PR adds some handling for `null` values in the Crash view header and widgets.

### Before
Screen Shot 2019-11-05 at 11 43 50 AM

### After 
Screen Shot 2019-11-05 at 11 44 22 AM

atd-vze/src/views/Crashes/Crash.js
```diff
@@ -117,29 +117,37 @@ function Crash(props) {
     <div className="animated fadeIn">
       <Row>
         <Col>
-          <h2 className="h2 mb-3">{`${primaryAddress} & ${secondaryAddress}`}</h2>
+          {primaryAddress && secondaryAddress ? (
+            <h2 className="h2 mb-3">{`${primaryAddress} & ${secondaryAddress}`}</h2>
+          ) : (
+            <h2 className="h2 mb-3">{`Location Not Yet Assigned`}</h2>
```
mateoclarke: I don't think this is the accurate way to describe it and might lead to confusion. A location association is different from having missing address fields. 

What if we did this instead:

```js
<h2 className="h2 mb-3">
  {`${primaryAddress ? primaryAddress : "PRIMARY ADDRESS MISSING"} & ${secondaryAddress ? secondaryAddress : "SECONDARY ADDRESS MISSING}`}
</h2>
```

mddilley: I'm glad you pointed this out, because it does make it sound like it has something to do with the Location view. I just fixed this in #412 so it will be clearer.

PR #428, opened 2019-11-08
Change `Fatality Flag` to `APD Fatality Flag`

### LINKED ISSUE #427: Change `Fatality Flag` to `APD Fatality Flag`
There are discrepancies in what APD classifies as a fatality vs what ATD classifies as a fatality. For example, APD may not classify a collision as a fatality because it is in a private drive, a suicide, etc. ATD still wants to track these as fatalities. 

The current state in VZE is that the Fatality Flag pulls the APD record. We need to change `Fatality Flag` to `APD Fatality Flag` so that editors have clarity on what classification they are looking at.

mddilley:
Closes #427

### DIFF
--- a/atd-vze/src/Components/DataTable.js
+++ b/atd-vze/src/Components/DataTable.js
@@ -98,7 +98,11 @@ const DataTable = ({
                           </td>
                           <td>
                             {isEditing ? (
-                              <form onSubmit={e => handleFieldUpdate(e)}>
+                              <form
+                                onSubmit={e =>
+                                  handleFieldUpdate(e, dataMap, field)
+                                }
+                              >
                                 {fieldUiType === "select" && (
                                   <Input
                                     name={field}
--- a/atd-vze/src/Components/DataTable.js
+++ b/atd-vze/src/Components/DataTable.js
@@ -37,7 +37,8 @@ const DataTable = ({
 
   const onButtonClick = (e, section) => {
     handleFieldUpdate && handleFieldUpdate(e, section, "button");
-    handleButtonClick && handleButtonClick(e, dataMap, data);
+    handleButtonClick &&
+      handleButtonClick(e, section.button.buttonFieldUpdate, data);
   };
 
   return (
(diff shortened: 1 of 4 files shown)


mddilley:
This PR extends `<DataTable />` to handle secondary field updates (ex. updating a field through the edit form also updates a secondary field behind the scenes). It also adds an optional button that can mutate data when clicked (in this case, `apd_human_update`). The former Fatalites/Injuries data table is also now split into two data tables.

@mateoclarke @sergiogcx I think one more step is required to finish this up. When clicking the "Revert to CRIS Data" button, I think it would make sense to revert the value of `apd_confirmed_death_count` to the value of `death_cnt` with a DB trigger. Right now, it changes the `apd_human_update` flag from "Y" to "N" from the frontend, but the user edited value remains. What do y'all think?

### No edits from a user
Screen Shot 2019-11-08 at 4 06 04 PM

### After user edit
Screen Shot 2019-11-08 at 4 05 41 PM

atd-vze/src/Components/DataTable.js
```diff
@@ -26,9 +35,18 @@ const DataTable = ({
...
+    handleFieldUpdate && handleFieldUpdate(e, section, "button");
+    handleButtonClick &&
+      handleButtonClick(e, section.button.buttonFieldUpdate, data);
+  };
+
   return (
     <>
       {dataMap.map((section, i) => {
+        const buttonCondition =
+          section.button && section.button.buttonCondition;
```
sergiogcx: After speaking with Mike we decided to remove the extra `section.button.buttonCondition` statement that was present at a previous commit.

atd-vze/src/Components/DataTable.js
```diff
@@ -26,9 +35,18 @@ const DataTable = ({
     setEditField("");
   };
 
+  const onButtonClick = (e, section) => {
```
mateoclarke: This implementation looks great! One recommendation that I think is a good UX pattern for reverting changes or deleting data is prompting at confirm dialog box. Could we add that to this feature?

https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm

PR #465, opened 2019-11-21
VZE: Update the Crash Location view on the crash record page

### LINKED ISSUE #457: VZE update the Crash Location view on the crash record page
Our customers are finding it difficult to see which records have locations which have been manually verified and/or updated. Currently the Crash Location view often shows “Crash Location (Unassigned)” above the map, which can be confusing to people. 

We need to implement two changes to make this view more intuitive:
- The Crash Location header should always include the following: `Crash Location (ID: )`. If there is a corresponding Crash Location ID associated with the crash, we should populate the ID within the parentheses. If there is no corresponding Crash Location ID associated with the crash, we should add `ID: unassigned`. See screenshots for comparison.

-There should be a new field under the Crash Location header for the crash location geospatial source. The correct source of the geographic information should be displayed in this field. For example, if the geographic information is sourced from CRIS , this field should display `Source: CRIS `. If a user manually updates the geographic information, this field should update to `Source: Manual Update`. If the geographic information is from the Geocoder, display `Source: Geocoder`. If there are No Primary Coordinates, display `Source: No Primary Coordinates`. Below is an example.

mddilley:
Closes #457

### DIFF
--- a/atd-vze/src/views/Crashes/Crash.js
+++ b/atd-vze/src/views/Crashes/Crash.js
@@ -25,6 +25,7 @@ import { crashDataMap } from "./crashDataMap";
 import "./crash.scss";
 
 import { GET_CRASH, UPDATE_CRASH } from "../../queries/crashes";
+import { GET_GEOCODERS } from "../../queries/lookups";
 
 const calculateYearsLifeLost = people => {
   // Assume 75 year life expectancy,
@@ -48,6 +49,8 @@ function Crash(props) {
   const { loading, error, data, refetch } = useQuery(GET_CRASH, {
     variables: { crashId },
   });
+
+  const { data: geocoderOptions } = useQuery(GET_GEOCODERS);
 
   const [editField, setEditField] = useState("");
   const [formData, setFormData] = useState({});
--- a/atd-vze/src/views/Crashes/Crash.js
+++ b/atd-vze/src/views/Crashes/Crash.js
@@ -78,12 +78,14 @@ function Crash(props) {
   };
 
   const convertGeocoderToName = geocoderID => {
-    const geocoders = geocoderOptions.atd_txdot_geocoders;
+    const geocoders = geocoderOptions.atd_txdot_geocoders || [];
     const primaryLat = data.atd_txdot_crashes[0]["latitude_primary"];
     const primaryLong = data.atd_txdot_crashes[0]["longitude_primary"];
 
-    const geocoderOption =
-      geocoders && geocoders.find(option => option.geocoder_id === geocoderID);
+    const geocoderOption = geocoders.find(
+      option => option.geocoder_id === geocoderID
+    );
+
     return primaryLat && primaryLong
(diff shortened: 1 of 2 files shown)


mddilley:
This PR adds a source field to the crash map card in the Crash view to show the user the source of the lat/lon. It also modifies the text surrounding either the location ID link or "unnassigned". The lat/lon source names are pulled from the geocoder lookup table in Hasura so the text varies a bit from the issue for manual edits ("Manual Q/A" vs. "Manual Update"). We can modify the lookup table to change this text if needed.

### No Primary Coordinates and unassigned location ID
Screen Shot 2019-11-22 at 12 21 26 PM

### Source: CRIS
Screen Shot 2019-11-22 at 12 21 09 PM

### Source: Here
Screen Shot 2019-11-22 at 12 17 15 PM

### Source: Manual Q/A
Screen Shot 2019-11-22 at 12 23 24 PM

atd-vze/src/views/Crashes/Crash.js
```diff
@@ -74,6 +77,20 @@ function Crash(props) {
     return geocoderAddressString;
   };
 
+  const convertGeocoderToName = geocoderID => {
```
mateoclarke: This certainly works as a method of matching a geocode ID to a geocode name. But I think the better way to do this would be through an association using GraphQL in the Hasura layer. 

The way to do this is to go to the Data tab in Hasura, find the table you want to query and then click the Relationships tab. In our case, we're already querying a single crash by its ID for this view so we want to add a nested relationship from crashes to geocoders via the `geocode_provider` column

Screen Shot 2019-11-25 at 6 08 23 PM

Once on that page, scroll to the bottom click "Configure" to add a new relationship manually. 

Screen Shot 2019-11-25 at 6 10 14 PM

The relationship type will be Object since its one to one. Give it a name like "geocode_method". The reference table will be our geocoder lookup table. And then match the related column names. Then click save.

Screen Shot 2019-11-25 at 6 26 59 PM

You would now be able to extend the existing `GET_CRASH` query instead of creating a new query and doing a manual join in the client like you currently are...

```graphql
{
  atd_txdot_crashes {
    geocode_method { 
      name
    }
  }
}
```

A couple gotchas... You'll may need to update the `select` permissions for the geocoder table. You also don't want to forget to make the same configuration changes in staging as you will in production. 

This update would also us to start using the geocoding name in the field data lower on the page as well. And in any other feature where we need to pull out the geocoder name instead of the ID in the future.

Screen Shot 2019-11-25 at 6 21 34 PM

I went ahead and added this relationship to staging. Could you pick it up from here and try to implement this feature to pull data via the crash query and also update the Geocode Provider field? And also make sure the same relationship is configured in production?

mddilley: I like this solution a lot better too. It is much easier to set this up in Hasura and eliminates more JS to maintain. I added this relationship to production and updated the mapping for pulling the geocoder name in PR #477

PR #480, opened 2019-11-27
Build map POCs

rgreinho:
Builds POCs to help decide how to render the information on the map.

This first map uses clustering to group all the fatality markers and
resize size based on the crash frequency. The color changes as well,
getting darker when the frequency increases.


rgreinho:
I found some interesting documentation to learn how to use the heatmap:

* [Official tutorial](https://docs.mapbox.com/help/tutorials/make-a-heatmap-with-mapbox-gl-js/)
* [Official example](https://docs.mapbox.com/mapbox-gl-js/example/heatmap-layer/)
* [Medium post](https://blog.mapbox.com/introducing-heatmaps-in-mapbox-gl-js-71355ada9e6c)

I adjusted some numbers based on what I read in these articles, and I think it looks better. 

I like the fact that at the default zoom level (set to 12), the heatmap groups the fatalities without covering everything, but still emphasizing the dangerous areas. 

Zooming down will show all the individual crashes as circles, which is very nice. Zooming up will eventually cover the full city, but the beauty of it is that we can adjust each property per zoom level! 

Therefore I think we can get rid of the clustered map and only use the heatmap.

We may want to ask the help of some designers to make the color look nicer.

sergiogcx:
Closing, some of the code here has been incorporated by @mddilley

PR #486, opened 2019-12-02
Fix gridtable search

### LINKED ISSUE #442: VZE: Search in GridTable breaks when submitting different searches (without clearing between)
To reproduce:
1. Go to https://localhost:3000/#/locations/5553969
2. Enter a search string in the search field ("mopac") and then select "Primary Address" as the search field. Then, search.
3. Without clearing, leave the same string in the search field, but change from "Primary Address" to "Secondary Address" and then search. Both filters apply to the query.
4. Clear the search filters, and the "Primary Address" filter remains in the query and cannot be cleared.

### After second search
Screen Shot 2019-11-14 at 5 54 00 PM

### After clearing
Screen Shot 2019-11-14 at 5 58 16 PM

mateoclarke:
Closes #442 

If possible, can y'all bang on this and make sure it's thoroughly tested? I tried to break it and I'm feeling pretty good about this fix but just in case...

### DIFF
--- a/atd-vze/src/Components/GridTable.js
+++ b/atd-vze/src/Components/GridTable.js
@@ -80,7 +80,7 @@ const GridTable = ({
   const [searchParameters, setSearchParameters] = useState(
     getSavedState("searchParameters") || {}
   );
-  const [collapseAdvancedFilters, setCollapseAdvacedFilters] = useState(
+  const [collapseAdvancedFilters, setCollapseAdvancedFilters] = useState(
     getSavedState("collapseAdvancedFilters") || false
   );
   const [filterOptions, setFilterOptions] = useState(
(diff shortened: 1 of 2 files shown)


atd-vze/src/Components/GridTable.js
```diff
@@ -330,6 +330,19 @@ const GridTable = ({
   if (searchParameters["column"] && searchParameters["value"]) {
     // We will need to be careful which operator we will be using depending
     // on whether the column is an integer or a string, etc.
+
+    // Create an array of the searchable filters so we can loop over them and delete
+    // any old search parameters from the 'where' query before adding new search parameters.
+    const searchableColumns = Object.keys(query.config.columns).filter(
+      columnKey => {
```
sergiogcx: The script works, but in the debugger I see this warning message:

```
Line 337:  Expected to return a value at the end of arrow function
```

Maybe there is a way to refactor in a way that does not show the warning in the future?

mateoclarke: ohh good catch, i can refactor.

PR #494, opened 2019-12-03
Compare Function

### LINKED ISSUE #200: VZD: ETL Process to handle updates on existing records
When a record gets updated, we need the ETL to:

- [x] Compare the existing record, if the record is exactly the same, then ignore the record insertion.
- [x] If the record is different, then use a weighing system to determine:
    - [x] What fields do we need to pay attention to? (Lat/long, address, case_id, ...)
    - [x] What fields do we not care about.
    - [x] Post the review on the proper table

### LINKED ISSUE #213: ETL: Implement record compare function
A record coming in from CRIS will need to be evaluated against a record from the database, both have to be the same format so they differences can be detected.

sergiogcx:
Closes #213 
Closes #200 

My apologies ahead of the reivew, it is a large number of changes, the challenge had some complexity to it.

** If it is too much work to test, feel free to just look at the syntax. **

To test, you will need to have the extract files in the data folder, then run this:

```
runetl ~/.ssh/atd-etl/etl.staging.env "app/process_hasura_import.py crash"
```

The output should show a new set of lines such as `Crash Review Request Inserted: 1234567` which indicates a record is different enough and has been inserted into a special table for further review.


atd-etl/app/process/queries.py
```diff
@@ -76,11 +76,17 @@ def search_charges(line):
     return line
 
 
-def search_units(line):
+def search_crash_query_full(crash_id, field_list):
     """
-    Generates a graphql query to search for a specific unit.
-    :param line: string - The raw CSV line
+    Generates a graphql query to search for a crash
+    :param crash_id: stirng - The Crash ID to search for.
```
mateoclarke: spelling on "stirng"

atd-etl/app/process/queries.py
```diff
@@ -76,11 +76,17 @@ def search_charges(line):
...
-    return line
\ No newline at end of file
+    return """
+        query search_crash_query {
+          atd_txdot_crashes(limit: 1, where: {crash_id: {_eq: %CRASH_ID%}}){
+            %FIELD_LIST%
+          }
+        }
+    """.replace("%CRASH_ID%", crash_id)\
+        .replace("%FIELD_LIST%", "\n            ".join(field_list))
```
mateoclarke: maybe this `.join()` should fall underneath in line with the `.replace`s

sergiogcx: good catch 👍  after review it should stay where it is, since we are using the output of join to replace the %FIELD_LIST% string.

PR #500, opened 2019-12-05
Incorrect values in VZE Dashboard

### LINKED ISSUE #499: Incorrect values in VZE Dashboard
For some reason, the serious injury and crash numbers in the VZE dashboard are incorrect. I checked in the crash page and the numbers are displaying correctly there. 

We were notified about this issue by the customer after we sent them the latest release of VZE. Can we look into why this is occurring?

mddilley:
Closes #499 

This PR adds an additional table query to the dashboard query aliased as `fatalities`. The issue arose when a filter, `apd_confirmed_fatality: { _neq: "N" }`, was added to the GraphQL query that populated the Fatalities, Serious Injuries, and Total Crashes widgets. This provided an accurate total for fatalities but it underrepresented serious injuries and total crashes. These widgets are now populated by separate queries with the appropriate filters.

### Updated dashboard
Screen Shot 2019-12-05 at 2 16 08 PM

### DIFF
--- a/atd-vze/src/queries/dashboard.js
+++ b/atd-vze/src/queries/dashboard.js
@@ -16,7 +16,7 @@ export const GET_CRASHES_YTD = gql`
         }
       }
     }
-    atd_txdot_crashes_aggregate(
+    seriousInjuriesAndTotal: atd_txdot_crashes_aggregate(
       where: {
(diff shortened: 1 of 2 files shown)


mddilley:
Thanks, y'all!

PR #510, opened 2019-12-09
Automated Geocoding

### LINKED ISSUE #318: VZD: Geocode new crash records that are received from latest import
There was a new batch of records imported in late summer 2019. We need to outline the approach we want to take in dealing with this set of problems.

We need to review the quality of our selected geocoder and determine if that meets our data quality standards.

### LINKED ISSUE #402: VZD: Implement Geocoding ETL + Special Handling for `MatchLevel`
Automate HERE geocoding ETL w/ Python + Airflow.

As described [here](https://github.com/cityofaustin/atd-vz-data/issues/378), we need to exclude geocoded records where the `MatchLevel` is not `intersection` or `houseNumber`.

### LINKED ISSUE #418: VZD: Automate crash geocoding
- Find out if Here returns County information (it does)
- Anything with no coordinates
    - maybe only geocode the address
    - Catching 
- Lock edit of lat/lon from geocoder if human has entered lat/lon based on primary lat/long columns.
    - Edge case/race condition
- Explore python libraries for address normalization and cleanup
    - https://github.com/datamade/usaddress
    - https://pypi.org/project/postal-address/

### LINKED ISSUE #495: Implement bounding box on HERE geocodes
Based on HERE geocode results, it doesn't look like we are using a bounding box. A bounding box could greatly improve the accuracy of our geocodes, and at minimum restrict out-of-bounds addresses/coordinates. For example, using "Cesar Chavez" would restrict results to Austin's Cesar Chavez Street only. 

Implement a bounding box for our HERE geocodes so that any addresses from CRIS will only return results in Austin, TX. Here are instructions for placing a bounding box over a mapview parameter:
https://developer.here.com/api-explorer/rest/geocoder/latitude-longitude-by-mapview-parameter

sergiogcx:
Closes #418 
Closes #495 
Closes #402 
Closes #318

To test, you can run this command:

```
runetl ~/.ssh/atd-etl/etl.staging.env app/process_hasura_geocode.py
```

That should geocode about 500 records in your staging environment.


atd-etl/app/process/helpers_hasura_geocode.py
```diff
@@ -0,0 +1,413 @@
...
+        "searchtext": address,
+    }
+
+    try:
+        # Make request to API Endpoint
+        return requests.get(
+            ATD_ETL_CONFIG["ATD_HERE_API_ENDPOINT"], params=parameters
+        ).json()
+        # coordinates = request.json()['Response']['View'][0]['Result'][0]['Location']['DisplayPosition']
+    except Exception as e:
```
mddilley: I know we talked about this earlier, but this is where the error msg `{"error": "Expecting value: line 1 column 1 (char 0)"}` comes from, right? Just want to make sure I understand the log.

sergiogcx: Correct, the .json method assumes there is a valid response object. If not it will raise an exception which gets caught as an error message in the lines below.

PR #526, opened 2019-12-16
ETL: Implement retry problem

### LINKED ISSUE #525: ETL: Import Retries
When inserting data, the ETL stops if it fails to insert any data and prints the problem record. This was done so that nothing would be overwritten; however, it seems that when doing unique checks, the RDS instance can take a little bit to respond, causing Hasura to abort the request, hence causing the script to stop.

I've implemented a simple loop that will try at least 5 times, and wait 5 seconds before re-attempting in order to let RDS catch up with the ETL script.

sergiogcx:
Closes #525 

Look at syntax, then you will have to trust me that it works ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,13 +8,13 @@ This folder hosts our API that securely downloads a private file from S3. It is 
 
 [more info](./atd-cr3-api/README.md)
 
-## atd-cris-capybara
+## atd-etl (Extract-Transform-Load)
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -8,13 +8,13 @@ This folder hosts our API that securely downloads a private file from S3. It is
 
 [more info](./atd-cr3-api/README.md)
 
-## atd-cris-capybara
+## atd-etl (Extract-Transform-Load)
```
mateoclarke: 👍

PR #530, opened 2019-12-17
VZE: Crash Details, saved lat/lon appears slightly different than static view

### LINKED ISSUE #314: VZE: Crash Details, saved lat/lon appears slightly different than what is shown before saving an edited
On the crash details page, when editing a lat/lon, the lat/lon that is saved appears to shift slightly from what is shown on the map before saving. Possibly a lat/lon truncation issue.

mddilley:
Closes #314 

This PR adjusts the styles of the `Pin` component that appears at the center of the map during editing and at the crash location when not editing. I added the styles suggested in an [open GH issue](https://github.com/uber/react-map-gl/issues/845) to make sure the static pin appeared at the correct coordinates at all zoom levels instead of shifting around. 

I also removed some hard-coded props on the `Marker` component in `CrashMap`. These were intended to combat the shifting but are no longer needed. The shadow was also adjusted to match the position of the pin.

### Position to save
Screen Shot 2019-12-18 at 11 37 27 AM

### Position saved (and coordinates saved to the DB)
Screen Shot 2019-12-18 at 11 37 43 AM
Screen Shot 2019-12-18 at 11 38 02 AM


atd-vze/src/views/Crashes/Maps/CrashMap.js
```diff
@@ -92,8 +92,6 @@ export default class CrashMap extends Component {
         <Marker
           latitude={this.props.data.latitude_primary}
           longitude={this.props.data.longitude_primary}
-          offsetLeft={-20}
-          offsetTop={-10}
```
mateoclarke: 🤣  too simple, haha

mddilley: Yep! 😆 I'm glad that it ended up being simpler than I thought.

PR #542, opened 2019-12-20
505 fatalities by time of day week

### LINKED ISSUE #505: VZV: Fatalities by time of day/day of week highlight table
Update the current highlight table being used for the static Vision Zero dashboard to allow users to view previous years of data. The user should have the ability to select and view fatality data from previous year that they want to view by tapping on tabs below the chart.

Organize the year selectors below the highlight table; something like this:
        2014     |     2015     |      2016    |    2017    |      2018    |     2019

tcwcodes:
Fixes #505 

Demo [here](https://gph.is/g/4AvpqJN).

Here is a working prototype of the fatalities by time of day and week module, which allows you to toggle through data going back five years. This uses D3js and a library called [Reaviz](https://reaviz.io/?path=/story/docs-intro--page).

There are a few minor differences between this and the static prototype. For example. this one doesn't include the numbers in each cell, you just hover over each cell to get that number (thanks to @mddilley for finding a solution to a tooltip styling issue within the Reaviz library). Using the tooltip instead of displaying a number within each cell keeps the module from getting too wide. @JaceDeloney, check out the demo GIF and let me know what you think! I used the same color scheme as the multiyear fatalities module to try to keep things consistent. Happy to make changes if necessary.

PS This PR also gets rid of the old years of life lost 5-year visualization.


atd-vzv/src/App.css
```diff
@@ -16,3 +16,7 @@ a:focus {
   text-decoration: none;
   transition: all 0.3s;
 }
+
+div.Tooltip-module_tooltip__omIxd {
+  background: rgb(54, 51, 51) !important;
```
mateoclarke: is there a better way to do this without throwing an `!important` bomb at the problem? Does the reaviz docs have any tips on customizing the css?

image

atd-vzv/src/views/dashboard/dashboard.js
```diff
@@ -3,8 +3,8 @@ import Fatalities from "./fatalities";
 import SeriousInjuries from "./seriousInjuries";
 import SeriousInjuryAndFatalCrashesByMonth from "./seriousInjuryAndFatalCrashesByMonth";
 import SeriousInjuryAndFatalCrashesByMode from "./seriousInjuryAndFatalCrashesByMode";
-import YearsOfLifeLost from "./yearsOfLifeLost";
 import FatalitiesMultiYear from "./FatalitiesMultiYear";
+import FatalitiesByTimeOfDayWeek from "./fatalitiesByTimeOfDayWeek";
```
mateoclarke: shouldn't we name these files with caps to be consistent with the component naming?

atd-vzv/src/views/dashboard/fatalitiesByTimeOfDayWeek.js
```diff
@@ -0,0 +1,215 @@
...
+    <Container>
+      <Row>
+        <h3>Traffic fatalities by time and day</h3>
+      </Row>
+      <Row>
+          <Heatmap
+            height={200}
+            width={400}
+            data={heatmapData}
+            series={<HeatmapSeries colorScheme={["#fee5d9", "#fcae91", "#fb6a4a", "#de2d26", "#a50f15"]} />}
```
mateoclarke: We should have our colors live in a separate js file as constants that can be imported anywhere into VZV. This will help us with consistency of our theme and color palette. 
ex: https://github.com/cityofaustin/atd-vz-data/blob/484_nonCR3_ui/atd-vzv/src/constants/colors.js

atd-vzv/src/views/dashboard/dashboard.js
tcwcodes: Updated all component filenames!

atd-vzv/src/views/dashboard/fatalitiesByTimeOfDayWeek.js
tcwcodes: Done - thanks!

atd-vzv/src/App.css
tcwcodes: This one is a little more complicated! To answer your first question, I think the CSS selector is specific enough to not have to use !important (I removed it and the viz tooltip is still showing an opaque background). I scoured the documentation and tried many ways to customize the component's tooltip element using props to no avail, and when that failed I e-mailed the developer and never got a response :( I found the code that sets these parameters in the library and it seemed like they just weren't working properly (at least in this environment), so I asked @mddilley for help overriding the particular bit of styling and he figured out how to do it (thanks again, Mike)! I'll submit the next PR without the !important bomb. Let me know if that works or if you want me to try to find a more elegant solution.

mateoclarke: Yeah Mike and I talked about it too briefly. Bummer the library doesn't have better hook-ins for custom styling and getting the job done with the `!important` makes me feel better for sure. Good stuff.

PR #576, opened 2019-12-30
Add new table for nonCR3 crashes

### LINKED ISSUE #484: VZE | New non-CR3 table UI
To complete #450, we need to implement a new table UI on the Locations Details for non-CR3s. The new table should be located in the Locations Data Section beneath the current CR3 table. A new table is needed because non-CR3 records (e.g., fender benders) have a different data model than CR3 records.

mateoclarke:
Closes #484 

Screen Shot 2020-01-02 at 12 32 02 PM

### DIFF
--- a/atd-vze/src/views/Crashes/crashGridTableParameters.js
+++ b/atd-vze/src/views/Crashes/crashGridTableParameters.js
@@ -66,6 +66,60 @@ export const crashGridTableColumns = {
   },
 };
 
+export const nonCR3CrashGridTableColumns = {
+  form_id: {
+    primary_key: false, // We say no here bc there is no page to link to
+    searchable: true,
+    sortable: true,
+    label_search: "Search by Crash Form ID",
+    label_table: "Crash Form ID",
+    type: "Int",
+  },
+  date: {
+    primary_key: false,
+    searchable: true,
+    sortable: true,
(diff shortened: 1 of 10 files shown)


atd-vze/src/views/Crashes/crashGridTableParameters.js
```diff
@@ -66,6 +66,60 @@ export const crashGridTableColumns = {
...
+    primary_key: false, // We say no here bc there is no page to link to
+    searchable: true,
+    sortable: true,
+    label_search: "Search by Crash Form ID",
+    label_table: "Crash Form ID",
+    type: "Int",
+  },
+  date: {
+    primary_key: false,
+    searchable: true,
```
mddilley: I think `searchable` needs to be `false` here. This is creating another search field in the `GridTable` that is blank and can break the query if selected.

Screen Shot 2020-01-05 at 10 23 05 AM

sergiogcx: Yep, line 80 to be specific seems to cause since there is no label defined.

I tested by turning the line to `false`, also tested to make sure the date ranges work fine when searching with that false value. It all seems to work fine:

mateoclarke:
good catch @mddilley! Thanks!

PR #587, opened 2020-01-03
VZV | Summary | Page Formatting

### LINKED ISSUE #549: VZV | Summary | Page Formatting
Implement a set of common rules for the spacing and positioning of quick views and charts using CoreUI. CoreUI is the design system that is being used in VZE. 

Below is an example of how the Summary page could look based on these common rules:

mddilley:
Closes #549


mddilley:
This PR adds rules within the `Summary` view to enforce common spacing and positioning of the child components. The key is a new class, `.summary-child`, and using reactstrap `Col` components along with `styled-components` to set padding. The content can then be defined Bootstrap grid classes to determine how many components per row at different breakpoints. The existing margin classes were removed from all of these components to start with a clean slate. A `SummaryCard` component was also added to keep visualizations equal height.

Some other small changes:
- `Dashboard` component was renamed as `Summary` to keep naming consistent between the code and the view. Other components were also renamed to stick to established naming conventions.
- Added placeholder VZ logo 
- Substituted more helpers, times, and urls in components with ones already defined in `helpers.js`, `time.js`, and `socrataQueries.js`
- Added buttons to navigation items in `Header` component
- Refactored `Header` and `Summary` with config arrays that define content of child components
 
### Screenshot
Screen Shot 2020-01-06 at 11 39 52 AM

atd-vzv/src/views/summary/FatalitiesByMode.js
```diff
@@ -13,7 +14,6 @@ const FatalitiesByMode = () => {
     { label: "Motorcycle", flag: "motorcycle_fl", color: "#de2d26" },
```
mateoclarke: I know you didn't add these here in this PR but we should probably extract those color hex IDs into a shared place.

atd-vzv/src/views/summary/Summary.js
```diff
@@ -0,0 +1,54 @@
...
+    title: "Serious Injury and Fatal Crashes by Mode"
+  },
+  { component: <YearsOfLifeLost />, title: "Year-to-Date Years of Life Lost" },
+  { component: <FatalitiesMultiYear />, title: "Traffic Fatalities by Year" }
+];
+
+const Summary = () => {
+  const StyledSummary = styled.div`
+    /* Set padding for all Summary children in grid that are Bootstrap columns and have .summary-child class */
+    .summary-child,
```
mateoclarke: 👍

PR #590, opened 2020-01-06
VZV | Map | General Page Formatting

### LINKED ISSUE #550: VZV | Map | General Page Formatting
Implement the following formatting on the map page:

There should be a menu component on the left side of the page that allows the user to select filters, overlays, and see additional information regarding the data that is displayed on the map.

Below is an example of how the Map page could look like

mddilley:
Part of #550 (formatting of controls in side menu will be added in PR for #567 to keep this PR smaller)

This PR adds the `Map` component (with static data points) and also establishes a store pattern using React Context. The idea behind the store is create a common place for state to be set and accessed globally within this React app. This enables getting and setting state of `SideDrawer` toggle and the future filters for `Map` data without having to pass them down with props or up with callbacks.

Styles for `Header` and `SideDrawer` components were updated to fill the space left over completely with the `Map` component and to enable scrolling for overflow in components other than the map. Better documentation was also added to explain how they interact based on screen size.

### Screenshot
Screen Shot 2020-01-07 at 1 38 47 PM


atd-vzv/src/utils/store.js
```diff
@@ -0,0 +1,15 @@
+import React, { useState } from "react";
+
+export const StoreContext = React.createContext(null);
```
mateoclarke: Nice use of React context 👏

PR #593, opened 2020-01-07
VZV | Map | Filter by Mode

### LINKED ISSUE #567: VZV | Map | Filter by Mode
As a user, I want to be able to filter crash data on the map by mode. 

The following filters should be available for the user to turn on/off:

- Pedestrian
- Bicycle
- Motorcycle
- Car

All mode filters should be turned on by default so that all crash data is displaying on the map when the user lands on the page.

mddilley:
Part of #550 and closes #567

This PR adds a `SideMapControl` component to contain the controls for filtering map data. Changes are set using React Context as state called `mapFilters`. Then, `mapFilters` from the Context store is a set as a dependency for `useEffect` in the `Map` component. Each time `mapFilters` updates, the `Map` component uses a helper, `createMapDataUrl`, to generate an updated API url and then refetches map data.

Also, the operator for stacking mode filters is currently set as `OR`, but it can be updated in the filters config as `AND` if needed.

### All modes selected
Screen Shot 2020-01-07 at 5 09 06 PM

### Car and bicycle selected
Screen Shot 2020-01-07 at 5 08 56 PM

### DIFF
--- a/atd-vzv/src/utils/store.js
+++ b/atd-vzv/src/utils/store.js
@@ -4,8 +4,10 @@ export const StoreContext = React.createContext(null);
 
 export default ({ children }) => {
   const [isOpen, setIsOpen] = useState(false);
+  const [mapFilters, setMapFilters] = useState([]);
 
   const store = {
+    mapFilters: [mapFilters, setMapFilters],
     sidebarToggle: [isOpen, setIsOpen]
   };
 
--- a/atd-vzv/src/views/nav/SideMapControl.js
+++ b/atd-vzv/src/views/nav/SideMapControl.js
@@ -1,9 +1,14 @@
 import React from "react";
 import { StoreContext } from "../../utils/store";
 
-import { Container, ButtonGroup, Button } from "reactstrap";
+import { Container, ButtonGroup, Button, Card } from "reactstrap";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
-import { faWalking, faBicycle, faCar } from "@fortawesome/free-solid-svg-icons";
+import {
+  faWalking,
+  faBicycle,
+  faCar,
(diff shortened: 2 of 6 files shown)


atd-vzv/src/utils/store.js
```diff
@@ -4,8 +4,10 @@ export const StoreContext = React.createContext(null);
 
 export default ({ children }) => {
   const [isOpen, setIsOpen] = useState(false);
+  const [mapFilters, setMapFilters] = useState([]);
 
   const store = {
```
mateoclarke: very cool, I like how this high level store is taking shape.

atd-vzv/src/views/nav/SideMapControl.js
```diff
@@ -0,0 +1,121 @@
...
+      );
+      setFilters(updatedFiltersArray);
+    } else {
+      const filter = modeParameters[filterName];
+      const filtersArray = [...filters, filter];
+      setFilters(filtersArray);
+    }
+  };
+
+  const handleAllFiltersClick = () => {
```
mateoclarke: I was kind of expecting that clicking "All" would visualize highlight all of the modes and the "All" button as well. Maybe that complicates the logic but its essentially the same crashes displayed as when you do this right?

Screen Shot 2020-01-10 at 9 25 51 AM

mddilley: That makes sense to me, but the way the logic works here is "All" includes the four modes selected above plus any other modes (train, etc.). It was definitely tough to figure out how this should work with the possibilities of using `OR` or `AND`. 

Right now, the above selection gives result that have pedestrian OR bicycle OR motor vehicle OR motorcycle. I'm open to any changes, and I don't think this will be that difficult to modify.

PR #602, opened 2020-01-10
Location view breaks when no CR3 or blueform crashes in 5 yr history

mddilley:
This PR modifies the `getLocationTotals` SQL fn in Hasura to create aggregate crash totals accurately even if there are 0 crashes found from either table in the date range queried.

Previously, finding no records in either table would return no rows for the aggregate. The two separate queries that are joined now return 0 for `total_crashes` and `total_est_comp_cost` in this case.

The SQL fn has been updated in Hasura staging and production.

### DIFF
--- a/atd-vzd/triggers/get_location_totals.sql
+++ b/atd-vzd/triggers/get_location_totals.sql
@@ -1,50 +1,56 @@
-CREATE OR REPLACE FUNCTION public.get_location_totals(
+CREATE
+OR REPLACE FUNCTION public.get_location_totals(
   cr3_crash_date date,
   noncr3_crash_date date,
   cr3_location character varying,
   noncr3_location character varying,
   cost_per_crash numeric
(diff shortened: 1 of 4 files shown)


mddilley:
Let me know too, @sergiogcx! I'd like to be able to do more than piece together SQL StackOverflow issues and apply them to a problem. 😆 Also, thanks for sharing about `coalesce` a while back. It came in handy for this one.

sergiogcx:
@mddilley @mateoclarke Most of the SQL i've learned through college. 

[I have this one for reference](https://www.udemy.com/course/postgresql-from-zero-to-hero/), but it is the only one I have. It has advanced topics like pg/sql stuff, subqueries, triggers, etc.

This one seems nice but basic, mostly focus on aggregations:
https://www.udemy.com/course/sql-and-postgresql-for-beginners/

PR #603, opened 2020-01-10
VZV | Map | Filter by Date Range

### LINKED ISSUE #569: VZV | Map | Filter by Date Range
As a user, I want to be able to filter crash data on the map by date. The default date range should be the rolling five years of previous data. The user should be able to select the filter button and choose a specific date range within that rolling five year period.

mddilley:
Closes #569

This PR adds the date range filter for the map data. A new component, `<SideMapControlDateRange />` contains the button to open and close the calendar and the calendar picker. A new piece of state, `mapDateRange`, was also added to the React Context store.

It also abstracts some of the color, time, and URL constants away from the components where they are used. Some of the helpers for counting records were also refactored with `.reduce()`.

### Date range button (calendar closed)
Screen Shot 2020-01-15 at 1 00 59 PM

### Date range button clicked (calendar open)
Screen Shot 2020-01-15 at 12 46 17 PM

### Mobile
Screen Shot 2020-01-15 at 12 46 44 PM


mddilley:
Thanks for the review, @sergio! Looks like master changed so I'll pull before merging. I'll remove the `x, y` variables too (left there to control where the map tooltip renders over the map, but serve no purpose now). Also, I'll look into the dependency warnings. I'd like to re-confirm that using `useEffect` as `componentDidMount` this way is best practice.

mddilley:
I addressed the linter warnings in the following ways.

1. `FatalitiesMultiYear` and `FatalitiesByTimeOfDay` - I eliminated dependencies by moving all function and variable definitions inside `useEffect`. Since they are wrapped in `useEffect`, they are no longer dependencies. If these functions/variables were needed outside the scope of `useEffect`, this would not have been a good solution.

2. `FatalitiesByMode` - I wrapped the dependency (`yearsArray`) with the `useCallback` hook (see [this](https://medium.com/@infinitypaul/reactjs-useeffect-usecallback-simplified-91e69fb0e7a3). 

[This](https://overreacted.io/a-complete-guide-to-useeffect/) is massive resource that I skimmed over that was recommended by Dan Abramov in [this](https://github.com/facebook/react/issues/14920) issue.

@tcwcodes FYI I made a few changes to some of the code you touched and mine as well to address linter warnings that Sergio mentioned above. Some good links on `useEffect` here too - I have much to learn on this!

tcwcodes:
> I addressed the linter warnings in the following ways.
> 
> 1. `FatalitiesMultiYear` and `FatalitiesByTimeOfDay` - I eliminated dependencies by moving all function and variable definitions inside `useEffect`. Since they are wrapped in `useEffect`, they are no longer dependencies. If these functions/variables were needed outside the scope of `useEffect`, this would not have been a good solution.
> 2. `FatalitiesByMode` - I wrapped the dependency (`yearsArray`) with the `useCallback` hook (see [this](https://medium.com/@infinitypaul/reactjs-useeffect-usecallback-simplified-91e69fb0e7a3)).
> 
> [This](https://overreacted.io/a-complete-guide-to-useeffect/) is massive resource that I skimmed over that was recommended by Dan Abramov in [this](https://github.com/facebook/react/issues/14920) issue.
> 
> @tcwcodes FYI I made a few changes to some of the code you touched and mine as well to address linter warnings that Sergio mentioned above. Some good links on `useEffect` here too - I have much to learn on this!

YES! Thanks @mddilley. This linter warnings were driving me crazy and I couldn't figure out how to get rid of them! I'll read up on those solutions for the future.

atd-vzv/src/views/summary/helpers/helpers.js
```diff
@@ -1,22 +1,21 @@
+import { lifespanYears } from "../../../constants/calc";
 // Helpers to handle Socrata responses for Summary view components
 
-export const calculateTotalFatalities = data => {
-  let total = 0;
-  data.forEach(record => (total += parseInt(record.death_cnt)));
-  return total;
-};
+export const calculateTotalFatalities = data =>
+  data.reduce(
```
mateoclarke: REDUCE!

PR #624, opened 2020-01-22
622 update month based on current month

### LINKED ISSUE #622: VZV | Summary | Traffic Fatalties by Year | Update month based on current month
There is a problem with the Traffic Fatalities by Year Viz. It is currently displaying: 
**“As of December, there have been 2 traffic fatalities in 2020.”**

The expected behavior is that this string updates every month based on the current month.

tcwcodes:
Fixes #622 

Fixes issue by only querying and rendering data for current year if it is past January. If this looks good, I can implement this fix on the other two VZV components.

Screen Shot 2020-01-22 at 5 40 18 PM

### DIFF
--- a/atd-vzv/src/constants/time.js
+++ b/atd-vzv/src/constants/time.js
@@ -6,6 +6,7 @@ export const oneYearAgo = moment()
   .subtract(1, "year")
   .format("YYYY-MM-DD");
 export const todayMonthYear = moment().format("-MM-DD");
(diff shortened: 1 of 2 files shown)


tcwcodes:
> Great job! This all makes sense to me. 👍

Yay, thanks! I'll implement similar fixes on the other two visualizations on Monday. @JaceDeloney, can you please make issues for those and assign them to me when you get a chance?

PR #625, opened 2020-01-23
VZV | Map | ASMP Street Level Overlay

### LINKED ISSUE #564: VZV | Map | ASMP Street Level Overlay
As a user, I want to be able to select and view ASMP Street Level designations so that we can see correlations between type of street and crashes. The user should only be able to display one overlay on the map at a time. All overlays should be inactive by default.

Here is the link to the ASMP street network rest service. It says it supports JSON: https://services.arcgis.com/0L95CJ0VTaxqcmED/arcgis/rest/services/TRANSPORTATION_asmp_street_network/FeatureServer/0

JS library that could be useful:
https://github.com/Esri/arcgis-to-geojson-utils

mddilley:
Closes #564 

This PR adds a pattern for adding overlays to the VZV map along with the ASMP street level overlay. I added a new component `SideMapControlOverlays` to contain the config for each overlay and to render button groups based on that config in a new card below the filters card. Also, another piece of state was added to the Context store to keep the selected overlay and any overlay options selected (ASMP street levels in this case). The `Map` component was also modified to render new map layers generated by some new helpers upon selecting an overlay.

### ASMP overlay not selected
Screen Shot 2020-01-22 at 9 22 50 PM

### ASMP overlay selected
Screen Shot 2020-01-22 at 9 22 57 PM

### ASMP overlay selected with some levels disabled
Screen Shot 2020-01-22 at 9 23 06 PM

### DIFF
--- a/atd-vzv/src/views/map/Map.js
+++ b/atd-vzv/src/views/map/Map.js
@@ -2,7 +2,7 @@ import React, { useState, useEffect } from "react";
 import { StoreContext } from "../../utils/store";
 import ReactMapGL, { Source, Layer } from "react-map-gl";
 import { createMapDataUrl } from "./helpers";
-import { dataLayer } from "./map-style";
+import { crashDataLayer, asmpDataLayer } from "./map-style";
 import axios from "axios";
 
 import { Card, CardBody, CardText } from "reactstrap";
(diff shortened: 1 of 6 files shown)


atd-vzv/src/views/map/Map.js
```diff
@@ -95,10 +96,15 @@ const Map = () => {
       onHover={_onHover}
     >
       {!!mapData && (
```
sergiogcx: 👍

PR #642, opened 2020-01-27
Fix time of day january issue

### LINKED ISSUE #641: VZV | Summary | Fix Jan edge case for Traffic fatalities by time of day component
Fix the same January edge case that was addressed in #622 & #623. We should only display information in VZV if we have a full month of crash data. For example: Since the full month of January 2020 is not over, we should not display January or 2020 anywhere in VZV.

tcwcodes:
Fixes #641 

Only shows current year tab if we are past January of current year. Otherwise, shows previous years.

Screen Shot 2020-01-27 at 11 51 47 AM

### DIFF
--- a/atd-vzv/src/constants/time.js
+++ b/atd-vzv/src/constants/time.js
@@ -2,15 +2,25 @@ import moment from "moment";
 
 // Time data
 export const today = moment().format("YYYY-MM-DD");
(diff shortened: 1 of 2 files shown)


tcwcodes:
Thanks! Merged

PR #645, opened 2020-01-27
640 fix jan edge case mode component

### LINKED ISSUE #640: VZV | Summary | Fix Jan edge case for Traffic fatalities by mode component
Fix the same January edge case that was addressed in #622 & #623. We should only display information in VZV if we have a full month of crash data. For example: Since the full month of January 2020 is not over, we should not display January or 2020 anywhere in VZV.

tcwcodes:
Fixes #640

### DIFF
--- a/atd-vzv/src/views/summary/FatalitiesByMode.js
+++ b/atd-vzv/src/views/summary/FatalitiesByMode.js
@@ -5,7 +5,13 @@ import { Bar } from "react-chartjs-2";
 import { colors } from "../../constants/colors";
 
 import { Container } from "reactstrap";
-import { thisMonth, thisYear, lastYear, lastMonth, lastDayOfLastMonth } from "../../constants/time";
+import {
+  thisMonth,
+  thisYear,
+  lastYear,
+  lastMonth,
+  lastDayOfLastMonth
+} from "../../constants/time";
 import { demographicsEndpointUrl } from "./queries/socrataQueries";
 
 const FatalitiesByMode = () => {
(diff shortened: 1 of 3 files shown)


atd-vzv/src/views/summary/FatalitiesByMode.js
```diff
@@ -56,8 +70,14 @@ const FatalitiesByMode = () => {
...
-    const url = `${demographicsEndpointUrl}?$limit=1&$order=crash_date DESC&$where=crash_date < '${thisYear}-12-31T23:59:59'`;
-
+    let url;
+    // If it is past January, set end of query to last day of previous month,
+    // else if it is January, set end of query to last day of last year
+    if (thisMonth > "01") {
+      url = `${demographicsEndpointUrl}?$limit=1&$order=crash_date DESC&$where=crash_date < '${thisYear}-${lastMonth}-${lastDayOfLastMonth}T23:59:59'`;
+    } else {
+      url = `${demographicsEndpointUrl}?$limit=1&$order=crash_date DESC&$where=crash_date < '${lastYear}-12-31T23:59:59'`;
+    }
```
mateoclarke: I think instead of doing the repetitive var setting for url inside the if/else here, you could just if else the endDate itself. Ex:

```js
let endDate = thisMonth > "01" 
   ? `${thisYear}-${lastMonth}-${lastDayOfLastMonth}T23:59:59` 
   : `${lastYear}-12-31T23:59:59`;

let url = `${demographicsEndpointUrl}?$limit=1&$order=crash_date DESC&$where=crash_date < ${endDate}`;
```

Maybe thats more a personal preferences that actually more readable.

tcwcodes: I like that! Definitely more eloquent. I'll make the change tomorrow morning and merge.

tcwcodes: Replaced this instance and two others with the ternary operation you suggested

PR #648, opened 2020-01-27
VZV | Map | Austin City Council Districts Overlay

### LINKED ISSUE #573: VZV | Map | Austin City Council Districts Overlay
As a user, I want to be able to see what crashes are occurring in each city council district. When a user selects this overlay, city council polygons should display with a key outlining what each polygon is.

The user should only be able to display one overlay on the map at a time. All overlays should be inactive by default.

mddilley:
Closes #573 

This PR adds the City Council districts overlay to the VZV map. Since this is a smaller layer than the other two layers implemented so far, the geojson was fetched from ArcGIS and rendered on the map.

### City Council Districts overlay
Screen Shot 2020-01-27 at 5 13 00 PM

### DIFF
--- a/atd-vzv/src/constants/colors.js
+++ b/atd-vzv/src/constants/colors.js
@@ -24,5 +24,15 @@ export const colors = {
   mapAsmp3: "#E60000",
   mapAsmp4: "#A50F15",
   mapAsmp5: "#1B519D",
-  mapHighInjuryNetwork: "#E60000"
+  mapHighInjuryNetwork: "#E60000",
+  mapCityCouncil1: " #ffffbf",
+  mapCityCouncil2: "#66c2a5",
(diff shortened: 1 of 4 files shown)


mateoclarke:
This looks great! My only feedback might be more a product question for @JaceDeloney... 

Do we want a legend for the Council Districts? I could see that showing conditionally in the sidebar when the layer is activated or as a traditional map legend in one of the corners. 

Alternatively, since this overlay is a geojson and not a tile layer like the others, we can do more interactive things with the layer. Maybe when you click on a Council polygon, you get a Popup component that tells you which district it is. Ex: https://docs.mapbox.com/mapbox-gl-js/example/popup-on-click/

Anyways, these are just random ideas, looks incredible! 🚢

PR #649, opened 2020-01-28
Fix delimiter issues in GridTableExport

mateoclarke:
This PR adds a block of string cleaning logic to the `formatData()` function.

After chatting with Xavier, the best solution for now sounds like a fix at the .csv data export level. We may decide in the future to do a bulk change to the values in the database instead.

### DIFF
--- a/atd-vze/src/Components/GridExportData.js
+++ b/atd-vze/src/Components/GridExportData.js
@@ -98,7 +98,31 @@ const GridExportData = ({ query, columnsToExport, totalRecords }) => {
       return flattenedRow;
     });
 
-    return flattenedData;
+    const cleanedAndFlattenedData = flattenedData.map(item => {
+      // Some fields are presenting issues when they are exported as .csv and
+      // imported into MSFT Excel, Acces & Power BI. For now, we're just going to
+      // modify the data for export to solve issues related to returns and double-quotes
+      // In the future, we may choose to clean these values at the database level.
+
+      const columnsToClean = [
+        "rpt_street_name",
+        "rpt_street_desc",
+        "rpt_sec_street_desc",
+        "rpt_sec_street_name",
+      ];
+
+      columnsToClean.map(col => {
+        if (item[col]) {
+          // remove return carriage and replace with space
+          item[col] = item[col].replace(/(\r\n|\n|\r|")/gm, " ");
+          // remove double-quotes
+          item[col] = item[col].replace(/"/g, "");
+        }
+      });
+      return item;
+    });
+
+    return cleanedAndFlattenedData;
   };
 
   return (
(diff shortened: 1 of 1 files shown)


atd-vze/src/Components/GridExportData.js
```diff
@@ -98,7 +98,31 @@ const GridExportData = ({ query, columnsToExport, totalRecords }) => {
...
     });
 
-    return flattenedData;
+    const cleanedAndFlattenedData = flattenedData.map(item => {
+      // Some fields are presenting issues when they are exported as .csv and
+      // imported into MSFT Excel, Acces & Power BI. For now, we're just going to
+      // modify the data for export to solve issues related to returns and double-quotes
+      // In the future, we may choose to clean these values at the database level.
+
+      const columnsToClean = [
```
mddilley: Good call on specifying which fields to avoid unnecessary cleanup and make this faster.

atd-vze/src/Components/GridExportData.js
```diff
@@ -159,6 +183,7 @@ const GridExportData = ({ query, columnsToExport, totalRecords }) => {
               className=""
               data={formatExportData(data[query.table])}
               filename={query.table + moment(Date.now()).format()}
+              // separator={"|"}
```
mddilley: Did changing the separator work out when you tested it?

atd-vze/src/Components/GridExportData.js
```diff
@@ -98,7 +98,31 @@ const GridExportData = ({ query, columnsToExport, totalRecords }) => {
...
+      // In the future, we may choose to clean these values at the database level.
+
+      const columnsToClean = [
+        "rpt_street_name",
+        "rpt_street_desc",
+        "rpt_sec_street_desc",
+        "rpt_sec_street_name",
+      ];
+
+      columnsToClean.map(col => {
```
mddilley: Just noticed this console warning when switching back to VZV. I think this would go away if `.map()` was changed to `.forEach()` here.
`Line 114:  Expected to return a value in arrow function  array-callback-return`

atd-vze/src/Components/GridExportData.js
mateoclarke: Thanks for catching this, meant to remove it. It does what you'd expect but requires a bit more configuration on the import side for whatever software you're using. Was pretty easy to figure out with Numbers but it doesn't solve the problem of returns.

Slacking you an example export with pipe delimiters.

atd-vze/src/Components/GridExportData.js
mateoclarke: thanks for this tip, makes sense after looking more into the linter warning.

PR #664, opened 2020-02-03
Added count fields to ETL as numeric

sergiogcx:


sergiogcx:
This is no longer needed, as another PR is taking care of it.

PR #667, opened 2020-02-03
VZV | Include "Other" modes in mode comparisons

### LINKED ISSUE #595: VZV | Include "Other" modes in mode comparisons
We need to include an additional "Other" mode category in all of our mode comparison visualizations so that the number of fatalities and serious injuries equal the overall fatality and serious injury totals.

We also need to create an issue to update the map mode filter to include an other category.

### LINKED ISSUE #613: VZV | Map | Remove limit from Socrata query and set type filter default to Injury
To avoid slowing down development, a limit of 1000 results was added to the Socrata query for map data. Need to remove this limit and also set the default selection for "Type" filter to "Injury". This filter also needs to be modified to show "Serious Injury" or "Fatality" only. Making these two changes at the same time will minimize the number of results rendered in the map data layer.

@JaceDeloney feel free to modify this issue. I made it so that I wouldn't forget about the limit that I set in place.

mddilley:
Closes #595 
Closes #613 

This PR updates both the map mode filters and "Fatalities by Mode" summary visualization with an "Other" mode which encompasses all other modes. "Other" modes for both are set in a new constant file, `filters.js`. 

All filter configs now have a `default` parameter to determine which filters are active on render (defined in #613 and #595). A rule of keeping one type filter active at all times was also added to keep users from fetching crashes without fatalities, serious injuries, or both. This same rule is applied to the mode filters to avoid showing a blank map with no crash points. 

I also created new time constants for the time window of data specified in #673 that are used in the map and mode visualization queries.

### Default filters with "Other" instead of "All"
Screen Shot 2020-02-05 at 1 49 34 PM

### Summary mode visualization with "Other"
Screen Shot 2020-02-05 at 4 48 10 PM

### DIFF
--- a/atd-vzv/src/constants/time.js
+++ b/atd-vzv/src/constants/time.js
@@ -2,8 +2,7 @@ import moment from "moment";
 
 // Time data
 
-// Last date of records that should be referenced in VZV
-// Show the last complete month of data
+// Last date of records that should be referenced in VZV (through last complete month of data)
 export const dataEndDate = moment()
   .subtract(1, "month")
   .endOf("month");
--- a/atd-vzv/src/constants/time.js
+++ b/atd-vzv/src/constants/time.js
@@ -1,12 +1,14 @@
 import moment from "moment";
 
-// Time data
-
 // Last date of records that should be referenced in VZV (through last complete month of data)
 export const dataEndDate = moment()
   .subtract(1, "month")
   .endOf("month");
 
+// Number of past years data to fetch
+export const rollingYearsOfData = 5;
+
+// Common time variables
 export const today = moment().format("YYYY-MM-DD");
 export const thisYear = moment().format("YYYY");
 export const oneYearAgo = moment()
@@ -29,8 +31,6 @@ export const lastMonthString = moment()
   .format("MMMM");
 
 // Map time data
-const rollingYearsOfData = 5;
-
 export const mapDataMinDate = new Date(
   moment()
     .subtract(rollingYearsOfData, "year")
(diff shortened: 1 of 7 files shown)


atd-vzv/src/constants/time.js
```diff
@@ -1,6 +1,21 @@
 import moment from "moment";
 
-// Time data
+// Number of past years data to fetch
+// 4 full years of data, plus data up to the last complete month of the current year
+export const rollingYearsOfData = 4;
```
mateoclarke: food for thought... Since this is a value that shouldn't change, a true constant variable, we might consider using the all cap convention to signal it should not be altered in the program. 

https://www.freecodecamp.org/news/when-to-capitalize-your-javascript-constants-4fabc0a4a4c4/

atd-vzv/src/views/nav/SideMapControl.js
```diff
@@ -38,76 +39,132 @@ const SideMapControl = () => {
...
+        default: false
       }
     }
   };
 
+  // Reduce all filters and set defaults as active on render
+  useEffect(() => {
+    // If no filters are applied (initial render), set all default filters
+    if (Object.keys(filters).length === 0) {
+      const initialFiltersArray = Object.entries(mapFilters).reduce(
```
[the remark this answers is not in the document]
mddilley: Thanks, Sergio! It was fun to write this, and I'm changing up the variable names to make what is being accumulated a bit clearer. Glad we talked through this yesterday!

atd-vzv/src/constants/time.js
mddilley: Yes, this is definitely the right choice here. I'll update before merge!

PR #675, opened 2020-02-06
VZV | Implement rules for displaying five previous years of data

### LINKED ISSUE #673: VZV | Implement rules for displaying five previous years of data
We have been requested to only show the five previous years of data in VZV. Below are the rules, which should apply to all data being presented in VZV (Summary page and Map page):

VZV should display all of the data for the oldest year. VZV should never display segments of the oldest year. So for example, today (Feb 5, 2020) we would show:

- All months for 2016
- All months for 2017
- All months for 2018
- All months for 2019
- January in 2020

As the year goes on we want to show the end date as the end of the previous month. In January 2021, VZV should display all of 2015 until we hit Feb 2021.

mddilley:
Closes #673 

This PR refactors the existing Summary and Map components to consider a 5 year time window set in `time.js` (four full years of previous data plus all data available for the current year). I created some high level time constants that define the earliest and latest dates for data pulled into VZV and then referenced them in the components.

The updates can be tested most easily by changing `ROLLING_YEARS_OF_DATA` or `dataStartDate` (number of months subtracted on line 19) in `/src/constants/time.js` and seeing the range of years in the map or visualizations respond.

This also resolves conflicts between the work started in this outdated branch and the merged enhancements that Tyler made around the data that populates the Summary view (Fatalities, Serious Injuries, or both).


atd-vzv/src/constants/time.js
```diff
@@ -15,27 +25,12 @@ export const dataEndDate = moment()
...
   .endOf("month");
 
-// Common time variables
-export const today = moment().format("YYYY-MM-DD");
-export const thisYear = moment().format("YYYY");
-export const oneYearAgo = moment()
-  .subtract(1, "year")
+// Summary time data
+export const summaryCurrentYearStartDate = dataEndDate
+  .clone() // Moment objects are mutable
```
mateoclarke: good to know about moment mutability!

atd-vzv/src/views/summary/FatalitiesByTimeOfDay.js
```diff
@@ -7,51 +7,25 @@ import { Nav, NavItem, NavLink, Row, Col, Container } from "reactstrap";
...
-    const dayOfWeekArray = [
-      "Sunday",
-      "Monday",
-      "Tuesday",
-      "Wednesday",
-      "Thursday",
-      "Friday",
-      "Saturday"
-    ];
+    const dayOfWeekArray = moment.weekdays();
```
mateoclarke: 🧹✨

atd-vzv/src/constants/time.js
mddilley: Yep, I felt kind of 🙃when I was trying to do math with a `moment` and kept mutating it and messing up existing references to it. Found [this](https://momentjs.com/guides/#/lib-concepts/mutability/) guide which I have never actually run into since it is separate from the docs.

atd-vzv/src/views/summary/FatalitiesByTimeOfDay.js
mddilley: I ❤️`Moment.js`

PR #683, opened 2020-02-14
VZV | Map | Dynamic Time of Day Chart

### LINKED ISSUE #660: VZV | Map | Dynamic Time of Day Chart
As a user viewing the VZV map, I want to see crashes broken down by time of day. 

Display crash data by the filters that are being used on the map. If I have filters turned on, only display crash data by time of day only for those filters. The default experience is to show crashes by time of day for all crashes.

Use the following time blocks:
12AM – 4AM
4AM – 8AM
8AM – 12PM
12PM – 4PM
4PM – 8PM
8PM – 12AM

The chart should look similar to NYC Vision Zero View

Ueer can select a time of day bar within the chart. Doing so would further filter the data displayed on the map by the time of day timeblock chosen.

To see all times, the user would select an all times button

mddilley:
Closes #660 

This PR adds an interactive chart breakdown of the filtered crashes in the map data by time of day. Each bar displays the percent of the total crashes show in the map as well as the number of crashes within a time window.

The chart bars are clickable and filter the map data to the time window represented by the bar clicked. The chart data reflects all filters except for the bar chart time window filters (otherwise it would display all other time windows as 0).

I'll run the updated ETL after merging, but the chart can be tested by updating line 8 & 9 in `socrataQueries.js` to:
```
export const crashEndpointUrl = `https://data.austintexas.gov/resource/3aut-fhzp.json`;
export const crashGeoJSONEndpointUrl = `https://data.austintexas.gov/resource/3aut-fhzp.geojson`;
```

I also made a small update to the `SideDrawer` so that the content scroll and the logo remains fixed (the logo was scrolling out of the window before).

### Time of Day chart (map filtered to "All Times")
Screen Shot 2020-02-19 at 1 54 13 PM

### Map filtered to a time window
image


atd-vzv/src/views/nav/SideMapTimeOfDayChart.js
```diff
@@ -0,0 +1,170 @@
...
+  // Get crash data without mapTimeWindow filter to populate chart
+  useEffect(() => {
+    const apiUrl = createMapDataUrl(crashEndpointUrl, mapFilters, dateRange);
+    !!apiUrl &&
+      axios.get(apiUrl).then(res => {
+        setChartData(res.data);
+      });
+  }, [dateRange, mapFilters]);
+
+  useMemo(() => {
```
mateoclarke: 🔥

PR #691, opened 2020-02-24
override bootstrap hover behavior

### LINKED ISSUE #689: Fix mobile styling on fatalities/serious injuries buttons
Need to make sure buttons are always the same size

tylerwhitson:
Fixes #689 

Uses styled components to override Bootstrap hover behavior rather than previous method of recreating Bootstrap styles from scratch. Looks and functions exactly as before, but retains Bootstrap responsive styles for improved mobile experience.

Screen Shot 2020-02-24 at 5 06 34 PM

### DIFF
--- a/atd-vzv/src/views/nav/CrashTypeSelector.js
+++ b/atd-vzv/src/views/nav/CrashTypeSelector.js
@@ -1,4 +1,5 @@
 import React, { useState, useEffect } from "react";
+import { Button, ButtonGroup } from "reactstrap";
 import styled, { css } from "styled-components";
 import { colors } from "../../constants/colors";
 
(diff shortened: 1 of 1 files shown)


atd-vzv/src/views/nav/CrashTypeSelector.js
```diff
@@ -1,4 +1,5 @@
 import React, { useState, useEffect } from "react";
+import { Button, ButtonGroup } from "reactstrap";
 import styled, { css } from "styled-components";
```
mddilley: Looks like this import is no longer used.

Screen Shot 2020-02-28 at 11 31 43 AM

tylerwhitson: Thanks! Removed

PR #705, opened 2020-03-04
361 sys viz

### LINKED ISSUE #361: VZV | Summary | Visualization | Fatalities and Serious injuries by CoA or TXDOT right of way
As a user I want to know the proportions of crashes occurring on city-owned right of way vs TXDOT right-of-way.

Display a pie chart that shows serious injuries and fatalities that occur on the City of Austin owned right of way vs TXDOT owned right of way.

City of Austin
TXDOT

tylerwhitson:
Fixes #361 

Currently using a test endpoint for the new visualization component CrashesBySystem.js. Once this is approved, I'll update the Socrata dataset and replace the endpoint. The viz just shows how many crashes took place on the TxDOT system vs. how many took place off the TxDOT system (Y and N, respectively). We could also use the private drive flag if we find that of use in the future, but per my conversation with @JaceDeloney this morning, I've left that off for now.

Apologies if this looks like a big PR, but the rest is really just updating the naming conventions for the other components to follow a specific schema based on which table/endpoint the component uses: EndpointByMetric. For example, CrashesByMonth uses the crash table and sorts the data by month, PeopleByDemographics uses the person table and sorts the data by demographics. We may want to consider renaming the Demographics Socrata dataset to People to get everything in line with the TxDOT schema, but it's not necessary.

sysviz


atd-vzv/src/views/summary/CrashesBySystem.js
```diff
@@ -0,0 +1,146 @@
...
+      getChartData();
+    }
+  }, [crashType]);
+
+  const createChartLabels = () => yearsArray().map(year => `${year}`);
+
+  // Tabulate crashes by flags in data
+  const getCategoryData = (flags, value) =>
+    yearsArray().map(year => {
+      return chartData[year].reduce((accumulator, record) => {
```
mateoclarke: 👍  Nice use of map/reduce here!

tylerwhitson:
Thanks, Mateo! Merging now

PR #708, opened 2020-03-06
VZE | Develop Admin UI (Get user, delete user, unblock user)

mddilley:
Closes #1718

This PR finishes out the User Details view with data from Auth0. It also adds buttons to delete a user or unblock a user (if they locked themselves out with too many password attempts). The unblock button only appears when user metadata field `blocked` is `true`.  I would love feedback on the fields in the User Details view. Not sure how useful some of them are, and maybe we need to ask the VZ team.

### User Details view with delete and unblock button
Screen Shot 2020-03-13 at 11 48 56 AM


atd-vze/src/views/Users/User.js
```diff
@@ -1,40 +1,159 @@
...
+      if (format === "string") {
+        formattedValue = user[key];
+      } else if (format === "bool") {
+        formattedValue = !!user[key] ? "Yes" : "No";
+      } else if (format === "time") {
+        formattedValue = moment(user[key]).format("MM/DD/YYYY, h:mm:ss a");
+      } else if (format === "object") {
+        const nestedKey = value.nestedKey;
+        formattedValue = user[key][nestedKey].join(", ");
+      }
```
mateoclarke: consider using `switch` instead of `else if` for these types of blocks.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch

mddilley: 👍I'll make this change before merging.

PR #713, opened 2020-03-09
2035 demographics whole numbers

tylerwhitson:
Fixes https://github.com/cityofaustin/atd-data-tech/issues/2035

This PR adds whole numbers to the tooltips for each category in the demographics visualization in addition to percentages. This gives the user valuable context and allows them so see how many fatalities and/or serious injuries occurred within a given category.

Screen Shot 2020-03-09 at 4 40 51 PM


atd-vzv/src/views/summary/PeopleByDemographics.js
```diff
@@ -113,7 +113,7 @@ const PeopleByDemographics = () => {
   const createChartLabels = () => yearsArray().map(year => `${year}`);
 
   // Tabulate fatalities by demographics in data
-  const getData = categoryValue =>
+  const getData = (categoryValue, getFullData) =>
```
mddilley: @mateoclarke shared [this article](https://www.samanthaming.com/tidbits/34-better-boolean-variable-names/) with me last year about naming boolean variables that could help out here. At first glance, I thought `getFullData` was a function passed as an argument.

Maybe `getFullData` could be renamed to `isWholeNumber` or something like that ? Definitely just my opinion and your comment on line 164 clarifies all of this.

atd-vzv/src/views/summary/PeopleByDemographics.js
```diff
@@ -188,7 +200,8 @@ const PeopleByDemographics = () => {
       hoverBackgroundColor: category.color,
       hoverBorderColor: category.color,
       label: category.label,
-      data: getData(category.categoryValue)
+      data: getData(category.categoryValue),
+      fullData: getData(category.categoryValue, true)
```
mddilley: Nice idea to sneak the data in this way! Very useful for the tooltip callback. 👏

atd-vzv/src/views/summary/PeopleByDemographics.js
```diff
@@ -154,8 +154,20 @@ const PeopleByDemographics = () => {
...
+        categoryTotal: categoryTotal,
+        overallTotal: overallTotal
+      }
+      // If getFullData is true, return the fullData Object for display in tooltips,
+      // else return percentage for chartJS to render the chart
+      const data = getFullData ?
+      fullData
+      : percentage
+
+      return data
```
mateoclarke: It seems like prettier isn't enabled in your code editor. If it was, there would be a `;` at the ends of lines 168 and 170. Can you check and make sure? Its important for consistency that the code we push to `master` gets automatically code formatted. Let me or @mddilley know if you need some help.

atd-vzv/src/views/summary/PeopleByDemographics.js
mateoclarke: maybe `isRoundedNumber`, then when you call it below in 203 & 204, you'd need to flip those. You may also consider using a default parameter value.
https://codeburst.io/understanding-default-parameters-in-javascript-5f3c4dcfc838
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Default_parameters

tylerwhitson: Thanks, y'all! I renamed the boolean to `isWholeNumber`and the associated object to `wholeNumbers`, I think this makes it cleaner!

atd-vzv/src/views/summary/PeopleByDemographics.js
tylerwhitson: Thanks!

atd-vzv/src/views/summary/PeopleByDemographics.js
tylerwhitson: It's enabled, I just forgot to apply the formatting before pushing. Thanks for catching that - will be more diligent in the future!

mateoclarke: sound good! Rémy also had a suggestion to automate prettier/linting on the VZE app, maybe its worth porting over those npm scripts to this directory too.

PR #715, opened 2020-03-10
Refactor auth to use Context

mddilley:
Closes part of #1820

Just a heads up, if you login using this branch, you will have to log out and then log back in before using the current staging or production VZE apps (since this code sets all user roles in a session and the current app relies on having only one role).

This PR refactors the existing auth flow to use the [Auth0 SPA SDK](https://github.com/auth0/auth0-spa-js) and enables us to expose user logged in state, permissions, etc from a context. Most of this code was set up using [this Auth0 guide](https://auth0.com/docs/quickstart/spa/react#configure-auth0) but some was adapted to make sure that Apollo is connected to Hasura before rendering protected views so GraphQL queries don't fail.

The one detail that is not ironed out yet is how user roles fit into restricting access and fetching data from Hasura. I'll work this out in the next PR.


atd-vze/src/App.js
```diff
@@ -28,179 +18,84 @@ const Register = React.lazy(() => import("./views/Pages/Register"));
...
+    const HASURA_ENDPOINT = process.env.REACT_APP_HASURA_ENDPOINT;
+
+    if (isAuthenticated && !!userClaims) {
+      const clientData = {
+        uri: HASURA_ENDPOINT,
+        headers: {
+          Authorization: `Bearer ${userClaims.__raw}`,
+          "x-hasura-role": "editor",
+        },
+      };
```
sergiogcx: The code looks a lot cleaner with hooks! 👍

atd-vze/src/App.js
```diff
@@ -28,179 +18,84 @@ const Register = React.lazy(() => import("./views/Pages/Register"));
...
-          : "";
-        window.location = prefix + "/#/login";
-      }
+const App = () => {
+  const {
+    loading,
+    loginWithRedirect,
+    isAuthenticated,
+    userClaims,
+  } = useAuth0();
```
sergiogcx: Oh, this is interesting with Auth0 SPA SDK

atd-vze/src/containers/DefaultLayout/DefaultHeader.js
```diff
@@ -19,64 +20,66 @@ const propTypes = {
...
-                <i className="fa fa-lock" /> Log Out
-              </DropdownItem>
-            </DropdownMenu>
-          </UncontrolledDropdown>
-        </Nav>
-      </React.Fragment>
-    );
-  }
-}
+              </Link>
```
sergiogcx: For some reason, the profile button stopped working for me in Chrome, though it does work in Firefox. If it works for you in chrome, ignore this comment.

mddilley: Awesome catch - I had the same issue. Looks like it was a problem with the `react-router` `Link` component being nested in this dropdown. I changed it back to the original code, and it is working reliably again.

atd-vze/src/auth/authContext.js
```diff
@@ -0,0 +1,95 @@
+import React, { useState, useEffect, useContext } from "react";
+import createAuth0Client from "@auth0/auth0-spa-js";
+
+export const Auth0Context = React.createContext();
```
mateoclarke: 👍

PR #716, opened 2020-03-13
Add Can component to restrict access based on permissions

mddilley:
Closes part of https://github.com/cityofaustin/atd-data-tech/issues/1820

This PR closes out the changes to the UI for the admin and read-only enhancements. It is a little large because I added the `Can` component wrapper around some components to control access (lots of lines didn't actually change). This component and the rules that feed it were set up with [this guide](https://auth0.com/blog/role-based-access-control-rbac-and-react-apps/).

I'd love feedback on how this is documented. Not every action defined in `rbac-rules.js` is present in the app, but I thought it documents what each role is able to do pretty well. Happy to screen share and talk through any questions.

### DIFF
--- a/atd-vze/src/containers/DefaultLayout/DefaultHeader.js
+++ b/atd-vze/src/containers/DefaultLayout/DefaultHeader.js
@@ -1,6 +1,7 @@
 import React from "react";
 import { NavLink } from "react-router-dom";
 import { useAuth0 } from "../../auth/authContext";
+import Can from "../../auth/Can";
 import {
   UncontrolledDropdown,
   DropdownItem,
@@ -21,7 +22,7 @@ const propTypes = {
 const defaultProps = {};
 
 const DefaultHeader = props => {
-  const { logout } = useAuth0();
+  const { logout, getRoles } = useAuth0();
   // eslint-disable-next-line
   const { children, ...attributes } = props;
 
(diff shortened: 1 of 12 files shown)


atd-vze/src/views/Users/Users.js
```diff
@@ -67,48 +70,54 @@ const Users = () => {
...
-                </Table>
-              ) : (
-                <Spinner className="mt-2" color="primary" />
-              )}
-            </CardBody>
-          </Card>
-        </Col>
-      </Row>
-    </div>
+    <Can
```
sergiogcx: Sweet!

atd-vze/src/containers/DefaultLayout/DefaultHeader.js
```diff
@@ -44,11 +45,17 @@ const DefaultHeader = props => {
...
+          roles={getRoles()}
+          perform="users:visit"
+          yes={() => (
+            <NavItem className="px-3">
+              <NavLink to="/users" className="nav-link">
+                Users
+              </NavLink>
+            </NavItem>
+          )}
+        />
```
sergiogcx: Awesome code, it's exciting to think how this is going to be scaled up for editing. 🙌

atd-vze/src/views/Users/User.js
```diff
@@ -114,45 +118,76 @@ const User = () => {
...
+              <Card>
+                <CardHeader>
+                  <strong>
+                    <i className="icon-info pr-1"></i>User ID: {id}
+                  </strong>
+                </CardHeader>
+                <CardBody>
+                  <Row className="align-items-center mb-3">
+                    <Col col="6" sm="4" md="2" xl className="mb-xl-0">
+                      <Can
```
sergiogcx: 👍 nested magic!

sergiogcx:
I noticed a problem when switching the role of the user, I did not enter the password and switched from read-only to editor, this is the output error message:

sergiogcx:
Also: Switching roles does not work, even after providing a password. Maybe not in the scope of this issue, just noticed though.

sergiogcx:
Esthetic wonkiness 🤷‍♂ not a huge deal, but maybe stuff you would like to know:

PR #717, opened 2020-03-19
VZV | Map | Data Range Picker Enhancements

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/2034

This PR replaces the existing date picker with `react-dates` ([link](https://github.com/airbnb/react-dates)). I set this picker up to pop out on devices 768px and up and to display in a full screen mode on devices 768px and below and default to our rolling time window. The "x" button to the right of the dates restores this default range when clicked. I wondered if a different icon might indicate this better since an "x" could also mean "clear".

This new picker brings in an additional styling library needed to theme the picker. This is largely because in order to have the picker appear above all other elements, a built-in portal feature was used. This means that any `styled-components` wrapper around the picker would render in different nodes in the DOM (leaving the styled component with nothing to target).

#### Pop-out picker in non-mobile
Screen Shot 2020-04-07 at 11 44 41 PM

#### Month/Year dropdown in non-mobile
Screen Shot 2020-04-07 at 11 44 49 PM

#### Mobile full screen picker 
Screen Shot 2020-04-07 at 11 45 43 PM


sergiogcx:
Not sure if this is fully tested, but it looks like the map overlaps on top of the picker:

atd-vzv/src/views/nav/SideMapControlDateRange.js
```diff
@@ -1,102 +1,226 @@
...
+      <FontAwesomeIcon
+        className="reset-button"
+        icon={faTimesCircle}
+        color={colors.info}
+        onClick={() => {
+          setStart(dataStartDate);
+          setEnd(dataEndDate);
+        }}
+      />
+    </StyledButtonContainer>
```
sergiogcx, line 223: Sheesh, that's a lot of properties for the DateRangePicker. It's looking really good though :) 👍

mddilley: I agree about the properties! The number of combinations made implementing this a bit tricky and overwhelming!  😅

PR #718, opened 2020-03-21
Fix element shifting in crash detail page

rgreinho:
Fixes the column properties making the elements move bellow each
others when their size exceeds the column size.

Fixes cityofaustin/atd-data-tech#1608


atd-vze/src/views/Crashes/Crash.js
```diff
@@ -241,21 +241,21 @@ function Crash(props) {
...
+                        setIsEditingCoords={setIsEditingCoords}
+                      />
+                    </>
+                  )}
               </CardBody>
             </Card>
           </div>
         </Col>
-        <Col>
+        <Col xs="12" md="6">
```
sergiogcx: Amazing, now it works like a champ! 👍
:)

📦 🚢

mateoclarke: boom_easy

PR #719, opened 2020-03-21
Add autoformat and autofix

rgreinho:
Leverages the powers of `prettier` and `eslint` to autoformat the JS code
as well as automatically fixing the issues detected by the linter.

The `npm-run-all` too is used to simplify running scripts sequentially.

The second commit reformats the code and fixes the linting error. The PR
is split to simplify reviewing the changes.

Ideally, the CI would run the `npm run lint` command to ensure the code base
stays consistent between the developers.


atd-vze/package.json
```diff
@@ -65,6 +68,11 @@
     "build": "react-scripts build && cp _redirects build && cp -r build editor && mv editor build",
     "build:production": "env-cmd -f .env.production npm run build",
     "build:staging": "env-cmd -f .env.staging npm run build",
+    "fix": "run-s fix:*",
+    "fix:format": "prettier --write \"src/**/*.js\" ",
+    "fix:javascript": "eslint --ext js,html --fix src",
+    "lint": "run-s lint:*",
+    "lint:javascript": "eslint --ext js,html src",
```
sergiogcx: I think it's cool to have the linter as a command option. If I recall correctly, we have been linting live in the console as it runs, am I correct @mddilley ? I see no harm adding these lines though

rgreinho: Oh! What command were you using before? The same ones? Or with different parameters? And simply from the command line?

mddilley: Since most of us have been using VS Code, we have been using the Prettier extension along with this [config file](https://github.com/cityofaustin/atd-vz-data/blob/master/atd-vze/.prettierrc). Looping @mateoclarke in on this too in case he has anything to add.

mateoclarke: Yup, Mike's right. Been relying on people to configure their text editor, which hasn't been 100% successful. I like this a lot @rgreinho! thanks!

PR #723, opened 2020-03-29
Upsert Records

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/2194

Code-only review 🙏


atd-etl/app/process/helpers_import.py
```diff
@@ -122,10 +145,22 @@ def generate_gql(line, fieldnames, file_type):
...
-                                     fields=fields)
+        # Clean field list
+        template_fields = [field.lower() for field in fieldnames if field.lower() not in filters[0][1]]
+        # Generate Template
+        template = generate_template(
+            name=query_name,
+            function=function_name,
+            fields=fields,
+            fieldnames=template_fields,
+            upsert=(file_type != "crash"),
```
mddilley: How are upserts for crash records handled? Or are they right now?

atd-etl/app/process/helpers_import.py
```diff
@@ -13,35 +13,58 @@
...
-    """
+    :param str name: The name of the graphql mutation
+    :param str function: The name of the graphql function
+    :param str fields: The value of the fields in graphql expression
+    :param str[] fieldnames: An array of strings containing the names of the columns
+    :param bool upsert: If true, adds upsert logic; false otherwise.
+    :param str constraint: The name of the constraint on_conflict
+    :return str:
+    """
+    if upsert:
```
mddilley: Thanks for opening my eyes to upserts in Hasura. I read the docs a little bit just now, and this is making me understand how you are controlling which fields will update even better. 👍

atd-etl/app/process/helpers_import.py
```diff
@@ -122,10 +145,22 @@ def generate_gql(line, fieldnames, file_type):
         fields = generate_fields_with_filters(line=line,
                                               fieldnames=fieldnames,
                                               filters=filters)
-
-        template = generate_template(name=query_name,
-                                     function=function_name,
-                                     fields=fields)
+        # Clean field list
+        template_fields = [field.lower() for field in fieldnames if field.lower() not in filters[0][1]]
```
mddilley: I'm out of the loop so I'm not sure why the fields need to be lowered. Could you help me understand more? Are the headers in the .csv uppercase?

sergiogcx: Yes, there are a couple reasons for this.

1. For some reason Hasura does not like camel-case for my fields 🤷‍♂it throws an error.
2. There are configurations settings containing column names in lower case in this script, some comparisons are made against those column names, and python is case sensitive when comparing strings. For example:

`filters[0][1]` contains all the needed to be removed.
`[field.lower() for field in fieldnames if field.lower() not in filters[0][1]]` in plain english: give me a lower-case array of all strings in fieldnames as long as they are not in the removed list.

atd-etl/app/process/helpers_import.py
sergiogcx: No upserts for crashes at the moment, we are going to assume the crash change needs to be reviewed by a human.

There are plains for having automated upserts if the crash has not been edited by a human, but not within the scope of this issue.

atd-etl/app/process/helpers_import.py
```diff
@@ -13,35 +13,58 @@
...
+              update_columns: [
+                %CONFLICT_FIELDS%
+              ]
+            }
+        """.replace(
+            "%CONFLICT_CONSTRAINT%",
+            constraint
+        ).replace(
+            "%CONFLICT_FIELDS%",
+            ",\n".join([f.lower() for f in fieldnames])
```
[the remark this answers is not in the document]
sergiogcx: Thanks Mike! 👍

atd-etl/app/process/helpers_import.py
sergiogcx: Thank you for your questions, I should make the code easier to read  or at least add some commentary

atd-etl/app/process/helpers_import.py
mddilley: Awesome, thanks!

atd-etl/app/process/helpers_import.py
mddilley: Thanks for answering them - still familiarizing myself with the whole picture so this was very helpful!

sergiogcx:
Issues fixed, merging to master 👍

PR #724, opened 2020-03-30
Changes to Records

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/1542

Code-only review 🙏


sergiogcx:
The code will work fine for all fields in the **important field list** which is the list it goes by to detect changes. The problem is that it ignores the formatting of the rest of the fields, and that shows once the records makes it in the queue:

PR #727, opened 2020-03-31
Revert "Add autoformat and autofix"

sergiogcx:
Reverts cityofaustin/atd-vz-data#719


sergiogcx:
It seems to fix the problem:

PR #737, opened 2020-04-07
Crashes ETL Process

sergiogcx:
Closes: https://github.com/cityofaustin/atd-data-tech/issues/1959

Once again, I'd appreciate a code-only review for now 🙏

### DIFF
--- a/atd-etl/app/process/helpers_import.py
+++ b/atd-etl/app/process/helpers_import.py
@@ -13,6 +13,7 @@ import io
 import json
 import re
 import datetime
+from dateutil import parser
 
 # Dependencies
 from .queries import search_crash_query, search_crash_query_full
@@ -355,7 +356,16 @@ def get_crash_record(crash_id):
     # First generate a query with a list of the columns we care about
     query = search_crash_query_full(
         crash_id=crash_id,
-        field_list=CRIS_TXDOT_COMPARE_FIELDS_LIST+["apd_human_update"]
+        field_list=(
+           CRIS_TXDOT_COMPARE_FIELDS_LIST +
+           [
+               "apd_human_update",
+               "geocode_provider",
+               "qa_status",
+               "updated_by",
+               "changes_approved_date"
+           ]
+        )
     )
 
     # Then try to run the query to get the actual record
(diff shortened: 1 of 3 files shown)


atd-etl/app/process/helpers_import.py
```diff
@@ -565,22 +568,37 @@ def insert_crash_change_template(new_record_dict, differences, crash_id):
...
     :return:
     """
-    if file_type == "crash":
-        fieldnames = [column.lower() for column in fieldnames]
-        crash_id = get_crash_id(line)
+    # Gather the field names
+    fieldnames = [column.lower() for column in fieldnames]
+
+    # If compare crash record is disabled, then exit.
+    if ATD_ETL_CONFIG["ATD_CRIS_IMPORT_COMPARE_FUNCTION"] != "ENABLED":
```
sergiogcx, line 582: I love that in python two merge two arrays all you need is the + sign. ✔️ 🙌

atd-etl/app/process/helpers_import.py
```diff
@@ -352,8 +353,10 @@ def get_crash_record(crash_id):
     :return: string
     """
     # First generate a query with a list of the columns we care about
-    query = search_crash_query_full(crash_id=crash_id,
-                                    field_list=CRIS_TXDOT_COMPARE_FIELDS_LIST)
+    query = search_crash_query_full(
+        crash_id=crash_id,
+        field_list=CRIS_TXDOT_COMPARE_FIELDS_LIST+["apd_human_update"]
```
sergiogcx, line 358: I love that in python two merge two arrays all you need is the + sign. ✔️ 🙌

atd-etl/app/process/helpers_import.py
```diff
@@ -565,22 +568,37 @@ def insert_crash_change_template(new_record_dict, differences, crash_id):
...
+        # Now calculate the differences
         differences = record_compare(record_new=record_new, record_existing=record_existing)
 
-        if len(differences) > 0:
+        # If there are no differences, just ignore...
+        if len(differences) == 0:
+            return False
+
+        # There are differences...
+        if record_existing["apd_human_update"] == "Y":
```
sergiogcx, line 601: In this line, if there are changes, and the crash record is updated by a human, then it will insert into the crash changes queue for review.

atd-etl/app/process/helpers_import.py
mddilley: Python magic 🎩🐇

atd-etl/app/process_hasura_import.py
```diff
@@ -74,16 +74,31 @@ def process_line(file_type, line, fieldnames, current_line, dryrun=False):
...
+            insert_crash = record_crash_compare(
+                line=line,
+                fieldnames=fieldnames,
+                crash_id=crash_id,
+                record_existing=record_existing
+            )
 
     # The record does not exist, insert.
-    else:
+    if file_type != "crash" or insert_crash:
```
sergiogcx, line 101: All other tables are upserted by default, if it is a crash then go by the value of `insert_crash` which returns True if the record needs to be upserted or created.

atd-vzd/migrations/migration_crashes_2020_03_07--0221.sql
```diff
@@ -0,0 +1,7 @@
+-- We need to drop this index Hasura will not see
+drop index idx_atd_txdot_crashes_crash_id;
+
+-- And add a primary key constraint that will be detected by Hasura
+alter table atd_txdot_crashes
+	add constraint atd_txdot_crashes_pkey
+		primary key (crash_id);
```
sergiogcx, line 7: The index for crashes needed removal, and a new constraint needed for Hasura to detect.

atd-etl/app/process/helpers_import.py
mateoclarke, line 601: I might be misunderstanding, but this reads to me like the only field we're checking as a proxy for knowing a record has been manually updated is `apd_human_update`. In the triggers, that only gets flipped to yes based on differences between `apd_confirmed_death_count` & `death_cnt`. 

Wouldn't we also want to check for `geocode_provider = 5` (aka Manual Q/A) so we flag human edits to coordinates too? Are there other cases?

Are there other clues for manual human edits?

mddilley: Cool, thanks for the explanation!

PR #739, opened 2020-04-09
VZV | Summary | Remove sidebar

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/2276

This PR removes the `SideDrawer` component from the `Summary` view for desktop viewports. The side bar is still need for navigation in mobile. I also added white space to the left and right of the content for desktop. As devices get smaller, the whitespace collapses to allow the content to fill smaller screens. I also updated the widget icons and text to stay equidistant through all viewports. Previously, icons and text would remain aligned but shift from left to right together. There is now a more uniform look regardless of the device. 

The header also fill the entire width on desktop along with showing the logo. I shifted the nav buttons to the right in order to allow room for the logo which duplicates some code in https://github.com/cityofaustin/atd-vz-data/pull/738 but I'll merge that one first and then resolve conflicts here.

### XL viewport
Screen Shot 2020-04-10 at 10 42 54 AM

### L viewport
Screen Shot 2020-04-10 at 10 43 11 AM

### M viewport
Screen Shot 2020-04-10 at 10 43 24 AM

### Smaller
Screen Shot 2020-04-10 at 10 43 41 AM


atd-vzv/src/views/nav/SideDrawer.js
```diff
@@ -55,12 +55,21 @@ const StyledDrawerHeader = styled.div`
 `;
 
 const StyledDrawer = styled.div`
+  /* Disable side drawer in desktop viewport */
+  #summary-side-drawer {
+    @media only screen and (min-width: ${responsive.bootstrapMedium}px) {
+      display: none;
+    }
+  }
```
sergiogcx, line 63: Clever stuff! I didn't realize you could do that with the `styled.div` method

mddilley:
@mateoclarke - I think that is a great idea for widgets when there is 1 card per row. I added that to the mobile responsive audit so that I can get this merged today. https://github.com/cityofaustin/atd-data-tech/issues/1811

atd-vzv/src/views/nav/SideDrawer.js
mddilley: Thanks - removing this side bar turned out to be a little more work than I thought. Hoping that having the CSS within the components makes it easier to see what styling was necessary.

PR #740, opened 2020-04-09
Add Prettier config and run on VZV

mddilley:
This PR adds and applies the same Prettier config used in VZE.


mateoclarke:
Thanks for spending the time to do this! 💅

PR #741, opened 2020-04-10
Update Record for CR3 PDF Processing

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/2201

Code-only review again 🙏


sergiogcx:
As far as why the trigger may not be a great idea after all (taken from the ZenHub issue):

```
For now the SQL trigger idea seems to be a bit of an overkill, one in the sense that any changes done in the UI do not mean that the CR3 has been updated, and secondly in the fact that the only time we really know the CR3 is updated is whenever a CRIS record comes through and it is an existing crash record.

Given these conditions, I think it is best to signal the change using Python code in the ETL
```

sergiogcx:
> Looks great to me! 👍 So, the only times that a CR3 is fetched are 1. On creation 2. If there is an update in CRIS. Do I have that right?

Correct @mddilley , any additional ways to request a new CR3 download can be implemented later anyway. This PR should at least cover the basics.

PR #743, opened 2020-04-10
VZV | Map | Custom polygon filter

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/1722

This PR adds a polygon filter to the Map component. The new filter applies to both the map data points and also the Time of Day chart component in the sidebar and works with all existing filters. The tool is limited to one polygon that is drawn and then deleted. The phases of drawing are styled to try to make it more intuitive (green dotted line leads to the point that will close the polygon) but I'm open to any suggestions. A polygon cannot be edited, but that feature could be added later.

I ended up adding the `mapbox-gl-draw` package to import the CSS needed for the editor control buttons. This could have also been added with a link tag in `index.html`. Any preference on this @mateoclarke?

### Drawing
Screen Shot 2020-04-16 at 3 21 01 PM

### Completed
Screen Shot 2020-04-16 at 3 21 13 PM

### Polygon with other layers and filters applied
Screen Shot 2020-04-16 at 3 21 41 PM


atd-vzv/src/views/map/MapPolygonFilter.js
```diff
@@ -0,0 +1,83 @@
...
+  };
+
+  const _renderDrawTools = () => {
+    // Disable draw button click handler and grey out button if a polygon is drawn
+    const isDisabled = features.length > 0;
+    return (
+      <StyledDrawTools>
+        <div className="mapboxgl-ctrl-top-right">
+          <div className="mapboxgl-ctrl-group mapboxgl-ctrl">
+            <button
```
mateoclarke, line 48: I wonder how much this matters, but I think it may be a good practice for a11y... 

The same way we give the button a `disabled` class for CSS updates, we can give the `<button>` element a `disabled` attribute for better semantic markup:
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled

atd-vzv/src/views/map/MapPolygonFilter.js
```diff
@@ -0,0 +1,83 @@
...
+          </div>
+        </div>
+      </StyledDrawTools>
+    );
+  };
+
+  return (
+    <>
+      <Editor
+        ref={(ref) => (_editorRef.current = ref)}
```
mateoclarke, line 69: 👍

atd-vzv/src/views/map/helpers.js
```diff
@@ -50,6 +51,9 @@ export const createMapDataUrl = (
     ? null
     : `${endpoint}?$limit=100000` +
         `&$where=crash_date between '${startDate}' and '${endDate}'` +
+        // if there is a polygon selected, add as filter
+        ((!!mapPolygon && ` AND within_polygon(point, '${mapPolygon}')`) ||
```
mateoclarke, line 55: ✨

atd-vzv/src/views/map/MapPolygonFilter.js
mddilley: This is definitely the right way to go! Added. ✅

atd-vzv/src/views/map/Map.js
```diff
@@ -19,6 +20,7 @@ import { faCompass, faCircle } from "@fortawesome/free-solid-svg-icons";
 import { colors } from "../../constants/colors";
 
 import "mapbox-gl/dist/mapbox-gl.css";
+import "@mapbox/mapbox-gl-draw/dist/mapbox-gl-draw.css"; // Get out-of-the-box icons
```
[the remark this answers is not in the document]
mddilley: Cool, I like it this way too. I had a bit of a hard time understanding where the CSS was coming from in the `react-map-gl-draw` example until I poked around in `index.html`. This makes it more obvious where these styles coming from. 👍

PR #750, opened 2020-04-15
Fix time of day tooltips

tylerwhitson:
Fixes issue with persisting tooltips by removing custom tooltips and reverting back to default library tooltip behavior. Also includes minor styling changes to make spacing consistent on all components.

Screen Shot 2020-04-15 at 1 50 11 PM


tylerwhitson:
Thanks!

PR #752, opened 2020-04-15
Added Index to city_id in crashes

sergiogcx:
This has been already implemented in staging and production, just added a migration file to keep track of structure.


sergiogcx:
> 👍
> 
> Just for my curiosity, what problem did we run into that caused the need for better performance for city_id queries?

A query view I am working on was running slow, and it depends heavily on city_id, so I checked whether the city_id had been indexed at all. It hadn't. So this should help a little bit more.

PR #755, opened 2020-04-16
added alert banner

tylerwhitson:
Fixes https://github.com/cityofaustin/atd-data-tech/issues/2094

Adds simple alert banner

Screen Shot 2020-04-16 at 5 46 15 PM

### DIFF
--- a/atd-vzv/src/views/summary/Summary.js
+++ b/atd-vzv/src/views/summary/Summary.js
@@ -6,8 +6,9 @@ import PeopleByDemographics from "./PeopleByDemographics";
 import CrashesByMode from "./CrashesByMode";
 import SummaryView from "./SummaryView";
 import SummaryCard from "./SummaryCard";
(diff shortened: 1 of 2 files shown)


atd-vzv/src/views/summary/Summary.js
```diff
@@ -30,6 +31,13 @@ const Summary = () => {
   return (
     <Container fluid>
       {/* Create whitespace on sides of view until mobile */}
+      <Row className="px-xs-0 mx-xs-0 px-lg-3 mx-lg-4 mt-4 mb-0">
```
mddilley, line 34: Thanks for making this snap out at the same breakpoints as the content below! This would have been easy to overlook! 🕵️

mateoclarke: 👍

tylerwhitson:
Thanks, y'all! This was an easy win :)

PR #760, opened 2020-04-21
Hide "Reports" from non-admins in nav & routing

mateoclarke:
Also update v.1.6.0 in footer

## Admin
Screen Shot 2020-04-21 at 4 51 25 PM

## Editor (can't see)
Screen Shot 2020-04-21 at 4 51 03 PM

### DIFF
--- a/atd-vze/src/_nav.js
+++ b/atd-vze/src/_nav.js
@@ -33,18 +33,6 @@ export const navigation = roles => {
       {
         divider: true,
       },
-      {
-        name: 'Reports',
-        url: '#',
-        icon: 'icon-chart',
-        children: [
-          {
-            name: "Inconsistent KSI Counts",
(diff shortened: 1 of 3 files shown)


atd-vze/src/_nav.js
```diff
@@ -55,10 +43,22 @@ export const navigation = roles => {
...
+          url: "/reports/inconsistent_ksi_counts",
+          icon: "icon-graph",
+        },
+      ],
+    },
   ];
 
   if (isAdmin(roles) || isItSupervisor(roles)) {
-    adminNavItems.forEach(item => nav.items.splice(1, 0, item));
+    adminNavItems.forEach(item => nav.items.splice(-1, 0, item));
```
mddilley, line 61: Thanks for changing the order here! 👍

PR #762, opened 2020-04-23
VZV | Map | Update hover box fields

mddilley:
Closes part of https://github.com/cityofaustin/atd-data-tech/issues/2329

This PR updates the fields included in map hover information box (Date/Time, Fatalities, Serious Injuries, Modes Involved, Crash ID). It also creates a separate mobile experience for the info box which shows a popup stemming from the crash point. The info box is also now a separate component.

Interactive layer IDs were also added to the Map component to enable the cursor to change when hovering over points in the GeoJSON crash point layers.

#### Desktop info box
Screen Shot 2020-04-24 at 3 00 52 PM

#### Mobile info box
Screen Shot 2020-04-24 at 3 01 23 PM

### DIFF
--- a/atd-vzv/src/views/map/CrashPointCard.js
+++ b/atd-vzv/src/views/map/CrashPointCard.js
@@ -1,6 +1,6 @@
 import * as React from "react";
 import { PureComponent } from "react";
-import { Card, CardBody, CardText } from "reactstrap";
+import { Card, CardBody, CardText, Table } from "reactstrap";
 import moment from "moment";
 
 export default class CrashPointCard extends PureComponent {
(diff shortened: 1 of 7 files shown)


atd-vzv/src/views/map/Map.js
```diff
@@ -99,8 +91,13 @@ const Map = () => {
   // Create ref to map to call Mapbox GL functions on instance
   const mapRef = useRef();
 
+  const { width } = useWindowSize();
+  const { bootstrapMedium } = responsive;
+  const isMobile = width <= bootstrapMedium;
```
mateoclarke, line 96: you should test this in a mobile browser but the chrome simulator had issues with the card width.

Also, the `x` close button wasn't working but a click off the card did.

Screen Shot 2020-04-28 at 2 06 54 AM

atd-vzv/src/views/map/CrashPointCard.js
```diff
@@ -0,0 +1,26 @@
...
+          <CardText>
+            Date/Time: {moment(info.crash_date).format("MM/DD/YYYY HH:MM A")}
+          </CardText>
+          <CardText>Fatalities: {info.death_cnt}</CardText>
+          <CardText>Serious Injuries: {info.sus_serious_injry_cnt}</CardText>
+          <CardText>
+            Modes Involved: {info.units_involved.split(" &").join(", ")}
+          </CardText>
+          <CardText>Crash ID: {info.crash_id}</CardText>
+        </CardBody>
```
mateoclarke, line 22: What would you think about this data presented in a two column table?
https://reactstrap.github.io/components/tables/

atd-vzv/src/views/map/Map.js
mddilley: Oh nice catch! I'll check out the "x" button in BrowserStack. For the card width, I'm not sure if there is a way to guarantee that the popup always opens 100% visible within the viewport since the same issue is present in the example in the docs. http://visgl.github.io/react-map-gl/examples/controls

Maybe we should brainstorm on this some more? The screenshot below captures the issues with both styles of info boxes in this PR when used on mobile (static or popup). Either controls are covered up by the static info box or the popup is off screen until dragged into center.

Screen Shot 2020-04-28 at 9 52 38 AM

atd-vzv/src/views/map/CrashPointCard.js
mddilley: I think that would look great! I'll test it out and see.

atd-vzv/src/views/map/Map.js
mddilley: I fixed the close button. I was missing a handler for the close button so I added that (`onClose={() => setSelectedFeature(null)}`). This was hard to notice because it turns out the existing close popup behavior was due to a bug in the click handler for the crash point layers.

Basically, the handler sets a `selectedFeature` if a GeoJSON feature exists where a click occurs. If not, it was setting `selectedFeature` as `undefined`. The logic to render the info box depends on a truthy value for `selectedFeature` so a setting a falsey value was "closing" the info box.

atd-vzv/src/views/map/CrashPointCard.js
mddilley: Thanks for the table suggestion because this looks much cleaner. 👍This was the look that I tried to go after with rows and columns but that didn't work out so well. 😅 

Screen Shot 2020-04-28 at 11 10 30 AM

Screen Shot 2020-04-28 at 11 10 55 AM

PR #772, opened 2020-04-30
VZV | Summary | "By Month" visualization doesn't plot if month total is 0

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/2249

This PR fixes a bug where the VZV By Month chart presented its time range of data based on the latest record pulled from Socrata. This caused a scenario where a month's total is not represented until the number of records with serious injuries, fatalities, or both pulled for a month surpasses 0. The code was updated to start each month's total at zero and then increment up. I also refactored the existing code to use `reduce` and avoid nested iteration.

#### Before
Screen Shot 2020-04-30 at 1 54 35 PM

#### After
Screen Shot 2020-04-30 at 1 54 23 PM

### DIFF
--- a/atd-vzv/src/views/summary/CrashesByMonth.js
+++ b/atd-vzv/src/views/summary/CrashesByMonth.js
@@ -35,7 +35,7 @@ const CrashesByMonth = () => {
 
   useEffect(() => {
     const calculateYearMonthlyTotals = (data) => {
-      // Data query is ordered by crash_date ASC so truncate dataset by month of latest record
+      // Determine if the data is a complete year or not and which records to process
       const isCurrentYear =
         data.length > 0 &&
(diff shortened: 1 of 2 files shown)


atd-vzv/src/views/summary/CrashesByMonth.js
```diff
@@ -35,58 +35,48 @@ const CrashesByMonth = () => {
...
-        });
-        cumulativeMonthTotal += monthTotal;
-        return cumulativeMonthTotal;
-      });
+        }
+        return acc;
+      }, monthTotalArray);
+
+      // Accumulate the monthly totals over the year of data
+      const accumulatedTotals = monthTotals.reduce((acc, month, i) => {
```
mateoclarke, line 74: fancy 💅

PR #776, opened 2020-05-05
adjust sizing so by month and by mode charts equal height

tylerwhitson:
Fixes https://github.com/cityofaustin/atd-data-tech/issues/2549

Adjusts heights of By Mode and By Month components so they are both the same.

Screen Shot 2020-05-05 at 3 28 03 PM


tylerwhitson:
> amazing, I was curious how hard this would be and wasn't expecting it to be solved with an options update. good stuff 👏
> 
> homer_easy_rubiks_cube

Thanks! Yeah, I tried a bunch of other stuff using Bootstrap classNames etc. before settling on this, but this seems to do the trick!

PR #777, opened 2020-05-06
VZV | Map | Make default date range for map YTD

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/2546

This PR updates the default date range for the Map date range picker to the full year preceding the end date of global VZV data date range.

#### Updated map date range
Screen Shot 2020-05-06 at 6 00 33 PM

### DIFF
--- a/atd-vzv/src/constants/time.js
+++ b/atd-vzv/src/constants/time.js
@@ -1,8 +1,10 @@
 import moment from "moment";
 
+// Set the sliding window of data that feeds VZV
 // Number of past years data to fetch
-// 4 full years of data, plus data up to the last complete month of the current year
 export const ROLLING_YEARS_OF_DATA = 4;
+// Number of months window slides in the past (to display most accurate data)
+export const MONTHS_AGO = 2;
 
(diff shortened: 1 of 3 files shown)


atd-vzv/src/constants/time.js
```diff
@@ -42,3 +42,8 @@ export const summaryLastYearEndDate = dataEndDate
 
 export const currentYearString = summaryCurrentYearStartDate.slice(0, 4);
 export const prevYearString = summaryLastYearStartDate.slice(0, 4);
+
+// Map time data
+export const mapStartDate = moment().subtract(1, "month").startOf("year");
```
mateoclarke, line 47: I was hung up at first why you would subtract a month, but then I remembered about the January use case. Good stuff. Maybe consider adding a comment here about that intended logic.

mddilley: That is a good idea. This also reminded me to update according to the global time range change in https://github.com/cityofaustin/atd-vz-data/pull/780. I'll update before merging. 👍

PR #779, opened 2020-05-06
Hasura event to automate location association

mateoclarke:


mateoclarke:
@sergiogcx Pw hashing we looked at this morning is now in this PR along with some try/except logic. Going to merge and try testing later this evening (perhaps). I'll log the times I start trying bulk updates in staging.

PR #787, opened 2020-05-20
VZV | Update header elements styling and behavior

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/2610

This PR updates the Header component styling and behavior as described in the issue. Changes have been reviewed with Jace. I also tested with BrowserStack to make sure the updated scroll behavior works across OSs and browsers (Firefox and Chrome in MacOS and Windows 10, 8, and 7).

I also rolled `/constants/drawer.js` into `/constants/responsive.js` since those variables all started falling under the same umbrella.

#### Large desktop
Screen Shot 2020-05-26 at 1 47 25 PM

#### Smaller desktop
Screen Shot 2020-05-26 at 1 47 43 PM

#### Mobile
Screen Shot 2020-05-26 at 1 47 55 PM

#### Mobile nav with icons
Screen Shot 2020-05-26 at 2 10 31 PM


atd-vzv/src/constants/nav.js
```diff
@@ -1,3 +1,7 @@
+import React from "react";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faChartBar, faMap } from "@fortawesome/free-solid-svg-icons";
```
mateoclarke, line 3: these nav buttons look so much nicer with simple icons and new outlines. 👍

mddilley: I agree! These details from feedback snapped into place nicely.

PR #788, opened 2020-05-21
VZV: Fixes the filter buttons in FireFox Windows

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/2706


mddilley:
Awesome stuff, @sergiogcx! Thanks for dissecting these and finding the problems. I tested the deploy preview on BrowserStack and noticed the date picker issue popping up with Windows 10, 8, 7 using Firefox and Chrome (using the latest versions of each). 

The type filters work perfect in all those combinations. I didn't realize how brittle that date reset button would become so feel free to talk to Jace if you want to take it a different direction. 👍

sergiogcx:
Thanks @mddilley , I've re-worked the lines and tested with BrowserStack, let me know if this fixes it for you or if it needs more adjustments.

PR #795, opened 2020-05-27
added race/ethnicity popover

tylerwhitson:
Fixes https://github.com/cityofaustin/atd-data-tech/issues/2686


atd-vzv/src/views/summary/PeopleByDemographics.js
```diff
@@ -317,6 +319,11 @@ const PeopleByDemographics = () => {
             >
               Race/Ethnicity
             </Button>
+            {activeTab === "prsn_ethnicity_id" ? (
```
mateoclarke, line 322: 👍

PR #796, opened 2020-05-28
2577 time of day firefox styling

tylerwhitson:
Fixes https://github.com/cityofaustin/atd-data-tech/issues/2577

Thanks to @mddilley for delving deep into the Reaviz code and discovering this solution! Thanks to @sergiogcx for teaching me a thing or two about DOM manipulation!

Firefox:
Screen Shot 2020-05-29 at 11 44 07 AM

### DIFF
--- a/atd-vzv/src/views/summary/CrashesByTimeOfDay.js
+++ b/atd-vzv/src/views/summary/CrashesByTimeOfDay.js
@@ -139,36 +139,6 @@ const CrashesByTimeOfDay = () => {
       });
   }, [activeTab, crashType]);
 
-  // useLayoutEffect(() => {
-  //   !!heatmapData[0] &&
-  //     // !!document.querySelector(
-  //     //   "div.m-0.p-0.container > div.h-auto.row > div > div > svg > g > g:nth-child(1)"
-  //     // ) &&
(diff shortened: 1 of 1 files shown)


atd-vzv/src/views/summary/CrashesByTimeOfDay.js
```diff
@@ -223,6 +226,18 @@ const CrashesByTimeOfDay = () => {
...
               />
             }
+            xAxis={
+              <LinearXAxis
+                type="category"
+                axisLine={null}
+                tickSeries={
+                  <LinearXAxisTickSeries
+                    line={null}
+                    label={<LinearXAxisTickLabel padding={navigator.userAgent.includes("Firefox") ? 15 : 5} />}
```
amcdnl, line 236: :( = what was the core issue here?

tylerwhitson: Thanks for checking in! The X axis labels were being partially covered up by the chart when viewed in Firefox (see the associated issue for a screenshot), so we added some additional padding for Firefox users. I noticed there is an [open bug report](https://github.com/jask-oss/reaviz/issues/121) for this problem and this seems to fix it.

PR #805, opened 2020-06-04
VZV | Map controls bug

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/2795

This PR addresses a bug in iOS Safari and other mobile browsers where a scroll bar would appear on the Map view and allow the controls to scroll out of view. 

After some trial and error, the fix described in [this post](https://allthingssmitty.com/2020/05/11/css-fix-for-100vh-in-mobile-webkit/) works as a solution for consistently filling the space below the header with the map (without a scrollbar).

#### On my iPhone X (with bottom controls)
Image from iOS

#### On my iPhone X (without bottom controls)
Image from iOS (1)

#### Galaxy S10 in Chrome
s10

#### Galaxy S4 in Chrome
s4

### DIFF
--- a/atd-vzv/src/App.css
+++ b/atd-vzv/src/App.css
@@ -1,6 +1,7 @@
 .App {
   display: flex;
   width: 100%;
+  height: 100vh;
   align-items: stretch;
 }
 
--- a/atd-vzv/src/App.css
+++ b/atd-vzv/src/App.css
@@ -1,8 +1,10 @@
 .App {
   display: flex;
   width: 100%;
-  height: 100vh;
   align-items: stretch;
+  min-height: 100vh;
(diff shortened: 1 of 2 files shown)


atd-vzv/src/App.css
```diff
@@ -2,6 +2,9 @@
   display: flex;
   width: 100%;
   align-items: stretch;
+  min-height: 100vh;
+  /* Mobile viewport iOS Safari bug fix for Content component (Map) */
```
mateoclarke, line 6: would it make sense to add in the commented code a link to that article you sent us as a reference?

mddilley: Definitely - added this for some context about the fix. 👍

PR #810, opened 2020-06-05
VZE Linter Clean-up

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/2913

We debated different options:

1. Keep the rule & fix the code
2. Keep the rule & add exceptions
3. Get rid of the linter rule entirely

Option 1 would have been the best, but sometimes the code needs lots of refactoring and it can make us wonder if these rules really make you write better code or it adds unnecessary complexity and code spaghetti.

Option 3 could introduce severe side-effect errors, the linter is helpful reminding you to check the code you are introducing in side-effects.

Upon further discussion, we decided with option 2, which is not great if we are going to add a lot of exceptions, but at least you can have the best of both worlds, knowing that you understand what code you are introducing into side-effects and you get feedback from the linter.

### DIFF
--- a/atd-vze/src/queries/gqlAbstract.js
+++ b/atd-vze/src/queries/gqlAbstract.js
@@ -319,15 +319,6 @@ gqlAbstractTableAggregateName (
   }
 
   /**
-   * Returns the type of the field as implemented in the field configuration
-   * @param {string} columnName - The name of the column
-   * @returns {string|null}
-   */
-  getType(columnName) {
-    return this.config["columns"][columnName]["type"] || null;
-  }
-
-  /**
    * Returns an array with key-value pairs
    * @param {string} section - the 'key' name of the section in the config object
    * @returns {[string, any][]}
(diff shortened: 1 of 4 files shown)


atd-vze/src/queries/gqlAbstract.js
```diff
@@ -318,15 +318,6 @@ gqlAbstractTableAggregateName (
...
 
-  /**
-   * Returns the type of the field as implemented in the field configuration
-   * @param {string} columnName - The name of the column
-   * @returns {string|null}
-   */
-  getType(columnName) {
-    return this.config["columns"][columnName]["type"] || null;
-  }
-
```
sergiogcx, line 329: This one is totally my bad, I did not realize I had written this method already. 🤦

atd-vze/src/views/Crashes/CrashChange.js
```diff
@@ -277,45 +262,11 @@ function CrashChange(props) {
 
     if (!newRecord) return [];
 
-    const selectable = Object.keys(newRecord).filter(field => {
+    return Object.keys(newRecord).filter(field => {
```
mddilley, line 265: 😎

PR #816, opened 2020-06-10
2601 ie compatibility 3

tylerwhitson:
Closes https://github.com/cityofaustin/atd-data-tech/issues/2601.

Both pages of VZV now function on IE11 without crashing the browser. There are still a few kinks related to styling and layout, but I'm not sure how deep we need to delve. The most significant remaining issue is that the crash type buttons on the map page overlap and may cause problems toggling - this may need to be a separate issue if it is worth pursuing.

@sergiogcx no need for you to review, but I wanted you to see that patch-package library Mateo discovered - super nifty!

IE11 on Windows 7 in BrowserStack:

Screen Shot 2020-06-10 at 4 51 15 PM

Screen Shot 2020-06-10 at 4 51 42 PM


atd-vzv/patches/reaviz+8.1.0.patch
```diff
@@ -0,0 +1,26 @@
+diff --git a/node_modules/reaviz/dist/index.js b/node_modules/reaviz/dist/index.js
```
mddilley, line 1: Cool stuff! Say goodbye to `reaviz` IE 11 issues 👋

tylerwhitson:
> Really happy to see this work out as it will be important for as many people as possible to have access to this resource. Also, nice perseverance with the BrowserStack workflow! I know it can be slow. 👏
> 
> Thanks for taking it to the next step of highlighting some issues with the styling in IE 11, too. We could add some IE-only styling to size the logo better. You led me to find an issue with the footer so maybe some work on this is justified.
> 
> Screen Shot 2020-06-11 at 10 12 12 AM

Thanks @mddilley !

PR #817, opened 2020-06-10
VZD 2963: Extend Automation of CircleCI SQS deployments

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/2963

The purpose of this PR is to extend the automation of the deployment of a queue with its accompanying Lambda function and Queue-To-Function mapping.

Currently, CircleCI assumes that the mapping and queue already exist. This is not ideal in the future and introduces the possibility of errors because of all the manual steps needed to make this happen. This PR should fix that.

### DIFF
--- a/atd-cr3-api/server.py
+++ b/atd-cr3-api/server.py
@@ -441,7 +441,7 @@ def associate_location():
 
     # Check if there is an event name provided, if not provide feedback.
     if incoming_event_name == "":
-        return {"statusCode": 403, "body": json.dumps({"message": "Forbidden Request"})}
+        return {"statusCode": 403, "body": json.dumps({"message": "Forbidden Request: Missing Event Name"})}
 
     # We continue the execution
     try:
--- a/atd-cr3-api/server.py
+++ b/atd-cr3-api/server.py
@@ -425,10 +425,11 @@ def user_delete_user(id):
         abort(403)
 
 
-@APP.route("/crashes/associate_location/", methods=["PUT", "POST"])
+@APP.route("/events/", methods=["PUT", "POST"])
 def associate_location():
     # Require matching token
-    incoming_token = request.headers.get('Hasura-Event-Api')
+    incoming_token = request.headers.get("HASURA_EVENT_API")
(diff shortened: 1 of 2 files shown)


atd-cr3-api/server.py
```diff
@@ -425,42 +425,54 @@ def user_delete_user(id):
...
     if hashed_events_api.hexdigest() != hashed_incoming_token.hexdigest():
-        return {
-            "statusCode": 403,
-            "body": json.dumps({'message': 'Forbidden Request'})
-        }
+        return {"statusCode": 403, "body": json.dumps({"message": "Forbidden Request"})}
+
+    # Check if there is an event name provided, if not provide feedback.
+    if incoming_event_name == "":
+        return {"statusCode": 403, "body": json.dumps({"message": "Forbidden Request"})}
```
mateoclarke, line 444: Should this error message include that the event name is missing? Just to be more descriptive for troubleshooting?

atd-cr3-api/server.py
```diff
@@ -425,42 +425,54 @@ def user_delete_user(id):
...
 
     # We continue the execution
     try:
-        sqs = boto3.client('sqs')
-        queue_url = HASURA_EVENTS_SQS_URL
+        sqs = boto3.client("sqs")
+        queue_url = (
+            HASURA_EVENTS_SQS_URL[0:48]
```
mateoclarke, line 450: can you add some commented notes on why we're trimming 48 characters? Maybe we set that as a constant with a explanatory variable name in case we need to update it in the future.

sergiogcx: The url is a constant that is composed by the region name, aws domain, and aws account number. We can assume the url will always follow that pattern whenever we create new queues.

atd-cr3-api/server.py
sergiogcx: Good call, just modified it 👍

atd-cr3-api/server.py
sergiogcx: Done, made comments.

PR #819, opened 2020-06-12
VZV | Map | Crash point popup does not always render its full width within the viewport

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/2491

This PR disables the `dynamicPosition` prop of the `<Popup />` that renders to show info about crashes or council district in the mobile map view. Instead, some logic was added to place the popup within the current device view based on the crash point's x-position.

I also made a new component to organize the content of the sidebar called `SidebarContent` with a new prop called `type` that is passed down and eventually interpolated in the date picker id's to fix the duplicate id console warnings that have been hanging around since the data picker is rendered in both the temporary and permanent MUI `Drawers`.

#### Crash point in left half of device view
Screen Shot 2020-06-24 at 5 28 56 PM

#### Crash point near center of device view
Screen Shot 2020-06-24 at 5 29 12 PM

#### Crash point in right half of device view
Screen Shot 2020-06-24 at 5 29 22 PM


atd-vzv/src/views/map/InfoBox/MapInfoBox.js
```diff
@@ -11,6 +16,13 @@ const MapInfoBox = ({
   type, // id of feature layer
 }) => {
   const popupInfo = selectedFeature && selectedFeature.properties;
+  const popupX = popupInfo.pixelCoordinates && popupInfo.pixelCoordinates.x;
+
+  const StyledPopup = styled.div`
+    .mapboxgl-popup-content {
+      ${setPopupPosition(popupX)}
+    }
+  `;
```
sergiogcx, line 25: 👍 amazing, thanks for showing us how to use styled components via `styled.div`

atd-vzv/src/views/map/InfoBox/infoBoxStyles.js
```diff
@@ -14,6 +17,28 @@ export const StyledMobileInfo = styled.div`
...
+    case canInfoBoxFitCentered:
+      return `left: ${popupX - halfInfoBoxWidth};`;
+    case isInfoBoxTooFarLeft:
+      return `left: ${halfInfoBoxWidth - popupX}px;`;
+    case isInfoBoxTooFarRight:
+      return `right: ${popupX + halfInfoBoxWidth - window.innerWidth}px;`;
+    default:
+      return null;
+  }
+};
```
sergiogcx, line 44: It works great! One thing I am not sure of is wether it needs to control the y-axis too. But clicking around the x-axis seems to be well taken care of.

This is what it looks like when I click on a low element within the browser:

mddilley: Great point, Sergio. I talked with Jace about this, and he thinks handling the x-axis only is good for now. Forgot to mention that in the PR. Thanks for taking a look at this in BrowserStack! 📱

atd-vzv/src/views/map/InfoBox/MapInfoBox.js
mateoclarke: clever 🧠

PR #824, opened 2020-06-17
Unit field triggers

mateoclarke:
Closes https://github.com/cityofaustin/atd-data-tech/issues/3038


mateoclarke:
@mddilley, @sergiogcx and I worked on this together today so I'm going to do an admin override merge. Just tagging y'all for visibility.

PR #828, opened 2020-06-17
VZ Events 1594: Finds Location for NonCR3 Record

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/1594

### DIFF
--- a/atd-vzd/migrations/migration_crashes.2020_06-17--0414.sql
+++ b/atd-vzd/migrations/migration_crashes.2020_06-17--0414.sql
@@ -1,0 +1,14 @@
+CREATE OR REPLACE FUNCTION find_location_for_noncr3_collision (id int)
+ RETURNS SETOF atd_txdot_locations
+ LANGUAGE sql
+ STABLE
+AS $function$
+SELECT atl.location_id, atl.shape FROM atd_apd_blueform AS aab
+  INNER JOIN atd_txdot_locations AS atl
+    ON ( 1=1
+        AND (atl.shape && st_setsrid(ST_POINT(aab.longitude, aab.latitude ), 4326))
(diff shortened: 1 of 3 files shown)


atd-vzd/migrations/migration_crashes.2020_06-17--0414.sql
```diff
@@ -0,0 +1,14 @@
+CREATE OR REPLACE FUNCTION find_location_for_noncr3_collision (id int)
+ RETURNS SETOF atd_txdot_locations
+ LANGUAGE sql
+ STABLE
+AS $function$
+SELECT atl.location_id, atl.shape FROM atd_apd_blueform AS aab
+  INNER JOIN atd_txdot_locations AS atl
+    ON ( 1=1
+        AND (atl.shape && st_setsrid(ST_POINT(aab.longitude, aab.latitude ), 4326))
```
mateoclarke, line 9: Wondering if it would make more sense for performance/indexing to have a position column for blueforms with a set SRID. Just a thought. Maybe if you agree we could make a new tech debt issue for it.

sergiogcx: Let me talk to Frank about this before I merge it

sergiogcx: What's happening here is that we are creating a NEW point, this new point does not have an SRID set when created. So this is actually proper implementation, and it is completely performant.

sergiogcx: Not sure where my last comment went, but I pointed out that this line creates a NEW point that does not have an srid. This is actually proper implementation, and it is completely performant.

PR #829, opened 2020-06-18
VZ Events: PRs for SQS

sergiogcx:
This deployment script aims to implement PRs for SQS using CircleCI and Python.

Current deployments follow this nomenclature pattern:

**{project}-{name of the queue}_{environment}**

Right now we only have two environments: production and staging (master). **The main idea is that we could switch {environment} for {branch}** which should allow us to list all queues in AWS by name, and _remove only the queues whose name do not end with production or staging_. This would protect production, and allow for testing of multiple PRs in SQS.

This presents one problem, how do we clean up after a PR has been merged to master or production? If we keep creating branches, we are going to have hundreds of sqs queues and lambda functions for branches that aren't even related.

To address this, every time a PR (any PR) is merged to master (or production) we are going to wipe out any SQS queues and accompanying Lambda functions that are not staging or production.

Unfortunately this does mean that if anyone merges a PR to master (even if unrelated). So we have a few options:

1. Remove only the PR that has been merged to master. This could work, but if the deletion process does not work all the time, we can end up with some SQS functions that somehow remain there. Cleanup may be required every now and then.

2. Remove all PRs at once from SQS, but check with GitHub first, if the PR status is closed then remove the PR in question.

3. Remove all PRs regardless. This is probably the most effective, but it will force the developers to re-deploy the SQS by making a dummy commit, or running CircleCI manually by re-running their PR process.

#### CircleCI vs GitHub Actions
It would be nice to keep things in CircleCI, but if managing PRs becomes difficult it may be possible to use GitHub Actions instead for SQS deployments only.

### DIFF
--- a/.circleci/aws-helper.sh
+++ b/.circleci/aws-helper.sh
@@ -4,33 +4,44 @@
 case "${CIRCLE_BRANCH}" in
   "production")
     export WORKING_STAGE="production";
-    ;;
+  ;;
+
+  "master")
+    export WORKING_STAGE="staging";
+  ;;
 
   *)
-    export WORKING_STAGE="staging";
-    ;;
+    export WORKING_STAGE="pr";
+  ;;
 esac
 
 # Deploys API to AWS
 function deploy_aws_lambda {
-    if [[ "${WORKING_STAGE}" == "" ]]; then
-        echo "No working stage could be determined."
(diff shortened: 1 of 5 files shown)


sergiogcx:
Closing, this PR is no longer needed.

PR #830, opened 2020-06-18
VZE 3072: Global Filters

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/2910

### DIFF
--- a/atd-vzd/migrations/migration_multiple_2020_06_19--0749.sql
+++ b/atd-vzd/migrations/migration_multiple_2020_06_19--0749.sql
@@ -1,0 +1,74 @@
+--
+-- We are going to add more indexing for better search performance
+--
+
+-- Crash
+create index atd_txdot_crashes_crash_fatal_fl_index
+	on atd_txdot_crashes (crash_fatal_fl);
+
+-- Units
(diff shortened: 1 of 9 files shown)


atd-vzd/migrations/migration_multiple_2020_06_19--0749.sql
```diff
@@ -0,0 +1,74 @@
+--
+-- We are going to add more indexing for better search performance
```
mateoclarke, line 2: gotta_go_fast

atd-vze/src/views/Crashes/Crashes.js
```diff
@@ -17,7 +18,7 @@ let queryConf = {
   columns: crashGridTableColumns,
   order_by: {},
   where: {
-    city_id: "_eq: 22",
+    ...crashesFullPurpose
```
mateoclarke, line 21: nice 💅

atd-vzd/migrations/migration_multiple_2020_06_19--0749.sql
sergiogcx: 🤣

PR #851, opened 2020-07-06
VZV | Map | Date picker updates

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/3222

This PR fixes several bugs with the map date range picker as well as updating the icon displayed to the right of the date range in the sidebar - calendar icon when at default range (not interactive) and a redo icon (instead of an X) for when a custom range is selected (clickable to reset to default range).

#### New default date input icon
Screen Shot 2020-07-08 at 5 35 17 PM

## Bugs
- In the mobile full screen date picker, focus would remain on the input fields below, a cursor would show through the calendar picker when attempting to choose a date, and the mobile keyboard would appear.
mobileKeyboardBug

- Interacting with the mobile full screen date picker would cause the zoom of the overall window to shift which created vertical and horizontal scroll bars and zoomed the map view which hid the map controls.
#### Before
calZoomBug

#### After
calZoomBugFix

- Text-align center style on the date inputs applied locally in dev but not in deployed app. This was fixed by adding styles to the ids of the inputs.
#### Before
inputBug

#### After
inputFixed

- The map view would crash when fetching a date range with no crash features. This bug stemmed from the Map component where the logic expected features to always be associated with the crash data fetched from Socrata. I added handling for no features.

### DIFF
--- a/atd-vzv/src/views/map/Map.js
+++ b/atd-vzv/src/views/map/Map.js
@@ -76,30 +76,29 @@ const Map = () => {
   useEffect(() => {
     // Sort and count crash data into fatality and injury subsets
     const sortAndCountMapData = (data) => {
-      const crashCounts = {};
-      const features = data.features.reduce(
-        (acc, feature) => {
-          crashCounts["injury"] =
-            (crashCounts["injury"] || 0) +
-            parseInt(feature.properties.sus_serious_injry_cnt);
+      const crashCounts = { injury: 0, fatality: 0 };
+      const features =
+        data.features &&
+        data.features.reduce(
+          (acc, feature) => {
+            crashCounts["injury"] += parseInt(
+              feature.properties.sus_serious_injry_cnt
+            );
+            crashCounts["fatality"] += parseInt(feature.properties.death_cnt);
 
-          crashCounts["fatality"] =
-            (crashCounts["fatality"] || 0) +
-            parseInt(feature.properties.death_cnt);
-
-          if (parseInt(feature.properties.sus_serious_injry_cnt) > 0) {
-            acc.injuries.features.push(feature);
+            if (parseInt(feature.properties.sus_serious_injry_cnt) > 0) {
+              acc.injuries.features.push(feature);
+            }
+            if (parseInt(feature.properties.death_cnt) > 0) {
+              acc.fatalities.features.push(feature);
+            }
+            return acc;
+          },
+          {
+            fatalities: { ...data, features: [] },
+            injuries: { ...data, features: [] },
           }
-          if (parseInt(feature.properties.death_cnt) > 0) {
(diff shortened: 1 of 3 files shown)


atd-vzv/src/views/nav/SideMapControlDateRange.js
```diff
@@ -189,27 +208,44 @@ const SideMapControlDateRange = ({ type }) => {
...
-      position: relative;
-      top: 5px;
-      right: 2px;
-      width: 22px;
-      height: 22px;
-      cursor: pointer;
+    /* Center start and end date inputs */
+    [id^="start_date_"],
+    [id^="end_date_"] {
+      text-align: center;
```
sergiogcx, line 214: Ha! This is so clever ⚡👍

atd-vzv/src/views/nav/SideMapControlDateRange.js
```diff
@@ -189,27 +208,44 @@ const SideMapControlDateRange = ({ type }) => {
...
         endDateId={`end_date_${type}`} // PropTypes.string.isRequired,
         startDate={start} // momentPropTypes.momentObj or null,
         endDate={end} // momentPropTypes.momentObj or null,
         onDatesChange={handleDateChange} // PropTypes.func.isRequired,
         focusedInput={focused} // PropTypes.oneOf([START_DATE, END_DATE]) or null,
-        onFocusChange={(focusedInput) => setFocused(focusedInput)} // PropTypes.func.isRequired,
+        onFocusChange={(focusedInput) => {
+          setFocused(focusedInput);
+          isMobile && document.activeElement.blur(); // Do not prompt the keyboard on mobile
+        }} // PropTypes.func.isRequired,
```
sergiogcx, line 247: Good call ✅  there is nothing more annoying than having a keyboard you don't need taking half of the screen.

atd-vzv/src/views/map/Map.js
```diff
@@ -76,30 +76,29 @@ const Map = () => {
...
-
-          crashCounts["fatality"] =
-            (crashCounts["fatality"] || 0) +
-            parseInt(feature.properties.death_cnt);
-
-          if (parseInt(feature.properties.sus_serious_injry_cnt) > 0) {
-            acc.injuries.features.push(feature);
+      const crashCounts = { injury: 0, fatality: 0 };
+      const features =
+        data.features &&
```
sergiogcx, line 81: That's a lot cleaner 👍

atd-vzv/src/views/nav/SideMapControlDateRange.js
mddilley: Thanks - this was a tricky one to figure out. I'm still not 100% why these styles coming from the picker library were being overridden when deployed.

PR #864, opened 2020-07-13
VZV | Summary | Update By Month/Year component

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/3261

This PR changes the title of the By Month/Year summary visualization to By Year. It also adds an info popover with additional content about the visualization.

The mobile version of the `InfoPopover` is also updated to be a modal to take full advantage of screen real estate on mobile devices. This simplifies the logic of the popover content with images since we can now use Bootstrap's `img-fluid` classname to determine the width of popover images within the `Modal` that auto-sizes to the viewport.

#### Desktop
Screen Shot 2020-07-14 at 2 06 40 PM

#### Mobile
Screen Shot 2020-07-14 at 2 07 02 PM


atd-vzv/src/Components/Popover/InfoPopover.js
```diff
@@ -56,6 +55,22 @@ const InfoPopover = ({ config }) => {
         <FontAwesomeIcon className="info-icon" icon={faInfoCircle} />
       </StyledInfoIcon>
     </Tippy>
+  ) : (
+    <span>
+      <FontAwesomeIcon icon={faInfoCircle} onClick={toggle} />
+      <Modal
+        isOpen={modal}
+        toggle={toggle}
+        zIndex={1305} // Set z-index to supercede SideDrawer and SideMapControlDateRange components
```
sergiogcx, line 64: This value seems trivial, but as long as it works 👍

atd-vzv/src/Components/Popover/InfoPopover.js
```diff
@@ -24,23 +29,17 @@ const InfoPopover = ({ config }) => {
     cursor: pointer;
   `;
 
-  const isMobile = useIsMobile();
-
   const content = <StyledPopover>{config.html}</StyledPopover>;
 
-  return (
+  return !isMobile ? (
```
sergiogcx, line 34: 💯  Very useful!

atd-vzv/src/Components/Popover/InfoPopover.js
mddilley: Good catch, yep, this isn't my favorite thing either. It becomes important when a modal is displayed at the same time as the mobile side control bar in the map view. Without this, the modal falls behind side bar and is not visible when clicking an info icon unless the side bar is closed (the MUI side bar starts its z-index at 1300 for a reason that I assume is important to MUI).

PR #865, opened 2020-07-13
Remove data quality Alert Banner

mateoclarke:
Closes https://github.com/cityofaustin/atd-data-tech/issues/2554


mateoclarke:
Good catch @mddilley, I'll go ahead and remove that one too!

PR #868, opened 2020-07-16
VZV | Map | Date picker opens mobile keyboard on Android devices/tablets

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/3354

When investigating the Android keyboard picker, I found that the virtual keyboard opened up on many tablets as well. I changed the condition for preventing the keyboard from opening in as many of these cases as possible but also prompt keyboard entry on desktops.

#### Before
Screen Shot 2020-07-16 at 1 36 36 PM

#### After
Screen Shot 2020-07-16 at 1 36 00 PM

### DIFF
--- a/atd-vzv/src/views/nav/SideMapControlDateRange.js
+++ b/atd-vzv/src/views/nav/SideMapControlDateRange.js
@@ -249,7 +249,7 @@ const SideMapControlDateRange = ({ type }) => {
         focusedInput={focused} // PropTypes.oneOf([START_DATE, END_DATE]) or null,
         onFocusChange={(focusedInput) => {
           setFocused(focusedInput);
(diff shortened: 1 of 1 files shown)


sergiogcx:
Thanks for working on this @mddilley , it does appear that the newer versions of iPad are still affected but I don't think it is a big issue, maybe it is a high-res thing that the isTablet function is missing. 🤷‍♂️  But in android it now works like a charm!




This is the view for an iPad Pro 11:

mddilley:
@sergiogcx, you are right about the virtual keyboard popping up on high-res tablets since they start to overlap with desktop territory. 👍 Thanks for testing this out! Maybe we can come up a more concrete idea of what should happen at what breakpoints or try to isolate tablets in a different way if it is important enough. cc @JaceDeloney

PR #870, opened 2020-07-17
VZE: Global Crash Data Export

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/3312


atd-vze/src/views/Locations/Location.js
```diff
@@ -76,6 +77,14 @@ function Location(props) {
     setEditField("");
   };
 
+
+  const downloadAllData = <>
```
mateoclarke, line 81: Could this be using the Button Component? Also the language "Download Global" might not be clear for users. Maybe something like "Download All Crashes" or "Download All-Time Crashes"

Screen Shot 2020-07-21 at 12 25 06 PM

mateoclarke: One particular thing, I didn't like that the text isn't clickable only the floppy disk icon.

PR #873, opened 2020-07-22
ETL: Fix time limit for VZV ETL 3323

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/3323

### DIFF
--- a/atd-etl/app/process/helpers_socrata.py
+++ b/atd-etl/app/process/helpers_socrata.py
@@ -14,6 +14,7 @@ import requests
 import json
 from copy import deepcopy
 from process.config import ATD_ETL_CONFIG
+from datetime import date
 
 # Dict to translate canonical modes to broader categories for VZV
 mode_categories = {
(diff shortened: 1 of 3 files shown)


sergiogcx:
As a quick comment, I have tested that the queries work, but I have not run a full insertion to Socrata.

atd-etl/app/process/helpers_socrata.py
```diff
@@ -311,3 +312,16 @@ def format_person_data(data, formatter_config):
     formatted_records = concatTimeAndDate(formatted_records)
 
     return formatted_records
+
+
+def get_date_limit():
+    """
+    Returns a string containing today's date 5 years ago in iso format: yyyy-mm-dd
```
mddilley, line 319: I think that this comment may need to be updated to describe the returned value (1st day of the previous month). I like this though as I have been using the day before this (last day of the month that is 2 months ago). Way easier to put this into words. 🎉

sergiogcx: Ah, good call. Let me update it. 🙏

PR #875, opened 2020-07-24
Temp record creation tool

mateoclarke:
Closes https://github.com/cityofaustin/atd-data-tech/issues/3431

This is the most bare bones implementation of Sergio's Temp Record Creator UI. I'm going to break up the index view of all temp records into another backlog issue for follow up: 
- [ ] Show all temp records in table below form
- [ ] Enable Delete action for crash record with cascade for related records

I'm wondering if we should add these features:
- [x] Add a date input field? Date seems pretty important for how a temp record effect counts.
- [ ] Add number of units field? Right now we default to only 1. But if we added the option for more, it might get confusing and difficult to associate the proper injury/fatalities. But maybe the unit modes are important for a temp record?
- [ ] Client side form validation? Right now, even a blank form should submit without error. Do we need to require a Case ID field or other client side form validation? **Yes, validate non nulls for Case ID & Date**

## Blank Form
Screen Shot 2020-07-27 at 2 27 00 AM

## Example Error State
Screen Shot 2020-07-27 at 2 27 32 AM

## Complete Form on success
Screen Shot 2020-07-27 at 2 28 08 AM

## Crash Details for new record(s)
Screen Shot 2020-07-27 at 2 31 34 AM

### DIFF
--- a/atd-vze/src/views/Tools/CreateCrashRecord.js
+++ b/atd-vze/src/views/Tools/CreateCrashRecord.js
@@ -1,0 +1,93 @@
+import React, { useState, useEffect } from "react";
+import {
+  Button,
+  Card,
+  CardBody,
+  CardHeader,
+  CardFooter,
+  Col,
+  Input,
+  Label,
+  Form,
+  FormGroup,
+  FormText,
+} from "reactstrap";
+
+import { withApollo } from "react-apollo";
+
+import "./CreateCrashRecord.css";
+
+const CreateCrashRecord = () => {
+  return (
+    <Card>
+      <CardHeader>Create New Crash Record Set</CardHeader>
+      <CardBody>
+        <Form action="" method="post" className="form-horizontal">
+          <FormGroup row>
+            <Col md="3">
+              <Label htmlFor="hf-email">Case ID</Label>
+            </Col>
+            <Col xs="12" md="9">
+              <Input
+                type="number"
+                id="hf-case-id"
+                name="hf-case-id"
+                placeholder="Enter Case ID..."
+                // autoComplete="email"
+              />
+              <FormText className="help-block">
+                Please enter the Case ID
+              </FormText>
+            </Col>
+          </FormGroup>
+          <FormGroup row>
+            <Col md="3">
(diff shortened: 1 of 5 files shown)


atd-vze/src/views/Tools/CreateCrashRecord.js
```diff
@@ -0,0 +1,284 @@
...
+        // default to 1000.
+        const highestCurrentTempId = res.data.atd_txdot_crashes[0]
+          ? res.data.atd_txdot_crashes[0].crash_id
+          : 1000;
+        setTempId(highestCurrentTempId + 1);
+      })
+      .catch(error => {
+        setFeedback({ title: "Error", message: String(error) });
+      });
+  });
```
sergiogcx, line 63: This is very clever! 💯

atd-vze/src/views/Tools/CreateCrashRecord.js
```diff
@@ -0,0 +1,284 @@
...
+        prsn_injry_sev_id: 1,
+      }`);
+    }
+    for (let index = 0; index < fatalityCount; index++) {
+      personObjects = personObjects.concat(`{
+        crash_id: ${tempId},
+        unit_nbr: 1,
+        prsn_injry_sev_id: 4,
+      }`);
+    }
```
sergiogcx, line 91: I had never thought it was so easy to bulk insert like this! 👍

atd-vze/src/views/Tools/CreateCrashRecord.js
```diff
@@ -0,0 +1,284 @@
...
+        // Find the highest existing temp crash record ID value.
+        // If there aren't and temp crash records,
+        // default to 1000.
+        const highestCurrentTempId = res.data.atd_txdot_crashes[0]
+          ? res.data.atd_txdot_crashes[0].crash_id
+          : 1000;
+        setTempId(highestCurrentTempId + 1);
+      })
+      .catch(error => {
+        setFeedback({ title: "Error", message: String(error) });
```
mddilley, line 61: Nice error handling for this form!

PR #885, opened 2020-07-30
VZE: Temp Record Search Production Patch

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/3531


atd-vze/src/views/Tools/CreateCrashRecord.js
```diff
@@ -111,6 +111,7 @@ const CreateCrashRecord = ({ client }) => {
               sus_serious_injry_cnt: $sus_serious_injry_cnt
               temp_record: true
               crash_date: $crash_date
+              city_id: 22
```
mateoclarke, line 114: Going to make sure this is an editable field with a variable that gets passed into the query when I make updates to the form.

PR #890, opened 2020-08-04
VZV | Map | Fix Chrome type filters buttons bug

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/3560

This PR updates the font size of the map type filter buttons and adds an `auto` class to the "All" button to get them to spill to another row if they don't fit. I also updated the Data link url (at the bottom of the Summary page) to reflect the updated URL for the Socrata crash dataset.

### Before
Screen Shot 2020-08-04 at 10 56 35 AM

### After 
Screen Shot 2020-08-04 at 10 52 10 AM

### Updated link
Screen Shot 2020-08-04 at 10 54 02 AM


atd-vzv/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "atd-vzv",
-  "version": "1.12.2",
+  "version": "1.12.3",
```
sergiogcx, line 3: Shouldn't we also be updating VZE?

mateoclarke: yeah

mddilley: Just updated this - 🙏 Thanks, y'all!

PR #895, opened 2020-08-06
Create Temp Record Form updates

mateoclarke:
Related to https://github.com/cityofaustin/atd-data-tech/issues/3524

This PR adds features to the form like multiple Unit records and client side validation. The index/table of existing temp records with delete feature will be tackled in a separate PR.

### New form initial state
Screen Shot 2020-08-11 at 9 40 53 AM

### New form with error messaging and client side validation
Screen Shot 2020-08-11 at 9 41 27 AM

### DIFF
--- a/atd-vze/src/views/Tools/CreateCrashRecord.js
+++ b/atd-vze/src/views/Tools/CreateCrashRecord.js
@@ -23,22 +23,34 @@ import "./CreateCrashRecord.css";
 import UnitsForm from "./UnitsForm";
 
 const CreateCrashRecord = ({ client }) => {
-  const [tempId, setTempId] = useState(1000);
-  const [caseId, setCaseId] = useState("");
-  const [successfulNewRecordId, setSuccessfulNewRecordId] = useState(null);
-  const [crashTime, setCrashTime] = useState(
-    moment(new Date()).format("HH:mm:ss")
-  );
-  const [crashDate, setCrashDate] = useState(
-    moment(new Date()).format("YYYY-MM-DD")
-  );
-  const [feedback, setFeedback] = useState(false);
-  const [primaryAddress, setPrimaryAddress] = useState("");
-  const [secondayAddress, setSecondaryAddress] = useState("");
-
   const unitsInitialState = [
     { unit_desc_id: 1, atd_fatality_count: 0, sus_serious_injry_cnt: 0 },
   ];
+
+  const formInitialState = {
+    tempId: 1000,
+    caseId: "",
+    crashTime: moment(new Date()).format("HH:mm:ss"),
+    crashDate: moment(new Date()).format("YYYY-MM-DD"),
+    primaryAddress: "",
+    secondayAddress: "",
+    units: [
+      { unit_desc_id: 1, atd_fatality_count: 0, sus_serious_injry_cnt: 0 },
+    ],
+  };
+
+  const [tempId, setTempId] = useState(formInitialState.tempId);
+  const [caseId, setCaseId] = useState(formInitialState.caseId);
+  const [successfulNewRecordId, setSuccessfulNewRecordId] = useState(null);
+  const [crashTime, setCrashTime] = useState(formInitialState.crashTime);
+  const [crashDate, setCrashDate] = useState(formInitialState.crashDate);
+  const [feedback, setFeedback] = useState(false);
+  const [primaryAddress, setPrimaryAddress] = useState(
+    formInitialState.primaryAddress
+  );
+  const [secondayAddress, setSecondaryAddress] = useState(
+    formInitialState.secondayAddress
+  );
 
   function unitsReducer(state, action) {
     const { type, payload, unitIndex } = action;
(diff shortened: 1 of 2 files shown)


atd-vze/src/views/Tools/CreateCrashRecord.js
```diff
@@ -63,55 +111,112 @@ const CreateCrashRecord = ({ client }) => {
...
+    setPrimaryAddress(formInitialState.primaryAddress);
+    setSecondaryAddress(formInitialState.secondayAddress);
+    unitFormDispatch({ type: "reset" });
   };
 
   const handleFormSubmit = e => {
     e.preventDefault();
+    setFeedback(false);
+
+    client
```
mddilley, line 126: This is a cool cascade of validation here!

atd-vze/src/views/Tools/UnitsForm.js
```diff
@@ -0,0 +1,132 @@
...
+                <Col xs="12" md="9">
+                  <Input
+                    type="select"
+                    id="unit-type"
+                    name="unit-type"
+                    placeholder="Enter Fatality Count..."
+                    value={unit.unit_desc_id}
+                    onChange={e =>
+                      handleUnitFormChange({
+                        type: "unit_desc_id",
```
mddilley, line 49: Took me a second to realize this was a default case in the `unitsReducer`. Works great but it might be clearer to add another action type called `updateUnitField` or something. I guess the payload would have to be an object like `{"unit_desc_id": e.target.value }`. Hmmm.

It could also just be me digging too deeply into the cobwebs of Redux patterns in my head.  🙃

atd-vze/src/views/Tools/CreateCrashRecord.js
```diff
@@ -14,22 +14,70 @@ import {
...
+  const [unitFormState, unitFormDispatch] = useReducer(
+    unitsReducer,
+    unitsInitialState
+  );
+
+  function isFieldInvalid(data) {
+    return !data;
+  }
 
   useEffect(() => {
```
mddilley, line 82: I didn't catch this the first time around, but, since this `useEffect` doesn't have a dependency array, it is querying for the highest temp record each time a form state is updated. If that isn't intentional, it could be limited to one request by passing `client` or an empty array (which causes a linter warning).

atd-vze/src/views/Tools/CreateCrashRecord.js
```diff
@@ -14,22 +14,70 @@ import {
...
+
+    const updatedState = [...state];
+    updatedState[unitIndex][type] = payload;
+    return updatedState;
+  }
+
+  const [unitFormState, unitFormDispatch] = useReducer(
+    unitsReducer,
+    unitsInitialState
+  );
```
sergiogcx, line 76: This is very cool! Thanks for showing me how to apply `useReducer` 👍

atd-vze/src/views/Tools/CreateCrashRecord.js
```diff
@@ -223,62 +369,88 @@ const CreateCrashRecord = ({ client }) => {
...
+                Please enter the Secondary Address
               </FormText>
             </Col>
           </FormGroup>
+
+          <UnitsForm
+            units={unitFormState}
+            handleUnitFormChange={unitFormDispatch}
+            client={client}
+          />
```
sergiogcx, line 425: Got it, we pass the dispatch function which has access to the state.

atd-vze/src/views/Tools/UnitsForm.js
```diff
@@ -0,0 +1,132 @@
...
+                        payload: e.target.value,
+                        unitIndex: i,
+                      })
```
sergiogcx, line 52: Basically `handleUnitFormChange` is an alias for `unitFormDispatch`, here we pass the type of event, the data (payload), and the array position (index) at which the state needs to be updated. 👍

atd-vze/src/views/Tools/CreateCrashRecord.js
```diff
@@ -63,55 +111,112 @@ const CreateCrashRecord = ({ client }) => {
...
+          {
+            atd_txdot_crashes(where: {case_id: {_eq: "${caseId}"}}) {
+              case_id
+            }
+          }
+        `,
+      })
+      .then(res => {
+        if (res.data.atd_txdot_crashes.length > 0) {
+          setFeedback({ title: "Error", message: "Case ID must be unique." });
```
sergiogcx, line 138: Cool, this works!

atd-vze/src/views/Tools/CreateCrashRecord.js
mateoclarke: nice, thanks for this tip. I didn't realize that!

atd-vze/src/views/Tools/UnitsForm.js
mateoclarke: yup, its pretty similar in theory to Redux actions and dispatch.

It's a convoluted metaphor but I like the cartoons here:
https://code-cartoons.com/a-cartoon-intro-to-redux-3afb775501a6

image


Another random image:
image

atd-vze/src/views/Tools/CreateCrashRecord.js
```diff
@@ -60,58 +108,115 @@ const CreateCrashRecord = ({ client }) => {
       .catch(error => {
         setFeedback({ title: "Error", message: String(error) });
       });
-  });
+  }, [client]);
```
mateoclarke, line 111: 👀  @mddilley, thanks for the tip!

PR #897, opened 2020-08-06
VZV | Small UI improvements

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/3574

This fixes the Summary widgets so that they remain equal height regardless of viewport width. The solution revolves around adding bootstrap height classes to the body and footer (75% height for the body and 25% for the footer).  Then, I added `flex-grow` to the styled.div that contains all the CSS styles for the widget elements. This makes sure that the card body and card footer can always take up their portion of the full card height correctly.

This also patches an issue with the chart legends in the By Demographics visualization. I added a redraw prop to force chart re-render on each component re-render.

@sergiogcx - I plan on reviewing these changes with @JaceDeloney early next week, and I'm including your text-size change idea from the other day. 🙏 **Update 8/10: Font size changed so say 👋 to words wrapping!**

### Improved widget responsiveness
Screen Shot 2020-08-07 at 3 24 26 PM
Screen Shot 2020-08-07 at 3 24 13 PM
Screen Shot 2020-08-07 at 3 23 58 PM

### Demographics legend bug fix
demoFix

### DIFF
--- a/atd-vzv/src/views/summary/PeopleByDemographics.js
+++ b/atd-vzv/src/views/summary/PeopleByDemographics.js
@@ -275,6 +275,7 @@ const PeopleByDemographics = () => {
       <Row>
         <Col>
           <HorizontalBar
+            redraw
             data={data}
             height={null}
             width={null}
--- a/atd-vzv/src/Components/Widgets/SummaryWidget.js
+++ b/atd-vzv/src/Components/Widgets/SummaryWidget.js
@@ -48,6 +48,10 @@ const SummaryWidget = ({
       bottom: 5px;
     }
 
+    .widget-header-text > h3 {
+      font-size: 1.2em;
+    }
+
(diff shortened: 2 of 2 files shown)


atd-vzv/src/views/summary/PeopleByDemographics.js
```diff
@@ -275,6 +275,7 @@ const PeopleByDemographics = () => {
       <Row>
         <Col>
           <HorizontalBar
+            redraw
```
mateoclarke, line 278: 🤔

mddilley: Since we talked about this in dev sync, leaving this here for future reference. https://github.com/jerairrest/react-chartjs-2/blob/3855af1f3bc5ec57199bc5af6aa7d61577e82206/src/index.js#L81

PR #898, opened 2020-08-06
VZV | Update map page to include previous 4 years plus YTD

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/3569

This PR updates the default map date range to the same date range that feeds the summary view. It also adds a select parameter to the Socrata query for the map to reduce size of the map data payload.

### DIFF
--- a/atd-vzv/src/views/map/helpers.js
+++ b/atd-vzv/src/views/map/helpers.js
@@ -1,3 +1,5 @@
+import { mapRequestFields } from "../summary/queries/socrataQueries";
+
 const convertDateToSocrataFormat = (date, suffix) =>
   date.format("YYYY-MM-DD") + suffix;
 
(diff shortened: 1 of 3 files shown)


atd-vzv/src/views/map/helpers.js
```diff
@@ -1,3 +1,5 @@
+import { mapRequestFields } from "../summary/queries/socrataQueries";
```
mateoclarke, line 1: I'm curious if you benchmarked to see how much this helped the payload. Did it have much of an effect?

mddilley: Good call. Yes, even though the additional years of data increased the payload overall, filtering down the columns returned reduced the payload quite a bit.

#### Before (current year data only)
Screen Shot 2020-08-10 at 12 28 05 PM

#### After (without selecting columns)
Screen Shot 2020-08-10 at 12 26 32 PM

#### After (selecting only necessary columns)
Screen Shot 2020-08-10 at 12 25 25 PM

PR #907, opened 2020-08-20
VZV | Update age brackets

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/2828

This PR updates the Demographics age visualization with new age ranges, and I refactored the code to use SoQL queries. I still need to QA each chart against production to make sure that all totals are accurate. The title of this visualization was also renamed to "By Demographics".

#### New ranges
Screen Shot 2020-08-21 at 4 54 27 PM


atd-vzv/src/views/summary/PeopleByDemographics.js
```diff
@@ -1,258 +1,284 @@
...
+      axios.all(requests).then((resArr) => {
+        const responses = resArr.reduce(
+          (acc, res) => ({ ...acc, ...{ [res.config.dataKey]: res.data } }),
+          {}
         );
-        setChartData(newData);
-      };
-      getChartData();
+
+        dispatch({
```
mateoclarke, line 136: i'm not sure what is the proper React convention, but just thinking it might make sense to call this `demoDataDispatch` or something to differentiate if you used `useReducer` for something else in  this component. Just a thought more than a suggestion since `useReducer` is still new for me.

atd-vzv/src/views/summary/PeopleByDemographics.js
```diff
@@ -1,258 +1,284 @@
...
+    categoryKey: "prsn_age",
+    categoryType: "range",
+    socrataCategories: [
+      [0, 15],
+      [16, 25],
+      [26, 35],
+      [36, 45],
+      [46, 55],
+      [56, 64],
+      [65, 200],
```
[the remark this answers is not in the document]
mddilley: Good question! I initially thought that I would be iterating through these ranges so I picked a reasonable max lifetime, but I never revisited after writing code to iterate through records. I'll update with something clearer.

atd-vzv/src/views/summary/PeopleByDemographics.js
mddilley: Nice thinking - I keep fighting an impulse to think of the state in `useReducer` as a catch all for state in a component for some reason. Good to have more thoughts pushing me away from that and thinking about keeping the code easy to read when state gets even more complex. Things could get buggy real fast without setting things apart in this way. 🐛

mddilley:
Checked the numbers against production, and everything matches up after one small bug fix.

PR #918, opened 2020-08-27
VZE: Temp Record Index & Delete

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/3652


sergiogcx:
Ah thanks @mateoclarke I definitely did not see it, let me add & rename the columns.

mddilley (commented):
This is such a great improvement for this process! One cool addition for the future would be seeing this table reflect a new record immediately after creating it in the form above. There really isn't a need because of the success message with the link to the new crash record, though.

PR #919, opened 2020-09-01
VZE: Bug in the chages search feature. #3845

sergiogcx:


sergiogcx:
Actually, this is a bad branch. Closing PR and doing it right.

PR #922, opened 2020-09-02
VZV | Map | Mode filters reset after changing type filter on touch devices

mddilley:
Closes https://github.com/cityofaustin/atd-data-tech/issues/3812

This PR updates the `SideMapControl` component and Context store with `useReducer` to handle the state of the mode filters. The bug where filters would reset stemmed from inconsistencies between local state in `SideMapControl` called `buttonFilters` and state in the Context store called `mapFilters`. `useEffect` was used to push changes from `buttonFilters` to `mapFilters`, but I noticed that, in certain conditions, they became out of sync. 

To fix this, I merged these two states and used a reducer to update the query syntax of the filters. `mapFilters` state managed by `useReducer` in the Context store is now the single source of truth for mode filters. This code was tested on my iPhone and multiple Android and iOS devices in BrowserStack.

I also changed the `span` wrappers around the mode filter buttons to `reactstrap` button links to make them keyboard accessible just like all of the other map controls.

fillterResetBugFix

### DIFF
--- a/atd-vzv/src/views/nav/SideMapControl.js
+++ b/atd-vzv/src/views/nav/SideMapControl.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useState, useReducer, useMemo } from "react";
+import React, { useEffect } from "react";
 import { StoreContext } from "../../utils/store";
 
 import SideMapControlDateRange from "./SideMapControlDateRange";
@@ -95,7 +95,6 @@ const createModeFilterSyntax = (isMapTypeSet, config) => {
 };
 
 export const mapFilterReducer = (mapFilters, action) => {
-  // TODO: Action - set mode filter
   const { type, payload } = action;
 
   switch (type) {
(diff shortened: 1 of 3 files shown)


atd-vzv/src/views/nav/SideMapControl.js
```diff
@@ -84,15 +85,46 @@ const StyledCard = styled.div`
...
+        syntax: createModeFilterString(isMapTypeSet, filter),
+      }));
+      return updatedModeFilters;
+    case "updateModeFilters":
+      const updatedFiltersArray = payload;
+      return updatedFiltersArray;
+    default:
+      return null;
+  }
+};
```
sergiogcx, line 120: Lots of great learning here, definitely code worth contemplating! 🙌

sergiogcx (approved):
Great work, learned a ton!

PR #923, opened 2020-09-02
VZ Events: City_id

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/3689

This is currently hooked up with VZD-Staging, to test on your end:
1. Find a crash that  is full purpose and has coordinates, copy the coordinates. And keep them in a safe notepad for later.
2. Find a crash that is not city_id 22 and is not full purpose, NOTE THE OLD CITY ID paste the coordinates and save.
3. Wait 10-20 seconds, see the magic happen (the city id will magically be updated to 22)

### DIFF
--- a/atd-events/crash_update_jurisdiction/app.py
+++ b/atd-events/crash_update_jurisdiction/app.py
@@ -238,6 +238,45 @@ def get_full_purpose_flag(crash_id: int) -> str:
         return "N"
 
 
+def get_city_id_from_db(crash_id: int) -> Optional[int]:
+    """
+    Attempts to find the jurisdiction flag of a crash by it's ID
+    :param int crash_id: The crash_id to be evaluated
+    :return str: Y or N
+    """
+    if not str(crash_id).isdigit():
+        print("It's bad")
+        return None
+
+    get_city_id_query = """
+        query getCityId($crash_id:Int!){
+            atd_txdot_crashes(where: {
+                crash_id: {_eq:$crash_id}
+            }){
+                city_id
+            }
+        }
+    """
+
+    try:
+        response = requests.post(
+            HASURA_ENDPOINT,
+            data=json.dumps(
+                {
+                    "query": get_city_id_query,
+                    "variables": {
+                        "crash_id": crash_id
+                    }
+                }
+            ),
+            headers=HEADERS
+        )
+        return response.json()["data"]["atd_txdot_crashes"][0]["city_id"]
+    except Exception as e:
+        print(f"LIttle error: {str(e)}")
+        return None
+
+
 def update_city_id(crash_id: int, city_id: int) -> dict:
     """
     Returns a dictionary and HTTP response from the GraphQL query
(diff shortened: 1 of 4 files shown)


atd-events/crash_update_jurisdiction/app.py
```diff
@@ -67,6 +69,46 @@ def get_jurisdiction_flag(data: dict) -> str:
         return "N"
 
 
+def get_city_id(data: dict) -> Optional[int]:
```
mddilley, line 72: Cool annotations here! I haven't seen this syntax before so it was fun looking into it. Is there any type checking here or is it for documentation? I read conflicting things so I thought that I would just ask. 👍

atd-events/crash_update_jurisdiction/app.py
```diff
@@ -67,6 +69,46 @@ def get_jurisdiction_flag(data: dict) -> str:
...
+    """
+    Compares two arrays of integers, and returns a set of unique common members
+    :param List[int] a: An array you want to compare against
+    :param List[int] b: The other array you want to compare against
+    :return Set[int]:
+    """
+    a_set = set(a)
+    b_set = set(b)
+
+    if a_set & b_set:
```
mddilley, line 106: I didn't realize that you could compare sets like this! Very handy way to find the common elements. 😎

atd-events/crash_update_jurisdiction/app.py
```diff
@@ -83,6 +125,45 @@ def load_data(record: str) -> dict:
...
+                    "query": find_crash_jurisdictions,
+                    "variables": {
+                        "crash_id": crash_id
+                    }
+                }
+            ),
+            headers=HEADERS
+        )
+        valid_jurisdictions = [5, 3, 7, 8, 10]
+        output_jurisdictions = list(map(lambda x: x['id'], response.json()["data"]["find_crash_jurisdictions"]))
```
mddilley, line 159: Cool lambda here! Gives me JS vibes, and I like it. 😄

mddilley (approved):
I tested as directed and saw the city ID update just as expected. 🙌  Thanks for listing out the steps to test it. Lots of learning here 🐍, and your tests helped me see how everything fits together - good stuff!

atd-events/crash_update_jurisdiction/app.py
sergiogcx: There is no type checking, only syntax. The only reason of these annotations is to help the developers understand what they are working with.

atd-events/crash_update_jurisdiction/app.py
sergiogcx: Yea, I had not been using sets before and I think I need to make more use of them going forward. They are a swiss-army knive for data processing

atd-events/crash_update_jurisdiction/app.py
mddilley: Awesome, thanks! They are very helpful. 👍

PR #925, opened 2020-09-04
VZE: Fix CR3 for temp records

sergiogcx:
For temp records, it shows an alert instead of a download button, even if there is a PDF available:

To test:
1. Find any temp record in the staging database. You should see the above Alert.
2. Change the cr3_stored_flag to 'N', which indicates there is a PDF in S3.
3. Change the temp_record flag to false, which indicates the record is considered temporary. You show see the download button.
4. Revert temp_record to true, the button will disappear, even if he cr3_stored_flag is set to 'Y'.

### DIFF
--- a/atd-vze/src/views/Crashes/CR3Record.js
+++ b/atd-vze/src/views/Crashes/CR3Record.js
@@ -26,7 +26,7 @@ function CR3Record(props) {
         {props.isTempRecord ? (
           <Alert color="warning">
             <strong>CR3 PDFs are not available for temporary records.</strong><br />
-            Using the crash id, check the{" "}
+            Using the case id, check the{" "}
             <a href={"https://cris.dot.state.tx.us/"} target={"_new"}>
               CRIS website
             </a>{" "}
--- a/atd-vze/src/views/Crashes/CR3Record.js
+++ b/atd-vze/src/views/Crashes/CR3Record.js
@@ -23,7 +23,16 @@ function CR3Record(props) {
     <Card>
       <CardHeader>Crash Report</CardHeader>
       <CardBody>
-        {props.isCr3Stored ? (
+        {props.isTempRecord ? (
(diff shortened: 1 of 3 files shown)


atd-vze/src/views/Crashes/CR3Record.js
```diff
@@ -23,7 +23,16 @@ function CR3Record(props) {
...
+        {props.isTempRecord ? (
+          <Alert color="warning">
+            <strong>CR3 PDFs are not available for temporary records.</strong><br />
+            Using the crash id, check the{" "}
+            <a href={"https://cris.dot.state.tx.us/"} target={"_new"}>
+              CRIS website
+            </a>{" "}
+            for the latest status of this crash.
+          </Alert>
+        ) : props.isCr3Stored ? (
```
mddilley, line 35: Nice - I like the chained ternaries. 😎

atd-vze/src/views/Crashes/CR3Record.js
```diff
@@ -23,7 +23,16 @@ function CR3Record(props) {
     <Card>
       <CardHeader>Crash Report</CardHeader>
       <CardBody>
-        {props.isCr3Stored ? (
+        {props.isTempRecord ? (
+          <Alert color="warning">
+            <strong>CR3 PDFs are not available for temporary records.</strong><br />
+            Using the crash id, check the{" "}
```
mddilley, line 29: Should this be case ID instead of crash ID?

mddilley (approved):
Tested locally, and this looks great! The temp records are really getting rounded out with details like this in VZE. 🙌

atd-vze/src/views/Crashes/CR3Record.js
sergiogcx: Good call, it should be the case id!

PR #927, opened 2020-09-17
VZE Table sorting

mateoclarke:
Closes https://github.com/cityofaustin/atd-data-tech/issues/3933
Closes https://github.com/cityofaustin/atd-data-tech/issues/3932

Screen Shot 2020-09-17 at 2 04 44 AM

Screen Shot 2020-09-17 at 2 05 33 AM


mddilley (commented):
I tested this locally, and everything works great but I ran into one issue on the Locations table in which the table sort parameters are duplicated in the GraphQL query if you update them and refresh the page in browser. Maybe a conflict with the local storage value?

I replicated by:
1. Go to Locations table
2. Click to sort by Total Est Comp. Cost and update a couple of times
3. Refresh in browser and get this query error

Screen Shot 2020-09-18 at 2 17 02 PM

atd-vze/src/queries/gqlAbstract.js
```diff
@@ -207,6 +207,10 @@ gqlAbstractTableAggregateName (
    */
   setOrder(key, syntax) {
     if (this.config && this.config["order_by"]) {
+      // First, RESET the order_by value.
+      // - Our assumption is that there should only by 1 order_by at a time.
+      this.config["order_by"] = {};
+      // Now, set new key, syntax pair for order_by
```
mateoclarke, line 213: This turned out to be the solution @mddilley. When updating the sort order, it was possible to create duplicate nested queries using slightly different syntax. 

ex:
```js
crashes_count_cost_summary: "{ est_comp_cost: desc}"
crashes_count_cost_summary: { est_comp_cost }: "desc"
```

Screen Shot 2020-09-20 at 9 16 57 PM

mddilley: Awesome, thanks for the screenshot and explanation. Tested on my end, and the bug is gone. 👍

mddilley (approved):
Good to 🚢 🙌

PR #930, opened 2020-09-24
Allow widgets and charts to re-render with query updates

mateoclarke:
Fixes https://github.com/cityofaustin/atd-data-tech/issues/3607

Sep-24-2020 17-10-46

### DIFF
--- a/atd-vze/src/Components/GridTable.js
+++ b/atd-vze/src/Components/GridTable.js
@@ -117,9 +117,10 @@ const GridTable = ({
   const [loadChartData, { data: chartData }] = useLazyQuery(chartQuery);
 
   // Update aggregate and chart queries with latest filters
-  // No dependencies added because we need to depend on changes in the query
-  // instance to propogate changes to children. This does not work properly so
-  // changes in the aggregateQuery and chartQuery are compared in if blocks.
+  //  Use `[query.gql]` for dependency list because we need to depend on changes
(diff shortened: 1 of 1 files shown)


mddilley (approved):
Thanks for sharing about this one yesterday. It was great to reflect on the initial code that we wrote with hooks, and I'm glad to see that you improved the original code and comments. If we pull this out as a Kickstand component, I think that it will be a good chance to reassess some of the logic even more.

I tested it locally, and everything looks great. 🙌

atd-vze/src/Components/GridTable.js
```diff
@@ -142,7 +143,7 @@ const GridTable = ({
         setChartQuery(updatedChartsQuery);
       }
     }
-  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+  }, [query.gql]); // eslint-disable-line react-hooks/exhaustive-deps
```
sergiogcx, line 146: We should probably remove the linter rule exception

sergiogcx (approved):
Fine with me, but I think that eslint exception needs to be removed if no longer needed

atd-vze/src/Components/GridTable.js
mateoclarke: it is needed. It expects a longer list of dependencies thats unnecessary imho.

PR #932, opened 2020-09-30
VZ ETL: Socrata Export Optional Time Constraint

sergiogcx:
Allows the script to run either with time constraint or without time constraint.

Closes https://github.com/cityofaustin/atd-data-tech/issues/4038


mateoclarke (approved):
🚢

PR #936, opened 2020-10-13
Edit unit details bug fix

mateoclarke:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4132

Oct-12-2020 21-20-34

### DIFF
--- a/atd-vze/package.json
+++ b/atd-vze/package.json
@@ -1,6 +1,6 @@
 {
   "name": "atd-vz-data",
-  "version": "1.15.1",
+  "version": "1.15.2",
   "homepage": "./",
   "description": "ATD Vision Zero Editor",
   "author": "ATD Data & Technology Services",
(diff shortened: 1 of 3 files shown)


mateoclarke:
After talking with Xavier in our dev sync and doing some more thinking, I realized that the reason this bug started happening is because Sergio was helping me deploy some database changes and he disabled the triggers on the units table and I forgot to reenable them after we were done.

I've reenabled the triggers and I've run this scripts on the prod DB to sync the data back the way it should:

There were 949 units that didn't have travel_direction ids before running this:
```sql
UPDATE
	atd_txdot_units
SET
	travel_direction = veh_trvl_dir_id
WHERE
	travel_direction IS NULL
```

There were 948 units that didn't have a movement_id before running this:

```sql
UPDATE
	atd_txdot_units
SET
	movement_id = 0
WHERE
	movement_id IS NULL
```

I checked to make sure all the units had `atd_mode_category` values which gets updated in the `atd_txdot_units_create_update` trigger and those were all NOT NULL.

mddilley (approved):
Tested this out on a temporary record in staging, and it works just as expected. Nice detail with the conditional text muting! ✨

PR #937, opened 2020-10-14
replaced CRIS Death Count column with ATD Death Count column

tylerwhitson:
Fixes https://github.com/cityofaustin/atd-data-tech/issues/4089

Screen Shot 2020-10-14 at 10 39 47 AM
Screen Shot 2020-10-14 at 10 50 29 AM


mddilley (commented):
👍

PR #939, opened 2020-10-21
4163 update export csv

tylerwhitson:
Fixes https://github.com/cityofaustin/atd-data-tech/issues/4163

- Removes unnecessary columns
- Reorders remaining columns as requested
- Renames `death_cnt` to `cris_death_cnt`

The only catch is there is a persisting bug that creates empty columns associated with nested objects and appends them at the end of the table. @JaceDeloney I plan to address this in a future issue/PR, but since this is marked as high-priority, we can merge this fix in the meantime.

Zip file with example CSV attached!

[atd_txdot_crashes2020-10-21T12_42_42-05_00.csv.zip](https://github.com/cityofaustin/atd-vz-data/files/5417817/atd_txdot_crashes2020-10-21T12_42_42-05_00.csv.zip)


atd-vze/src/Components/GridExportData.js
```diff
@@ -68,15 +68,18 @@ const GridExportData = ({ query, columnsToExport, totalRecords }) => {
...
+          // Ignore __typename (contains table name which is already in filename)
           return;
         } else if (Array.isArray(columnValue)) {
           // If value is array, recursive call and handle objects in array
           flattenRow(columnValue, flattenedRow);
         } else if (typeof columnValue === "object" && columnValue !== null) {
           // If value is object, recursive call and handle k/v pairs in object
           flattenRow(columnValue, flattenedRow);
+        } else if (columnName === "death_cnt") {
+          // Rename death_cnt column to cris_death_cnt at the request of VZ Team
```
mateoclarke, line 81: @JaceDeloney, is this rename something that should be happening only for the CSV export? Should we be doing this at a system/database level?

mateoclarke (approved):
🚢 


pumpkin_dance

atd-vze/src/Components/GridExportData.js
mateoclarke: nvm, @tylerwhitson let me know its only for CSVs

PR #940, opened 2020-10-21
adds fix to delete unnecessary columns

tylerwhitson:
Fixes https://github.com/cityofaustin/atd-data-tech/issues/4163 once and for all

Parses out nested objects to create CSV with desired columns and order (as @mateoclarke originally suggested)

Examples CSV:

[atd_txdot_crashes2020-10-21T19_23_58-05_00.csv.zip](https://github.com/cityofaustin/atd-vz-data/files/5419307/atd_txdot_crashes2020-10-21T19_23_58-05_00.csv.zip)

### DIFF
--- a/atd-vze/src/Components/GridExportData.js
+++ b/atd-vze/src/Components/GridExportData.js
@@ -131,7 +131,7 @@ const GridExportData = ({ query, columnsToExport, totalRecords }) => {
       // push resulting column names to columnsToExportParsed array
       if (line === "") {
         return;
-      } else if (line.includes(" ")) {
+      } else if (line.includes(" { ")) {
         const nestedKeyArray = line.split(" { ");
         const nestedKeyString = nestedKeyArray[nestedKeyArray.length - 1];
         const nestedKey = nestedKeyString.split(" ")[0];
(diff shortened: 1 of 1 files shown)


atd-vze/src/Components/GridExportData.js
```diff
@@ -122,10 +119,44 @@ const GridExportData = ({ query, columnsToExport, totalRecords }) => {
...
+
+    // Create array of columns that should be displayed in the table
+    let columnsToExportParsed = [];
+
+    columnsToExport.split("\n").forEach(line => {
+      // Parse out unnecessary text and nesting keys from columnsToExport string,
+      // push resulting column names to columnsToExportParsed array
+      if (line === "") {
+        return;
+      } else if (line.includes(" ")) {
```
mateoclarke, line 134: Would this work the same if we were looking for the line to include a curly bracket character instead of an empty space? Just thinking that a trailing space added by accident to the query in the future could break things and be really hard to troubleshoot. An opening or closing bracket seems better to me.

mateoclarke (approved):
Looks great, just that one suggestion to make the includes less brittle.

pedro

atd-vze/src/Components/GridExportData.js
tylerwhitson: Good call, added an opening curly bracket to line 134. Thanks!

PR #942, opened 2020-10-28
updated pop estimate table and created constant with estimates

tillyw:
Fixes https://github.com/cityofaustin/atd-data-tech/issues/4215

Replaces population estimate image in information popover with table. Refactored code to stores values for table and By Population module in a constant so they only need to be changed in one location.

Screen Shot 2020-10-27 at 8 38 37 PM


atd-vzv/src/constants/populationEstimates.js
```diff
@@ -0,0 +1,26 @@
+export const populationEstimates = [
```
mddilley, line 1: It might make the references more readable to use an object here for `populationEstimates` instead of an array of objects. That way you could reference estimates with something `populationEstimates["years"]["2020"]` instead of relying on arbitrary indexes. Just a suggestion since this works as it is.

Maybe something like:
```
{
    "years": {
       "2020": 989233,
    },
    "sourceURL": <the url>,
    "sourceString": <the string>,
}
```

atd-vzv/src/constants/populationEstimates.js
```diff
@@ -0,0 +1,26 @@
...
+    year: "2017",
+    estimate: 937065,
+  },
+  {
+    year: "2016",
+    estimate: 913917,
+  },
+  {
+    sourceURL: "http://www.austintexas.gov/sites/default/files/files/Planning/Demographics/Austin_Forecast_2020_pub.pdf",
+    sourceString: "Ryan Robinson, City Demographer, Department of Planning, Cit of Austin. December 2019."
```
mddilley, line 24: Small typo here - the "y" on City

mddilley (approved):
This is a great improvement on the hard-coded values and the static image that was here before. Great thinking on setting up a config to make future updates a breeze. 🙌

atd-vzv/src/constants/populationEstimates.js
mateoclarke: good eye Mike!

mateoclarke (commented):
👍

PR #943, opened 2020-10-28
4188 preview instance

tillyw:
Fixes https://github.com/cityofaustin/atd-data-tech/issues/4187
Fixes https://github.com/cityofaustin/atd-data-tech/issues/4188

Adds REACT_APP_VZV_ENVIRONMENT env variable
Adds conditional statements so that when Netlify feeds in that env variable as PREVIEW, the data range is set to go up to the last day of the previous month and the endpoint is set to the test Socrata dataset

Screen Shot 2020-10-29 at 11 09 16 AM
Screen Shot 2020-10-29 at 11 09 33 AM
Screen Shot 2020-10-29 at 11 10 19 AM


atd-vzv/public/index.html
```diff
@@ -49,7 +49,7 @@
       content="https://visionzero.austin.gov/viewer/summary_img.png"
     />
     <!-- DO NOT ALTER "BELOW" THIS LINE FOR AUTOMATED ENVIRONMENT SNIPPETS -->
-    <!--VZV_HEADER_SNIPPETS-->
+<!-- VZV STAGING SNIPPETS -->
```
mateoclarke, line 52: Why did this update? Given the all caps warning comments, I'm hesitant to merge before reverting this code back

atd-vzv/public/index.backup.html
```diff
@@ -0,0 +1,69 @@
+<!DOCTYPE html>
```
mateoclarke, line 1: Doesn't seem like this file should've been committed either.

mateoclarke (changes requested):
Two files seem out of place and I don't think we should include those changes. Other than that, all good.

PR #947, opened 2020-11-11
VZE: Make sure VZE only shows level 4 and below #4452

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4452


frankhereford (approved):
This looks like it's correct to me. I double checked the equivalent WHERE clause using QGIS to visualize the results, and `(level_1 > 0 OR level_2 > 0 OR level_3 > 0 OR level_4 > 0)` returns only the level 1-4 set. I also confirmed, I believe, that the locations page of the staging VZE is doing what's expected after searching for level 1-4 polygon IDs and level 5 polygon IDs in the staging instance of VZE and got the expected results.

mateoclarke (approved):
🚢

PR #951, opened 2020-12-01
VZV | Update Socrata queries with times

mddilley:
This PR patches a bug where Socrata queries had a date range that was exclusive of the first and last date in the range. I updated these queries with time parameters. Below is a screenshot of the fatalities By Race/Ethnicity for start of January to end of October 2020 that now matches the overall total (78).

Screen Shot 2020-11-30 at 6 24 18 PM


sergiogcx (approved):
Good to go! ⏲️

mateoclarke (commented):
Really impressive that you were able to get to the source of this bug. Appreciate you digging deep on this one @mddilley!!! 🙏

PR #954, opened 2020-12-02
Fix invalid date bug by passing a fallback value

mateoclarke:
Closes https://github.com/cityofaustin/atd-data-tech/issues/2439

Dec-01-2020 19-26-10


sergiogcx (approved):
Thanks for taking care of this!

mddilley (commented):
This is a really nice fix! 🙌

PR #958, opened 2020-12-18
update netlify.toml for deploy preview context

tillyw:
Updates netlify.toml to set NODE_ENV to "staging" when in deploy preview context.


mateoclarke (approved):
🙏  ✨  🚢

PR #959, opened 2020-12-21
VisionZero ETL | The ETL fails to download the ZIP Files #4759

sergiogcx:
Closes https://github.com/cityofaustin/atd-data-tech/issues/4759


mddilley (approved):
Thanks for documenting the bug in the issue. 🙏 This change makes sense to me!

mateoclarke (commented):
👍

PR #963, opened 2021-01-05
Modify the message related to unavailable CR3 diagrams/narratives

frankhereford:
The intent is to not imply that a CR3 is not available when it just
may not have been processed yet. The message is made more general
to try to cover both when a CR3 is just not processed yet, or when
it was not possible for a diagram / narrative to have been extracted.


mateoclarke (approved):
👍

mateoclarke:
Thanks for this update @frankhereford! I'm going to go ahead and merge this to master and cherry-pick the commit to prod as well.

PR #967, opened 2021-01-09
Add new `investigator_narrative_ocr` field

mateoclarke:
https://github.com/cityofaustin/atd-data-tech/issues/4871

The database updates and permissions have already been added in staging & production Postgres and Hasura. This PR also removes the narrative field from the "Details" section of the crash page since it now has its own dedicated card.

### DIFF
--- a/atd-vzd/migrations/migration_crashes_2021_01_08--1946.sql
+++ b/atd-vzd/migrations/migration_crashes_2021_01_08--1946.sql
@@ -1,0 +1,12 @@
+--
+-- Adds column for narratives that won't be automatically overwritten by crash
+-- record updates
+--
+ALTER TABLE "public"."atd_txdot_crashes" ADD COLUMN "investigator_narrative_ocr" text;
+
+UPDATE
+	atd_txdot_crashes
+SET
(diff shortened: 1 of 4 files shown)


frankhereford:
This looks great to me - thank you for doing this. I'm processing the missing data, and it should be ready in 3-4 hours.

frankhereford (approved):
Thanks Mateo - I ack the the new '_ocr' suffix, and your fix looks great.

atd-vzd/migrations/migration_crashes_2021_01_08--1946.sql
```diff
@@ -0,0 +1,12 @@
...
+-- record updates
+--
+ALTER TABLE "public"."atd_txdot_crashes" ADD COLUMN "investigator_narrative_ocr" text;
+
+UPDATE
+	atd_txdot_crashes
+SET
+	investigator_narrative_ocr = investigator_narrative
+WHERE
+	investigator_narrative IS NOT NULL
```
sergiogcx, line 12: This UPDATE part is a good idea, we may have to run it again after the ETL is re-enabled.

sergiogcx (approved):

PR #970, opened 2021-01-26
prevent movement with id = 0 from being counted as null

tillyw:
Fixes https://github.com/cityofaustin/atd-data-tech/issues/5063

Prevents movements with id = 0 from being counted as null. I don't think this will break the initial intent of this logical statement of checking for null values and displaying them as blank, but would like to confirm that.

Now behaves as expected for unit records with movement of "Unknown":

https://user-images.githubusercontent.com/35410637/105905088-a6798d00-5fe7-11eb-978a-546314e3d641.mov

### DIFF
--- a/atd-vze/src/views/Crashes/RelatedRecordsTable.js
+++ b/atd-vze/src/views/Crashes/RelatedRecordsTable.js
@@ -69,7 +69,7 @@ const RelatedRecordsTable = ({
 
   const formatValue = (data, field) => {
     let fieldValue = data[field];
-    
+
     if (typeof data[field] === "object") {
       fieldValue =
         data[field] && data[field][fieldConfig.fields[field].lookup_desc];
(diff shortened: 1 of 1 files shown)


mddilley (approved):
Works great, Tilly! Another falsey bug squashed 🐛  🙌 

I tested against master branch using https://localhost:3000/#/crashes/17539181 and the dropdown populates as expected with this change. Also checked to make sure that a `null` value for `movement_id` still works with https://localhost:3000/#/crashes/18034580

mateoclarke (approved):
slayin' annoying bugs one line at a time. We should've done this earlier.

obama_tear_gratitude

tillyw:
> slayin' annoying bugs one line at a time. We should've done this earlier.
> 
> obama_tear_gratitude

Feels so good! One-liners can be just as good as PRs as they can be as jokes. Thanks, y'all!

PR #972, opened 2021-01-28
5020 by pop 2020 ratio feb 1

tillyw:
Fixes https://github.com/cityofaustin/atd-data-tech/issues/5020

Issue was due to time range end constant that feeds the by Pop and By Year & Month charts. It's due to the same edge case we encountered last year where the calculation for Feb. 1 is looking at Jan. 31 minus one month (Dec. 31, 2020) and minus one year (Dec. 31, 2019), so actually setting the end date to December 2019 instead of 2020. Adding a month to the end date solves the problem for both charts.

Screen Shot 2021-01-28 at 5 10 37 PM
Screen Shot 2021-01-28 at 5 10 48 PM

### DIFF
--- a/atd-vzv/src/constants/time.js
+++ b/atd-vzv/src/constants/time.js
@@ -60,7 +60,6 @@ export const fiveYearAvgStartDate = dataEndDate
   .startOf("year")
   .subtract(5, "year")
   .format("YYYY-MM-DD");
-
 export const fiveYearAvgEndDate = dataEndDate
   .clone()
   .add(1, "month")
--- a/atd-vzv/src/constants/time.js
+++ b/atd-vzv/src/constants/time.js
@@ -62,6 +62,13 @@ export const fiveYearAvgStartDate = dataEndDate
   .format("YYYY-MM-DD");
 export const fiveYearAvgEndDate = dataEndDate
   .clone()
+  .subtract(1, "year")
+  .endOf("year")
+  .format("YYYY-MM-DD");
+// Unique variable for the byPop chart that prevents the edge case
+// where the Feb. 1 query ends a year earlier than intended
+export const fiveYearAvgEndDateByPop = dataEndDate
+  .clone()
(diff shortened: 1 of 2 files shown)


mateoclarke:
Hey Tilly, I think the link to the issue is unrelated. Could you update the link so I can review the issue?

tillyw:
> Hey Tilly, I think the issued you linked above is unrelated. Could you update the link so I can review the issue?

Sorry about that! I accidentally deleted the last 0 in the URL and it became a completely different issue!

mddilley (commented):
This change works great for the By Population chart, Tilly! It's super cool to see your refactor of this component at work with the new year of data. 🙌  🎊 

I did notice a small bug that this change introduces in the five year avg. queries in the `CrashesByYear` component. The addition of one extra month to `fiveYearAvgEndDate` can push it into the next year in some cases and makes it a 6 year query. Happy to talk through this because it was hard for me to wrap my head around. The screenshots are 
from adding `console.log(avgQueries.fatalities);` on line 54 of `CrashesByYear`.

Before this change
Screen Shot 2021-02-01 at 5 15 53 PM

After this change
Screen Shot 2021-02-01 at 5 15 42 PM

tillyw:
Thanks for catching this, @mddilley! I got a little too enticed by the idea of a one-line fix and overlooked the bug this would have introduced. As we discussed, I'm introducing a new variable in time.js exclusively for this. PR is updated and ready for review and merging once approved @mateoclarke.

mddilley (approved):
Looks great to me! Glad that we had the chance to look back at some of these constants and discuss 🙏

tillyw:
> Looks great to me! Glad that we had the chance to look back at some of these constants and discuss 🙏

Me too, thanks @mddilley! Merging now

mateoclarke:
Looks good, thanks for the attention to detail in the review @mddilley

PR #973, opened 2021-02-03
#5182 update 2020 pop

tillyw:
Fixes https://github.com/cityofaustin/atd-data-tech/issues/5182

Updates population numbers for 2020 and 2021 using City of Austin Full Purpose Population column in [this doc](https://www.austintexas.gov/sites/default/files/files/Planning/Demographics/austin_forecast_2021_pub.pdf).

I did not see any changes to the 2019 figures, so I left them as they are.


mateoclarke (changes requested):
Can you also update the `sourceURL` & `sourceString` values? With the new PDF link and the new sources. It might also be helpful to either update the variable name or add a comment that signals that we're using the CoA Full Purpose Pop figure in case its ever a point of confusion down the line.

tillyw:
> Can you also update the `sourceURL` & `sourceString` values? With the new PDF link and the new sources. It might also be helpful to either update the variable name or add a comment that signals that we're using the CoA Full Purpose Pop figure in case its ever a point of confusion down the line.

@mateoclarke Apologies for overlooking that! I updated the sourceURL and the sourceString. I also added a note in the sourceString as to which column we are using for the figures, and reformatted with Prettier.

mateoclarke (approved):
Austin-Skyline-2013-2019

PR #975, opened 2021-02-26
VZV | Show partial year data in By Population visualization

mddilley:
After some discussion around the By Population visualization, we decided it makes more sense to show the latest data available in this chart and no longer wait until a full year of data is available.

To test how this will look on March 1st, take a look at the deploy preview:

https://deploy-preview-975--atd-vzv-preview.netlify.app/

or update `MONTHS_AGO` in `time.js` in the local environment to:

```js
export const MONTHS_AGO = process.env.REACT_APP_VZV_ENVIRONMENT === "PREVIEW" ? 1 : 1;
```

### Before
Screen Shot 2021-02-26 at 5 00 51 PM

### After
Screen Shot 2021-02-26 at 5 00 42 PM

### DIFF
--- a/atd-vzv/src/constants/time.js
+++ b/atd-vzv/src/constants/time.js
@@ -6,7 +6,7 @@ export const ROLLING_YEARS_OF_DATA = 4;
 // Number of months window slides in the past (to display most accurate data)
 // Accommodate for a preview instance that provide and extra month of data
 export const MONTHS_AGO =
-  process.env.REACT_APP_VZV_ENVIRONMENT === "PREVIEW" ? 1 : 1;
+  process.env.REACT_APP_VZV_ENVIRONMENT === "PREVIEW" ? 1 : 2;
 
 // Create array of ints of last n years
 export const yearsArray = () => {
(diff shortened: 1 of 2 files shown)


mateoclarke:
Surprised to see the number so low from 2021 since it should be a rate similar to the rest. I'm going to try to look into it more.

mateoclarke:
Ok, so I think our assumption was that we could include the ratio for the current year because it would be a consistent comparison against previous years. This would be true if we were calculating the ratio by month and averaging the monthly ratios over the course of the year to come up with the final ratio for each year. 

Unfortunately, that's not how the calculation currently works. We calculate the ratio of crashes using a year of data. So if we decided to show 2021 data, it would only include one month of data and the ratio will be about one-twelfth of the accurate ratio value. I think we need to revisit the idea of displaying the ratio for the current year. If we calculate by month and avg, it could work. But I'm not sure and we should discuss before decided our next steps.

mddilley:
These changes were discussed in the VZ check-in on 3/23/2021. Decided to keep this visualization as is or take a different direction if updated. Closing.

mateoclarke:
Closing this after the discussion we had today in our Monthly Meeting with VZ Team

PR #999, opened 2021-06-24
Jg 5971 crash info popup box disappears

jgabitto:
- Closes https://github.com/cityofaustin/atd-data-tech/issues/5971
- Make desktop and mobile same for usability
- Make Popover text selectable and copyable
- Add copy button for crash id in popover
- Add animated icon on selected marker

### DIFF
--- a/atd-vzv/src/views/map/AnimatedIcon.js
+++ b/atd-vzv/src/views/map/AnimatedIcon.js
@@ -1,0 +1,36 @@
+import React from "react";
+import { CanvasOverlay } from "react-map-gl";
+
+const AnimatedIcon = ({ location, paint }) => {
+  const { x, y } = location;
+  const _redraw = ({ width, height, ctx, project }) => {
+    const coordinates = project([x, y]);
+
+    var size = 66.6667;
+    var duration = 1000;
+    var t = (performance.now() % duration) / duration;
+
+    var radius = (size / 2) * 0.3;
+    var outerRadius = (size / 2) * 0.7 * t + radius;
+
+    // Draw the outer circle.
+    ctx.clearRect(0, 0, width, height);
+    ctx.beginPath();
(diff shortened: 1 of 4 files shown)


patrickm02L:
@jgabitto Nice work on getting the pop-up to stay, and figuring out how to indicate which dot is selected. I didn't consider having a pulsing dot to differentiate which point was selected but it seems like a novel solution. My only concern is that it may be distracting and off-putting. Perhaps, keeping the halo effect around the selected dot with a white border around the main dot would work too. Also, like that when you zoom out the selected dot gets bigger and the pop-up remains the same size.

patrickm02L (approved):
This solution works nicely. Prefer a non-pulsing halo around the selected dot versus the pulsing dot - just being nit-picky. Push it.

atd-vzv/src/views/map/AnimatedIcon.js
```diff
@@ -0,0 +1,36 @@
...
+  const _redraw = ({ width, height, ctx, project }) => {
+    const coordinates = project([x, y]);
+
+    var size = 66.6667;
+    var duration = 1000;
+    var t = (performance.now() % duration) / duration;
+
+    var radius = (size / 2) * 0.3;
+    var outerRadius = (size / 2) * 0.7 * t + radius;
+
```
sergiogcx, line 15: We do not use var, my linter is throwing warnings. For variables, we use either `const` for that will not change, and `let` for values that will change. These all seem constants, it would be great help to us if these can be made consts so that we can be consistent with our conventions.

sergiogcx (approved):
Love this magic! Thanks Jorge! 🙏 🙌 🚀 🚢

atd-vzv/src/views/map/AnimatedIcon.js
jgabitto: Yes I will change them to `const`s!

PR #1001, opened 2021-06-28
Make list order of travel modes consistent

jgabitto:
- Closes https://github.com/cityofaustin/atd-data-tech/issues/6391
- Displays consistent list order for All, Fatalities, and Serious Injuries categories in By Travel Mode chart.

### DIFF
--- a/atd-vzv/src/views/summary/CrashesByMode.js
+++ b/atd-vzv/src/views/summary/CrashesByMode.js
@@ -136,19 +136,14 @@ const CrashesByMode = () => {
 
   // Sort mode order in stack and apply colors by averaging total mode fatalities across all years in chart
   const sortAndColorModeData = (modeData) => {
-    const averageModeFatalities = (modeDataArray) =>
-      modeDataArray.reduce((a, b) => a + b) / modeDataArray.length;
-    const modeDataSorted = modeData.sort(
-      (a, b) => averageModeFatalities(b.data) - averageModeFatalities(a.data)
-    );
-    modeDataSorted.forEach((category, i) => {
+    modeData.forEach((category, i) => {
       const color = chartColors[i];
       category.backgroundColor = color;
       category.borderColor = color;
       category.hoverBackgroundColor = color;
       category.hoverBorderColor = color;
     });
-    return modeDataSorted;
+    return modeData;
   };
 
   // Create dataset for each mode type, data property is an array of fatality sums sorted chronologically


patrickm02L (approved):
Crashes by Mode fix is so nice. Push it!

atd-vzv/src/views/map/AnimatedIcon.js
```diff
@@ -0,0 +1,36 @@
...
+  const { x, y } = location;
+  const _redraw = ({ width, height, ctx, project }) => {
+    const coordinates = project([x, y]);
+
+    var size = 66.6667;
+    var duration = 1000;
+    var t = (performance.now() % duration) / duration;
+
+    var radius = (size / 2) * 0.3;
+    var outerRadius = (size / 2) * 0.7 * t + radius;
```
sergiogcx, line 14: Be sure to update your code here too! 🙌 🙏

atd-vzv/src/views/map/InfoBox/MapInfoBox.js
```diff
@@ -3,73 +3,74 @@ import { Popup } from "react-map-gl";
...
-  type, // id of feature layer
-}) => {
-  const popupInfo = selectedFeature && selectedFeature.properties;
-  const popupX = popupInfo.pixelCoordinates && popupInfo.pixelCoordinates.x;
+const MapInfoBox = React.memo(
+  ({
+    selectedFeature,
+    setSelectedFeature,
+    type, // id of feature layer
+  }) => {
```
sergiogcx, line 13: This is very cool! 👀  Out of curiosity, is there any particular reason we are using memo? Are we trying to improve on render performance?

atd-vzv/src/views/summary/CrashesByMode.js
```diff
@@ -136,19 +136,14 @@ const CrashesByMode = () => {
 
   // Sort mode order in stack and apply colors by averaging total mode fatalities across all years in chart
   const sortAndColorModeData = (modeData) => {
-    const averageModeFatalities = (modeDataArray) =>
-      modeDataArray.reduce((a, b) => a + b) / modeDataArray.length;
-    const modeDataSorted = modeData.sort(
-      (a, b) => averageModeFatalities(b.data) - averageModeFatalities(a.data)
-    );
-    modeDataSorted.forEach((category, i) => {
+    modeData.forEach((category, i) => {
```
sergiogcx, line 139: Perfect, and thank you! 🙌

atd-vzv/src/views/map/Map.js
```diff
@@ -58,6 +59,7 @@ const Map = () => {
   const [cityCouncilOverlay, setCityCouncilOverlay] = useState(null);
   const [isMapDataLoading, setIsMapDataLoading] = useState(false);
   const [crashCounts, setCrashCounts] = useState(null);
+  const [, setPointData] = useState(null);
```
sergiogcx, line 62: This is interesting, we create a state but we don't use it. To me this seems like a function that just triggers a change in the state to re-render via useEffect?

atd-vzv/src/views/map/InfoBox/MapInfoBox.js
jgabitto: So I did this because the parent component is being rerendered because of the animated icon and I was not able to access the copy and close button of this child component. With React.memo, it essentially prevents this component from being rerendered. That's my understanding of how this works, but I could be wrong.

atd-vzv/src/views/map/Map.js
jgabitto: Yeah so the map wouldn't rerender unless I updated a state value. It kind of feels hacky, but if I don't try and update a state value, the animated icon won't have the 'animated' effect. If you look at where I'm calling `setPointData`, I'm passing in an empty object. From what I read somewhere, because it's an object, Mapbox interprets it as a new value and the whole map gets rerendered causing the animated icon to animate. If you change the empty object to a value such as a number or string, the map doesn't rerender. I'm not sure how correct what I'm saying is, but that's what I found on google.

sergiogcx: @jgabitto It's not hacky it's genious! 👍 💯  Who knows if this is best practice but at least it is _working_ code. 😄  I love that code, but if you don't mind to add some commentary as to why that code was written that way so anyone who has to work on it will know what it's doing and why.

atd-vzv/src/views/map/InfoBox/MapInfoBox.js
sergiogcx: @jgabitto that's smart! Glad you were able to resolve it with memo 📝 🙌

sergiogcx (approved):
🚀 🚢

mateoclarke (approved):

mateoclarke:
@jgabitto, I'm going to go ahead and merge this one into `master` branch for you.

PR #1002, opened 2021-07-01
Address issue in get_location_totals() function

frankhereford:
The subquery used to count the number of Non-CR3 crashes
was using count(1) to count the number of crashes, but was
used in the context of a set of aggregate results, therefore
always coming up with 1 or the number of non-CR3 crashes, which
ever was higher.


sergiogcx (approved):
Thank you! 🙏  peace has been restored...

PR #1003, opened 2021-07-01
Re-write get_location_totals()

frankhereford:
Sergio and I were talking about this function earlier, and I wanted to see if I could offer another framing of the query to see if it could be made easier on the database and also correct the problem I had introduced earlier. What do you all think of this implementation?    

>     My earlier commit today introduced a behavior where the use of the
>     query as submitted was very expensive for the DB. It is not suitable for
>     production.
> 
>     I would like to propose this re-write of the function. This is,
>     essentially, two very simply queries which can each use the indexes in
>     place for atd_apd_blueform and atd_txdot_crashes. Both queries only can
>     return a single row, as all of the selected items are aggregate function
>     results. These two queries are defined as CTE tables containing one row
>     each, so the primary query can simply JOIN the two "tables", which
>     concatenates the fields into one row. Here, you can add the columns you
>     need to add and simply pass through the others.
> 
>     One thing to note, this version does allow the invocation of different
>     locations for CR3 crashes vs. non-CR3 crashes, where the previous query
>     would return nothing in that case. Since the arguments are both
>     specified, I think this is the intended behavior. Also, like the
>     previous query, a location_id is returned in the results, and this query
>     returns the CR3 location_id and not the non-cr3 one.

### DIFF
--- a/atd-vzd/triggers/get_location_totals.sql
+++ b/atd-vzd/triggers/get_location_totals.sql
@@ -1,30 +1,61 @@
-CREATE OR REPLACE FUNCTION public.get_location_totals(cr3_crash_date date, noncr3_crash_date date, cr3_location character varying, noncr3_location character varying, cost_per_crash numeric)
- RETURNS SETOF atd_location_crash_and_cost_totals
- LANGUAGE sql
- STABLE
-AS $function$
-with
-  noncr3 as (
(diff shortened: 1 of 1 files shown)


sergiogcx (approved):
Thanks again! 🚀 🚢

PR #1005, opened 2021-07-05
Change formatting for comp-cost and economic-cost to currency

frankhereford:
These value are shown in the GridTable for Non-CR3 crashes on the
location view.


sergiogcx (approved):
Let's do it!

PR #1006, opened 2021-07-05
Modify VZD export for Socrata to exclude private drive crashes

frankhereford:


sergiogcx (approved):
Looks good! Thanks for fixing this! 🙏

PR #1007, opened 2021-07-12
Add trigger to set austin_full_purpose

mateoclarke:
Closes https://github.com/cityofaustin/atd-data-tech/issues/6351

The intent of this PR is to make the Austin Full Purpose column more accurate for the needs of the VZ team and to include more crashes in the universe that counts towards metrics (like those in VZV).

The logic change is that if crashes have Austin City ID 22 and no coordinations (aka null position), we should assume they are Austin Full Purpose. 

This PR has not yet been tested on staging or implemented in production. First I wanted to make sure this is the right place for the code to live.


sergiogcx (approved):
Thanks for taking care of this @mateoclarke , I trust it has been tested in Staging and saved to production already.

PR #1009, opened 2021-07-13
Patch for Production: Location Totals Correction

sergiogcx:
Applies a front-end correction that makes use of an adjustment in the backend


frankhereford (approved):
Thank you for cherry picking these into production!

PR #1013, opened 2021-07-16
Update to lambda function to associate SVRD, directional crashes to polygons

frankhereford:
This PR includes the creation of a new view (cr3_nonproper_crashes_on_mainlane) which can be used to determine if and which polygon a crash should be associated to if it fits the following criteria:

- Incorrectly geocoded to be on the main-lane of a level 5 road
- Correctly tagged in the CRIS data to indicate that it's on a SVRD
- Contains a directionality indicator in the user provided name of the service road `/\s?([nsew])[arthous]*\s??b(oun)?d?/`
- Has a polygon such that that polygon can be found within a 130 degree cone emanating in the cardinal direction 90 degrees out of phase of the direction of travel within 500 linear feet under SRID 2277

This table needs to be exposed for SELECT via Hasura.

Also, the changed to the python code uses the := assignment (walrus) operator which was introduced in python 3.8 (Oct-2019). Its use eliminates a Hasura call, but I can code it out if that's too recent a python version for the lambda call.

Closes cityofaustin/atd-data-tech#6700

### DIFF
--- a/atd-events/crash_update_location/app.py
+++ b/atd-events/crash_update_location/app.py
@@ -11,7 +11,7 @@ from typing import Optional
 HASURA_ADMIN_SECRET = os.getenv("HASURA_ADMIN_SECRET", "")
 HASURA_ENDPOINT = os.getenv("HASURA_ENDPOINT", "")
 HASURA_EVENT_API = os.getenv("HASURA_EVENT_API", "")
-HASURA_SSL_VERIFY = True# Set False for local hasura with self-generated SSL cert
+HASURA_SSL_VERIFY = True # Set False for local hasura with self-generated SSL cert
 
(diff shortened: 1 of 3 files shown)


sergiogcx (approved):

patrickm02L (approved):

frankhereford:
I apologize for the post-approval commit, but I wanted to resolve build test problem. It looks like dependbot bumped our urllib3 8 days ago putting us in place where the previously specified version of requests and the new version of urllib3 were in a unresolvable state. Bumping requests up to 2.26.0 fixes this and the build tests are looking good again.

johnclary:
Just seeing this after re-reading the v1.21 release notes. This is phenomenal, @frankhereford!

PR #1014, opened 2021-07-19
Allow searching for crashes outside AFP / city_id 22

frankhereford:
Being able to search for crashes outside of the Austin Full Purpose jurisdiction has been requested by the VZ division.

This PR has a soft requirement of the forthcoming PR, #1007, but will function without it. 

This is accomplished in this PR by removing the crashesFullPurpose portion of the abstracted graphql query and adding a new advanced filter that allows for, when set to on, to remove the requirement for a crash to have austin_full_purpose = 'Y'. This ("on" => "remove requirement") capability is added to gqlAbstract.js to prevent having to write the search option in an awkward negative, such as "Remove restriction for being inside Austin." 

Closes cityofaustin/atd-data-tech#6702


sergiogcx (approved):
It works! 📦 🚢 🚀

atd-vze/src/queries/gqlAbstract.js
```diff
@@ -429,8 +429,13 @@ gqlAbstractTableAggregateName (
...
           for (let [key, syntax] of this.getEntries(filterItem)) {
-            // If enabled, add to the list or remove it from the query.
-            if (filtersState[filter.id]) {
+            // The invert_toggle_state property can be used to control if the condition is applied
+            // when the switch is 'on', the default, or when the switch is off.
+            // Based on the switch state and invert_toggle state, 
+            // add to the list or remove it from the query. 
+            if (filter.hasOwnProperty('invert_toggle_state') 
+                ? !filtersState[filter.id] 
+                : filtersState[filter.id]) {
```
sergiogcx, line 438: This is one of those edge cases where I cringe but I have enough peace to use it... Ideally libraries are agnostic to specific keywords but every once in a while you do need to have keywords that allow you to keep a state, in this case I can definitely see `invert_toggle_state` as being necessary to implement this functionality and making it available throughout the framework. 👍

frankhereford: Thanks for your thoughts on this! 

I had originally modified the code to where you would define a "where" key/value pair if you wanted to have a condition added to the graphql when the switch was turned on and a "where_false" key/value pair if you wanted to have a condition added to the graphql when the switch was turned off. 

It felt like a kludge and the code in gqlAbstract wasn't to great either, so I refactored it this morning and felt like it was good enough to offer to the group. 

In the future, I would be eager to learn how this sort of thing would be done most cleanly and return to this code to bring it up to speed. Again - thanks for your review and your observations on the implementation.

mateoclarke:
Hey I had some feedback on this one I wanted to share but this already got merged. Can we discuss?

frankhereford:
Yes, for sure. I merged this too early. Sergio let me know that I needed to, at least, have a developer and a PM give approval - I appreciate him letting me know. 

If there are problems with this, I am happy to back it out and I am looking forward to your thoughts - thank you!

atd-vze/src/queries/gqlAbstract.js
mateoclarke: To me, this seems to add unneeded complexity. Instead of having this extra code and required configuration for "Remove Austin Full Purpose" could we just invert the name of the label? It could be the phrasing of the label be inverted? Maybe I'm misunderstanding the problem.

PR #1016, opened 2021-07-23
Production Patch: Upgrade python runtime to 3.8 for SQS

sergiogcx:
One of the event handlers is having problems with the walrus operator, as referenced in this bug report:

https://github.com/cityofaustin/atd-data-tech/issues/6759

Rather than refactoring the code and re-testing it is easier to upgrade the runtime version to python's 3.8 which is currently supported in AWS Lambda:

### DIFF
--- a/atd-vze/package.json
+++ b/atd-vze/package.json
@@ -1,6 +1,6 @@
 {
   "name": "atd-vz-data",
-  "version": "1.21.0",
+  "version": "1.21.1",
   "homepage": "./",
   "description": "ATD Vision Zero Editor",
   "author": "ATD Data & Technology Services",
--- a/atd-events/README.md
+++ b/atd-events/README.md
@@ -2,7 +2,8 @@
 
 An event handler is basically a python script that
 will run every time we submit a message to AWS Simple
-Queue Service (SQS).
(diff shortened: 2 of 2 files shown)


mateoclarke:
Can you reference the issue in your PR description? https://github.com/cityofaustin/atd-data-tech/issues/6759

atd-vze/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "atd-vz-data",
-  "version": "1.21.0",
+  "version": "1.21.1",
```
mateoclarke, line 3: 👍

.github/workflows/aws-lambda-helper.sh
```diff
@@ -79,7 +79,7 @@ function deploy_event_function {
         --role $ATD_VZ_DATA_EVENTS_ROLE \
         --handler "app.handler" \
         --tags "project=atd-vz-data,environment=${WORKING_STAGE}" \
-        --runtime python3.7 \
+        --runtime python3.8 \
```
mateoclarke, line 82: ➕  🐍

mateoclarke (approved):
🚢

PR #1017, opened 2021-08-04
Mark For Re-Download Button

sergiogcx:
WIP

Demo:
https://deploy-preview-1017--atd-vze-staging.netlify.app/#/

This is what it looks like atm:

https://user-images.githubusercontent.com/5282430/128257589-632e97fb-f6bb-47e6-b957-c8f2710f3871.mp4

### DIFF
--- a/atd-vze/package-lock.json
+++ b/atd-vze/package-lock.json
@@ -1,8 +1,22794 @@
 {
   "name": "atd-vz-data",
-  "version": "1.20.0",
-  "lockfileVersion": 1,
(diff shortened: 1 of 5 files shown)


mateoclarke:
I think we decided this solution wasn't ideal and we wanted to get more input from Xavier about how this would fit into his workflow. I appreciate you taking the time to present this proof of concept, but I think we should make sure to have an issue created and a solution identified from the product side before we ask you to work on something that take away from other priorities.

PR #1018, opened 2021-08-09
Script to restore archived versions of CR3s on S3

frankhereford:
Included is a script which intends to do the following:

- Accept a list of crashes in the form of a JSON file on disk. Please find a file suitable for testing the program and the file intended to fix the set of crashes identified earlier [in this gist](https://gist.github.com/frankhereford/19017a97a628d2204a9a74adcb2afeaa). 
- Iterates over the archived versions of the CR3 file in S3, newest to oldest
- Inspects the files using libmagic (the library behind unix's `file`) to find a valid PDF
- Builds an accurate cr3_file_metadata jsonb object for the first found PDF
- Restores the PDF and updates the CR3 file metadata.

Existing, unrelated flags that may be in that JSONB object previous to restoration are preserved, and an additional flag is added to indicate that the restore operation was preformed on the crash. Both database operations are performed using GraphQL via Hasura. 

Closes cityofaustin/atd-data-tech/issues/6866


patrickm02L (approved):
Appreciate the thorough notes in the code. Ship it!

atd-toolbox/s3_restore_cr3_pdfs.py
```diff
@@ -0,0 +1,270 @@
+#!/usr/bin/python3
+
+from __future__ import print_function
+
+import re
+import os
+import sys
+import json
+import boto3
+import magic
```
sergiogcx, line 10: Brilliant! I've never worked with magic, it's cool to see it in action. 🧙 ✨

atd-toolbox/s3_restore_cr3_pdfs.py
```diff
@@ -0,0 +1,270 @@
...
+import argparse
+import datetime
+import requests
+from operator import attrgetter
+from botocore.config import Config
+
+# https://stackoverflow.com/questions/5574702/how-to-print-to-stderr-in-python
+def eprint(*args, **kwargs):
+    print(*args, file=sys.stderr, **kwargs)
```
sergiogcx, line 19: Neat 💯  I need to take your example and make my ETL scripts to print to stderr more often

atd-toolbox/s3_restore_cr3_pdfs.py
```diff
@@ -0,0 +1,270 @@
...
+    eprint("Please set environment variables AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY")
+    sys.exit(1)
+
+
+# verify that environment Hasura variables were available and have populated values
+try:
+    assert(HASURA_ADMIN_KEY is not None and HASURA_ENDPOINT is not None)
+except:
+    eprint("Please set environment variables HASURA_ADMIN_KEY and HASURA_ENDPOINT")
+    sys.exit(1)
```
sergiogcx, line 101: It never occurred to me we can use the assert function like this (outside of a test) 🥇

atd-toolbox/s3_restore_cr3_pdfs.py
```diff
@@ -0,0 +1,270 @@
...
+        # not really magic; the underlying library of the `file` command on unix is called libmagic
+        # use libmagic to figure out what kind of file the file is
+        mime_type = magic.Magic(flags = magic.MAGIC_MIME_TYPE).id_buffer(previous_version)
+        encoding = magic.Magic(flags = magic.MAGIC_MIME_ENCODING).id_buffer(previous_version)
+
+        if mime_type != 'application/pdf':
+            print("Skipping version " + obj.get('VersionId') + " because it is a " + mime_type)
+            continue
+        else:
+            print("Version " + obj.get('VersionId') + " is acceptable for restoration because it is a " + mime_type)
```
sergiogcx, line 212: Where the real magic is! 🧙 Would you mind setting up a requirements file @frankhereford ? so that way we know what libraries we need to install to run the code? You may be able to collect the libraries for this script using pip freeze.

```
pip freeze > requirements.txt 
```

sergiogcx (approved):
Tremendous code, learning a lot from it! it's so impressive you coded a complicated version scanner in such a short amount of time!

atd-toolbox/s3_restore_cr3_pdfs.py
frankhereford: Yea 100%, that is a great idea. Thanks for the suggestion.

frankhereford: I installed a venv, got the script running in that and then froze out the requirements with pip. Thanks again for the suggestion! I'll be sure to do any future python scripts in a venv if they are stand-alone like this one; it's so the only way to go.

atd-toolbox/s3_restore_cr3_pdfs.py
johnclary: TIL! for what it's worth the `logging` [stream handler](https://docs.python.org/3/library/logging.handlers.html#streamhandler) is probably the more conventional approach.

frankhereford: Thanks John - I will check that out and use it in the future.

atd-toolbox/s3_restore_cr3_pdfs.py
johnclary: @frankhereford FYI, the [PEP 8 guidelines](https://www.python.org/dev/peps/pep-0008/#imports) on module imports. To make the dependencies super clear, the best practice is to import 3rd party libs after built-ins. I even add a linebreak between built-in/3rd-party/local imports but that is probably controversial.

atd-toolbox/s3_restore_cr3_pdfs.py
johnclary: good call—and thanks @frankhereford !

atd-toolbox/s3_restore_cr3_pdfs.py
frankhereford: That's really helpful to have the guidelines document to reference. I also like the new line to group them into logical sections. Thanks for that advice, I'm going to adopt it in the future.

atd-toolbox/s3_restore_cr3_pdfs/s3_restore_cr3_pdfs.py
```diff
@@ -0,0 +1,268 @@
+from __future__ import print_function
```
johnclary, line 1: @frankhereford are you running this in a python 2.x environment? i'm curious why this import is needed.

atd-toolbox/s3_restore_cr3_pdfs/s3_restore_cr3_pdfs.py
```diff
@@ -0,0 +1,268 @@
...
+
+    s3 = boto3.client(
+            's3', 
+            aws_access_key_id = ACCESS_KEY, 
+            aws_secret_access_key = SECRET_KEY, 
+            config = aws_config
+            )
+    prefix = ('production' if args.production else 'staging') + '/cris-cr3-files/'
+    s3.list_objects(Bucket = bucket, Prefix = prefix)
+except:
```
johnclary, line 131: @frankhereford the tract/catch pattern is good. for debugging, it's probably worthwhile to print the actual exception message before you exit. You can extend your handler like:

```python
try:
    # do stuff
except Exception as e:
    eprint("Unable to complete call to S3; check AWS credentials")
    # print string representation of Exception msg
    eprint(str(e))
    sys.exit(1)
```

atd-toolbox/s3_restore_cr3_pdfs/s3_restore_cr3_pdfs.py
```diff
@@ -0,0 +1,268 @@
...
+            metavar = 'crashes.json')
+
+    argparse.add_argument("--i-understand",
+            help = 'Do not ask the user to acknoledge that this program changes the state of S3 objects and the database.',
+            action = 'store_true')
+
+    args = argparse.parse_args()
+except:
+    # a stderr log is not needed, argparse croaks verbosly
+    sys.exit(1)
```
johnclary, line 65: @frankhereford @sergiogcx i assume the idea is that Airflow will pickup the exit code (`1`) and handle alerts as needed? I've never actually attempted this in a docker container but it makes perfect sense :)

sergiogcx: Airflow should pick it up, most processes that have an exception or an error exit with a code greater than 0. By default Airflow will run the DAG through (as it finishes the execution), and depending on how it's written it will trigger the `on_failure_callback `, which in our case it drops a message in Slack.

atd-toolbox/s3_restore_cr3_pdfs/s3_restore_cr3_pdfs.py
frankhereford: Good call; It's actually not needed. I picked up that line from the page I mentioned on line 15, and it isn't needed on the 3.9.5 version that I'm using to develop this. 

I'd like to take your feedback that you've given and update the program to implement it (including striking that line) - it's really helpful direction and I'm grateful for your time looking at it.

atd-toolbox/s3_restore_cr3_pdfs/s3_restore_cr3_pdfs.py
frankhereford: I wasn't actually think of this program ever running in Airflow. If it needs to, I'd like to move it into the ETL directory in the repo and test it out in the context of docker.

As it stands, it's a one-off script to fix a known set of crashes that have had their CR3 file over-written with an HTML page on the S3 bucket where they are stored. FWIW, here are [some JSON](https://gist.github.com/frankhereford/19017a97a628d2204a9a74adcb2afeaa) files - one for testing and one that defines that set of crashes.

atd-toolbox/s3_restore_cr3_pdfs/s3_restore_cr3_pdfs.py
frankhereford: Got it - thanks! I'll do that for the places where that pattern is used.

atd-toolbox/s3_restore_cr3_pdfs/s3_restore_cr3_pdfs.py
johnclary: thank you both. i wasn't sure of the deployment/use context!

atd-toolbox/s3_restore_cr3_pdfs/s3_restore_cr3_pdfs.py
johnclary: 👍 @frankhereford and i defer to whatever y'all think is the right amount of investment on this. just wanted to make sure you had the feedback. thanks for taking it.

mateoclarke:
Great stuff Frank! 👏

PR #1021, opened 2021-08-11
Refactor lambda function to find service road polygons via a function instead of a view

frankhereford:
This PR intends to close cityofaustin/atd-data-tech/issues/6763. 

It contains a new SQL function based on the logic contained in the existing view `cr3_nonproper_crashes_on_mainlane` but is reworked to only evaluate a single crash, as specified as an argument, instead of many crashes which occurs when the view was used. This optimization allows a crashes' potential location_id to be found much quicker so the lambda function doesn't get terminated.

The python cProfile module reports that the function which makes the call to Hasura to see if a service road polygon is appropriate and to retrieve its id now takes .6 seconds, when tested against the staging database.

### DIFF
--- a/atd-events/crash_update_location/app.py
+++ b/atd-events/crash_update_location/app.py
@@ -71,11 +71,12 @@ def is_crash_nonproper_and_directional(crash_id: int) -> str:
         return False
 
     check_nonproper_polygon_query = """
-        query has_alt_polygon($crash_id: Int) {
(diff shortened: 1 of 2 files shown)


patrickm02L (approved):
Ship It!

sergiogcx (approved):
Be sure to apply to both staging and production 📦 🚀

mateoclarke (commented):
👍

PR #1025, opened 2021-09-02
Promote v.1.22.0 to production

frankhereford:
This PR is for testing of forthcoming release [v1.22.0 Enfield Road](https://github.com/cityofaustin/atd-vz-data/releases/tag/v.1.22.0).


frankhereford:
## λ function
#### Testing
1) Execute this query to get a list of crashes of interest
```
select c.crash_id, c.crash_date, c.surface_street_polygon 
from cr3_nonproper_crashes_on_mainlane c
order by c.crash_date desc
```
2) Pick one which has a non-null surface street polygon and query the centroid out of Hasura with: 

```
query get_centroid($location_id: String) {
  atd_txdot_locations_with_centroids(where: {location_id: {_eq: $location_id}}) {
  centroid
  }
}

{ "location_id": "BADD00DAD"}
```
3) Grab that XY pair and nudge the crash from the first step over [an inch or two](http://wiki.gis.com/wiki/index.php/Decimal_degrees). For example, adjust digits out past the 7th digit of precision, like this:

```
mutation update_crash_position($crashId: Int!, $longitude: float8, $latitude: float8) {
  update_atd_txdot_crashes(where: {crash_id: {_eq: $crashId}},
  _set:{latitude_primary:$latitude, longitude_primary: $longitude}) {
    affected_rows
    }
  }

{
  "crashId": <crash id from above>,
  "longitude": -97.67342928675309,
  "latitude": 30.3265447525600
}
```
4) If things are working as intended, then you should see the crash jump locations (much more than you moved it) to the centroid of a nearby polygon and become associated to that polygon. 

5) Repeat 1-4, but this time with a crash that does not have a `surface_street_polygon` from the query in step 1, and use the latitude, longitude stored in the `position` field, which you can access easily from the VZE. It should not jump, nor get a polygon ID associated to it.

6) Now move this crash off to a service road and make sure it gets a polygon. 

7) And finally, move it back outside of a polygon on the center of the level 5 road and make sure it looses its polygon association. You can move it to the center of a city block if you need to test this more easily, just be sure to geolocate it accurately when you're done.

patrickm02L (approved):
Everything looks good. Super happy to launch this beast!

PR #1026, opened 2021-09-02
Script to nudge crashes so the lambda function moves them to a service road location

frankhereford:
This PR contains a one-off script and also includes a README.md file for the atd-toolbox directory. 

The README describes the new script as:

> This is a Python script which will use a database view to find the set of crashes which:
> * Are on a level 5 centerline,
> * Have CR3 data which indicates that the crash occurred on a service road,
> * Have a directionality which gives a hint which way off the centerline the crash should be moved.
> 
> For these crashes, the script will add 0.00000001 decimal degrees to the latitude of the crash, which is an immaterial movement in geographic terms, but it will cause the lambda function tied to crash-updates to execute. It is this lambda function which will figure out what location to move the crash to and will do so.


mateoclarke (approved):
japan-toolbox

patrickm02L (approved):
0.00000001!

PR #1027, opened 2021-09-07
Setup local dev environment

mateoclarke:
Closes https://github.com/cityofaustin/atd-data-tech/issues/7240


frankhereford:
I'm going to close this. I think we have some other efforts along this line that will supercede this work. Thanks!

PR #1030, opened 2021-10-03
Resolve "824 crashes have $0 comprehensive cost"

frankhereford:
This pull request intends to close cityofaustin/atd-data-tech/issues/6875, but it is not a currently approved solution.

@patrickm02L If you're available, can we sync on this on Monday in advance of the VZ meeting coming up?

This is a potential solution for the 800 crashes which do not have a comprehensive cost. This solution would need approval from the VZ stakeholders because it changes the ground-state, default comprehensive cost value at the same cost of "non-injury" crash.

If deployed, it would need to use a query derived from the following to set the `est_comp_cost_crash_based` column to the default value for crash.

```
with persons as (
    select person_id, crash_id
    from atd_txdot_person
    union all
    select primaryperson_id as person_id, crash_id
    from atd_txdot_primaryperson
    )
--select crashes.crash_id, count(persons.person_id) as person_count
select crashes.crash_id, persons.person_id
from atd_txdot_crashes crashes
join atd_jurisdictions j on (ST_Intersects(j.geometry, crashes.position) and j.id = 5)
left join persons on (persons.crash_id = crashes.crash_id)
where crashes.est_comp_cost_crash_based = 0
and crash_date > now() - '5 year'::interval
order by person_id asc
--group by crashes.crash_id
```
After investigating ~30 of the crashes which are affected, I have observed the following:

- A small handful (<3%) of the crashes affected are those which have a person in one of the persons tables, but that person died after the 30 day window following the crash. This makes them a "non-atd" fatality, and there is not a comprehensive cost defined for that injury outcome.

- The vast majority of the crashes affected simply do not have people listed in the persons tables, nor on the CR3. It's an issue stemming from the initial issuance of the CR3 from the investigating officer. Consequently, there are no people to have injuries in the computation, and a non-zero, default value may be preferred.

### DIFF
--- a/atd-vzd/migrations/migration_update_audit_log_trigger_2021-10-03--1721.sql
+++ b/atd-vzd/migrations/migration_update_audit_log_trigger_2021-10-03--1721.sql
@@ -1,3 +1,5 @@
+-- Augment the atd_txdot_crashes_update_audit_log() function to compute and store the crash based comprehensive cost
+-- This updated version of the function is also found in atd-vzd/triggers/atd_txdot_crashes_updates_audit_log.sql
 create or replace function atd_txdot_crashes_updates_audit_log() returns trigger
     language plpgsql
 as
@@ -16,8 +18,8 @@ BEGIN
         INSERT INTO atd_txdot_change_log (record_id, record_crash_id, record_type, record_json, updated_by)
         VALUES (old.crash_id, old.crash_id, 'crashes', row_to_json(old), NEW.updated_by);
     END IF;
-            
-    -- COPIES THE CITY_ID INTO ORIGINAL_CITY_ID FOR BACKUP 
+
+    -- COPIES THE CITY_ID INTO ORIGINAL_CITY_ID FOR BACKUP
     IF (TG_OP = 'INSERT') THEN
             NEW.original_city_id = NEW.city_id;
     END IF;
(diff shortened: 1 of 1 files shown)


frankhereford:
In order to deploy this change, we would need to update the `audit_log` function on the database, and then run this query:

```
update atd_txdot_crashes set est_comp_cost_crash_based = 50000 where crash_id in (
with persons as (
    select person_id, crash_id
    from atd_txdot_person
    union all
    select primaryperson_id as person_id, crash_id
    from atd_txdot_primaryperson
    )
select crashes.crash_id
from atd_txdot_crashes crashes
left join persons on (persons.crash_id = crashes.crash_id)
where crashes.est_comp_cost_crash_based = 0
and person_id is null
)
```

This query is intentionally omitting the 16 crashes from the last 5 years in the Austin Full Purpose jurisdiction which contain only non-ATD fatalities. I have emailed the list of said crashes to Lewis, per our conversation this morning, and we can adjust this query to follow his direction on those if it's to include them at the $50 comprehensive cost level, or write up an issue for a more complicated fix that places them in their own comprehensive cost bracket.

patrickm02L (approved):
Stellar notes. Ship it!

frankhereford:
Memorializing direction from Lewis Leff this morning: I was able to quickly recap the circumstances which occured to create this set of crashes having no comprehensive cost. Lewis approved of the plan to assign crashes with no people, and therefor no injuries, the value of a non-injury crash, which currently is set at $50,000 as per the `atd_txdot__est_comp_cost_crash_based` table.

mateoclarke (approved):
🚢 💨

frankhereford:
@patrickm02L I put this in ready-to-deploy on the kanban board; please advise when / if / how you'd like this change to be put in the database to make it live. The changes have been made on staging.

patrickm02L:
@frankhereford go ahead and make it live, and close out the issue.

PR #1037, opened 2021-11-17
Enable optional chaining & use it on locations page

frankhereford:
This PR intends to close issue [#7735](https://github.com/cityofaustin/atd-data-tech/issues/7735).

In the above bug, it was noted that the locations page for location `83E78D0D49`, [here](https://visionzero.austin.gov/editor/#/locations/83E78D0D49), would produce a WSOD. This was the result of a temporary crash record which was bearing a `null` value in the `fhe_collsn_id` field in the `atd_txdot_crashes` table. The temporary crash record was created correctly; there is no way for the user to input the manner of the collision. As a result, any temporary records would render the location page in which they landed to be unusable. 

This PR upgrades the `react-scripts` package which in turn allows its dependency packages in the `babel` family to update. Babel 7.8.0 is the minimum version which supports optional chaining, and it was released on 1/11/2020.

### DIFF
--- a/atd-vze/package-lock.json
+++ b/atd-vze/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "atd-vz-data",
-  "version": "1.21.1",
+  "version": "1.22.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


sergiogcx (approved):

patrickm02L (approved):
No more WSOD!!!

frankhereford:
Thanks for the super fast reviews y'all! 🙏

mateoclarke (commented):
Tested it in staging and it looks good! 👍

PR #1051, opened 2022-01-12
Update Crash Changes business logic

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/4582

- [ ] Add 2 new columns for killed and seriously injured field to the changes index page table
- [ ] Double check that VZE rejects any records that were previously manually QA/updated.
- [ ] Move fields between tables

Adding the two new columns involved adding a SQL View to the database. This has been executed on the staging database but needs to be executed in production after PR approval. 

Line 678 in helpers_import.py skips records that were already human QA'd.

## Testing
**URL to test:** 
https://deploy-preview-1051--atd-vze-staging.netlify.app/#/changes

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-etl/app/process/helpers_import.py
+++ b/atd-etl/app/process/helpers_import.py
@@ -509,12 +509,10 @@ def clean_up_record(record):
 def record_compare(record_new, record_existing):
     """
     Compares two dictionaries. It uses the CRIS_TXDOT_COMPARE_FIELDS_LIST
-    to determine what fields are important enough, and returns True if
-    there is one important difference. Returns False if none of the fields
-    present any differences.
+    to determine what fields are important enough, and returns list of differences
     :param dict record_new: The new object being parsed from csv
     :param dict record_existing: The existing object, as parsed from an HTTP query
-    :return: bool
+    :return: list
     """
     differences = []
     record_new = clean_up_record(record_new)
@@ -549,6 +547,8 @@ def insert_crash_change_template(new_record_dict, differences, crash_id):
     """
     Generates a crash insertion graphql query
     :param new_record_dict: dict - The new record as a dictionary
+    :param differences
+    :param str crash_id
     :return: string
     """
     # Turn the dictionary into a character-escaped json string
--- a/atd-etl/app/process/helpers_import.py
+++ b/atd-etl/app/process/helpers_import.py
@@ -679,7 +679,7 @@ def record_crash_compare(line, fieldnames, crash_id, record_existing):
           return False, "Skipping changes on manually QA/updated record"
         elif important_update:
             affected_rows = 0
-            if compare_enabled: # what happens if this is false? silent skip?
+            if compare_enabled:
                 mutation_template = insert_crash_change_template(
                     new_record_dict=record_new,
                     differences=differences,
--- a/atd-etl/app/process/helpers_import.py
+++ b/atd-etl/app/process/helpers_import.py
@@ -634,9 +634,11 @@ def is_important_update(differences):
 def record_crash_compare(line, fieldnames, crash_id, record_existing):
     """
     Hook that finds an existing record, and compares it with a new incoming record built from a raw csv line.
-    :param str  line: The raw csv line
+    :param str line: The raw csv line
     :param str[] fieldnames: The strings to be used as headers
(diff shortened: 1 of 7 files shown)


atd-etl/app/process/helpers_import.py
```diff
@@ -663,19 +665,21 @@ def record_crash_compare(line, fieldnames, crash_id, record_existing):
...
-        # There are differences, and it can update unless
+        # There are differences, and it can automatically update unless
         # either human_updated or important_update is true
         compare_enabled = (
             ATD_ETL_CONFIG["ATD_CRIS_IMPORT_COMPARE_FUNCTION"] == "ENABLED"
         )
-        if human_updated or important_update:
+        if human_updated:
+          return False, "Skipping changes on manually QA/updated record"
+        elif important_update:
```
chiaberry, line 680: Here is my change in accordance with the issue: 

> So that the query that populates this table checks whether the status is manually QA-ed and blocks incoming changes so it doesn't show up on the table.

atd-vze/src/views/Crashes/crashChangesGridTableParameters.js
```diff
@@ -1,28 +1,48 @@
 export const crashChangesGridTableColumns = {
-    record_id: {
-        primary_key: true,
-        searchable: true,
-        sortable: true,
-        label_search: "Search by Record ID",
-        label_table: "Record ID",
-        type: "int",
+  record_id: {
```
chiaberry, line 2: I ran prettier on this file which changed all the spacing

atd-etl/app/process/helpers_import.py
chiaberry: > Accept changes to lat/long changes by default (if not manually updated previously), and don't include those records in this index.

lat/long isn't included as one of the important fields when checking for important update.

patrickm02L (approved):
Reviewed new columns added, and fields moved to respective new tables. Everything checks out. SHIPIT! 

https://user-images.githubusercontent.com/5566727/149586463-02cf4ec7-f422-4263-82fb-a4a81968902a.mp4

atd-etl/app/process/helpers_import.py
frankhereford: > > Accept changes to lat/long changes by default (if not manually updated previously), and don't include those records in this index.
> 
> lat/long isn't included as one of the important fields when checking for important update.

I think those two fields may be actually included in the list of important fields; [line 355 & 356](https://github.com/cityofaustin/atd-vz-data/blob/4582-crash-logic/atd-etl/app/process/helpers_import_fields.py#L355-L356) in `helpers_import_fields.py`.

I think they may need to be pulled out and handled with a special case bit of logic in [helpers_import.py](https://github.com/cityofaustin/atd-vz-data/blob/4582-crash-logic/atd-etl/app/process/helpers_import.py#L509).

chiaberry: They are included in the list of differences to look for, but the important fields just checks for one of two: https://github.com/cityofaustin/atd-vz-data/blob/4582-crash-logic/atd-etl/app/process/helpers_import.py#L630

frankhereford (approved):
🚢 🚢 🚢 

I've tested this using a local instance of the DB, Hasura, VZE and the ETL image. For those who may want to execute further tests, please let me know if you're interested in the docker-compose file that ties these together into a stack or a set of modified CSV scripts that you can run through `process_hasura_import.py.`

image

This is a screenshot showing a fake crash I crafted which I have subsequently reprocessed with changing data, as if we were getting updates from CRIS.

atd-vzd/migrations/migration_view_atd_txdot_crashes_2022_01_14--1307.sql
```diff
@@ -0,0 +1,26 @@
+CREATE 
+OR replace view atd_txdot_changes_view AS WITH changes_formatted AS 
+(
+   SELECT
+      *,
+      atc.record_json::jsonb ->> 0 AS record_json_as_object
```
mateoclarke, line 6: 👏  Glad this view approach was the way to go!

mateoclarke (approved):
🚢 💨 ✨

PR #1057, opened 2022-02-16
Add Lighting Condition OCR Script as a new VZ Toolbox Entry

frankhereford:
## Associated issues

This script was used to close: https://github.com/cityofaustin/atd-data-tech/issues/8475. 

While it is unlikely that we'll need to OCR this particular field out of CR3s again, it certainly possible that we'll need to OCR some other data in the CR3 PDFs in the future, and this script will be a handy reference or jumping off point on that work.

This program was pulled together hastily to address an urgent data integrity question. It currently lacks meaningful error handling and documentation. This PR is intended to hold a reference to the branch open so we don't delete it without getting those enhancements done and merging them into `master`. 

## Testing

---
#### Ship list
- [ ] Code reviewed


frankhereford:
While this script was successful, I no longer thing it bears enough merit to polish up and get into master. Any future event that causes us to do one-off OCR on a unusual field in CR3s would benefit from this script, perhaps, but it would stake customization to get it done. I want to just deal with that off-chance when it comes -- we can always dig this up or really, remaking it for some arbitrary field will all be about the same amount of work.

PR #1070, opened 2022-04-28
Add imported_at as a field to record when a crash record is created in the VZDB

frankhereford:
## Associated issues

This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/9000. 

Upon acceptance of this PR into the codebase, I will execute the migration in staging and then production. This action will be recorded in a comment to this PR. I've tagged myself as the assignee responsible for this execution.

## Testing

One can test this by spinning up a local copy of the VZDB and executing the `ALTER` statement. I would be happy to provide  anyone some `docker-compose.yml` content if it would be helpful.

FWIW, I have run the command against a local copy of the DB populated with yesterday's backup, so it may be sufficient to simply read the code for sanity and make a determination. 

---
#### Ship list
- [x] Code reviewed


patrickm02L (approved):
Looking forward to knowing when they records get delivered (imported)!

mateoclarke (approved):
datetime-timestamp

frankhereford:
🏁 

I have added the column `imported_at` as type timestamp to staging and to production. 

I did do this in two statements however, as follows. The result configuration does match the PR'd migration exactly however. 

```
ALTER TABLE atd_txdot_crashes ADD COLUMN imported_at timestamp DEFAULT null;
ALTER TABLE atd_txdot_crashes ALTER COLUMN imported_at SET default now();
```

This isn't really here nor there, it just keeps 2022-04-29 from having a jillion records on file as imported today. Everything that was imported prior to running that pair of `alter`s has an `imported_date` of `null`. We can back fill that with best guesses if needed, but I don't see a justification for the work.

PR #1079, opened 2022-06-29
9497 create new note

roseeichelmann:
## Associated issues

Closes cityofaustin/atd-data-tech#9497

## Testing
**URL to test:** 

This can be tested locally or as a Netlify preview (https://deploy-preview-1079--atd-vze-staging.netlify.app/#/dashboard)

**Steps to test:**

1. Navigate to Crashes tab and select any crash ID.
2. Test adding a new note to the Notes table.
3. Table should repopulate with your new note, your email, and the current date. The input text box should automatically have cleared.
4. Go to crash details page for crash ID #17727381 to see what the Notes table looks like when populated with multiple text entries and longer paragraphs. (https://deploy-preview-1079--atd-vze-staging.netlify.app/#/crashes/17727381)

---
#### Ship list
- [x] Code reviewed


atd-vze/src/queries/notes.js
```diff
@@ -2,7 +2,7 @@ import { gql } from "apollo-boost";
 
 export const GET_NOTES = gql`
   query FindNotes($crashId: Int) {
-    notes(where: { crash_id: { _eq: $crashId } }) {
+    notes(where: {crash_id: {_eq: $crashId}}, order_by: {date: desc}) {
```
mateoclarke, line 5: 👍 ⬇️

atd-vze/src/views/Crashes/Notes.js
```diff
@@ -1,42 +1,95 @@
...
+    addNote({
+      variables: {
+        note: newNote,
+        crashId: crashId,
+        userEmail: userEmail
+      }
+    }).then(response => {
+      setNewNote("");
+      refetch();
+    });
```
mateoclarke, line 46: It's not required but it would be a best practice to use `.catch` to watch for any errors in the mutation process. after the `.then()` closes, you could add `).catch( error => console.error(error)`.

This could at least help a dev triage if a bug is reproducible.

More info of error handling in JS:
- https://lucymarmitchell.medium.com/using-then-catch-finally-to-handle-errors-in-javascript-promises-6de92bce3afc
- https://javascript.info/promise-error-handling

mateoclarke (approved):
nice-michael-scott

patrickm02L (approved):
It looks good. Really good job! 🤘

https://user-images.githubusercontent.com/5566727/177997218-d72d2c7b-e0f8-4f05-b156-d4593410e074.mp4

atd-vze/src/queries/notes.js
```diff
@@ -13,3 +13,20 @@ export const GET_NOTES = gql`
...
+    }) {
+      returning {
+        crash_id
+        text
+        date
+        user_email
+      }
+    }
+  }
+`;
```
frankhereford, line 32: It's super not a big deal, but if you want, can you put a line feed at the end of the file? It's a hold-over convention from the 80s, but it persists as part of the POSIX standard.

frankhereford (approved):
Rose, this looks awesome! 🙏 

Thank you for adding this feature; I know the VZ team will put it to good use. I can imagine tons of times when we would have loved to have this back when I worked in that group.

The code looks great, and I have tested it on the netlify deployment. Thank you again.

atd-vze/src/queries/notes.js
mateoclarke: SO link on Frank's suggestion here:
https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

roseeichelmann: interesting, i absolutely will! thanks for this info :)

atd-vze/src/views/Crashes/Notes.js
```diff
@@ -1,42 +1,96 @@
...
+                    onChange={e => setNewNote(e.target.value)}
+                  />
+                </td>
+                <td>
+                  <Button
+                    type="submit"
+                    color="primary"
+                    onClick={(e) => {
+                      handleAddNoteClick();
+                    }}
```
mddilley, line 88: Small note here: This syntax is declaring an anonymous function that is called on click which then calls the `handleAddNoteClick` function. 

The anonymous function can be removed, and the code would run the same.
`<Button onClick={(e) => { handleAddNoteClick()}} />`
is equivalent to
`<Button onClick={handleAddNoteClick} />`

atd-vze/src/views/Crashes/Notes.js
mateoclarke: I don't see your update. Maybe try to push again?

roseeichelmann: image

You dont see this?

mateoclarke: Thank you, I do now. And I'm not sure if that chaining will work as is. Here is what I think you need:

```js
    }).then(response => {
      setNewNote("")
      refetch();
    }).catch(error => console.error(error));
```

As you have it written, the catch would run after `setNewNote()` errors (assuming its a Promise). You want it chained after the `.then()` errors.

roseeichelmann: Oooh okay, I was unsure about the chaining. that makes a lot more sense. thank you mateo!

atd-vze/src/views/Crashes/Notes.js
roseeichelmann: Yeah that's much more readable, thanks for this comment! not sure why i did that to begin with, probably just saw that syntax elsewhere and copied the pattern. still getting used to react functions

PR #1083, opened 2022-07-27
9518 recommendation component

roseeichelmann:
## Associated issues

Closes https://github.com/cityofaustin/atd-data-tech/issues/9518

## Testing
**URL to test:** 

This can be tested locally or as a Netlify preview https://deploy-preview-1083--atd-vze-staging.netlify.app/#/dashboard

**Steps to test:**

1. Click on Crashes tab.
2. Go to crash 18800909 (https://deploy-preview-1083--atd-vze-staging.netlify.app/#/crashes/18800909)
3. New Fatality Review Board Recommendations section should be sitting above notes section with data for Coordination Partner (Central Area Engineering Office), Status (Closed - No Action Recommended), and a note. 
4. Go to any other crash.
5. If there is no recommendations data for the crash the section should be empty except for the title and the labels "Coordination Partner:" and "Status:"

---
#### Ship list
- [x] Code reviewed

### DIFF
--- a/atd-vze/src/views/Crashes/Recommendations.js
+++ b/atd-vze/src/views/Crashes/Recommendations.js
@@ -16,6 +16,18 @@ const Recommendations = ({ crashId }) => {
   if (loading) return "Loading...";
   if (error) return `Error! ${error.message}`;
 
+  // if crash has data in recommendations table, display it
+  const displayData = (field) => {
+    if (data.recommendations[0]) {
+        if (field === "partner") {
+            return data.recommendations[0].atd__coordination_partners_lkp.description
+        } else if (field === "status") {
+            return data.recommendations[0].atd__recommendation_status_lkp.description
+        } else if (field === "note") {
+            return data.recommendations[0].text
+        }}
+  }
+
   const fieldConfig = recommendationsDataMap[0];
 
     return(
@@ -26,27 +38,19 @@ const Recommendations = ({ crashId }) => {
                     <tr style={{width: "100%"}}>
                         <td style={{width: "35%"}}>
                             <b>{fieldConfig.fields.coordination_partner_id.label}</b>
-                            {/* i feel like this part of my code where i check if there is a record for the current
-                            crash is repetive, would welcome more efficient methods so i dont have to write the
-                            same conditional three times */}
-                            {data.recommendations[0] &&
-                                data.recommendations[0].atd__coordination_partners_lkp.description
-                            }
+                            {displayData("partner")}
                         </td>
                         <td style={{width: "65%"}}>
                             <b>{fieldConfig.fields.recommendation_status_id.label}</b>
-                            {data.recommendations[0] &&
-                                data.recommendations[0].atd__recommendation_status_lkp.description
-                            }
+                            {displayData("status")}
                         </td>
                     </tr>
                     <tr>
                         <td colspan={2}>
-                            {data.recommendations[0] &&
-                                data.recommendations[0].text
-                            }
+                            {displayData("note")}
                         </td>
                     </tr>
+                
                 </Table>
             </CardBody>
             <CardFooter>
--- a/atd-vze/src/views/Crashes/Recommendations.js
+++ b/atd-vze/src/views/Crashes/Recommendations.js
@@ -16,41 +16,40 @@ const Recommendations = ({ crashId }) => {
   if (loading) return "Loading...";
   if (error) return `Error! ${error.message}`;
 
-  // if crash has data in recommendations table, display it
-  const displayData = (field) => {
-    if (data.recommendations[0]) {
-        if (field === "partner") {
-            return data.recommendations[0].atd__coordination_partners_lkp.description
-        } else if (field === "status") {
-            return data.recommendations[0].atd__recommendation_status_lkp.description
-        } else if (field === "note") {
-            return data.recommendations[0].text
-        }}
-  }
+  const fieldConfig = recommendationsDataMap;
+  const recommendation = data?.recommendations?.[0];
 
-  const fieldConfig = recommendationsDataMap[0];
+  // if crash has data in recommendations table, display it
+  const displayData = (data, {lookupOptions, key}) => {
+    if (recommendation) {
+        if (lookupOptions) {
+            return data[lookupOptions][key];
+        } else {
+            return data[key];
+        }
+    }
+  };
 
     return(
         <Card>
-            <CardHeader>{fieldConfig.title}</CardHeader>
-            <CardBody>
+            <CardHeader>Fatality Review Board Recommendations</CardHeader>
+            <CardBody style={{padding:"5px 20px 20px 20px" }}>
                 <Table>
-                    <tr style={{width: "100%"}}>
-                        <td style={{width: "35%"}}>
-                            <b>{fieldConfig.fields.coordination_partner_id.label}</b>
-                            {displayData("partner")}
+                    <tr style={{width: "100%", "border-top": "0px"}}>
+                        <td style={{width: "35%", "border-top": "0px"}}>
+                            <b>Coordination Partner: </b>
+                            {displayData(recommendation, fieldConfig.fields.coordination_partner_id)}
                         </td>
-                        <td style={{width: "65%"}}>
-                            <b>{fieldConfig.fields.recommendation_status_id.label}</b>
-                            {displayData("status")}
+                        <td style={{width: "65%", "border-top": "0px"}}>
+                            <b>Status: </b>
+                            {displayData(recommendation, fieldConfig.fields.recommendation_status_id)}
                         </td>
                     </tr>
                     <tr>
                         <td colspan={2}>
-                            {displayData("note")}
+                            {displayData(recommendation, fieldConfig.fields.text)}
                         </td>
                     </tr>
-                
                 </Table>
             </CardBody>
             <CardFooter>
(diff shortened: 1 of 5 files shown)


rebeccabrand (changes requested):
The interaction looks good to me! However, I find the fine line above the text to be distracting visually. @patrickm02L is this standard for the design system or could we review this?

Screen Shot 2022-07-29 at 3 28 19 PM

atd-vze/src/views/Crashes/Recommendations.js
```diff
@@ -0,0 +1,62 @@
...
+
+// declare fatality review board recommendations component
+const Recommendations = ({ crashId }) => {
+
+  // fetch data from database using graphQL query
+  const { loading, error, data, refetch } = useQuery(GET_RECOMMENDATIONS, {
+    variables: { crashId },
+  });
+
+  if (loading) return "Loading...";
```
johnclary, line 16: do y'all use a loading spinner elsewhere in VZ? would be a nice touch to add one.

atd-vze/src/views/Crashes/Recommendations.js
```diff
@@ -0,0 +1,62 @@
...
+  if (error) return `Error! ${error.message}`;
+
+  // if crash has data in recommendations table, display it
+  const displayData = (field) => {
+    if (data.recommendations[0]) {
```
johnclary, line 21: since you access it a lot, i suggest you assign `data.recommendations[0]` to a variable (as you do with `fieldConfig`)

atd-vze/src/views/Crashes/Recommendations.js
```diff
@@ -0,0 +1,62 @@
...
+  // fetch data from database using graphQL query
+  const { loading, error, data, refetch } = useQuery(GET_RECOMMENDATIONS, {
+    variables: { crashId },
+  });
+
+  if (loading) return "Loading...";
+  if (error) return `Error! ${error.message}`;
+
+  // if crash has data in recommendations table, display it
+  const displayData = (field) => {
```
johnclary, line 20: this function is being constructed on every render. it's a teen tiny performance hit, but you could wrap it in a `useCallback` hook to optimize it ([docs](https://reactjs.org/docs/hooks-reference.html#usecallback)).

atd-vze/src/views/Crashes/Recommendations.js
```diff
@@ -0,0 +1,62 @@
...
+    return(
+        <Card>
+            <CardHeader>{fieldConfig.title}</CardHeader>
+            <CardBody>
+                <Table>
+                    <tr style={{width: "100%"}}>
+                        <td style={{width: "35%"}}>
+                            <b>{fieldConfig.fields.coordination_partner_id.label}</b>
+                            {displayData("partner")}
+                        </td>
```
johnclary, line 42: i like that you're driving this with a config, but fwiw with just three cells i'd be fine hardcoding it (and it may be easier to grok)

```javascript
          <td style={{width: "35%"}}>
              <b>Fatality Review Board Recommendations</b>
               {data.recommendations[0].atd__coordination_partners_lkp.description}
          </td>
```

atd-vze/src/views/Crashes/recommendationsDataMap.js
```diff
@@ -0,0 +1,20 @@
+export const recommendationsDataMap = [
```
johnclary, line 1: unless you're planning to expand this and want to `map()` over it, it's fine to make this an object instead of an array.

atd-vze/src/views/Crashes/recommendationsDataMap.js
```diff
@@ -0,0 +1,20 @@
...
+      title: "Fatality Review Board Recommendations",
+  
+      fields: {
+        coordination_partner_id: {
+          label: "Coordination Partner: ",
+          lookupOptions: "atd__coordination_partners_lkp",
+        },
+        recommendation_status_id: {
+          label: "Status: ",
+          lookupOptions: "atd__recommendation_status_lkp"
```
johnclary, line 12: i notice you're not using `lookupOptions`. if you wanted to, you might rewrite `displayData` like this:

```javascript
  const displayData = (data, { lookupOptions, key }) => {
    if (lookupOptions) {
      return data[lookupOptions][key];
    } else {
      return data[key];
    }
  };
```
and add the key you want to access to you config

```javascript
export const recommendationsDataMap = {
  title: "Fatality Review Board Recommendations",

  fields: {
    coordination_partner_id: {
      label: "Coordination Partner: ",
      lookupOptions: "atd__coordination_partners_lkp",
      key: "description",
    },
    recommendation_status_id: {
      label: "Status: ",
      lookupOptions: "atd__recommendation_status_lkp",
      key: "description",
    },
    text: {
      label: "Note",
      key: "text",
    },
  },
};
```

and finally, the JSX

```javascript
 <td style={{width: "65%"}}>
      <b>{fieldConfig.fields.recommendation_status_id.label}</b>
      {displayData(recommendation, fieldConfig.fields.recommendation_status_id)}
  </td>
```

☝️ i haven't tested this, and what you have is already fine, but it gets at the idea of letting your config help you write less code, and centralizing as much of the hard-coded stuff as possible in one place (your config).

johnclary (approved):
I'm not set up to run VZ locally, but this looks great. Nice work! 🚢 🚢 🚢

atd-vze/src/views/Crashes/Recommendations.js
johnclary: ...and if there's a chance that `data` will be `undefined`, or that it will not have any recommendations, use nullish chaining

```
const rec = data?.recommendations?.[0]
```

atd-vze/src/views/Crashes/Recommendations.js
```diff
@@ -0,0 +1,62 @@
...
+        }}
+  }
+
+  const fieldConfig = recommendationsDataMap[0];
+
+    return(
+        <Card>
+            <CardHeader>{fieldConfig.title}</CardHeader>
+            <CardBody>
+                <Table>
```
johnclary, line 37: do you need to handle when there this is no recommendation? If so, conditionally returning a message inside the card body prob makes the most sense, along with only conditionally rendering the `Table` component if there is a recommendation.

atd-vze/src/views/Crashes/Recommendations.js
roseeichelmann: good call, definitely will assign it to a variable and use nullish chaining

atd-vze/src/views/Crashes/Recommendations.js
roseeichelmann: oo i havent heard of `useCallback` before, ill look into that

atd-vze/src/views/Crashes/Recommendations.js
roseeichelmann: yeah i was unsure what would be best practice. at this point it feels like a datamap isnt necessary for this component, am i wrong?

atd-vze/src/views/Crashes/recommendationsDataMap.js
roseeichelmann: okay this answers my previous question about whether a data map is necessary. i like this, its a looot cleaner. thank you!!

atd-vze/src/views/Crashes/Recommendations.js
roseeichelmann: next up i will be adding editing capabilities to the component, so the table component will need to be rendered even if there is no recommendation (https://github.com/cityofaustin/atd-data-tech/issues/9519). so i think it makes sense to just leave it as is for now?

atd-vze/src/views/Crashes/Recommendations.js
johnclary: 👍 either way is fine

atd-vze/src/views/Crashes/recommendationsDataMap.js
johnclary: yep, and if you had many more columns to deal with, that's when you'd want to start thinking about a custom table cell component, something like

```javascript
```

atd-vze/src/views/Crashes/Recommendations.js
johnclary: ah, I see. humble opinion, but the right approach is probably to move your labels into a proper header proper header row (reactstrap may have a component for it)
```html
 <thead><tr>Coordination Partner<th>Status</th><th>Note</th></tr></thead>
```
and in cases where there is no recommendation, render the "add recommendation" form in a row below it.

patrickm02L:
> The interaction looks good to me! However, I find the fine line above the text to be distracting visually. @patrickm02L is this standard for the design system or could we review this?
> 
> Screen Shot 2022-07-29 at 3 28 19 PM

Good catch. I believe the line above the text is the default design state. I agree we don't really need it, and can lose it.

roseeichelmann:
> 

Should I remove both lines or just the one above the row for Coordination Partner/Status?

roseeichelmann:
> The interaction looks good to me! However, I find the fine line above the text to be distracting visually. @patrickm02L is this standard for the design system or could we review this?
> 
> Screen Shot 2022-07-29 at 3 28 19 PM

Screenshot from 2022-08-01 12-41-38

image

Here are two different options, which do you like better from a design standpoint?

patrickm02L:
@roseeichelmann the one with the line underneath the Coordination Partners and Status check. It gives it a little bit of separation which helps prioritize action.

patrickm02L (approved):
Looks super clean! Ship it! 🤘

mateoclarke (approved):
🚢💨✨

PR #1087, opened 2022-07-28
Add SQL files describing the migration and creation (same action today) of `notes` and updating stored metadata

frankhereford:
There is not an issue related to this PR. This work was created from conversation during today's VZ Launch Party meeting.

This PR intends to capture the database changes which were a part of @roseeichelmann's recent work enhancing the crash page. Included here are:

* A schema definition in the form of a SQL create statement
* A migration-like file, indicating that that CREATE has been run in staging and production
* An updated copy of the `graphql-engine` metadata. This file had not previously been updated in the last 3 years, so I'd welcome consideration about if this is where I should have written it into the repo and if we want to begin capturing this. I am inclined to say 👍, we should.


patrickm02L (approved):
Thanks for knocking this out! 🤘

mateoclarke (approved):
Looks good! 🚢💨✨

Should we patch this to `production` branch though? Since these changes are already live?

roseeichelmann (approved):
Looks great! Thanks Frank for explaining this to me

PR #1088, opened 2022-07-28
Prevents drawing polygon from crashing VZV

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/9714

The issue was caused by the "copied" popover in the info window trying to attach itself to an element of the DOM that didn't exist. Updated that component to use hooks and a ref and that fixed the issue.

Screen Shot 2022-07-28 at 3 27 29 PM
Screen Shot 2022-07-28 at 3 27 58 PM

## Testing
**URL to test:** 
https://deploy-preview-1088--atd-vzv-staging.netlify.app/

**Steps to test:**
Go to map view, create a polygon. You should see the analysis and not a wsod.
If you click on a point in the map and then click the copy icon you should see a little "copied" popover pop up.

---
#### Ship list
- [x] Code reviewed

### DIFF
--- a/atd-vzv/src/views/map/InfoBox/InfoCard.js
+++ b/atd-vzv/src/views/map/InfoBox/InfoCard.js
@@ -1,5 +1,4 @@
-import * as React from "react";
-import { PureComponent } from "react";
+import React, { useRef, useState } from "react";
 import {
   Card,
   Table,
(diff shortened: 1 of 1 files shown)


atd-vzv/src/views/map/InfoBox/InfoCard.js
```diff
@@ -17,56 +16,52 @@ const StyledTableRow = styled.tr`
   }
 `;
 
-export default class InfoCard extends PureComponent {
-  render() {
-    const { content } = this.props;
+const InfoCard = ({ content }) => {
```
mddilley, line 19: I did a quick search about PureComponents since it has been a while since I've thought of them. The bug fix is much more valuable than the small performance optimization that a PureComponent would give here so this is great. [This](https://blog.logrocket.com/what-are-react-pure-functional-components/) is the blog post that I found about it.

mddilley (approved):
I tested this locally and in the deploy preview, and it works beautifully! Polygon tool restored! 🗺️ ✨

rebeccabrand (approved):
UX interaction looks good on my front! Thanks so much.

roseeichelmann (approved):
Looks great, thanks tilly <3

patrickm02L (approved):
Fantastic! 🤘

https://user-images.githubusercontent.com/5566727/182142899-04f928eb-708f-4777-9ac6-c01aa86c7135.mp4

mateoclarke (approved):
🚢💨✨

PR #1095, opened 2022-08-24
AFD & EMS DB change requests and Prefect Flows

mateoclarke:
## Intent & Associated issues

This PR has a sibling here: https://github.com/cityofaustin/atd-prefect/pull/38. Please review it also, if you don't mind. 🙏 

This PR adds / updates the `afd__incidents` and `ems__incidents` tables in the VZDB. It should address the following:

- Closes cityofaustin/atd-data-tech#9075
- Closes cityofaustin/atd-data-tech#9076
- Closes cityofaustin/atd-data-tech#9201
- Closes cityofaustin/atd-data-tech#7985
- Closes cityofaustin/atd-data-tech#7969
- Part 1 of cityofaustin/atd-data-tech#9077

It also adds the two Prefect Flows that will automate imports of AFD & EMS incident data. These flows are referenced in this companion PR to the `atd-prefect` repo: https://github.com/cityofaustin/atd-prefect/pull/38

## A note on triggers

This PR makes use of triggers to calculate derived data. I put these columns at the end of the table, but that is not really a convention, just something I did. These values can be overwritten, and the DB will not complain if you do so. However, next time any of the fields which contribute to the derived values changes, the entire set of derived values will be recalculated. 

⭐ I am very eager to discuss this pattern, and I'd really like to get it right, as I think it has a lot of potential to be reused both in VZ and in Moped. 

### Hard Mode

This is another one that would require a tester to spin up the dockerized agent for VZ. Please feel free to do so, and I would be happy to provide files to help anyone get a local VZ stack going.

### Easy Mode

Upon discussion with @patrickm02L & @chiaberry, we came to agreement that because this adds new tables and does not change existing ones, I could put this into production via Prefect, and the read replica could be used to validate its performance.

For both tables, `ems__incidents` and `afd__incidents`, I have imported 90 days of data. I'll do a full import (which isn't the fastest process) once this merges, and I install it all on `atd-data03`. 

Please review this data, if you wish, via the VZ Read Replica in the production database.

## Other thoughts

@patrickm02L Do you want an issue to document the actual install on `atd-data03` or should that just be part of getting to push the merge button?

---

#### Ship list
- [x] Code reviewed


atd-etl/prefect/afd_ems_incidents/.gitignore
```diff
@@ -0,0 +1 @@
+ems_reference.py
```
frankhereford, line 1: Remove this file

atd-vzd/schema/afd__incidents.sql
```diff
@@ -0,0 +1,65 @@
...
+after insert on public.afd__incidents
+for each row execute procedure afd_incidents_trigger();
+
+create trigger afd_incidents_trigger_update
+after update on public.afd__incidents
+for each row 
+WHEN (false 
+  or OLD.geometry IS DISTINCT FROM NEW.geometry
+  or old.ems_incident_numbers is distinct from new.ems_incident_numbers
+  or old.call_datetime is distinct from new.call_datetime)
```
frankhereford, line 64: This causes the trigger to only update on real changes to a certain set of columns. Same in the other SQL file.

atd-vzd/schema/afd__incidents.sql
```diff
@@ -0,0 +1,65 @@
...
+    ems_incident_number_2 integer default null,
+    call_date date default null,
+    call_time time default null 
+);
+
+CREATE OR REPLACE FUNCTION afd_incidents_trigger()
+  RETURNS trigger AS
+$$
+BEGIN
+  update afd__incidents set
```
frankhereford, line 31: When the trigger runs, it essentially does so only to have this query execute on the row in question.

atd-vzd/schema/ems__incidents.sql
```diff
@@ -0,0 +1,147 @@
...
+for each row 
+WHEN (false 
+  or old.geometry IS DISTINCT FROM new.geometry
+  or old.apd_incident_numbers is distinct from new.apd_incident_numbers
+  or old.mvc_form_extrication_datetime is distinct from new.mvc_form_extrication_datetime)
+execute procedure ems_incidents_trigger();
+
+
+
+-- Column Comment
```
frankhereford, line 105: Mateo had built out these comments, and I think they are really helpful. This would be a good pattern to carry forward when possible in new tables.

atd-vzd/schema/ems__incidents.sql
```diff
@@ -0,0 +1,147 @@
...
+	mvc_form_safety_devices text NULL,
+	mvc_form_seat_row_number text NULL,
+	mvc_form_vehicle_type text NULL,
+	mvc_form_weather text NULL,
+	pcr_additional_agencies text NULL,
+	pcr_transport_priority text NULL,
+	pcr_patient_acuity_initial text NULL,
+	pcr_patient_acuity_final text NULL,
+	unparsed_apd_incident_numbers text NULL,
+	apd_incident_numbers integer[] NULL,
```
frankhereford, line 49: We can get 0, 1, or more incident numbers here, so we're storing them in an array. Also keeping the unparsed value, just in case.

atd-vzd/schema/afd__incidents.sql
```diff
@@ -0,0 +1,65 @@
+DROP TABLE public.afd__incidents;
+
+-- Table Definition
+CREATE TABLE "public"."afd__incidents" (
+    id serial primary key,
+    incident_number integer,
+    crash_id integer default null,
+    unparsed_ems_incident_number text default null,
+    ems_incident_numbers integer[] default null,
```
frankhereford, line 9: We can get 0, 1, or more incident numbers here, so we're storing them in an array. Also keeping the unparsed value, just in case.

atd-etl/prefect/afd_ems_incidents/ems_incident_upload.py
```diff
@@ -0,0 +1,440 @@
...
+    """
+    Find the most recently updated file in the bucket. This will be the newest email.
+    :return: string
+    """
+
+    session = boto3.Session(
+        aws_access_key_id=AWS_ACCESS_KEY_ID,
+        aws_secret_access_key=AWS_SECRET_ACCESS_KEY,
+    )
+    s3 = session.resource("s3")
```
frankhereford, line 105: These `boto3` session make it thread safe which is needed if we want to do dask-enabled parallelism down the road.

atd-etl/prefect/afd_ems_incidents/afd_incident_upload.py
```diff
@@ -0,0 +1,279 @@
+#!/usr/bin/env python
+
+"""
+Name: AFD Incident Uploads
+Description: This flow uploads AFD Incident Response CSVs (AFD Contact: Gus). 
+    The data is emailed to atd-afd-incident-data@austinmobility.io daily ~ 3:30AM. From there it
```
frankhereford, line 6: This actually comes in around 6:30 AM local

atd-etl/prefect/afd_ems_incidents/afd_incident_upload.py
```diff
@@ -0,0 +1,279 @@
...
+import shutil
+import pandas
+from datetime import datetime, timedelta
+import psycopg2
+import psycopg2.extras
+from prefect.backend import get_key_value
+
+pp = pprint.PrettyPrinter(indent=4)
+
+kv_data = get_key_value(key="Vision Zero Development")
```
Charlie-Henry, line 29: Suggestion to put this in a task that way you won't have to re-register your flow if any of your key values get updated. Could also accept a parameter to change it from dev to production?

Something like this?
```
@task
def get_env_vars(env):
    if env == "prod":
       environment_variables = get_key_value(key="Vision Zero")
   else:
      environment_variables = get_key_value(key="Vision Zero Development")
    return environment_variables

```

atd-etl/prefect/afd_ems_incidents/afd_incident_upload.py
```diff
@@ -0,0 +1,279 @@
...
+
+with Flow("AFD Import ETL") as flow:
+    record_age_maximum = Parameter("record_age_maximum", default=False)
+    timestamp = get_timestamp()
+    newest_email = get_most_recent_email()
+    attachment_location = extract_email_attachment(newest_email)
+    uploaded_token = upload_attachment_to_S3(attachment_location, timestamp)
+    data = create_and_parse_dataframe(attachment_location)
+    upload_token = upload_data_to_postgres(data, record_age_maximum)
+    clean_up_token = clean_up(attachment_location, upstream_tasks=[upload_token])
```
Charlie-Henry, line 275: Your flow looks so clean compared to mine ✨

atd-etl/prefect/afd_ems_incidents/ems_incident_upload.py
```diff
@@ -0,0 +1,440 @@
...
+import psycopg2
+import psycopg2.extras
+from psycopg2 import Error
+from prefect.storage import GitHub
+from prefect.run_configs import UniversalRun
+from prefect.backend import get_key_value
+
+pp = pprint.PrettyPrinter(indent=4)
+
+kv_data = get_key_value(key="Vision Zero Development")
```
Charlie-Henry, line 32: Same as afd incident comment

atd-etl/prefect/afd_ems_incidents/ems_incident_upload.py
```diff
@@ -0,0 +1,440 @@
...
+        f"{location}/attachment.csv",
+        f"{EMS_S3_ARCHIVE_PREFIX}upload-{timestamp}.csv",
+    )
+    return True
+
+
+@task
+def create_and_parse_dataframe(location):
+    # Parse CSV into pandas dataframe
+    data = pandas.read_csv(f"{location}/attachment.csv", header=0, on_bad_lines='skip',  encoding='ISO-8859-1')
```
Charlie-Henry, line 155: Is it a big deal that we could be losing some data here with `on_bad_lines='skip'`? Based on Pandas docs it looks like it's probably dropping rows that have too many commas and therefore columns?

frankhereford: You're 100% right, yea, this will cause a records to be skipped if they are malformed. My experience is that ~ 1 / 20,000 records from EMS are malformed, and it's not really obvious if there is a way to repair them. 

That said, the low rate of malformed records makes me think this is an OK concession. We could emit something like a warning to make them more visible in Prefect, but I'd want to spin that off as a new issue. @patrickm02L

atd-vzd/schema/ems__incidents.sql
```diff
@@ -0,0 +1,147 @@
...
+  or old.apd_incident_numbers is distinct from new.apd_incident_numbers
+  or old.mvc_form_extrication_datetime is distinct from new.mvc_form_extrication_datetime)
+execute procedure ems_incidents_trigger();
+
+
+
+-- Column Comment
+COMMENT ON COLUMN "public"."ems__incidents"."pcr_key" IS 'Unique identifier for the patient care record in the EMS Data Warehouse. Can be used to uniquely identify records in this dataset';
+COMMENT ON COLUMN "public"."ems__incidents"."incident_date_received" IS 'The date that the incident was received by EMS. This could be the date that the EMS call taker took the call, or when it was transferred to EMS from another agency';
+COMMENT ON COLUMN "public"."ems__incidents"."incident_time_received" IS 'The time that the incident was received by EMS. This could be the time that the EMS call taker took the call, or when it was transferred to EMS from another agency';
```
Charlie-Henry, line 108: Probably need to confirm to be certain but looks like it is in local time

Charlie-Henry (approved):

atd-etl/prefect/afd_ems_incidents/afd_incident_upload.py
frankhereford: I'm not sure if we should make it switch between production or not, but I 100% agree that the KV retrieval should be pulled into a task. I'll update the PR with that - 🙏

atd-etl/prefect/afd_ems_incidents/afd_incident_upload.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+@task
+def clean_up(path):
+    # Clean up the temp location
+    shutil.rmtree(path)
+
+
+with Flow("AFD Import ETL") as flow:
+    record_age_maximum = Parameter("record_age_maximum", default=False)
```
frankhereford, line 268: This parameter is of type `int`, the default should be `90`.

atd-vzd/schema/afd__incidents.sql
```diff
@@ -0,0 +1,65 @@
...
+      left join atd_jurisdictions jurisdiction on (jurisdiction.jurisdiction_label = 'AUSTIN FULL PURPOSE')
+      where incidents.id = new.id),
+    location_id = (
+      select locations.location_id
+      from afd__incidents incidents
+      join atd_txdot_locations locations on (incidents.geometry && locations.shape and ST_Contains(locations.shape, incidents.geometry))
+      where incidents.id = new.id),
+    latitude = ST_Y(afd__incidents.geometry),
+    longitude = ST_X(afd__incidents.geometry),
+    ems_incident_number_1 = afd__incidents.ems_incident_numbers[1],
```
frankhereford, line 44: Arrays are 1-indexed in PG because some people like to watch the world burn. 🔥

atd-vzd/schema/ems__incidents.sql
frankhereford: It is, for sure. I bet I'm storing it in a UTC column. Thanks for pointing that out!

atd-vzd/schema/afd__incidents.sql
mddilley: Cool - thanks for the explanation here. I haven't worked with PG triggers too much so it is nice to see how to manage them likes this.

mddilley (approved):
All looks great to me! I agree that exploring triggers (especially with the conditions like used here) is worthwhile. I know that I have much to learn there. 😀 

Confirmed the following below by looking at the VZD read replica, and the values in the 90 day import look great too 👍 I also see the data in the two S3 buckets.

`afd__incidents` table
- `austin_full_purpose` column ✅ 
- `location_id` column ✅ 
- `latitude` and `longitude` columns ✅ 
- `ems_incident_number` ➡️  `ems_incident_number_1` and `ems_incident_number_2` with `-` removed ✅ 
- `call_datetime` ➡️  `call_date` and `call_time` ✅ 

`ems__incidents` table
- `austin_full_purpose` column ✅ 
- `location_id` column ✅ 
- `mvc_form_extrication_time` ➡️  `mvc_form_date` and `mvc_form_time` ✅ 
- `apd_incident_numbers` ➡️  `apd_incident_number_1` and `apd_incident_numbers_2` ✅ 

Both
- `crash_id` ✅

patrickm02L (approved):
Co-signing based on the Dev reviews. 🤘

atd-vzd/schema/afd__incidents.sql
chiaberry: rude

atd-vzd/schema/afd__incidents.sql
chiaberry: same -- thanks for the explanation. Why is it WHEN (false or x or y or z) and not just WHEN (x or y or z)? Convention?

atd-etl/prefect/afd_ems_incidents/afd_incident_upload.py
```diff
@@ -0,0 +1,279 @@
...
+DB_DATABASE = None
+AWS_ACCESS_KEY_ID = None
+AWS_SECRET_ACCESS_KEY = None
+AFD_S3_SOURCE_BUCKET = None
+AFD_S3_ARCHIVE_BUCKET = None
+AFD_S3_SOURCE_PREFIX = None
+AFD_S3_ARCHIVE_PREFIX = None
+
+# Retrieve the db configuration
+if False:
```
chiaberry, line 45: I'm confused by this, why if False?

atd-etl/prefect/afd_ems_incidents/afd_incident_upload.py
chiaberry: setting it as False gets a full import? so we default to a full import?

atd-etl/prefect/afd_ems_incidents/afd_incident_upload.py
chiaberry: oh, is this so someone can toggle between getting the config out of the kv store instead of local environment?

chiaberry (approved):

atd-vzd/schema/afd__incidents.sql
frankhereford: @chiaberry wrote: 
> Why is it WHEN (false or x or y or z) and not just WHEN (x or y or z)? Convention?

This is really just a Frank thing, but I like each of the three conditions that make up the larger `where` clause to line-up and have the same keywords around them. I think this better indicates that they are part of a working-group, and you can also comment them in and out without having to worry about if you have the right keywords.

I don't see this used by many folks except myself, so if you all think it's not a great pattern, I would be happy to change it.

atd-etl/prefect/afd_ems_incidents/afd_incident_upload.py
frankhereford: Thank you for catching that! I updated the PR to have a default of `0` for both flows' value for the `record_age_maximum` int parameter. This will make the default behavior to do a full import.

PR #1100, opened 2022-09-09
Adds street labels on top of nearmap tiles

johnclary:
## Associated issues
[#10149](https://github.com/cityofaustin/atd-data-tech/issues/10149)

PR #1106, opened 2022-09-22
Update the MapGL configuration when in edit mode for a crash location

frankhereford:
## Associated issues

This PR intends to close [issue 10174](https://github.com/cityofaustin/atd-data-tech/issues/10174).

## Testing
**URL to test:** 😞 👎 

This one is a trick to test. Nearmap checks the `Referrer` header for a known set of domains, and I don't think our netlify PR builds or `localhost` can be one of them. At least not without first working with CTM, which I don't think we should do for this.

### Testing approaches

* It may be worth comparing this to a recent PR #1100 by @johnclary which actually was where this work and technique were first completed. This PR borrows almost entirely from his work. His previous PR, linked above, has made it to production. You be the judge, but I think a code review can be enough for this minor change.

* Alternatively, you can fire up the app locally and change the raster tile layer definition to use this XYZ tileserver `https://tiles.frankhereford.io/tile/{z}/{x}/{y}.png` which does not check referrers. Then overload the `isDev` boolean to see how things work. You still have to imagine what the imagery would look like as the tile layer just the OSM, but you **can** see that it pulls tiles as expected.

* Go `l33t hax0r` and [overload your referrer header](https://chrome.google.com/webstore/detail/referer-control/hnkcfpcejkafcihlgbojoidoihckciin?hl=en) to convince NearMap that you're coming from the production VZ app. We have a valid, paid key, this isn't a bad thing.

## Additional Considerations
* @johnclary @mddilley I bumped a version in a unrelated style definition, and I wanted to double check with you that that is a good idea.
* ~Everyone, I am removing the toggle between satellite & street layers. It's not working. @patrickm02L, I recommend this become an issue. Do you agree? Let me know, I'd be happy to write it and offer up solutions.~ Nope, it works fine in production.

---

#### Ship list
- [x] Code reviewed

### DIFF
--- a/atd-vze/src/helpers/map.js
+++ b/atd-vze/src/helpers/map.js
@@ -9,6 +9,8 @@ export const LOCATION_MAP_CONFIG = {
       type: "raster",
       tiles: [
         `https://api.nearmap.com/tiles/v3/Vert/{z}/{x}/{y}.jpg?apikey=${NEARMAP_KEY}`,
+        // testing basic raster tile server for TX, US
+        // "https://tiles.frankhereford.io/tile/{z}/{x}/{y}.png",
       ],
       tileSize: 256,
     },
--- a/atd-vze/src/helpers/map.js
+++ b/atd-vze/src/helpers/map.js
@@ -10,7 +10,7 @@ export const LOCATION_MAP_CONFIG = {
       tiles: [
         `https://api.nearmap.com/tiles/v3/Vert/{z}/{x}/{y}.jpg?apikey=${NEARMAP_KEY}`,
         // testing basic raster tile server for TX, US
-        // "https://tiles.frankhereford.io/tile/{z}/{x}/{y}.png",
+        //`https://tiles.frankhereford.io/tile/{z}/{x}/{y}.png`,
       ],
       tileSize: 256,
     },
--- a/atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
+++ b/atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
@@ -202,7 +202,7 @@ class CrashEditCoordsMap extends Component {
             <Pin size={40} color={pinColor} isDragging={isDragging} animated />
           </Marker>
           <MapStyleSelector>
-              <ButtonGroup className="float-right">
+            <ButtonGroup className="float-right">
               <Button
                 active={mapStyle === "satellite-streets"}
                 id="satellite-streets"
--- a/atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
+++ b/atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
@@ -6,6 +6,8 @@ import MapGL, {
   Marker,
   NavigationControl,
   FullscreenControl,
+  Source,
+  Layer,
 } from "react-map-gl";
 import "mapbox-gl/dist/mapbox-gl.css";
 import Geocoder from "react-map-gl-geocoder";
@@ -20,6 +22,8 @@ import { Button, ButtonGroup } from "reactstrap";
 import Pin from "./Pin";
 import { setPinColor } from "../../../styles/mapPinStyles";
 import { CrashEditLatLonForm } from "./CrashEditLatLonForm";
+
+import { LOCATION_MAP_CONFIG } from "../../../helpers/map";
 
 const TOKEN = process.env.REACT_APP_MAPBOX_TOKEN;
 
@@ -151,6 +155,7 @@ class CrashEditCoordsMap extends Component {
       isDragging,
     } = this.state;
     const geocoderAddress = this.props.mapGeocoderAddress;
+    const isDev = window.location.hostname === "localhost";
 
     return (
       <div>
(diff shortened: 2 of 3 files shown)


atd-vze/src/helpers/map.js
```diff
@@ -9,6 +9,8 @@ export const LOCATION_MAP_CONFIG = {
       type: "raster",
       tiles: [
         `https://api.nearmap.com/tiles/v3/Vert/{z}/{x}/{y}.jpg?apikey=${NEARMAP_KEY}`,
+        // testing basic raster tile server for TX, US
+        //`https://tiles.frankhereford.io/tile/{z}/{x}/{y}.png`,
```
frankhereford, line 13: Feels a little strange to leave a comment in the code here. Do you think this is useful or better removed?

atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
```diff
@@ -159,7 +164,11 @@ class CrashEditCoordsMap extends Component {
           ref={this.mapRef}
           width="100%"
           height="350px"
-          mapStyle={`mapbox://styles/mapbox/${mapStyle}-v9`}
+          mapStyle={
+            isDev
+              ? "mapbox://styles/mapbox/satellite-streets-v11"
+              : LOCATION_MAP_CONFIG.mapStyle
+          }
```
frankhereford, line 171: Lifted directly from @johnclary's work - thank you for that!

atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
```diff
@@ -180,31 +189,18 @@ class CrashEditCoordsMap extends Component {
...
+              <Source {...LOCATION_MAP_CONFIG.sources.aerials} />
+              <Layer {...LOCATION_MAP_CONFIG.layers.aerials} />
+              {/* show street labels on top of other layers */}
+              <Layer {...LOCATION_MAP_CONFIG.layers.streetLabels} />
+            </>
+          )}
           <Marker latitude={markerLatitude} longitude={markerLongitude}>
             <Pin size={40} color={pinColor} isDragging={isDragging} animated />
           </Marker>
-          <MapStyleSelector>
```
frankhereford, line 186: This is the removed toggle between base layers that was not working. I recommend we backlog this via a new issue. @patrickm02L

atd-vze/src/views/Crashes/Maps/CrashMap.js
```diff
@@ -84,7 +84,7 @@ export default class CrashMap extends Component {
         height="100%"
         mapStyle={
           isDev
-            ? "mapbox://styles/mapbox/satellite-streets-v9"
+            ? "mapbox://styles/mapbox/satellite-streets-v11"
```
frankhereford, line 87: @johnclary Here's that "unrelated" version bump - does this seem good to set to 11, since that's what used in the other style definitions for the display and edit versions of the map?

atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
```diff
@@ -151,6 +155,7 @@ class CrashEditCoordsMap extends Component {
       isDragging,
     } = this.state;
     const geocoderAddress = this.props.mapGeocoderAddress;
+    const isDev = window.location.hostname === "localhost";
```
frankhereford, line 158: You can set this value manually to `false` to cause the conditional style application later to be applied - useful for testing if you want to see it working.

frankhereford:
@patrickm02L wrote:
> Does Nearmap have a street view? If not, would it make sense to keep the street view from Mapbox? We should ask Xavier or Joel how often then use the Street view in Edit mode.

You know, I've made a mistake here. This wasn't working on my local deployment, but I see that that is a result of my deployment, not the app itself. The feature is working in production, and I should not have removed it. I'm going to revert that commit and amend the PR. Thank you!

atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
frankhereford: This should not have come out; I'm going to restore it.

patrickm02L (approved):
Looks good! Thanks for the demo. Glad we could keep the streetmap view in edit mode! 🤘

atd-vze/src/helpers/map.js
johnclary: doesn't bother me!

atd-vze/src/views/Crashes/Maps/CrashMap.js
johnclary: yep 100% - i think it's safe to assume we want Mapbox's latest and greatest opinions on basemap styles. glad you caught that!

johnclary (approved):
Have not tested, but it passes the eyeball test! 🚢

atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
johnclary: pretty sure i copied Chia...

mddilley (approved):
Looks great! Thanks for updating this - I know the VZ team will love it. I also did not test this or go `1337 h4x0r`❤️, but I did compare to the last PR and approve!

chiaberry (approved):

PR #1107, opened 2022-09-22
9995 refactor notes component

roseeichelmann:
## Associated issues
Closes cityofaustin/atd-data-tech/issues/9995

This PR refactors Location Notes & Crash Notes to use the same reusable Notes component. Prior to this PR there were separate Notes components for Crash Notes and Location Notes which was unnecessary because they were doing the exact same thing aside from using different variables. The trick was finding a way for the query/mutation calls in the new Notes component to be reusable for both Crashes and Locations despite having slightly different variables (the location_notes table uses a location_id while crash_notes uses a crash_id).

## Testing
**URL to test:** 
https://deploy-preview-1107--atd-vze-staging.netlify.app/#/dashboard

**Steps to test:**
1. Go to any crash details page.
2. Add a note, edit it, cancel, edit again, save, then delete the note. Ensure it's all working as expected.
3. Repeat for any location details page.

---
#### Ship list
- [x] Code reviewed


atd-vze/src/Components/Notes/Notes.js
```diff
@@ -10,17 +10,18 @@ import {
...
 import moment from "moment";
-import { notesDataMap } from "./notesDataMap";
-import {
+import { notesDataMap } from "./notesDataMap.js";
+import { useAuth0, isReadOnly } from "../../auth/authContext";
+
+// declare a notes component
+const Notes = ({
+  crashLocationId,
+  tableName,
```
roseeichelmann, line 19: The Notes component accepts tableName and various query/mutation functions as parameters because the Location Notes and Crash Notes pull from different tables in the DB.

roseeichelmann: I'm curious if anyone would have approached this differently or can think of a more obvious way to make the component reusable between crashes/locations.

atd-vze/src/Components/Notes/Notes.js
```diff
@@ -10,17 +10,18 @@ import {
...
 } from "reactstrap";
 import moment from "moment";
-import { notesDataMap } from "./notesDataMap";
-import {
+import { notesDataMap } from "./notesDataMap.js";
+import { useAuth0, isReadOnly } from "../../auth/authContext";
+
+// declare a notes component
+const Notes = ({
+  crashLocationId,
```
mddilley, line 18: Just thinking out loud here - this id name could be updated to be more general like `recordId` or something - not a huge deal at all. Naming is always tricky, and I am no authority on it!

mddilley (approved):
Nice refactor! I'm still able to add, edit, and remove both crash and location notes. I love seeing all that code cut out ✂️ and watching a component grow up 🌳.  🎉

atd-vze/src/Components/Notes/Notes.js
roseeichelmann: Yeah coming up with that variable name was tricky, I like your idea of making it more general!

atd-vze/package-lock.json
```diff
@@ -1,12 +1,12 @@
...
   "name": "atd-vz-data",
-  "version": "1.25.0",
+  "version": "1.27.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "atd-vz-data",
-      "version": "1.25.0",
+      "version": "1.27.1",
```
johnclary, line 9: out of curiosity—are y'all skipping versions?

chiaberry: I've found that if no one has npm-installed between versions, the package-lock won't update and so its version in github doesn't match the package.json

chiaberry (approved):

atd-vze/src/Components/Notes/Notes.js
```diff
@@ -197,7 +197,7 @@ const Notes = ({ crashId }) => {
               </tr>
             )}
             {/* iterate through each row in notes table */}
-            {data.crash_notes.map(row => {
+            {data[tableName].map(row => {
```
tillyw, line 200: :tada:

patrickm02L (approved):
Good to go! 🤘

PR #1109, opened 2022-09-22
10151 alphabetize partners

roseeichelmann:
## Associated issues
closes cityofaustin/atd-data-tech/issues/10151

This PR alphabetizes the Coordination Partner dropdown in the FRB Recommendations section while keeping "None" at the top.

## Testing
**URL to test:** 
https://deploy-preview-1109--atd-vze-staging.netlify.app/#/crashes/18857822

**Steps to test:**
1.  Click on the Coordination Partner dropdown and ensure that the list starts with "None" and then is ordered alphabetically A-Z.

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vze/src/views/Crashes/Recommendations/Recommendations.js
+++ b/atd-vze/src/views/Crashes/Recommendations/Recommendations.js
@@ -68,6 +68,12 @@ const Recommendations = ({ crashId }) => {
       .catch(error => console.error(error));
   };
 
+  // Function to alphabetize an array by a certain field
+  const alphabetizeArray = (array, field) => {
+    array.sort((a, b) => a[field].localeCompare(b[field]));
+    return array;
+  };
+
   return (
     <Card>
       <CardHeader>Fatality Review Board Recommendations</CardHeader>
@@ -89,7 +95,10 @@ const Recommendations = ({ crashId }) => {
                     onOptionClick={
                       doesRecommendationRecordExist ? onEdit : onAdd
                     }
-                    options={data.atd__coordination_partners_lkp}
+                    options={alphabetizeArray(
+                      data.atd__coordination_partners_lkp,
+                      "description"
+                    )}
                     field={"coordination_partner_id"}
                   />
                 </div>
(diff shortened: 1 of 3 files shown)


atd-vze/src/views/Crashes/Recommendations/Recommendations.js
```diff
@@ -68,6 +68,12 @@ const Recommendations = ({ crashId }) => {
       .catch(error => console.error(error));
   };
 
+  // Function to alphabetize an array by a certain field
+  const alphabetizeArray = (array, field) => {
+    array.sort((a, b) => a[field].localeCompare(b[field]));
+    return array;
+  };
+
```
mddilley, line 76: An alternative to sorting in the JavaScript could be to add an `order_by` clause to the `atd__coordination_partners_lkp` part of the `GET_RECOMMENDATIONS` query. This would serve up the ordered results directly to the front end. I'll let others weigh in, but I think that less code is a perk. 

You can try that out in the [Hasura console](https://vzd-staging.austinmobility.io/console/api-explorer) - GraphiQL is super helpful with building queries like this.

mddilley (approved):
Test steps check out! Looks great! ✨💪

atd-vze/src/views/Crashes/Recommendations/Recommendations.js
roseeichelmann: Thanks for the feedback Mike! That's actually what I originally tried doing but it wasn't working so I switched gears, but I just attempted it again a little differently and got it to work! I'll push those changes because I agree, less code is a perk.

mddilley: Anytime - glad it worked out! 🎉

roseeichelmann: If you're curious, what I was attempting to do the first go around was adding the `order_by` clause under the `recommendations` part of the  query, so I was getting an array of recommendations sorted alphabetically by their coordination partners rather than sorting the query results for `atd__coordination_partners_lkp` directly.

johnclary (approved):
🌬️ ⛵

PR #1114, opened 2022-10-04
replace moment.js with date-fns

roseeichelmann:
## Associated issues
closes cityofaustin/atd-data-tech/issues/9592

This issue removes the deprecated moment.js dependency in VZE and replaces it with [date-fns](https://date-fns.org/). I am confident that my changes are correct throughout most of the codebase _except_ two instances in which I was unable to test if they worked which I describe in the comments below. Any feedback on those would be greatly appreciated 🙏

## Testing
**URL to test:** 
This can be tested locally or with a Netlify preview: https://deploy-preview-1114--atd-vze-staging.netlify.app/#/dashboard

**Steps to test:**
1. Make sure the Fatality/Years of Life Lost/Serious Injuries widgets look right on the dashboard.
2. Click on Crashes tab, make sure dates are formatted correctly (Crash Date column, date filter components)
3. Click on a Crash, make sure all dates are there and formatted. (Notes section Date, Details section with rows such as Last Updates, Crash Date, Crash Time)
4. Click on Locations tab and click on a Location. Make sure all dates are there and formatted correctly (Last Update row in Details section, Date in Notes section, Crash Date column in CR3 Crashes section, Crash Date column in Non-CR3 section). _**I have additional code questions having to do with the LocationMap in my comments**_
5. Click on Changes tab, ensure dates are correct (date filter components and Create Date column).
6. Click on Tools tab and Create Crash Record, Crash Time and Crash Date should be populated with current time/date.
7. Click on Users tab, ensure Created and Last Login columns have formatted dates.
8. If possible click on a User and make sure those dates are correctly formatted, ~~**_however when I try to click on any User the page loads forever, so I couldn't test this. This may be a bug we need to fix? More in comments_**~~ Frank confirmed that it does look formatted correctly, however I'm going to investigate why this doesn't load for me

---
#### Ship list
- [x] Code reviewed

### DIFF
--- a/atd-vze/src/Components/GridExportDataButton.js
+++ b/atd-vze/src/Components/GridExportDataButton.js
@@ -119,7 +119,7 @@ const GridExportDataButton = ({
 
     return cleanedAndFlattenedData;
   };
-  // fa-2x ml-2 mt-1
+
   return (
     <>
       <Button color="primary" onClick={toggleModalAndExport}>
--- a/atd-vze/src/Components/GridExportDataButton.js
+++ b/atd-vze/src/Components/GridExportDataButton.js
@@ -1,7 +1,7 @@
 import React, { useState } from "react";
 import { withApollo } from "react-apollo";
 import { useLazyQuery } from "@apollo/react-hooks";
-import moment from "moment";
+import { formatISO } from "date-fns";
 import { CSVLink } from "react-csv";
 import {
   Button,
@@ -21,7 +21,12 @@ import { AppSwitch } from "@coreui/react";
 
 const exportWarningLimit = 4000;
 
-const GridExportDataButton = ({ query, columnsToExport, totalRecords, label }) => {
+const GridExportDataButton = ({
+  query,
+  columnsToExport,
+  totalRecords,
+  label,
+}) => {
   const [isModalOpen, setIsModalOpen] = useState(false);
 
   // Use .queryCSV to insert columnsToExport prop into query
@@ -114,7 +119,7 @@ const GridExportDataButton = ({ query, columnsToExport, totalRecords, label }) =
 
     return cleanedAndFlattenedData;
   };
-// fa-2x ml-2 mt-1
+  // fa-2x ml-2 mt-1
   return (
     <>
       <Button color="primary" onClick={toggleModalAndExport}>
@@ -169,7 +174,7 @@ const GridExportDataButton = ({ query, columnsToExport, totalRecords, label }) =
             <CSVLink
               className=""
               data={formatExportData(data[query.table])}
-              filename={query.table + moment(Date.now()).format()}
+              filename={query.table + formatISO(Date.now())}
             >
               <Button color="primary" onClick={toggleModal}>
                 Save
--- a/atd-vze/src/views/Crashes/crashGridTableParameters.js
+++ b/atd-vze/src/views/Crashes/crashGridTableParameters.js
@@ -131,59 +131,70 @@ export const crashGridTableAdvancedFilters = {
   groupInjuries: {
     icon: "cab",
     label: "Deaths & Injuries",
-    filters: [{
+    filters: [
+      {
         id: "dni_atd_deaths",
         label: "ATD Fatality Crashes",
         filter: {
-          where: [{
-            or: {
-              atd_fatality_count: "_gt: 0",
+          where: [
+            {
+              or: {
+                atd_fatality_count: "_gt: 0",
+              },
             },
-          }, ],
+          ],
         },
       },
       {
         id: "dni_cris_deaths",
         label: "CRIS Fatality Crashes",
         filter: {
-          where: [{
-            or: {
-              death_cnt: "_gt: 0",
+          where: [
+            {
+              or: {
+                death_cnt: "_gt: 0",
+              },
             },
-          }, ],
+          ],
         },
       },
       {
         id: "dni_apd_deaths",
         label: "APD Confirmed Fatality Crashes",
         filter: {
-          where: [{
-            or: {
-              apd_confirmed_death_count: "_gt: 0",
+          where: [
+            {
+              or: {
+                apd_confirmed_death_count: "_gt: 0",
+              },
             },
-          }, ],
+          ],
         },
       },
       {
         id: "dni_serious_injuries",
         label: "Suspected Serious Injury Crashes",
         filter: {
-          where: [{
-            or: {
-              sus_serious_injry_cnt: "_gt: 0",
+          where: [
+            {
+              or: {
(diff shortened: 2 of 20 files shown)


atd-vze/src/views/Crashes/crashGridTableParameters.js
```diff
@@ -131,71 +131,84 @@ export const crashGridTableAdvancedFilters = {
   groupInjuries: {
     icon: "cab",
     label: "Deaths & Injuries",
-    filters: [{
+    filters: [
+      {
         id: "dni_atd_deaths",
```
roseeichelmann, line 136: This is all formatting with prettier 💅

atd-vze/src/views/Locations/LocationMap.js
```diff
@@ -81,13 +81,15 @@ export default class LocationMap extends Component {
...
     this.setState({ viewport });
   };
 
-  getLatestAerialTimestamp = (timestampArray) => timestampArray.slice(-1)[0];
+  getLatestAerialTimestamp = timestampArray => timestampArray.slice(-1)[0];
 
-  convertNearMapTimeFormat = (date) => moment(date).format("MM/DD/YYYY");
+  convertNearMapTimeFormat = date => {
+    format(parseISO(date), "MM/dd/yyyy");
+  };
```
roseeichelmann, line 92: I'm assuming here that `date` is a string, hence the `parseISO` function for parsing strings, but if it's not then this won't work. I couldn't confirm the data type despite all my best console.log efforts because the code was unreachable? I'm not sure how to even get the `convertNearMapTimeFormat` function to run. Does anyone have any insight on what data type `date` is?

atd-vze/src/views/Users/User.js
```diff
@@ -67,17 +67,17 @@ const User = () => {
...
+      switch (attributeFormat) {
         case "string":
           formattedValue = user[key];
           break;
         case "bool":
           formattedValue = !!user[key] ? "Yes" : "No";
           break;
         case "time":
-          formattedValue = moment(user[key]).format("MM/DD/YYYY, h:mm:ss a");
+          formattedValue = format(parseISO(user[key]), "MM/dd/yyyy, h:mm:ss a");
```
roseeichelmann, line 80: This is another instance where I couldn't test if my change worked because there's a bug in VZE where if I click on a user from the Users tab, the page just loads forever. Once again I'm assuming `user[key]` is a string, I'd like to confirm this but I'm not sure how 🤔

roseeichelmann: Update:  @frankhereford checked and it looks right on his end. Still need to figure out why these pages don't load for me, could have something to do with permissions even though I do have admin permissions?

atd-vze/src/views/Crashes/crashGridTableParameters.js
chiaberry: 💅🏼

atd-vze/src/views/Locations/LocationMap.js
mddilley: I took myself down a little rabbit hole here because I couldn't remember how this timestamp is retrieved. The format that comes back here is `/Date(1416873600000)/` which moment.js really surprisingly handles well 🤷, but it is a milliseconds timestamp wrapped with some extra stuff (see the link below).

For `date-fns`, TIL that you can escape characters in the source time string using single quotes when using the [parse](https://date-fns.org/v2.29.3/docs/parse) method. So, I think this works:
```
format(parse(date, "'/Date('T')/'", new Date()), "MM/dd/yyyy");
```

Could you add this link (https://docs.nearmap.com/display/ND/Nearmap+TMS+Integration#NearmapTMSIntegration-Attributes) as a comment somewhere in this code? I should have done that when this code was first written, but I think that I was too excited to get the timestamp for the aerial imagery from the API at all. 😄

atd-vze/src/Components/GridExportDataButton.js
```diff
@@ -114,7 +119,7 @@ const GridExportDataButton = ({ query, columnsToExport, totalRecords, label }) =
 
     return cleanedAndFlattenedData;
   };
-// fa-2x ml-2 mt-1
+  // fa-2x ml-2 mt-1
```
mddilley, line 122: This could be removed since it looks like it was left behind sometime in the past. 🧹 🙏

atd-vze/src/views/Users/User.js
mddilley: Looks great to me as well 👍 Also noticed that the AM is now capitalized just like it is elsewhere in the app so that is a nice improvement.
Screen Shot 2022-10-11 at 11 56 39 AM
Screen Shot 2022-10-11 at 11 56 33 AM

mddilley (commented):
This looks great! I compared this branch to staging, and everything looks the same or better. 🙌  I left one comment about the NearMap timestamp format question that you had. I'll approve after that small change which all stems from my old code. 🙏

atd-vze/src/views/Locations/LocationMap.js
mddilley: Also forgot to mention that I tested this locally by replacing the API call with using the helper on a sample return (like `/Date(1416873600000)/`) and then setting that value to state

mddilley (approved):
Awesome stuff!! Thank you for adding in those comments. 🙏

roseeichelmann:
> Awesome stuff!! Thank you for adding in those comments. pray

@mddilley is the aerial timestamp actually showing up for you on the map? its not for me so i think there may still be a problem with the parsing/formatting:

image

roseeichelmann:
Versus what it looks like in production:

image

mddilley:
@roseeichelmann Really good call on that! That is definitely confusing but expected. The API call that is made to get that timestamp uses the same NearMap API token as the aerial tiles that don't work locally or in deploy previews. The dynamic Netlify preview urls aren't on the [approved referrer list](https://docs.nearmap.com/display/ND/Manage+API+Applications) to use that token, and CTM manages that for us so we can't easily update it ourselves.

There is a issue tracking this because it affects Moped as well. https://github.com/cityofaustin/atd-data-tech/issues/6008#issuecomment-1247179900

The best we can do for now is use a mock date locally to test out that the helpers format it correctly and show it in the UI.

roseeichelmann:
@mddilley Ooo okay that makes so much sense. I tested locally with a mock date and it works, thanks so much for your insight on this!

patrickm02L (approved):
Pretty! 🤘

frankhereford (approved):
🚢 🚀 💯 

Rose, thank you for getting moment out of there! This is awesome work!

I've reviewed the code and tested the deployment provided in the initial comment. Everything looks ship-shape to me. Thank you again!

PR #1118, opened 2022-10-17
10462 migrate recommendation partners

roseeichelmann:
## Associated issues

Closes cityofaustin/atd-data-tech/issues/10462

This PR includes a script that migrates `recommendation_id`'s and their corresponding `partner_id`'s from the `recommendations` table to the new `recommendations_partners` table. The pattern for this code was copied from @johnclary 's [work](https://github.com/cityofaustin/atd-moped/blob/10474-jc-team-table-edits/moped-toolbox/migrate_personnel/run.py) in a moped migration, so thank you to John for that 🙏 This will be the final step in changing the database relationship to many to many between recommendations and partners.

I have run it with the staging database and it worked just as expected, so whenever it comes time to update prod, this script is ready to go!

- [ ] A clean-up todo after this is done is to delete the coordination_partner_id column in the recommendations table.

#### Ship list
- [x] Code reviewed
- [x] PM reviewed

### DIFF
--- a/atd-toolbox/coordination_partners_migration/secrets.py
+++ b/atd-toolbox/coordination_partners_migration/secrets.py
@@ -1,0 +1,10 @@
+HASURA_AUTH = {
+  "hasura_graphql_endpoint": {
+    "staging": "<redacted>",
+    "prod": "<redacted>",
+  },
+  "hasura_graphql_admin_secret": {
+    "staging": "<redacted>",
+    "prod": "<redacted>",
+  },
+}
--- a/atd-toolbox/coordination_partners_migration/coordination_partners_migration.py
+++ b/atd-toolbox/coordination_partners_migration/coordination_partners_migration.py
@@ -1,0 +1,95 @@
+import argparse
+import requests
+
+from secrets import HASURA_AUTH
+
+def make_hasura_request(*, query, variables, env):
+  """Fetch data from hasura
+  Args:
+    query (str): the hasura query
+    env (str): the environment name, which will be used to acces secrets
+  Raises:
+    ValueError: If no data is returned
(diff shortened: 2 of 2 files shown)


johnclary (approved):
Have not tested, but it looks good! 🚢

patrickm02L (approved):
Co-signing based on what I heard in Dev Sync 10/18, and John's review! 🤘

atd-toolbox/coordination_partners_migration/secrets.py
```diff
@@ -0,0 +1,10 @@
+HASURA_AUTH = {
+  "hasura_graphql_endpoint": {
+    "staging": "<redacted>",
+    "prod": "<redacted>",
```
frankhereford, line 4: Thank you for redacting these, and I am sorry for being so late to the party in time to leave you a helpful review. In the future, if you have a file that contains secrets, I'd consider the pattern of putting that file into your `.gitignore` file so it just can't get put into the repo. Then, add a file like `secrets.py_template` with the shape of the file but the secrets blanked out so people know what the file they need looks like but have to provide their own secrets.

With this pattern, it's too easy for one of us to unredact it and then push it up to the repo.

chiaberry: thats helpful feedback Frank -- we have this pattern of a secrets file in moped-toolbox and I get paranoid that im accidentally going to push up the secrets to the repo when I'm testing against staging or prod

roseeichelmann: thats a great tip Frank and it makes a lot of sense. i will definitely take up that pattern in the future!

PR #1145, opened 2022-12-23
Add a fallback value to By Population year ratios

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/11056

This PR sets a fallback value for the to yearly ratios in the **By Population** visualization similar to how we fall back to `No data available` when you open the info icon in this visualization. It also adds the 2023 Austin Full Purpose figure and makes sure the other ones use that column as well.

## Testing
**URL to test:** 
https://deploy-preview-1145--atd-vzv-staging.netlify.app/

**Steps to test:**
**Hard to test this since we've added an estimate for 2023 in the population config**
I tested this locally by updating the variable `yearRatio` on [line 122 of CrashesByPopulation.js](https://github.com/cityofaustin/atd-vz-data/blob/0f5010c253b5020fba4a7fcc4bc8bd78e6d44b3c/atd-vzv/src/views/summary/CrashesByPopulation.js#L122) with `undefined`. That will make every ratio figure in the visualization turn into `-`.

A preview of what this will look like can be seen in https://github.com/cityofaustin/atd-data-tech/issues/11056#issue-1509589925 (look for the **Possible fix** heading).

---
#### Ship list
- [x] Code reviewed


atd-vzv/src/Components/Popover/popoverConfig.js
```diff
@@ -2,8 +2,13 @@ import React from "react";
...
 
-const formatNumber = (num) => {
-  return num.toString().replace(/(\d)(?=(\d{3})+(?!\d))/g, "$1,");
+/**
+ * Format a number with commas
+ * @param {Number} num - a number that needs commas added
+ * @returns {String} - a string of the number with commas added
+ */
+const addCommasToNumber = (num) => {
+  return num ? num.toString().replace(/(\d)(?=(\d{3})+(?!\d))/g, "$1,") : "";
```
mddilley, line 11: Not sure what fallback value we want here, but the deploy preview broke when `undefined` was coming through function so I wanted to do something about it.

atd-vzv/src/views/summary/CrashesByPopulation.js
```diff
@@ -95,7 +95,10 @@ const CrashesByPopulation = () => {
       </Row>
       <Row>
         <Col>
-          <CrashTypeSelector setCrashType={setCrashType} componentName="CrashesByPopulation"/>
+          <CrashTypeSelector
+            setCrashType={setCrashType}
+            componentName="CrashesByPopulation"
+          />
```
mddilley, line 101: Prettier 💅

chiaberry (approved):
yay for fallbacks!

atd-vzv/src/Components/Popover/popoverConfig.js
johnclary: just want to point out that we handle this w/o regex in Moped using [`Intl.NumberFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat).

```javascript
> new Intl.NumberFormat('en-US').format(123567890)
// '123,567,890'
```

johnclary (approved):
glad to see this patched :)

atd-vzv/src/Components/Popover/popoverConfig.js
johnclary: ...and actually if you don't specify the locale, i think the number format is [localized automatically](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#locale_identification_and_negotiation) - so we might actually want to opt for this in case someone is browsing our website....from India?

roseeichelmann (approved):
noice 😎

atd-vzv/src/Components/Popover/popoverConfig.js
johnclary: btw @patrickm02L if this is scheduled to fail again over the weekend (which i think you mentioned was likely) — i am all for merging as-is!

patrickm02L (approved):
Good to go 🤘

PR #1146, opened 2022-12-30
Cascade delete crash notes when parent record is deleted

roseeichelmann:
## Associated issues
Closes cityofaustin/atd-data-tech/issues/11030

Upon deleting a temp crash and creating a new temp crash that reused the same crash_id (temp crash_id's are created by incrementing by 1 from the highest crash_id), the notes from the first temp crash were carrying forward into the new crash because the crash_note for that crash_id was not deleted in the DB.

There are no code changes for this PR but I wanted to open it so y'all could test the fix I made for the issue above and also so I can outline the changes I made to the DB. @frankhereford and I had a discussion about whether I needed to make the usual SQL files for these minor database updates and want to have a discussion with the rest of the dev team about the necessity of these files/whether we should continue this practice considering we haven't been consistent with them and haven't used them much (to our knowledge?)

Changes:
- Added the following foreign key - crash_id → atd_txdot_crashes . crash_id - crash_notes_crash_id_fkey with:
    - Cascade on delete violation
    - Cascade on update violation
- Added an object relationship: crash_notes . crash_id  → atd_txdot_crashes . crash_id
- Added an array relationship: crash_notes . crash_id  → atd_txdot_crashes . crash_id

## Testing
**URL to test:** 
You can test this in VZE staging (https://visionzero-staging.austinmobility.io/editor/#/tools/create_crash_record)

**Steps to test:**

1. Navigate to tools -> create crash record
2. Create a new crash record, make note of the crash_id that was generated
3. Add a new note to that temp crash record
4. Go back to tools -> create crash record, scroll down and delete the temp crash you created
5. Wait a minute to give the database time to make the deletes
6. Create another crash record, it should have the same crash_id but the note you made earlier shouldn't be present in the crash details page for the new crash.
7. You could also be checking the hasura staging console to double-check that the crash in the atd_txdot_crashes table and the note in the crash_notes table is deleted.

Once this PR is approved I will make the same changes in the hasura prod console. Also, to fix Xavier's specific issue with crash 1026, I can delete that crash note in the DB.

---
#### Ship list
- [ ] Code reviewed


patrickm02L (approved):
The cascading delete works like a charm! ✨ 

**Video highlights**
- :12 Create temp Crash Record
- :35 Write a note
- :50 Delete temp crash record
- 1:36 Create another Crash Record
- 1:53 No new note **CHECK**
- 2:09 Delete temp crash record

https://user-images.githubusercontent.com/5566727/210383006-a7ea9887-bc90-411f-b1e2-022afcd6ecc7.mp4

frankhereford (approved):
It may be my weak memory, but I don't recall us having a conversation as a group to remove the requirement that we keep up with a mock-migration files in the VZ repo. I think we need to have that discussion so people can offer pros and cons to making this policy change. It would be a good time to ask people to put their approval or not on this PR too.

# Proposal related to this PR
I would like to co-opt this PR, and use it to record the group's agreement or lack thereof to adopt the policy that we do not need to keep those migration-like files in the repo. I propose that an approval on this PR indicates your support for those changes. Also, under this proposal, this PR bears more weight than just the changes you outlined for this one use case, and as a result, I'd like to see a majority of the dev team's approval on it + Patrick's as the PM. @patrickm02L please adjust your approval if needed based on this change of context.

# Review of the work

I looked at the FK directives you've put in place via a SQL client, and they look good. Thank you Rose!

I think these observations should be spun-off into (a) new issue(s) based on @patrickm02L's read.

1) I observed that when I create (well, re-created) temp crash record 1027 in the staging environment, I observed that there are change log entries which persisted from the crashes earlier existence before it was deleted. You can see Patrick's changes here on a crash I created today (1/11/2023):

image

I recommend that we treat change log entries in the same way - when a crash is deleted, any change log entries that exist should also be deleted in the cascade. As it stands, these change log entries are referring to an incarnation of the crash prior to the one I created.

frankhereford:
Related question, but outside the scope of my review for @patrickm02L --  What was the final decision on how we should handle ID'ing crashes, ie, are we doing to forgo using `crash_id` as our primary key and use an auto-incrementing value, and then track the crash ID as an attribute of a crash with our internal ID?

Such a change would be invisible to our users, but it would require a non-trivial refactor, but it would also eliminate these types of issues going forward.

chiaberry (approved):

PR #1150, opened 2023-01-03
Add `docker compose` & python tooling, dockerize VZ stack

frankhereford:
## Associated issues

This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/10650. 

## Quick-start

1) Fill out a `.env` file based on `env_template`. Frank can give you one too to work from.
1) In the root of the VZ repo, you'll find a program called `vision-zero`. It is a python program which takes a number of commands similarly to the way the `hasura-cluster` program works for the moped database, except that it controls the entire stack. Documentation for the tool can be found in the root [README](https://github.com/cityofaustin/atd-vz-data/tree/ldm_development_environment#local-development).
1) This script should be executed in a [virtual environment ](https://docs.python.org/3/library/venv.html) and requires a single dependency `shtab` as noted in the `requirements.txt`. It can produce command completion scripts which are optional, but helpful.
1) If, from the root of the VZ repo, you can `./vision-zero`, then you should be ready to go.

## Testing

Perform common tasks, such as:

* Build or rebuild images used in the stack: `./vision-zero build`
* Replicate the production database to your local database: `./vision-zero replicate-db`
  * You must be on the VPN or have your public IP in the access lists for the VZ read replica RDS instance
  * Be patient, this is at least a 15 minute process (the first time each day) when you skip the changelog, and can be an hour easily if you do, using the `-c / --include-change-log-data` option.
  * The script defaults to not including the change log data.
  * The script will cache these database images as raw SQL in the `./atd-vzd/snapshots/` directory, so clean them out periodically with `./vision-zero remove-snapshots`
  * These caches are reused for a given day, so the first time you run this command, you have to download the data, and all subsequent invocations /that day/, you are reloading that same cached copy. 
  * RAM disk support - optional feature. If you don't use this, the script will use your hard drive, which is "normal."
    * The `vision-zero` application recognizes the `-r / --ram-disk` option for applicable operations, which if used from DB spin-up, will result in a database container backed with RAM for disk. This may make DB population and read intensive queries immeasurably faster at the expense of non-volatile storage. It also may not; there situations where the extra RAM pressure may make the configuration non-optimal. By default, the configuration is to not use a RAM disk, which is generally safe and predictable. 
* Individually start and stop services:
  * `./vision-zero graphql-engine-up`
  * `./vision-zero graphql-engine-down`
  * `./vision-zero db-up`
  * `./vision-zero db-down`
* Start or stop the VZE (available on localhost:3000 or VZV available on localhost:3001)
  * `./vision-zero vze-up`
  * `./vision-zero vze-down`
  * `./vision-zero vzv-up`
  * `./vision-zero vzv-down`
* Start a `psql` session on your local database: `./vision-zero psql`
* Start a `bash` shell on a container with additional tooling to interoperate with the database and stack: `./vision-zero tools-shell`

---
#### Ship list
- [ ] Code reviewed


atd-vze/package.json
```diff
@@ -67,7 +67,7 @@
     "react-scripts": "^3.4.4"
   },
   "scripts": {
-    "start": "export HTTPS=true && react-scripts start",
+    "start": "export HTTPS=false && react-scripts start",
```
frankhereford, line 70: I'm disabling SSL here. The previous configuration uses a unsigned (snake-oil) certificate and I am unsure of the value of making developers acknowledge this, when the network traffic in question is not traversing the internet.

atd-vzv/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM node:16-bullseye
+RUN apt-get update
+RUN apt-get install -y vim aptitude magic-wormhole
+WORKDIR /root/
+RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh | bash
+RUN npm install -g npm-check-updates
+WORKDIR /root/atd-vzv
+#COPY . /root/atd-vzv
+#RUN npm install
```
frankhereford, line 9: I toyed with the idea of copying the state of the app at image build time into the docker image, so it could stand-alone. With the app on there, you can still bind-mount over it, and get the behavior we're used to, but it makes build times muuuch longer, and it also introduces a host of confusions when things are not working just right. 

I'd be curious if there's interest in this. Maybe something to experiment more with if we ever want to host our own web server on ECS instead of netlify, which seems very unlikely.

docker-compose-docker-volume.yml
```diff
@@ -0,0 +1,15 @@
+version: '3.7'
+services:
+  postgis:
+    container_name: visionzero-postgis
+    build: atd-vzd
+    env_file:
+      - .env
```
frankhereford, line 7: The form of this file is found in `env_template`

atd-vze/package.json
mddilley: 👍 I remember this being significant but I have no idea why. I found the PR where is was added, and I [reacted with an emoji to that change](https://github.com/cityofaustin/atd-vz-data/pull/110/files#r314878942). None of which is helpful. 😆 The app starts up just fine either way so I'm supporting it since it only applies to local dev anyways.

atd-vzv/Dockerfile
mddilley: I'm a bit torn about local dev of VZE and VZV with the extra Docker layer. I could be missing something so eager to hear other opinions, but my thoughts are:
- extra complexity
- you have to attach to the container to see status of the dev server, compile warnings, etc. which I depend on a lot
- npm start by itself is pretty straightforward and is the way we consistently start up other Create React App bootstrapped apps

requirements.txt
```diff
@@ -0,0 +1 @@
+shtab==1.5.7
```
mddilley, line 1: I had to add `python-dotenv` here to get the `build` command to run successfully since it is used in the `vision-zero` program

vision-zero
```diff
@@ -0,0 +1,330 @@
...
+        stop(args)
+    elif args.command == "replicate-db":
+        dbUp(args)
+        replicateDb(args)
+    elif args.command == "remove-snapshots":
+        removeSnapshots(args)
+
+
+## function to list the files in a directory and delete them
+def removeSnapshots(args):
```
mddilley, line 49: I also had an issue with running function with the command. I added my output below, but it works fine if I run from the command line without `subprocess`. Another tester could help see if this is a problem with my environment.

`FileNotFoundError: [Errno 2] No such file or directory: 'rm -rvf atd-vzd/snapshots/*.sql'`

mddilley (commented):
Frank, all of this is looking great! 🚗 🚀 Really liking the up and down commands for the different services! RR replication is also a really nice thing to have baked in already. 🙌 

I might need some help testing this some more because everything works for me up until I try to start the `psql` session. I get `psql: error: connection to server at "postgis" (172.24.0.2), port 5432 failed: FATAL:  database "visionzero" does not exist` even though the DB container is up. This could be an issue with my local setup, but I'm not sure. I added some comments and will pick back up on this when I figure it out.

atd-vzv/Dockerfile
frankhereford: ## TLDR

* Mike is right, and there isn't an obviously compelling reason to dockerize the apps.
* They still work the old way in this configuration, so I want to give it a whirl -- using docker for the VZV/VZE is optional.

## The deets

@mddilley's observations here are each entirely correct and well considered. I don't have a problem to point to that is explicitly solved by pushing the VZV and VZE down into a docker container, except maybe just the intent of "dockerize all the things" which is far from a cogent and valid reason. Dockerizing the app isn't really solving a problem we're bumping into and is more like a design philosophy, which is totally hand-wavy. 👋🌊

But, that said, I would like to propose that it stay in, at least for the time being, because the way its setup doesn't preclude the use of running the environments on the bare metal outside of docker. The code is bind-mounted into the containers, so you can both change into the root directory of either app and `npm start` them up, or you can use a docker container to run them, if that's your preference. I can see a world where a developer running on windows needs to get these spun up, and suddenly that docker environment is looking pretty enticing. (I'm reaching here, but I think it's a valid argument.)

I do see the concern that we'll make some change to the apps that would break the docker builds, but I don't feel like this will be a common occurrence. If it does happen, it probably means we've done something "interesting," and fixing the issue will likely be a revealing, learning experience. I propose that keeping the docker images working can be my responsibility and will not block a PR that otherwise would be accepted from being merged. This will help me learn if this is worth it or not, and we can go from there.

requirements.txt
frankhereford: Sure enough - great catch. Patched -- thanks!

vision-zero
frankhereford: Another awesome catch, thank you Mike. I have no idea why this was working for me previously, but it sure wasn't this morning and it was exhibiting the same error you described.

I scraped the `rm` invocation and am using the `os` library now with a loop. Feels more pythonic, maybe?

atd-vze/package.json
frankhereford: Oh Mike, that is the hint I needed. Peeking at that PR, i see this addition: `REACT_APP_MAPBOX_TOKEN="PRIVATE TOKEN GOES HERE"`. 

I bet mapbox won't serve up background tiles if we mix and match HTTPS & HTTP. Mapbox also won't serve them up without overloading the referrer, so I am not entirely sure there is any benefit to enabling local SSL, unless we all agree that we're going to cohesively fib to mapbox in our request headers. Thanks for this - we know how to turn it back on if we need to, and I'm totally open to conversations about why we should leave it on anyway, if anyone feels strongly on the subject.

frankhereford:
Mike, thank you for your testing and thoughtful comments. I think I've addressed everything, except for your comment here:

> I might need some help testing this some more because everything works for me up until I try to start the psql session. I get psql: error: connection to server at "postgis" (172.24.0.2), port 5432 failed: FATAL:  database "visionzero" does not exist even though the DB container is up.

I'm wondering, is it possible that you may have run the `./vision-zero psql` command before you did a `./vision-zero replicate-db`? A brand-new, fresh DB container won't have a `visionzero` database, and the `psql` command expects there to be one. If this problem persists, I think I'll have to ask you if you can demo the issue for me, and I'll do my best to get it working. 

Thanks!!

atd-events/README.md
```diff
@@ -91,7 +91,7 @@ $ pip freeze > requirements.txt
 
 #### New PRs
 First, it is recommended that you create a PR. Whenever a PR is created
-CircleCI will begin creating the python bundles for Lambda and SQS and it
+a GitHub Action flow will begin creating the python bundles for Lambda and SQS and it
```
chiaberry, line 94: thank you for updating this

atd-vze/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM node:16-bullseye
+RUN apt-get update
+RUN apt-get install -y vim aptitude magic-wormhole
+WORKDIR /root/
+RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh | bash
+RUN npm install -g npm-check-updates
+WORKDIR /root/atd-vze
+#COPY . /root/atd-vze
+#RUN npm install
```
chiaberry, line 9: are these commented out lines left in here so we have a different way to install?

atd-vze/src/views/Tools/ToolsUploadNonCR3.js
```diff
@@ -163,7 +163,9 @@ const ToolsUploadNonCR3 = () => {
     }
 
     if (!isValidCoord(record["longitude"], record["latitude"])) {
-      errors.push("Invalid coordinate pair");
+      errors.push(
```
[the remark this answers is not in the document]
frankhereford: Yes - thank you for pointing this out. I'll merge `master` into `layer-data-model` right now, which should make this diff once again reflect just the changes I put in for the dev environment.

requirements.txt
```diff
@@ -0,0 +1,2 @@
+shtab
+python-dotenv
```
chiaberry, line 2: should these be pinned to a specific version, or it doesnt matter?

atd-vze/src/views/Tools/ToolsUploadNonCR3.js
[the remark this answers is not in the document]
frankhereford: This is done, and the PR's diff looks right again. Sorry for that extra noise complicating review!

requirements.txt
frankhereford: We can pin them in an abundance of caution, but given this very simple set of dependencies, I'd be interested in trying out seeing if we can avoid dependabot PRs on this `requirements.txt` file by letting them float. I think @johnclary may be able to offer us direction on what he's found most reliable in the past in his python experience.

chiaberry: okay i figured these were small potatoes, but just wondered.

atd-vze/Dockerfile
frankhereford: They are there as a reminder that we have an option of bundling a snapshot of the application into the docker image, and then being able to bind-mount the actual local copy of the files over that. This would be something we might do if we ever were to want to have more control over the web-server that serves them up, which is currently netlify. Like if we wanted to move it to ECS, for example.

I think thought that they are not helpful right now in there, and doubly so without comments at least explaining what they are. I am going to remove them - thank you @chiaberry.

requirements.txt
johnclary: having been bitten by completely untethered package versions recently, i would say to always pin unless you have a good reason not to. major versions is usually safe:

```
python-dotenv==0.*
```

frankhereford: Thanks @johnclary - I appreciate your input. I'll pin these to their major versions here in a just a few while the VZDB RR updates.

frankhereford: This is done - thank you both!

atd-vzv/Dockerfile
chiaberry: I agree with Mike that I can see myself missing the compile warnings and status of the dev server, I found myself waiting to see the feedback one usually gets from `npm start` so its really good for you to point out Frank that we can still use the npm start to get them going. 

I need to dig into this more, but if we are running the backend locally, changing into the root directory and then using`npm start` still points to the staging environment right?

frankhereford: That's a great question Chia. 🙏🏻 I believe this would be controlled by the `.env` file in place, and I think it would be helpful for me to add a `.env.local` similarly to how we have `.env.staging` and `.env.production`, for the VZE. I'll do that. Thanks!

frankhereford:
@mddilley wrote:

> psql: error: connection to server at "postgis" (172.24.0.2), port 5432 failed: FATAL:  database "visionzero" does not exist

Mike, Chia pointed out something which I think caused this problem. I had the DB name for the local VZ db set as `visionzero` in the template .env file. This doesn't work, and it turns out that it needs to be set to `atd_vz_data` so the `pg_dump` output from the RR works as expected. I've pushed a commit to fix this in the template, but it may still persist in your RR.

I'm not 100% sure this is what caused this, but it does seem like a strong possibility.

chiaberry (approved):
First off I want to thank you for this big step forward in our local development! It's great that we will be able to run the backend locally. 

I've been able to run the editors and viewers, and am currently testing the replicate-db command. The way I envision I'd use this would be to run a database locally and use `npm start` to run the vze, but set it so its using the local db and not the staging one. I think it would entail adding a .env.local to the atd-vze and adding an `npm start:local` command? But this feels out of scope of this issue. An enhancement after this gets merged. 

I'd like to see the quick start information you included in the PR description to be included in the README, and explicit steps for a first time user to be able to spin this up. I am more than happy to open a PR against this or the layer-data-model branch after this one is merged.

chiaberry:
I opened an issue to jot down future enhancements to this https://github.com/cityofaustin/atd-data-tech/issues/11401, so this can be merged and any of us can pick up the issue to work on enhancements. or even treat them like snackoos and pick up parts of it. 

It is so great that we are now able to run vzd locally! 🎉

atd-vzv/Dockerfile
mddilley: Y'all are on it! I didn't even think of what it would take to get this hooked up to a separate local server for the frontend. Agreed that optional is the way to go! 🙏 🚀

mddilley:
@frankhereford Thanks for the giving me more info about the .env - that was part of the problem that I was facing. 

I noticed that the DB container wasn't staying up on my machine. That is why the `replicate-db` command wasn't working for me because the DB would start and then immediately stop so the dump couldn't load into anything. Did some digging and ended up `docker system prune`ing. That did the trick, and I was off to the races. 🏇

mddilley (approved):
Thanks again for looking the issues that were happening on my machine. Hopefully, it didn't cause too much confusion. I just retested every command, and it all works great! This is the foundation to something big - very exciting! 🙏 🚀 🚀 🚀

roseeichelmann (changes requested):
like we discussed over slack, because I have the latest version of docker, my machine doesn't use `docker-compose` and instead uses the `docker`  command itself, which is a change docker has been migrating to over the last couple of years. this is causing the scripts to fail for me because they rely on the `docker-compose` command. another thing to consider is i run Ubuntu 22.04.1 LTS and have an intel chip unlike the rest of the team which may or may not affect my ability to test as well.

chiaberry:
Maybe we should all as a team make it so we are able to run `docker compose` especially since

> From the end of June 2023 Compose V1 won’t be supported anymore and will be removed from all Docker Desktop versions.
Source https://docs.docker.com/compose/reference/

vision-zero
```diff
@@ -0,0 +1,339 @@
+#!env python
```
roseeichelmann, line 1: oo another thing to note is i had to change this shebang to `#!/usr/bin/env python` for it to work on my machine. im not really sure why i need to use an absolute path for it to work and yall dont tho :0

frankhereford: that is a good question! i would have expected `source`ing the `activate` script out of the venv would have set those paths. Thanks for this extra piece of info though - I can't wait to iron out these wrinkles tomorrow.

patrickm02L (approved):
Approved based on Devs approvals and discussion in 2/8/23 Sprint Planning. ✨

vision-zero
frankhereford: I addressed this issue through testing and experimentation on an x86 machine. Thank you Rose!

roseeichelmann (approved):
works like a charm now! thanks for adding support to linux machines and other docker versions. this is a huge glow-up for VZ 🌟

vision-zero
```diff
@@ -0,0 +1,361 @@
...
+
+    if docker_compose_bundled:
+        print("Using `docker compose`")
+        return ["docker", "compose"]
+    elif docker_compose_stand_alone:
+        print("Using `docker-compose`")
+        return ["docker-compose"]
+    elif not docker_compose_stand_alone and not docker_compose_bundled:
+        print("Neither `docker-compose` nor `docker compose` is available.")
+        exit()
```
chiaberry, line 42: ✨

frankhereford:
I want to thank you all again for all the really careful testing and feedback that you provided. This task had a pretty amorphous set of acceptance criteria -- to provide an environment for people to work in -- and you each helped a ton in seeking that state. 🙏🏻🙏🏻🙏🏻

PR #1156, opened 2023-01-13
remove all sql database migrations cruft

roseeichelmann:
## Associated issues

No associated issues, but this PR is a follow up to #1146 and our dev sync discussion on 1/12 where there was consensus among devs to adopt the policy of not keeping the migration-like SQL files for database changes in the repo. This PR removes all of those old SQL files/folders in the repo. This PR can also be a place for further discussion or questions on this topic.

---
#### Ship list
- [ ] Code reviewed


frankhereford (approved):
# 🧹🧹🧹

Rose and I spoke about this briefly, and while this PR leaves it pretty barren in there, the forthcoming work in the [VZ orchestration PR](https://github.com/cityofaustin/atd-vz-data/pull/1150) will put some stuff in there again.

Thanks for this Rose!

mddilley (commented):
All about getting these migration-like files out of here 🧹, but I'm hesitant about clearing the slate before we switch to the LDM. I missed the dev sync discussion about this so keep that in mind, and I can totally be swayed. 👍 

I think this is mostly coming from my inexperience with the triggers and views and a hypothetical situation where I would need to look into one of them quickly between now and when LDM is merged. I'm probably overlooking something or overthinking it, but that is my first instinct.

Also, I think it would be good to keep the Moped pattern of having version controlled views. And maybe that it would be good to spread that to triggers?

PR #1157, opened 2023-01-13
Patch `production` with VZV NaN fix from PR 1145

frankhereford:
## Associated issues

none

## Cherry-picking notes

This did involve resolving a merge conflict between these commits and `production`, but it was a straightforward process - we wanted everything incoming in the commits as they better organize the JSX in `popoverConfig.js`.

## Testing
URL: https://visionzero-staging.austinmobility.io/viewer/ will demonstrate the changes compile, and I have spun it up locally as well. This doesn't test the underlying issue at hand, because that would require putting together some curated test data in the staging dataset, which is a whole other ball of wax. 

I recommend that we accept this as-is, as it was reviewed in PR 1145, and I have full confidence in @mddilley's work.


chiaberry (approved):
thank you!!

patrickm02L (approved):
Awesome!

PR #1161, opened 2023-01-18
"Confirm Save" modal is activated when trying to use "X" button to close "Success" modal

roseeichelmann:
## Associated issues
cityofaustin/atd-data-tech/issues/11210

Linked in the issue above is a video of the bug. ~I thought it made the most sense to just get rid of the "X" buttons at the top of the modals because the other buttons already serve the function of exiting or proceeding to the next modal. Having both just unnecessarily complicates things imo? But if anyone feels strongly about leaving in the "X" buttons I can bring them back!~

My opinion has been changed after reading [this](https://ux.stackexchange.com/questions/90153/are-close-and-x-on-a-popup-redundant) discussion about modal redundancy/user experience so I decided to bring the "X" buttons back in, super curious about what y'all think on this topic!

## Testing
**URL to test:** 
https://deploy-preview-1161--atd-vze-staging.netlify.app/#/tools/upload_non_cr3

1. Go to the upload Non-CR3 page.
2. You can upload [this](https://github.com/cityofaustin/atd-vz-data/files/10449521/non_cr3_template.csv) csv to test
3. Click the "Save to Database" button then click continue on the "Confirm Save" modal
4. Click "Ok" on the "Success" modal
5. Try again but this time click "x" and/or "Cancel" on the "Confirm Save" modal
6. Try again but click away on the screen to exit the modals.

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vze/src/views/Tools/ToolsUploadNonCR3.js
+++ b/atd-vze/src/views/Tools/ToolsUploadNonCR3.js
@@ -710,7 +710,7 @@ const ToolsUploadNonCR3 = () => {
         toggle={toggleModalSaveConfirm}
         className={"modal-primary"}
       >
-        <ModalHeader toggle={toggleModalSaveConfirm}>Confirm Save</ModalHeader>
+        <ModalHeader>Confirm Save</ModalHeader>
         <ModalBody>
           <p>
             Only valid records will be processed, if you have any invalid
@@ -736,7 +736,7 @@ const ToolsUploadNonCR3 = () => {
 
       <Modal
         isOpen={modalFeedback}
-        toggle={toggleModalSaveConfirm}
+        toggle={() => setModalFeedback(false)}
         className={"modal-primary"}
       >
         <ModalHeader>{feedback["title"]}</ModalHeader>
--- a/atd-vze/src/views/Tools/ToolsUploadNonCR3.js
+++ b/atd-vze/src/views/Tools/ToolsUploadNonCR3.js
@@ -710,7 +710,7 @@ const ToolsUploadNonCR3 = () => {
         toggle={toggleModalSaveConfirm}
         className={"modal-primary"}
       >
-        <ModalHeader>Confirm Save</ModalHeader>
+        <ModalHeader toggle={toggleModalSaveConfirm}>Confirm Save</ModalHeader>
         <ModalBody>
           <p>
             Only valid records will be processed, if you have any invalid
@@ -739,7 +739,9 @@ const ToolsUploadNonCR3 = () => {
         toggle={() => setModalFeedback(false)}
         className={"modal-primary"}
       >
-        <ModalHeader>{feedback["title"]}</ModalHeader>
+        <ModalHeader toggle={() => setModalFeedback(false)}>
+          {feedback["title"]}
+        </ModalHeader>
         <ModalBody>{feedback["message"]}</ModalBody>
         <ModalFooter>
           <Button color="secondary" onClick={() => setModalFeedback(false)}>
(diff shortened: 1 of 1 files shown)


atd-vze/src/views/Tools/ToolsUploadNonCR3.js
```diff
@@ -710,7 +710,7 @@ const ToolsUploadNonCR3 = () => {
         toggle={toggleModalSaveConfirm}
         className={"modal-primary"}
       >
-        <ModalHeader toggle={toggleModalSaveConfirm}>Confirm Save</ModalHeader>
```
roseeichelmann, line 713: this removes the "X" button at the top

atd-vze/src/views/Tools/ToolsUploadNonCR3.js
```diff
@@ -736,12 +736,10 @@ const ToolsUploadNonCR3 = () => {
 
       <Modal
         isOpen={modalFeedback}
-        toggle={toggleModalSaveConfirm}
+        toggle={() => setModalFeedback(false)}
```
roseeichelmann, line 739: this function was causing the other modal to return when clicking away on the screen because `modalSaveConfirm` was getting set to `true`. now exiting by clicking away actually closes the modal and the other one won't return (yay no infinite close loop)

atd-vze/src/views/Tools/ToolsUploadNonCR3.js
```diff
@@ -736,12 +736,10 @@ const ToolsUploadNonCR3 = () => {
...
         className={"modal-primary"}
       >
-        <ModalHeader toggle={toggleModalSaveConfirm}>
-          {feedback["title"]}
-        </ModalHeader>
+        <ModalHeader>{feedback["title"]}</ModalHeader>
```
roseeichelmann, line 742: gets rid of the "X" button on the second "Success" modal

roseeichelmann:
Also I found this conversation to be really interesting about whether having both an "X" and a "Cancel" button on a popup is redundant. I think I have been convinced that it's best to leave the "X" because users are used to it so I'm actually gonna bring them back for now, super curious what yall think lol!

atd-vze/src/views/Tools/ToolsUploadNonCR3.js
```diff
@@ -736,10 +736,10 @@ const ToolsUploadNonCR3 = () => {
 
       <Modal
         isOpen={modalFeedback}
-        toggle={toggleModalSaveConfirm}
+        toggle={() => setModalFeedback(false)}
         className={"modal-primary"}
       >
-        <ModalHeader toggle={toggleModalSaveConfirm}>
+        <ModalHeader toggle={() => setModalFeedback(false)}>
```
roseeichelmann, line 742: same fix as above but for the "X" button to close

frankhereford:
@roseeichelmann wrote:
> both an "X" and a "Cancel" button on a popup is redundant

I never really thought about it, but yea, I think I agree with that assertion. But ... I don't think that many people have thought about it, and while it's technically correct, based on what prompts the user to bail on the interaction will color which interface element they look for.

That said, I agree with your idea to provide both to the user, and I actually would need some convincing to get excited about removing one or the other. I think your route is great.

patrickm02L (approved):
Works as expected. No more wacky close loop! 💫 

https://user-images.githubusercontent.com/5566727/213459780-438811d1-ca8b-4e4b-916f-3d4a836e6ed3.mp4

mddilley (approved):
Works great! Thanks for sharing the UX discussion. Found myself thinking about how I'm "X" button user and how I could have been conditioned by spam popups. 🤯  Love it! 🚢

johnclary (approved):
yayyy thanks for fixing this! apparently i'm an "X" user too 😊  🙌

chiaberry (approved):
count me in the "X" group too

PR #1163, opened 2023-01-20
Last column in By Time of Day visualization that should be hidden is hiding inconsistently

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/11214

This patches a bug where a column populated with the maximum single day total of fatailities, serious injuries, or both (depending on the option chosen) would appear as the 25th column in the By Time of Day visualization. 

The fix was to use a callback ref as a way to inspect the refs of rendered cells one by one and hide them if they are placeholder cells. All within the component lifecycle instead of using vanilla JS. Here is the [article about the callback ref pattern](https://tkdodo.eu/blog/avoiding-use-effect-with-callback-refs#usecallback-to-the-rescue).

## Testing
**URL to test:** 
https://deploy-preview-1163--atd-vzv-staging.netlify.app/

**Steps to test:**
1. Go to the preview and look at the By Time of Day visualization
2. Keep refreshing the page and looking to see a 25th column appears with the same color as the max in the legend at the bottom
3. If you see it, let me know that the 👻 has returned

---
#### Ship list
- [x] Code reviewed

### DIFF
--- a/atd-vzv/src/views/summary/CrashesByTimeOfDay.js
+++ b/atd-vzv/src/views/summary/CrashesByTimeOfDay.js
@@ -157,21 +157,21 @@ const CrashesByTimeOfDay = () => {
   }, [maxForLegend, heatmapData, crashType]);
 
   // Hide placeholder cells
-  useEffect(() => {
-    const heatmapChildCellNumbers = [171, 172, 173, 174, 175, 176, 177];
+  // useEffect(() => {
+  //   const heatmapChildCellNumbers = [171, 172, 173, 174, 175, 176, 177];
 
-    let cellsToHide = heatmapChildCellNumbers.map((num) =>
-      document.querySelector(
-        `#demographics-heatmap > div > svg > g > g:nth-child(${num})`
-      )
-    );
+  //   let cellsToHide = heatmapChildCellNumbers.map((num) =>
+  //     document.querySelector(
+  //       `#demographics-heatmap > div > svg > g > g:nth-child(${num})`
+  //     )
+  //   );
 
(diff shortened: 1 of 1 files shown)


atd-vzv/src/views/summary/CrashesByTimeOfDay.js
```diff
@@ -33,6 +33,73 @@ const hourBlockArray = [...Array(24).keys()].map((hour) =>
   moment({ hour }).format("hhA")
 );
 
+/**
```
mddilley, line 36: These two big blocks of green and red are to move some helpers outside of the component body and add some documentation. I had to add some parameters to a couple of them and then pass them where they are called.

atd-vzv/src/views/summary/CrashesByTimeOfDay.js
```diff
@@ -199,7 +208,10 @@ const CrashesByTimeOfDay = () => {
       </Row>
       <Row>
         <Col>
-          <CrashTypeSelector setCrashType={setCrashType} componentName="CrashesByTimeOfDay"/>
+          <CrashTypeSelector
```
mddilley, line 211: 💅 only

chiaberry (approved):
code upgrade and I cant see the phantom column anymore! thank you for fixing this

johnclary (approved):
wow—this is such a cool pattern!

> All ref props are just functions!

🤯 💡 ⚛️

thanks for the TIL. looks great and nice to see the tidying as well 🧹 🙌

roseeichelmann (approved):
the 👻 is gone!

PR #1164, opened 2023-01-25
Create MVP of FTT Index

roseeichelmann:
## Associated issues
Closes cityofaustin/atd-data-tech/issues/11190

We decided to split the above issue into multiple issues due to it's complexity, so this PR is primarily responsible for the front end UI of the FTT index and the initial creation of the `fatalities` table in the DB.

I populated the table with mock data so it is **not** accurate and you will notice the year and the date columns don't match up; that's fine I just wanted to be able to see what the table looks like with data and test pagination.

## Testing
**URL to test:** 
https://deploy-preview-1164--atd-vze-staging.netlify.app/#/fatalities

**Steps to test:**
1. Go to the new Fatalities section in the Nav bar, it is only visible to Admin access & above.
2. The following columns should be displayed in the table: Year, Crash ID, Case ID, Law Enforcement #, YTD Fatal Crash, YTD Fatalities, Crash Date, Location, Victim Name, Current FRB Status, and FRB Recommendation
3. Set the results to 10 per page so you can test pagination.
4. Clicking on the crash ID will take you to that crash details page
5. Not focusing on implementation of search or sort at this time so you don't need to test that, though I did turn on sorting for most columns.

---
#### Ship list
- [ ] Code reviewed


atd-vze/src/routes.js
```diff
@@ -161,6 +162,12 @@ const routes = roles => [
     name: "Crash Details",
     component: Crash,
   },
+  {
```
mddilley, line 165: This path needs one more protection to keep read-only and editors from visiting it. Line 189 in here has some role checks as an example. 

The good thing is that the query fails if you visit this page as a read-only or editor thanks to the Hasura permissions 🙌 , but adding the role checks will redirect to the dashboard view if anyone types the path in the address bar.

Also, I had forgotten they existed, but there are some test accounts in 1Password for each role that are called **VZE Test User: Editor**, etc if you ever need them. They came in handy for me to test this. 🧰

mddilley (commented):
This is looking great! Thanks for mocking up some data, and I'm excited to see this get built up 🧱 by 🧱. Test steps check out, and everything looks super in the new table. I left a couple comments, but I think the small change to role access is the most important.

patrickm02L (approved):
We've already talked about this, but I wanted to capture it here for the record. 

Let's replace the 'filled' Fatalities font awesome icon with one that is outlined. I know we can't get the outlined version from font awesome with out having a pro account, so here are some options from google icons. 

[Ecg heart](https://fonts.google.com/icons?selected=Material%20Symbols%20Outlined%3Aecg_heart%3AFILL%400%3Bwght%40400%3BGRAD%400%3Bopsz%4048)
[Cardiology](https://fonts.google.com/icons?selected=Material%20Symbols%20Outlined%3Acardiology%3AFILL%400%3Bwght%40400%3BGRAD%400%3Bopsz%4048)

chiaberry:
@patrickm02L Rose and I worked on it some the other night, and it is not straightforward at all to use a google icon in the coreUI component. The way the icons are used in the component make it really difficult to choose one not in existing libraries. I'm going to see if there are comparable outlined icons or if there are other ways to import custom icons into the component, but maybe other team members have ideas?

patrickm02L:
@chiaberry @roseeichelmann, thanks for looking into a better icon. I don't think we should spend any more effort trying to get the icon perfect. Let's use the regular heart icon - there's an outlined version https://fontawesome.com/icons/heart?s=regular&f=classic

atd-vze/src/views/Fatalities/Fatalities.js
```diff
@@ -0,0 +1,31 @@
...
+import { withApollo } from "react-apollo";
+
+import { subYears } from "date-fns";
+import GridTable from "../../Components/GridTable";
+import gqlAbstract from "../../queries/gqlAbstract";
+import { fatalityGridTableColumns } from "./fatalityGridTableParameters";
+
+// Our initial query configuration
+let queryConf = {
+  table: "fatalities",
```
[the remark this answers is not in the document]
mddilley: Just thought about it a little more, but maybe we use `refetch` in some places to freshen up the data in a parent component like the crash and location tables. It would be hard to do that here since these two views are pretty disconnected.

roseeichelmann:
@mddilley thank you so much for your in depth review and for catching those bugs! i fixed both of the issues you pointed out, should be good to go now.
and @patrickm02L i switched the nav icon to the outlined heart <3

atd-vze/src/routes.js
roseeichelmann: oo yeah i need to use the test user more option. thanks for reminding me about it!

frankhereford (approved):
Rose, I think this looks great. I have tested it via your netlify deployment, and reviewed your code, and it all looks fantastic. I was also really encouraged by the very positive response you received when demonstrating this in the recent sprint review with our stakeholders. 🥇

This is probably best a spin off issue type thing, but I wanted to observe that it may be more helpful that when one sorts by "YTD Fatal Crashes" or "YTD Fatalities," that you may want to overload that request from the user to be `sort by crash_datetime [asc|desc], ytd_fatal_crash_number [asc|desc]` so that the "YTD" part of their meaning continues to be meaningful. I suspect people will be more interested in those sequences of numbers when they are grouped by year, and less likely to be looking for all the, say, 8th crashes of each various year. This is such a minor thing - it may be just as helpful and a ton easier to disable sorting on those columns and rely on the user to figure out that they should sort by the crash_date to get the YTD columns into sequence.

roseeichelmann:
@frankhereford thats a great point! i think for now it makes sense to just disable sorting on that column. thanks for thinking about that!

PR #1171, opened 2023-02-06
Check file type in CR3 download script before updating crash record and storing file in S3

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/8711

This PR updates the CR3 download script to check the mime type of the file downloaded for each crash before storing them and updating the DB to show that it has been processed. This will help us avoid seeing errors in VZE where a valid CR3 pdf cannot be shown for a record that appears to have one. Open to any and all feedback about the logging. 🙏 

## Testing
**URL to test:** 
Local

**Steps to test:**
It isn't easy to make a real non-pdf CR3 download happen so we can simulate it by running the script locally and making the script think a non-pdf has been downloaded. To do this, follow the steps below. You can also see the screenshots below to see the output for invalid and valid CR3 files below without running this locally.

1. Using the production read replica, find a crash record ID that has a non-pdf file stored in S3 by running this statement:
```sql
SELECT
	*
FROM
	atd_txdot_crashes
WHERE
	cr3_file_metadata->>'mime_type' = 'text/html'
```
2. ⚠️ **Very carefully** ⚠️  (or ping me if you would like me to handle this part - happy to take this on and give you a crash record to look for when running the script), connect to the production VZ database and update one of the records that you found to make the script see it as non-processed using this statement:
```sql
UPDATE
	atd_txdot_crashes
SET
	cr3_stored_flag = 'N',
	cr3_file_metadata = NULL,
	cr3_ocr_extraction_date = NULL
WHERE
	crash_id = <the crash record that you found>
```
4. Now that the script will recognize this crash ID as not processed yet, modify [this line](https://github.com/cityofaustin/atd-vz-data/blob/494c482940e721ed4d3db83c02e904f1b5393c4b/atd-etl/app/process/helpers_cr3.py#L157) to read as follows:
```python
is_file_pdf = False
```
5. This will make the file type appear to be non-pdf so we can test that path of the code. You can run the script by following [these steps](https://app.gitbook.com/o/-LzDQOVGhTudbKRDGpUA/s/-M4Ul-hSBiM-3KkOynqS/vision-zero-crash-database/cr3-download-script).
6. You should see the script log that the file upload and crash record update was skipped for this record and also see a list containing this crash ID print at the end of the script.
7. Now that you have tested an invalid CR3 pdf, update [this line](https://github.com/cityofaustin/atd-vz-data/blob/494c482940e721ed4d3db83c02e904f1b5393c4b/atd-etl/app/process/helpers_cr3.py#L157) back from the change made above.
8. Run the script again and the valid CR3 will upload and the crash record will be updated to show that it is processed.
 
### Script handling a non-pdf CR3 download ❌
Screenshot 2023-02-06 at 11 31 23 AM

### Script handling a pdf CR3 download ✅ 
Screenshot 2023-02-06 at 11 12 24 AM

---
#### Ship list
- [x] Code reviewed

### DIFF
--- a/atd-etl/Dockerfile
+++ b/atd-etl/Dockerfile
@@ -13,7 +13,7 @@ RUN apk add bash build-base python3-dev libffi-dev openssl-dev \
 #RUN apk update && apk add build-base python3-dev libffi-dev openssl-dev && pip3 install magic-wormhole
 
 RUN pip install cryptography==3.3.2 splinter selenium requests \
-    awscli web-pdb sodapy boto3 mail-parser
+    awscli web-pdb sodapy boto3 mail-parser filemagic==1.6
 
 WORKDIR /app
 
--- a/atd-etl/Dockerfile
+++ b/atd-etl/Dockerfile
@@ -13,7 +13,7 @@ RUN apk add bash build-base python3-dev libffi-dev openssl-dev \
 #RUN apk update && apk add build-base python3-dev libffi-dev openssl-dev && pip3 install magic-wormhole
 
 RUN pip install cryptography==3.3.2 splinter selenium requests \
-    awscli web-pdb sodapy boto3 mail-parser filemagic==1.6
+    awscli web-pdb sodapy boto3 mail-parser python-magic==0.4.27
 
 WORKDIR /app
 
--- a/atd-etl/Dockerfile
+++ b/atd-etl/Dockerfile
@@ -12,8 +12,8 @@ RUN apk add bash build-base python3-dev libffi-dev openssl-dev \
 # the following is helpful when developing using this image but is not needed for a production build
 #RUN apk update && apk add build-base python3-dev libffi-dev openssl-dev && pip3 install magic-wormhole
 
-RUN pip install cryptography==3.3.2 splinter selenium requests \
-    awscli web-pdb sodapy boto3 mail-parser python-magic==0.4.27
+RUN pip install cryptography==3.3.* splinter==0.* selenium==4.* requests==2.* \
+    awscli==1.* web-pdb==1.* sodapy==2.* boto3==1.* mail-parser==3.* python-magic==0.*
 
 WORKDIR /app
 
--- a/atd-etl/Dockerfile
+++ b/atd-etl/Dockerfile
@@ -12,8 +12,8 @@ RUN apk add bash build-base python3-dev libffi-dev openssl-dev \
 # the following is helpful when developing using this image but is not needed for a production build
 #RUN apk update && apk add build-base python3-dev libffi-dev openssl-dev && pip3 install magic-wormhole
 
-RUN pip install cryptography==3.3.* splinter==0.* selenium==4.* requests==2.* \
-    awscli==1.* web-pdb==1.* sodapy==2.* boto3==1.* mail-parser==3.* python-magic==0.*
+COPY requirements.txt /tmp/
+RUN pip install -r /tmp/requirements.txt
 
 WORKDIR /app
 
--- a/atd-etl/app/process/helpers_cr3.py
+++ b/atd-etl/app/process/helpers_cr3.py
@@ -154,10 +154,9 @@ def process_crash_cr3(crash_record, cookies):
 
         download_path = download_cr3(crash_id, cookies)
         is_file_pdf = check_if_pdf(download_path)
-        print(f"File is pdf: {is_file_pdf}")
 
         if not is_file_pdf:
-            print("File is not a pdf - skipping upload and update")
+            print(f"File {download_path} is not a pdf - skipping upload and update")
             return
         else:
             upload_cr3(crash_id)
--- a/atd-etl/app/process/helpers_cr3.py
+++ b/atd-etl/app/process/helpers_cr3.py
@@ -141,7 +141,7 @@ def check_if_pdf(file_path):
     return file_type == "application/pdf"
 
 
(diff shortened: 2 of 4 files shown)


atd-etl/app/process/helpers_cr3.py
```diff
@@ -16,6 +16,7 @@
 import time
 from http.cookies import SimpleCookie
 
+import magic
```
johnclary, line 19: ✨

atd-etl/app/process_cris_cr3.py
```diff
@@ -59,11 +62,15 @@ def wait(int):
 
 with concurrent.futures.ThreadPoolExecutor(max_workers=1) as executor:
```
johnclary, line 63: cool to see concurrency here—although y'all are keeping it capped at 1 worker?

atd-etl/app/process_cris_cr3.py
```diff
@@ -59,11 +62,15 @@ def wait(int):
...
     for crash_record in crashes_list:
-        executor.submit(process_crash_cr3, crash_record, CRIS_BROWSER_COOKIES)
+        executor.submit(process_crash_cr3, crash_record, CRIS_BROWSER_COOKIES, skipped_uploads_and_updates)
 
 print("\nProcess done.")
 
+if len(skipped_uploads_and_updates) > 0:
```
johnclary, line 69: we're in empty objects are falsey land:

``` python
if skipped_uploads_and_updates:
    # do something great
```

atd-etl/Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN apk add bash build-base python3-dev libffi-dev openssl-dev \
 #RUN apk update && apk add build-base python3-dev libffi-dev openssl-dev && pip3 install magic-wormhole
 
 RUN pip install cryptography==3.3.2 splinter selenium requests \
-    awscli web-pdb sodapy boto3 mail-parser
+    awscli web-pdb sodapy boto3 mail-parser python-magic==0.4.27
```
johnclary, line 16: might as well pin the rest of these or take the plunge for a `requirements.txt` ?

johnclary (approved):
I have not tested this 🤪 but it looks great to me. TIL `libmagic` 🐇 ✨

atd-etl/app/process/helpers_cr3.py
chiaberry: 🪄

atd-etl/app/process_cris_cr3.py
frankhereford: That concurrency feature has been disabled as long as I can remember - I think the original author coded it up but it did not perform well in practice. As it stands, with just one worker, it is generally able to finish its daily run in 5-10 minutes, so I think we will be in the green in terms of execution duration for the foreseeable future.

mddilley: Thanks for that context, Frank! 🙏  It was fuzzy to me and I decided on the "code around it" approach, but this is making me wonder if should trim it out since it did take a little bit of thinking to see exactly how the process helpers were being called - especially since it isn't being used. What do y'all think?

atd-etl/app/process_cris_cr3.py
mddilley: Nice, TIL! I'll update this 👍

atd-etl/Dockerfile
mddilley: Yeah, this is making me think about [this](https://github.com/cityofaustin/atd-vz-data/pull/1150#discussion_r1097734519) that I just saw. Maybe worth a short dev sync discussion so we handle these the same way.

atd-etl/app/process_cris_cr3.py
chiaberry: I think if we arent using it, we should remove it/trim it. Especially if it just complicated understanding what is happening

frankhereford: Trimming it gets my vote, but I do think also that it works solidly as a rock as is too, so I'm totally fine either way. 👍

atd-etl/app/process/helpers_cr3.py
frankhereford: I have always loved that `file` is backed by libmagic. You can even add your own magic in `/etc/magic` if you work with uncommon file types, but I've never had to do that.

frankhereford (approved):
magic

Mike, this is fantastic, you might even say magic. This will save both us and the VZ team plenty of headache going forward. I'm super grateful for this enhancement.

frankhereford:
The team has approved [the original hack](https://github.com/cityofaustin/atd-vz-data/pull/1048), but I'll wait to merge it until you merge this, unless you'd like otherwise. I'll merge promptly when you do this one, so we can get your work into production.

frankhereford:
@mddilley wrote:

> Can y'all give me a quick 👍 or 👎 on [this commit](https://github.com/cityofaustin/atd-vz-data/pull/1171/commits/5aadfee9d2ba700ebea228192ebcafd0e0adacc9).

Looks perfect, wouldn't change a thing. Thanks for bringing the maintainability to our airflow ETL container!

johnclary:
+1 the requirements look great—glad you caught that issue with the crypto lib 🚢

patrickm02L (approved):
Approved based on Devs approvals. 🪄

PR #1175, opened 2023-02-09
manually adjust 2022 pedestrian fatalities

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/11391

this simple helper function is a temporary measure to reconcile the data displayed in VZV with the data the VZ Team is using based on their analyses. ultimately the data responsible for the discrepancy will be reconfigurable via the Layer Data Model, at which point this function will need to be removed. by its nature, this code goes against established patterns by hardcoding (in this case incrementing) changes to specific fields in the data display, but i did not see an alternative and did not want to spend too much time on this since it is a temporary measure. however, I welcome any comments to improve this code!

## Testing
**URL to test:** 
https://deploy-preview-1175--atd-vzv-staging.netlify.app/

**Steps to test:**
- compare prod and deploy preview
- 2022 pedestrian fatalities in deploy preview should be 136 (all) and 48 (fatalities), incremented by one over prod
- all other numbers should be unchanged

before
Screen Shot 2023-02-09 at 12 52 25 PM

after
Screen Shot 2023-02-09 at 12 52 38 PM

before
Screen Shot 2023-02-09 at 12 52 47 PM

after
Screen Shot 2023-02-09 at 12 53 18 PM

before
Screen Shot 2023-02-09 at 12 53 30 PM

after (no change)
Screen Shot 2023-02-09 at 12 53 38 PM

---
#### Ship list
- [x] Code reviewed

### DIFF
--- a/atd-vzv/src/views/summary/CrashesByMode.js
+++ b/atd-vzv/src/views/summary/CrashesByMode.js
@@ -134,6 +134,15 @@ const CrashesByMode = () => {
       }, 0);
     });
 
+  const adjustModeData = (modeData) => {
+    // TEMPORARY MEASURE!!
+    // Manually adjust mode data displayed for 2022 pedestrian fatalities to match VZ Team's analysis
+    if (crashType.name === "fatalitiesAndSeriousInjuries" || crashType.name === "fatalities") {
+      modeData[1].data[3] = modeData[1].data[3] + 1
+    }
+    return modeData;
+  }
+
   // Sort mode order in stack and apply colors by averaging total mode fatalities across all years in chart
   const sortAndColorModeData = (modeData) => {
     modeData.forEach((category, i) => {
@@ -143,7 +152,7 @@ const CrashesByMode = () => {
       category.hoverBackgroundColor = color;
       category.hoverBorderColor = color;
     });
-    return modeData;
+    return adjustModeData(modeData);
   };
 
   // Create dataset for each mode type, data property is an array of fatality sums sorted chronologically


atd-vzv/src/views/summary/CrashesByMode.js
```diff
@@ -134,6 +134,15 @@ const CrashesByMode = () => {
       }, 0);
     });
 
+  const adjustModeData = (modeData) => {
+    // TEMPORARY MEASURE!!
```
frankhereford, line 138: ❤️ Thank you for this!

atd-vzv/src/views/summary/CrashesByMode.js
```diff
@@ -134,6 +134,15 @@ const CrashesByMode = () => {
...
+    // Manually adjust mode data displayed for 2022 pedestrian fatalities to match VZ Team's analysis
+    if (crashType.name === "fatalitiesAndSeriousInjuries" || crashType.name === "fatalities") {
+      modeData[1].data[3] = modeData[1].data[3] + 1
```
frankhereford, line 141: The path of least resistance is what makes the river crooked.

I want to reiterate my support for this measure in as much as we follow up on finding the real, root cause. Thank you Tilly for this work, and to everyone for our discussion where we continue to value generalized solutions that fully comprehend and address discrepancies in calculations.

frankhereford (approved):
I really like how you even wrapped the temporary solution in a function instead of burying it in the JSX or something similar. Thank you!

Box this up and ship it! 📦🚛

atd-vzv/src/views/summary/CrashesByMode.js
frankhereford: Another thought - I think this will offset "last year" and not 2022 explicitly. In this particular case, I think that is fine, as this is temporary, but we'll be scratching our heads pretty good if we don't address this before 2024.

tillyw: Thanks, @frankhereford !! re: your comment about "last year" - that was my main reservation with this code, but since the data is calculated as current year minus a year, minus two years, etc., this was the only way i could think of to do this. I could implement a failsafe that disables this when current year is 2024, but i thought the chances of us getting there before we implement a long-term fix were pretty slim. still, happy to do so if that makes everyone more comfortable.

frankhereford: You're way ahead of me! I think this is totally fine to have a ticking "reminder" that we need to address it. 

Makes me wish we could put a "become bright read and start flashing at a certain date" reminder on issues.. GitHub, are you out there? Can you hear me? 

Thanks Tilly!!

chiaberry (approved):
I agree with Frank's review, this is cleanly done and I super appreciate the comments throughout. 

@patrickm02L should this be a patch to prod, or wait until next official VZ release for it to make it to production?

patrickm02L:
This should be a patch to prod. Next release will be closer to March or April.

patrickm02L (approved):
This is good to go. Thanks for providing the images in the PR for quick scanning! 🚢

PR #1178, opened 2023-02-21
VZ Development Environment fast-track into master

frankhereford:
This PR contains the same content as [PR 1150](https://github.com/cityofaustin/atd-vz-data/pull/1150). The intent is to not have to wait until the Layer Data Model collector branch is entirely complete to get these changes into `master`. This is done in support of @roseeichelmann's upcoming PR in which she has used this tooling to provide a local development environment and thus to may benefit from not having to including the complexity of jumping around in various branches in her testing instructions.

---
#### Ship list
- [ ] Code reviewed


chiaberry (approved):
Such a useful set up, I agree with getting it into `master` so we can use it to test @roseeichelmann 's PR (and future db PRs)

PR #1179, opened 2023-02-23
Database -- automate insertion into fatalities table

roseeichelmann:
## Associated issues
Closes cityofaustin/atd-data-tech/issues/11337

## Testing
**URL to test:** 
You will test this using Frank's new local development steps to replicate the production db onto your local machine and then add my DB triggers.

**Steps to test:**
1. Checkout this branch locally
2. In the root of the branch follow instructions in the README to start up a local instance of the VZ stack
3. Make sure you un-comment [this](https://github.com/cityofaustin/atd-vz-data/blob/e09b9ad73c9c5e5880ad9757975b930ec01fbec9/atd-vze/.env#L2) line and comment out the other one, I also had to change the "https" to "http" or else I get an Network connection error when trying to load data in the VZE.
4. You will need to add the `is_deleted` column to the `fatalities` table. It is a boolean and set it to default = false.
5. Add the new fatalities [view](https://github.com/cityofaustin/atd-vz-data/blob/3c9c5dfdce272759a01fb059198d15d06908f059/atd-vzd/views/view_fatalities.sql) to your local DB. Theres a SQL editor in the hasura console if you wanna use that, or use DBeaver.
6. Open up your localhost:8080 instance of the hasura console. Go to the new `view_fatalities` view you created and manually add a relationship to the recommendations table like so:
Screenshot from 2023-03-08 12-42-14
7. Copy and paste the triggers and trigger functions from [this](https://github.com/cityofaustin/atd-vz-data/files/10926497/fatalities_db_automation.txt) file into the hasura SQL editor or DBeaver. This will add the triggers and functions into the DB and their respective tables.
8. Now go to your localhost:3000 instance of the VZE. Go to the Fatalities tab, you will see an empty table.
9. In your local graphql engine test adding fatal entries to the `atd_txdot_primaryperson` or `atd_txdot_person` tables where the `prsn_injry_sev_id = 4`. This will add a fatality for that person to the `fatalities` table. If the fatality also occurred within the Austin jurisdiction, the fatality will appear in the `view_fatality` and thus in the front end Fatalities table in VZE.
10. Test deleting the entry you made in the `primary_person` or `person` table, this should also delete the entry in the `fatalities` table.
11. Deleting the crash from the `atd_txdot_crashes` table will also delete the record in the `fatalities` table.
---
#### Ship list
- [x] Code reviewed


atd-vzd/schema/atd_txdot_person.sql
```diff
@@ -67,15 +64,18 @@ CREATE TABLE public.atd_txdot_person (
     updated_by character varying(64),
     person_id integer NOT NULL,
     is_retired boolean DEFAULT false,
-    years_of_life_lost integer DEFAULT 0 NOT NULL
+    years_of_life_lost integer GENERATED ALWAYS AS (
+CASE
+    WHEN (prsn_injry_sev_id = 4) THEN GREATEST((75 - prsn_age), 0)
+    ELSE 0
+END) STORED
```
roseeichelmann, line 71: a lot of these changes arent mine this schema just hasnt been updated in this folder in awhile.

atd-vzd/schema/atd_txdot_person.sql
```diff
@@ -87,69 +87,147 @@ CREATE SEQUENCE public.atd_txdot_person_person_id_seq
...
+CREATE TRIGGER atd_txdot_person_audit_log BEFORE UPDATE ON public.atd_txdot_person FOR EACH ROW EXECUTE FUNCTION public.atd_txdot_person_updates_audit_log();
+
+ALTER TABLE public.atd_txdot_person DISABLE TRIGGER atd_txdot_person_audit_log;
+
+
+--
+-- Name: atd_txdot_person atd_txdot_person_fatal_insert; Type: TRIGGER; Schema: public; Owner: visionzero
+--
+
+CREATE TRIGGER atd_txdot_person_fatal_insert AFTER INSERT ON public.atd_txdot_person FOR EACH ROW WHEN ((new.prsn_injry_sev_id = 4)) EXECUTE FUNCTION public.fatality_insert();
```
roseeichelmann, line 220: here is the trigger i added!

atd-vzd/schema/atd_txdot_primaryperson.sql
```diff
@@ -89,60 +89,137 @@ CREATE SEQUENCE public.atd_txdot_primaryperson_primaryperson_id_seq
...
+CREATE TRIGGER atd_txdot_primaryperson_audit_log BEFORE UPDATE ON public.atd_txdot_primaryperson FOR EACH ROW EXECUTE FUNCTION public.atd_txdot_primaryperson_updates_audit_log();
 
 
--- Completed on 2019-10-15 13:46:38 CDT
+--
+-- Name: atd_txdot_primaryperson atd_txdot_primaryperson_fatal_insert; Type: TRIGGER; Schema: public; Owner: visionzero
+--
+
+CREATE TRIGGER atd_txdot_primaryperson_fatal_insert AFTER INSERT ON public.atd_txdot_primaryperson FOR EACH ROW WHEN ((new.prsn_injry_sev_id = 4)) EXECUTE FUNCTION public.fatality_insert();
+
```
roseeichelmann, line 221: this is the trigger i added

atd-vzd/triggers/fatality_insert.sql
```diff
@@ -0,0 +1,29 @@
...
+        location text;
+    BEGIN
+    	SELECT TO_CHAR(crash_date,'yyyy')
+            FROM atd_txdot_crashes
+            WHERE crash_id = NEW.crash_id
+            INTO year;
+        SELECT CONCAT_WS(' ', rpt_block_num, rpt_street_pfx, rpt_street_name, '(', rpt_sec_block_num, rpt_sec_street_pfx, rpt_sec_street_name, ')')
+            FROM atd_txdot_crashes
+            WHERE crash_id = NEW.crash_id
+            INTO location;
```
roseeichelmann, line 18: this creates the new location field that the VZ team wants in the fatality table

atd-vzd/schema/atd_txdot_person.sql
```diff
@@ -87,69 +87,147 @@ CREATE SEQUENCE public.atd_txdot_person_person_id_seq
...
 --
--- TOC entry 5625 (class 2620 OID 2346816)
--- Name: atd_txdot_person atd_txdot_person_audit_log; Type: TRIGGER; Schema: public; Owner: atd_vz_data
+-- Name: atd_txdot_person atd_txdot_person_audit_log; Type: TRIGGER; Schema: public; Owner: visionzero
 --
 
-CREATE TRIGGER atd_txdot_person_audit_log BEFORE UPDATE ON public.atd_txdot_person FOR EACH ROW EXECUTE PROCEDURE public.atd_txdot_person_updates_audit_log();
+CREATE TRIGGER atd_txdot_person_audit_log BEFORE UPDATE ON public.atd_txdot_person FOR EACH ROW EXECUTE FUNCTION public.atd_txdot_person_updates_audit_log();
+
+ALTER TABLE public.atd_txdot_person DISABLE TRIGGER atd_txdot_person_audit_log;
```
mddilley, line 213: Not related to your changes, but it is interesting that we have this trigger defined and then immediately disabled here. 

Makes me wonder if this is more cruft that @frankhereford would like a look at. Maybe we already know about this or there is more to it that I'm seeing.

atd-vzd/triggers/fatality_insert.sql
```diff
@@ -0,0 +1,29 @@
...
+    LANGUAGE plpgsql
+AS $function$
+    DECLARE
+    	year integer;
+        location text;
+    BEGIN
+    	SELECT TO_CHAR(crash_date,'yyyy')
+            FROM atd_txdot_crashes
+            WHERE crash_id = NEW.crash_id
+            INTO year;
```
mddilley, line 14: Since this year is created from `crash_date` in the crashes table, I'm curious if this could be handled in the front end to keep us from storing the year info twice in the DB. Not sure what our options are in the React code so disregard if this is the best way. 👍

The location field makes sense to me since that is more complex.

mddilley (approved):
This works so nicely, and thanks for taking your time because it is giving the rest of us the chance to learn along the way too. The test steps worked great, and I was able to trigger creates and deletes in the new table. 🎉 I added a couple questions, but nothing that should hold this one back! 🚀 Also, super glad to see the local environment already paying off 🙌

atd-vzd/triggers/fatality_insert.sql
mddilley: I'm thinking about this some more now that I submitted review, but should we think about how crash records edits could cause the data in this table to drift from the source? Thinking about `year` and `location` specifically.

roseeichelmann:
I just updated this PR to check if the crash is within AFP before adding to the fatality table. This PR should be good to go now!

atd-vzd/triggers/fatality_insert.sql
```diff
@@ -0,0 +1,26 @@
...
+CREATE OR REPLACE FUNCTION fatality_insert() 
+    RETURNS trigger
+    LANGUAGE plpgsql
+AS $function$
+    DECLARE
+    	year integer;
+        location text;
+        crash_position geometry;
+    BEGIN
+        SELECT TO_CHAR(crash_date,'yyyy'), CONCAT_WS(' ', rpt_block_num, rpt_street_pfx, rpt_street_name, '(', rpt_sec_block_num, rpt_sec_street_pfx, rpt_sec_street_name, ')'), position
```
johnclary, line 12: if y'all use this same concatenation elsewhere, it might be best to add it as a generated column on the `atd_txdot_crashes` table.

atd-vzd/schema/fatalities.sql
```diff
@@ -0,0 +1,122 @@
...
+    crash_id integer,
+    location text,
+    victim_name text,
+    law_enforcement_num integer,
+    ytd_fatal_crash integer,
+    ytd_fatality integer,
+    person_id integer,
+    primaryperson_id integer,
+    year text
+);
```
johnclary, line 38: If each row must have either a `person_id` or `primaryperson_id` (and not both?), you could consider adding a [check constraint](https://www.postgresql.org/docs/9.0/ddl-constraints.html#AEN2385) that will enforce that rule. 

A `comment` on the table describing it's purpose would be a nice touch, too.

atd-vzd/triggers/fatality_insert.sql
```diff
@@ -0,0 +1,26 @@
...
+            INTO year, location, crash_position
+            FROM atd_txdot_crashes
+            WHERE crash_id = NEW.crash_id;
+    	IF (TG_TABLE_NAME = 'atd_txdot_primaryperson' AND ST_CONTAINS((SELECT geometry FROM atd_jurisdictions WHERE atd_jurisdictions.id = 5), crash_position)) THEN
+    	    INSERT INTO fatalities (crash_id, primaryperson_id, year, location)
+    		    VALUES (NEW.crash_id, NEW.primaryperson_id, year, location);
+    	ELSIF (TG_TABLE_NAME = 'atd_txdot_person' AND ST_CONTAINS((SELECT geometry FROM atd_jurisdictions WHERE atd_jurisdictions.id = 5), crash_position)) THEN
+    		INSERT INTO fatalities (crash_id, person_id, year, location)
+    		    VALUES (NEW.crash_id, NEW.person_id, year, location);
```
johnclary, line 21: there are quite a few more columns in the `fatalities` table. will those be populated users through VZE?

johnclary (approved):
@roseeichelmann this is amazing work! SQL functions are really tough, and you've made it look easy here.

I ran through your test steps and everything worked exactly as expected. 

Please ignore me if y'all don't want to rehash old conversation, but I do have some doubts about trying to keep the location, year, and position data in sync with the fatalities table.

My instinct for "normalized data" here would be to:

1. Use triggers only to insert new records into the fatality table 
2. Remove year, position, and location from the fatalities table and instead use a view to join up those columns - as well as to filter out fatalities outside of the jurisdiction of concern.

Not only does that remove the potential issues with data drifting/de-syncing from the reference tables, it also makes future tweaks to the data model much easier to manage. E.g. if users want to change to geospatial bounds of the query, or if they want to change the concatenation format of that location description. 

I know the biggest drawback to the view-based approach will be performance, but maybe we could get by with a materialized view that updates when new falities occur? 

One last thought - it might be good to have an `is_deleted` flag on the fatalities table, for edge cases when a person injury severity is demoted? I don't know if/when this would happen.

I'm hitting this with ✅ approved because it works great and I don't know the whole background on this feature.

chiaberry (approved):
I really appreciate the thorough testing instructions. And I agree with John, you made this all look easy!

atd-vzd/schema/fatalities.sql
roseeichelmann: ah yeah thats a great point, i will add that!

roseeichelmann: and correct its either not both

atd-vzd/triggers/fatality_insert.sql
roseeichelmann: TIL generated columns

roseeichelmann:
@johnclary thanks for the super thoughtful review! i agree that the location and year columns aren't ideal. when i originally made the fatalities table i thought about having those columns be handled on the front end like mike suggested but i didnt know how to do that with the [GridTable](https://github.com/cityofaustin/atd-vz-data/blob/master/atd-vze/src/Components/GridTable.js) component we use to generate the Fatalities table in the UI. With the GridTable you just give it the columns from the table in the DB you want it to display and it doesn't seem to have functionality for formatting/doing calculations on the data. im also not sure how to incorporate a materialized view into the query of the fatalities table thats fed to the GridTable. i did address possible desyncing in [this PR](https://github.com/cityofaustin/atd-vz-data/pull/1184) but you're right, doing it this way still leaves room for error in the trigger process and makes it more complicated to adjust those columns like how the `location` is concatenated. in dev sync today we talked about going forward with this current implementation and maybe having those database updates you described as future enhancements once we get this initial feature out the door? but definitely want to explore your suggestions

mddilley (approved):
Thanks for updating this and generating a lot of discussion here, Rose! It is interesting to see history repeat itself in finding our limits when it comes to displaying the data that we have in the way that we want to see it. Definitely has me think about how to make that better. I tested the steps again locally, and they all work great. 🚀 

I like John's feedback, and I think it should be incorporated into future updates if we need to keep this moving. It definitely feels right especially after demoing the pitfalls of the mode metadata field in the crash table today.

roseeichelmann:
i talked to john yesterday and he showed me how i would use a view in this situation and it totally clicked for me, so I'm gonna adjust this feature to implement the view now rather than later! thanks for the helpful feedback yall :)

roseeichelmann:
this PR is ready for review once again with updated instructions :-) 

the "files changed" tab isnt rly accurately representing the current state of what files are different from branch 11190_mvp_ftt_index bc ive done so much commit reverting and branch rebasing. for example `view_fatalities.sql` doesnt exist in branch 11190_mvp_ftt_index but the changelog in this PR is acting like that file was only modified when it was completely added. so hopefully it isnt too confusing for yall to look at the changes.

the fatalities page now pulls from a view instead of the fatalities table which will be much easier to manage and keep data from desyncing. i also added an `is_deleted` field to the fatalities table for soft-deletes and the view will only pull fatalities where `is_deleted = false.`

atd-vze/src/views/Fatalities/Fatalities.js
```diff
@@ -13,7 +13,7 @@ let queryConf = {
       fetchPolicy: "no-cache",
     },
   },
-  table: "fatalities",
+  table: "view_fatalities",
```
roseeichelmann, line 16: using the view instead of the fatalities table now

atd-vzd/schema/fatalities.sql
```diff
@@ -0,0 +1,128 @@
...
+
+
+ALTER TABLE public.fatalities OWNER TO visionzero;
+
+--
+-- Name: CONSTRAINT either_primaryperson_or_person ON fatalities; Type: COMMENT; Schema: public; Owner: visionzero
+--
+
+COMMENT ON CONSTRAINT either_primaryperson_or_person ON public.fatalities IS 'Each fatality must have either a primary person or person ID but not both';
```
roseeichelmann, line 46: @johnclary added a constraint... idk if this is what you meant by comment but TIL the COMMENT command

atd-vzd/views/view_fatalities.sql
```diff
@@ -0,0 +1,45 @@
+-- view of fatalities in jursidiction ID with crash attributes joined
```
johnclary, line 1: type here: "...in jurisdiction ID 5 (austin full purpose)..."

atd-vzd/views/view_fatalities.sql
```diff
@@ -0,0 +1,45 @@
...
+WHERE
+    ST_CONTAINS((
+        SELECT
+            geometry FROM atd_jurisdictions
+        WHERE
+            atd_jurisdictions.id = 5),
+        fatalities_plus_crash.position)
+AND 
+    fatalities_plus_crash.is_deleted = false
+);
```
johnclary, line 45: to recap recent conversation—if we were confident in the `atd_txdot_crashes.austin_full_purpose` column, we could do away with this spatial filter. also be advised that this filter uses `atd_jurisdictions.id = 5` while the `austin_full_purpose` column includes [a few additional jurisdictions](https://github.com/cityofaustin/atd-vz-data/blob/53ddfccd3c24b81a9a35ed4af766d68ca02cd971/atd-events/crash_update_jurisdiction/app.py#L158).

atd-vzd/schema/fatalities.sql
```diff
@@ -0,0 +1,128 @@
...
+
+--
+-- Name: fatalities; Type: TABLE; Schema: public; Owner: visionzero
+--
+
+CREATE TABLE public.fatalities (
+    id integer NOT NULL,
+    crash_id integer,
+    victim_name text,
+    law_enforcement_num integer,
```
johnclary, line 31: is `law_enforcement_num` unique to the individual person, or to the crash?

atd-vzd/schema/fatalities.sql
johnclary: just want to point out that this table is already in staging/prod. so it will need to be dropped and re-added and re-tracked in hasura.

also, this sql is missing the `is_deleted` column.

johnclary:
@roseeichelmann one more thing: since you're already updating all of the files in the `/schema` folder with the latest snapshot—you might as well go ahead and update the [hasura metadata](https://github.com/cityofaustin/atd-vz-data/blob/master/atd-vzd/hasura/metadata.json) json as well?

atd-vzd/schema/fatalities.sql
roseeichelmann: this is a new field the VZ team wants added that they will be adding in the UI themselves and i dont have much insight into it yet. buuut i think its unique to a crash not a person because the place they want to eventually add functionality to add/edit is on a crash page. so now you have me thinking that this column can be removed from the fatalities table and instead needs to exist in the crashes table and the view will pull from it

johnclary: 👍 yep—that sounds right to me!

chiaberry: I went back to check their sharepoint, and the law_enforcement_num is taking the place of what they currently call "APD Fatal Crash" which is a number that APD assigns to crashes, like their YTD crash number. (Not YTD Fatality, so I agree crashes table sounds right)

atd-vzd/schema/atd_txdot_crashes.sql
```diff
@@ -219,84 +216,209 @@ CREATE TABLE public.atd_txdot_crashes (
...
+    austin_full_purpose character varying(1) DEFAULT 'N'::character varying NOT NULL,
+    original_city_id integer,
+    atd_fatality_count integer,
+    temp_record boolean DEFAULT false,
+    cr3_file_metadata jsonb,
+    cr3_ocr_extraction_date timestamp with time zone,
+    investigator_narrative_ocr text,
+    est_comp_cost_crash_based numeric(10,2) DEFAULT 0,
+    imported_at timestamp without time zone DEFAULT now(),
+    law_enforcement_num integer
```
roseeichelmann, line 240: added this law_enforcement_num column

chiaberry (approved):

atd-vzd/views/view_fatalities.sql
```diff
@@ -0,0 +1,32 @@
+-- view of fatalities in jurisdiction ID 5 (austin full purpose) with crash attributes joined
+CREATE OR REPLACE VIEW view_fatalities AS (
```
frankhereford, line 2: This is really not a big deal, so please disregard if you want. I'm wondering what you think about naming the view with something that's more focused on what the data inside is? Maybe something like `fatalities_in_afp` or maybe `ftt_fatalities`. I'm not sure these are the best names, so take this with a grain of salt.

atd-vzd/views/view_fatalities.sql
```diff
@@ -0,0 +1,32 @@
...
+        f.id,
+        f.crash_id,
+        f.victim_name,
+        f.ytd_fatal_crash,
+        f.ytd_fatality,
+        f.person_id,
+        f.primaryperson_id,
+        TO_CHAR(crashes.crash_date,
+            'yyyy') AS year,
+        CONCAT_WS(' ',
```
frankhereford, line 13: TIL `CONCAT_WS()`! I have needed this about a billion times and always did a messy `CONTACT()`. This is great.

atd-vzd/schema/atd_txdot_person.sql
frankhereford: Good question! I wonder if this SQL was taken from a `pg_dump` which will embed the enabling and disabling of triggers as needed so that the ultimate final result after the entire output is what the DB was when the export was made. Like, disabling triggers before all the `insert`s and then re-enabling them. 

In this case, I think it will leave the database with that trigger disabled, so this line should probably be removed. Does that seem right @roseeichelmann?

atd-vzd/schema/atd_txdot_crashes.sql
```diff
@@ -219,84 +216,209 @@ CREATE TABLE public.atd_txdot_crashes (
...
 );
 
 
-ALTER TABLE public.atd_txdot_crashes OWNER TO atd_vz_data;
+ALTER TABLE public.atd_txdot_crashes OWNER TO visionzero;
```
frankhereford, line 244: I would advise removing all of these ownership `ALTER`s. I don't think the intent is to shift around ownership, and I am pretty sure we use totally different role names in production vs local.

That said, I bet they will fail harmlessly if left in.

atd-vzd/schema/atd_txdot_crashes.sql
```diff
@@ -20,15 +18,14 @@ SET row_security = off;
...
-SET default_with_oids = false;
+SET default_table_access_method = heap;
 
 --
--- TOC entry 219 (class 1259 OID 2345633)
--- Name: atd_txdot_crashes; Type: TABLE; Schema: public; Owner: atd_vz_data
+-- Name: atd_txdot_crashes; Type: TABLE; Schema: public; Owner: visionzero
 --
 
 CREATE TABLE public.atd_txdot_crashes (
```
frankhereford, line 27: I'd start each of these table definitions right here on the `CREATE`.

atd-vzd/schema/atd_txdot_crashes.sql
```diff
@@ -219,84 +216,209 @@ CREATE TABLE public.atd_txdot_crashes (
...
+
+CREATE INDEX atd_txdot_crashes_sus_serious_injry_cnt_index ON public.atd_txdot_crashes USING btree (sus_serious_injry_cnt);
+
+
+--
+-- Name: atd_txdot_crashes_temp_record_index; Type: INDEX; Schema: public; Owner: visionzero
 --
 
-CREATE UNIQUE INDEX idx_atd_txdot_crashes_crash_id ON public.atd_txdot_crashes USING btree (crash_id);
+CREATE INDEX atd_txdot_crashes_temp_record_index ON public.atd_txdot_crashes USING btree (temp_record);
```
frankhereford, line 384: This freaked me out for a second because of the unique index is removed here. But, it is fulfilled by the setting of `crash_id` as the primary key, which will get its own `btree` index.

atd-vzd/schema/atd_txdot_crashes.sql
```diff
@@ -219,84 +216,209 @@ CREATE TABLE public.atd_txdot_crashes (
...
 
 
 --
--- TOC entry 5634 (class 2620 OID 2346813)
--- Name: atd_txdot_crashes atd_txdot_crashes_audit_log; Type: TRIGGER; Schema: public; Owner: atd_vz_data
+-- Name: atd_txdot_crashes atd_txdot_crashes_audit_log; Type: TRIGGER; Schema: public; Owner: visionzero
 --
 
-CREATE TRIGGER atd_txdot_crashes_audit_log BEFORE INSERT OR UPDATE ON public.atd_txdot_crashes FOR EACH ROW EXECUTE PROCEDURE public.atd_txdot_crashes_updates_audit_log();
+CREATE TRIGGER atd_txdot_crashes_audit_log BEFORE INSERT OR UPDATE ON public.atd_txdot_crashes FOR EACH ROW EXECUTE FUNCTION public.atd_txdot_crashes_updates_audit_log();
```
frankhereford, line 391: TIL the difference here. "Procedures do not return a function value; hence CREATE PROCEDURE lacks a RETURNS clause." Very cool.

atd-vzd/schema/fatalities.sql
```diff
@@ -0,0 +1,128 @@
...
+-- Name: fatalities_id_seq; Type: SEQUENCE; Schema: public; Owner: visionzero
+--
+
+CREATE SEQUENCE public.fatalities_id_seq
+    AS integer
+    START WITH 1
+    INCREMENT BY 1
+    NO MINVALUE
+    NO MAXVALUE
+    CACHE 1;
```
frankhereford, line 59: I don't think this is going to be a problem, but one thing that is a gotcha that I've run into before is if you somehow have this table around with records in it, and then reset its sequence to dole out IDs that end up causing inserts to fail because of a primary key or unique constraint on the ID field.

Here, where you're creating the table just above, I think you can assume that this sequence is created at the same time and will start with 1.

frankhereford (approved):
Rose, I put some comments in there with suggestions on ways to shrink the SQL files down by trimming out things that may have come out of `pg_dump` that don't help as much in a single table definition. 

I tested this on an extract of data from today and then via use in the VZE and through `INSERT`s and `DELETE`s. It worked just perfectly.

Thank you for writing up such good testing instructions. That was helpful as can be. Also, thank you to everyone who has provided thoughtful feedback on this PR! This has been a great read.

atd-vzd/schema/atd_txdot_person.sql
roseeichelmann: honestly im not sure, i just did pg_dumps of these tables bc i saw they havent been updated in a long time and figured i might as well. now im thinking we should do pg_dumps of the production db and commit them after the release of this feature and all the needed db changes in prod are made. idk why my `pg_dump` would disable this trigger though?

atd-vzd/schema/fatalities.sql
roseeichelmann: huh interesting, yeah i just used the hasura console to make this table and used the default -> nextval selection for the `id` and this is how hasura automatically writes the sql for that i guess. i cant really think of how the sequence would get reset though?

atd-vzd/views/view_fatalities.sql
roseeichelmann: yeah that makes sense to me too. i was trying to keep in pattern with some of the naming of other views i saw in the db but theres not really a set convention, and specifying ftt or in afp would be more descriptive.

atd-vzd/schema/atd_txdot_person.sql
frankhereford: I think what happens is that `pg_dump` is working in logical phases, sort of like this, because it's operating in a bigger context of "my job is to create the whole DB in a certain state from whole cloth."

* Enter into create the schema mode
  * Create the table
  * Setup indexes, constraints, triggers, etc for the table
* Exit create schema mode
* Enter into populate data mode
  * Disable triggers
  * Insert the data, without the triggers firing for each record
  * Re-enable triggers
* Exit populate data mode

I think when you captured these files out of `pg_dump`'s output, you may have caught the first part of populate data mode, which has that disable statement. I think you can cut off where you're pulling from the `pg_dump` a little earlier and be good. I'd be super happy to sync if that would be helpful - I am probably explaining this poorly.

atd-vzd/schema/atd_txdot_crashes.sql
frankhereford: Rose and I got to sync a few moments ago, and she explained to me that the form of this file is actually matching the other files in this folder, and we looked at it, and she helped me realize that all of the stuff above the CREATE is actually fine, and we agreed that `pg_dump` knows what it's doing. I think my above suggestion isn't helpful should be struck. Thank you @roseeichelmann!

atd-vzd/views/view_fatalities.sql
johnclary: For posterity—my previous comment is incorrect. Refer to [this thread](https://austininnovation.slack.com/archives/CM9MK950S/p1678398208097029) for more background on the austin_full_purpose field.

patrickm02L (approved):
Adding my rubber stamp based on the Dev reviews! 🚢

atd-vze/src/views/Fatalities/Fatalities.js
```diff
@@ -25,12 +25,8 @@ let queryConf = {
 
 const fatalitiesQuery = new gqlAbstract(queryConf);
 
-const minDate = subYears(new Date(), 10);
```
mddilley, line 28: Little tiny thing - just noticed that removing this line leaves an unused import behind which causes a compile warning.

mddilley (approved):
Tested all of the steps and the new view is a go! 🚀 🚢  🎉

atd-vzd/views/view_fatalities.sql
roseeichelmann: yeah i was stoked to find this function it made my life a lot easier!

atd-vze/src/views/Fatalities/Fatalities.js
roseeichelmann: good catch! i removed that import in the new PR i just opened so that should clear up that compile warning. thank you!

PR #1181, opened 2023-03-01
 Fill fatalities table with data from existing crash information

roseeichelmann:
## Associated issues
Closes cityofaustin/atd-data-tech/issues/11486

## Testing
**URL to test:** 
Test this with a local VZ DB environment setup

**Steps to test:**
1. Checkout this branch locally
2. Run the queries in [fatalities_backfill.sql](https://github.com/cityofaustin/atd-vz-data/pull/1181/files) within your local VZ DB to populate the fatalities table with data
3. All persons and primary persons with a `prsn_injry_sev_id` of 4 should be present in the fatalities table.

Not sure if putting this script in the atd-toolbox folder long-term makes sense but just put it here for now for y'alls ease of testing.

---
#### Ship list
- [ ] Code reviewed


atd-toolbox/fatalities_backfill/fatalities_backfill.sql
```diff
@@ -0,0 +1,13 @@
+-- This is a one time script to populate the fatalities table with historical data
+
+INSERT INTO fatalities (crash_id, primaryperson_id, location, year)
+SELECT atd_txdot_primaryperson.crash_id, primaryperson_id, CONCAT_WS(' ', rpt_block_num, rpt_street_pfx, rpt_street_name, '(', rpt_sec_block_num, rpt_sec_street_pfx, rpt_sec_street_name, ')'), TO_CHAR(crash_date,'yyyy')
+FROM atd_txdot_primaryperson
+JOIN atd_txdot_crashes ON atd_txdot_primaryperson.crash_id = atd_txdot_crashes.crash_id
+WHERE prsn_injry_sev_id = 4 AND ST_CONTAINS((SELECT geometry FROM atd_jurisdictions WHERE atd_jurisdictions.id = 5), position);
```
roseeichelmann, line 7: the `ST_CONTAINS` checks if crash was in austin

chiaberry (approved):
ran the script, also hadnt deleted a crash from testing the first trigger PR so I inadvertently tested the unique key constraint.

I like keeping it in the toolbox so if there are any questions, its easy to find what the script was that filled the table.

atd-toolbox/fatalities_backfill/fatalities_backfill.sql
mddilley: I didn't catch this in the triggers PR, but is there a reason to use `ST_CONTAINS` with this condition as opposed to the `austin_full_purpose` flag? Not being critical but want to know for my own understanding. 🙏

mddilley (approved):
Tested this out on my local stack, and all looks in order! The utility of this table is really coming together when looking at it with real data. It is going to be so useful to the VZ team - thanks for making this happen! 🙏🚢

I know that this is largely up to the VZ team, but I noticed that it can be hard to tell rows apart when multiple fatalities happen in a single crash like crash ID `18391851`. I started wondering if it would be useful to add the person ID, but maybe that will be less important when the names come into this view. Just a thought and no changes needed.

roseeichelmann:
update to this PR: 
its basically the same but i removed the austin jurisdiction check bc now the view handles filtering out fatalities outside of austin

mddilley:
@roseeichelmann Thanks! 🙏 I have a small concern (small because LDM is on the way and the improvements you mentioned) about straying from the Austin full purpose flag since [Xavier is using it in his queries](https://github.com/cityofaustin/atd-data-tech/issues/11340#issuecomment-1410961486). Maybe we need to loop Xavier in on this change or maybe all this work will come together at the same time and make this a non-issue. Just something to keep in mind.

johnclary (changes requested):
@roseeichelmann this sql looks right to me, but when i test this against a prod replica i hit an error because it seems we have at least one primary person record that references a crash ID (1023) that doesn't exist in the crashes table :/

```
insert or update on table "fatalities" violates foreign key constraint "fatalities_crash_id_fkey".Key (crash_id)=(1023) is not present in table "atd_txdot_crashes".
````

**Comment by johnclary** [comment]
+1 what Mike said—i think it'd be good to double-check with Xavier and team on the spatial filter, which is now being applied [here](https://github.com/cityofaustin/atd-vz-data/pull/1179#discussion_r1131258982).

**Comment by johnclary** [comment]
please ignore the error message [i flagged](https://github.com/cityofaustin/atd-vz-data/pull/1181#pullrequestreview-1333225359). that was crufty data i created myself 🤦

**Review by johnclary — APPROVED** [review]
🚢

**Comment by tillyw** [comment]
i ran this script on the prod read replica and ran into the error that everyone has been discussing. i don't know what the ultimate solution will be for these records but when i delete record `1023` from the db everything runs smoothly and the fatalities table is populated as expected!

**Comment by roseeichelmann** [comment]
to address @tillyw and @johnclary comments, the problem stemmed from the fact that when a temp crash is created that has a fatality or serious injury associated with it, a record to count that fatality/injury is created in the primary person table. this is so that the VZ team can keep accurate counts of deaths/injuries before the official data from CRIS comes in. when it does, they go ahead and delete the temp crash. however, associated person records are not deleted on temp crash delete, so this crufty data just sits around. this code surfaces that error because the fatality table expects a foreign key relationship to the crashes table for each record but temp crash 1023 doesnt exist in the crashes table anymore. the solution is to 1. delete the crufty person records and 2. add a foreign key relationship between the persons table and crashes table that cascades crash deletes so that the delete of a temp crash for example will automatically delete the person record. ive done both of these steps in staging and prod already so this error wont surface anymore. thanks for the help yall!

**Comment by tillyw** [comment]
> to address @tillyw and @johnclary comments, the problem stemmed from the fact that when a temp crash is created that has a fatality or serious injury associated with it, a record to count that fatality/injury is created in the primary person table. this is so that the VZ team can keep accurate counts of deaths/injuries before the official data from CRIS comes in. when it does, they go ahead and delete the temp crash. however, associated person records are not deleted on temp crash delete, so this crufty data just sits around. this code surfaces that error because the fatality table expects a foreign key relationship to the crashes table for each record but temp crash 1023 doesnt exist in the crashes table anymore. the solution is to 1. delete the crufty person records and 2. add a foreign key relationship between the persons table and crashes table that cascades crash deletes so that the delete of a temp crash for example will automatically delete the person record. ive done both of these steps in staging and prod already so this error wont surface anymore. thanks for the help yall!

this is perfect ✨ thank you @roseeichelmann !!

File: atd-toolbox/fatalities_backfill/fatalities_backfill.sql [hunk 2]
```diff
@@ -0,0 +1,11 @@
...
+
+INSERT INTO fatalities (crash_id, primaryperson_id)
+SELECT crash_id, primaryperson_id
+FROM atd_txdot_primaryperson
+WHERE prsn_injry_sev_id = 4;
+
+INSERT INTO fatalities (crash_id, person_id)
+SELECT crash_id, person_id
+FROM atd_txdot_person
+WHERE prsn_injry_sev_id = 4;
```
frankhereford, line 11: Super not important, but if add a new line at the end of the file, GitHub won't flag it as missing and it will be posix compliant.

frankhereford (approved):
# 💯 🎯 

Thanks Rose!

johnclary:
> add a foreign key relationship between the persons table and crashes table

Nice! I'm glad you were able to add that constraint without errors.  🚢 🚢 🚢

patrickm02L (approved):
Adding my concurrence based on Dev reviews. 🚀

mddilley:
Dropping a note that I retested this before moving on to the next one - still works great! 🙏

roseeichelmann:
also @mddilley i was thinking a similar thing about showing the person id bc like you said it is hard to differentiate when its from the same crash. this may not be as big of a deal when names are added but im not sure there will always be names for each fatality at all times. definitely something to bring up to the VZ team! maybe once these PRs are pushed into the next release the VZ team will be able to give us critiques/let us know if adding person ids would be helpful.

atd-toolbox/fatalities_backfill/fatalities_backfill.sql
roseeichelmann: fixed :)

PR #1188, opened 2023-03-13
 Address injury severity change for a person

roseeichelmann:
## Testing
**URL to test:** 

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vzd/triggers/update_fatality_soft_delete.sql
+++ b/atd-vzd/triggers/update_fatality_soft_delete.sql
@@ -1,0 +1,40 @@
+CREATE OR REPLACE FUNCTION update_fatality_soft_delete()
+    RETURNS TRIGGER
+    LANGUAGE plpgsql
+AS $function$
+    BEGIN
+    	IF (TG_TABLE_NAME = 'atd_txdot_primaryperson') THEN
+        	IF (NEW.prsn_injry_sev_id = 4) THEN
+        	    IF (EXISTS (SELECT FROM fatalities WHERE primaryperson_id = NEW.primaryperson_id)) THEN
+                    UPDATE fatalities f
+                    SET is_deleted = false
+                    WHERE (f.primaryperson_id = NEW.primaryperson_id);
+                ELSE
+                    INSERT INTO fatalities (crash_id, primaryperson_id)
+    		        VALUES (NEW.crash_id, NEW.primaryperson_id);
+    		    END IF;
+            ELSE
+                UPDATE fatalities f
+                SET is_deleted = true
+                WHERE (f.primaryperson_id = NEW.primaryperson_id);
+            END IF;
+        ELSIF (TG_TABLE_NAME = 'atd_txdot_person') THEN
+            IF (NEW.prsn_injry_sev_id = 4) THEN
+                IF (EXISTS (SELECT FROM fatalities WHERE person_id = NEW.person_id)) THEN
+                    UPDATE fatalities f
+                    SET is_deleted = false
+                    WHERE (f.person_id = NEW.person_id);
+                ELSE
+                    INSERT INTO fatalities (crash_id, primaryperson_id)
+    		        VALUES (NEW.crash_id, NEW.primaryperson_id);
+    		    END IF;
+            ELSE
+                UPDATE fatalities f
+                SET is_deleted = true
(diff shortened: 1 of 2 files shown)


frankhereford (approved):
This works great, and I think the code also is right on the money. I wouldn't change a thing.

Thank you for the note in your instructions to include the setup of the DB from the two previous PRs. 

I can't wait to see this all in this release - it's a huge enhancement. Thank you @roseeichelmann.

atd-vzd/triggers/update_fatality_soft_delete.sql
```diff
@@ -0,0 +1,50 @@
...
+                -- if the fatality already exists then undo the soft-delete
+        	    IF (EXISTS (SELECT FROM fatalities WHERE primaryperson_id = NEW.primaryperson_id)) THEN
+                    UPDATE fatalities f
+                    SET is_deleted = false
+                    WHERE (f.primaryperson_id = NEW.primaryperson_id);
+                -- if the fatality doesnt exist yet add it to the table
+                ELSE
+                    INSERT INTO fatalities (crash_id, primaryperson_id)
+    		        VALUES (NEW.crash_id, NEW.primaryperson_id);
+    		    END IF;
```
johnclary, line 23: this works well :) 

if you wanted to get fancy, you could handle both the `if (exists)...else` block with an [upsert](https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-upsert/) into the fatalities table. in that case you'd insert but set `is_deleted = false` on conflict.

johnclary (changes requested):
it's tempting to try and DRY out this code by using variable names for the tables and primary key columns, but in my limited pgsql experience there's not a clean way to do this without `EXECUTE` statements. i really like what you've got! 

that bug with the mixed-up column names is an easy fix, then this is ready to gooo 🚢

atd-vzd/triggers/update_fatality_soft_delete.sql
```diff
@@ -0,0 +1,50 @@
...
+        -- else if table was person then do the same thing
+        ELSIF (TG_TABLE_NAME = 'atd_txdot_person') THEN
+            IF (NEW.prsn_injry_sev_id = 4) THEN
+                IF (EXISTS (SELECT FROM fatalities WHERE person_id = NEW.person_id)) THEN
+                    UPDATE fatalities f
+                    SET is_deleted = false
+                    WHERE (f.person_id = NEW.person_id);
+                ELSE
+                    INSERT INTO fatalities (crash_id, primaryperson_id)
+    		        VALUES (NEW.crash_id, NEW.primaryperson_id);
```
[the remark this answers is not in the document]
roseeichelmann: i sureee did, thanks for catching that! nothing like another set of eyes 👁️

atd-vzd/triggers/update_fatality_soft_delete.sql
roseeichelmann: TIL upsert, that looks a lot cleaner than a whole bunch of nested if/elses. implemented!

roseeichelmann:
@johnclary fixed the mixed up column names and implemented upserts ✨. yeah i was definitely tempted to DRYify this function but the more i looked into it the more it seemed like it would go against the nature of SQL and a lot of consensus online seemed to be that its best to [avoid dynamic SQL](https://www.sommarskog.se/dynamic_sql.html#dyntablenames) when not necessary.. but i am very new to this so im really not sure what would be best practice in this situation 🤔

mddilley (approved):
Nice stuff here!! 👏 I followed the test steps and observed the updates to the fatalities table in the UI and the Hasura console as I flipped back and forth on the injury severity. Also nice to see upserts on the DB level. I've only ever done that through some kind of abstraction so TIL as well. 🛳️✨

johnclary:
> the more i looked into it the more it seemed like it would go against the nature of SQL and a lot of consensus online seemed to be that its best to [avoid dynamic SQL](https://www.sommarskog.se/dynamic_sql.html#dyntablenames) when not necessary

that's a super helpful read @roseeichelmann. thank you! 🚢

johnclary (approved):

patrickm02L (approved):
🚢 Ship it!

PR #1195, opened 2023-03-17
Update VZ Github Actions so that we can manually trigger them

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/11752

This will help us manually trigger these in the future when rotating secrets. 🔐 I don't think it is important to be able to manually trigger the SQS clean up workflow so I left it as is, but let me know if any of y'all disagree! 

See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch and https://github.com/cityofaustin/atd-moped/blob/main/.github/workflows/atd_moped_api.yml#L20

## Testing
**URL to test:** 
n/a

**Steps to test:**
n/a

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/.github/workflows/atd_sqs_build.yml
+++ b/.github/workflows/atd_sqs_build.yml
@@ -15,6 +15,10 @@ on:
       - master
     paths:
       - "atd-events/**"
+  workflow_dispatch:
+    inputs:
+      description:
+        default: ""
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


roseeichelmann (approved):
nice, thanks for addressing this! im honestly pretty unfamiliar with github workflows so i was able to skim the docs and get a better understanding of what they are/see how we use them in our repository.

frankhereford (approved):
action

This looks great Mike, thank you for enabling us to trigger them at will!

patrickm02L (approved):
Thanks for the link to the docs about how to manually trigger! 💯

PR #1196, opened 2023-03-18
Create tailored view for locations listing page

frankhereford:
## Associated issues & Intent

This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/11755.

This PR eliminates the use of `view_location_injry_count_cost_summary` being joined against `atd_txdot_locations` via a `graphql-engine` relationship. The combination of the way `graphql-engine` executed the join along with additional `join` complexity in the view lead to low query performance.

This PR introduces a new view, `locations_with_crash_injury_counts`, allowing for more carefully crafted SQL be used to query the information needed to display this table.

## Deployment

* [ ] View created
* [ ] Indexes added
* [ ] `graphql-engine` tracked
* [ ] Permissions granted

## Testing

https://deploy-preview-1196--atd-vze-staging.netlify.app/#/locations

### Steps

Load up the locations listings page, and 

🤔 did it load at all?
🤔 did it load quickly enough, on the first load?
🤔 did it load quickly on subsequent loads of the page?
🤔 can it be searched?
🤔 can it be sorted?

Then review the code and check out the view, and 

🤔 Does this make sense?
🤔 Could this be done quicker?

Finally, check out the staging `graphql-engine` [console](https://vzd-staging.austinmobility.io/console), and

🤔 Do I like how the permissions are set for this view?
🤔 Does it need any relationships?

## Performance

* Maybe still suffers from slowness on the first load?
  * This is oddly hard to test or reproduce
  * Working to capture this in an explain analyze to demonstrate cause
  * Suspected mitigations for cold data are to improve disk speed and improve cache size / configuration
    * `t4g` class instance to get fast EBS access
    * `medium` class instance for more RAM
  * 🎙️Please share your observations

🎉 🪤 I finally caught one: https://explain.dalibo.com/plan/4h66bd3h2d88g177#plan/node/9 and it sure tells the tale.

* This was taken on a `db.t4g.medium` read replica, querying the staging data that had not been used in 12+ hours.
* The operation that wrecks performance on cold data is a `Bitmap Heap Scan` which is parsing over the combined sets of crashes fulfilling the matching `where` clauses. 
* This operation reads data at 5 MB/s for 48 seconds. 
  * Makes me think this is CPU bound, because I'd expect it to be able to pull data in much faster than that. I know I can get near gigabit/s speeds off the cluster with `pg_dump` over the AWS network to the bastion, so the disk has to be at least that fast (~100 MB/s).
* The above mitigations still feel like good solutions, especially giving it more RAM to let it hold onto this cache.
* I'm going to try a multicolumn index specifically targeting this operation.

### Locations Detail Page

We have talked about using this view to speed up the locations detail page, which is also super slow. I looked at it in my inspector, and I found that it's not just one slow query, it's four or five. To me, this means that applying this same fix wouldn't give any immediate benefit, as there would be other slow queries, and we could address them all in a spin-off issue if it's worth it.

#### Things we can do to help
* [Upgrade to graphql-engine 2+](https://github.com/cityofaustin/atd-data-tech/issues/6988)
* Give the RDS cluster more RAM 
  * It's `t3.small` now, and I think we should use a `t4g.medium`.
  * Moped, which has < 1% of the data, runs on `t2.medium`
* Run the cluster on a AWS graviton processor
  * Faster EBS performance (lower disk latency)
  * Offset some of the cost from the instance upgrade

---

#### Ship list
- [ ] Code reviewed


atd-vzd/views/locations_wiwth_crash_injury_counts.sql
roseeichelmann, line 1: tinyy little typo on this file

atd-vzd/views/locations_wiwth_crash_injury_counts.sql
```diff
@@ -0,0 +1,58 @@
...
+    from cris_crashes
+    -- this unions the sets, over location_id, such that a location is represented if it has either/or/and a cris and blueform crash(s).
+    full outer join apd_crashes
+      on (cris_crashes.location_id = apd_crashes.location_id)
+    )
+  -- finally, this is a list of locations after all, so query them and join the crash data we've built up.
+  select -- think of this as step 3
+    locations.description,
+    locations.location_id,
+    coalesce(crashes.crash_count, 0) as crash_count, -- graphql abstract makes dealing with nulls difficult, so coalesce to 0.
```
roseeichelmann, line 49: TIL!

atd-vzd/views/locations_wiwth_crash_injury_counts.sql
roseeichelmann: also testing file comments hehe pretty cool!

roseeichelmann (approved):
thanks for hopping on this so quickly frank! i reviewed the code on the new view you created and it looks cleannnn. your view is such huge improvement! this whole thing has been fascinating, i too have no idea how the locations page has ever been loading fine (albeit slow) in prod.
first time around in testing the locations page did load, it didn't take super long but probably at least 10 seconds? 
the page can be reloaded with ease after the first load ✔️
it can be searched and sorted ✔️
code makes sense, can't think of a way it could be done quicker ✔️
permissions look good to me for the view ✔️
i dont think any relationships need to be added ✔️
thanks for the super helpful comments in the view it made it easier to follow along, also that explain analyze was really interesting, its cool you were able to capture that! i dont rly have any helpful comments on how it could be made faster but john's idea in the slack discussion to use a materialized view sounds good to me! 💫💫

atd-vzd/views/locations_wiwth_crash_injury_counts.sql
```diff
@@ -0,0 +1,58 @@
...
+    coalesce(crashes.total_est_comp_cost, 0) as total_est_comp_cost,
+    coalesce(crashes.fatalities_count, 0) as fatalities_count,
+    coalesce(crashes.serious_injury_count, 0) as serious_injury_count
+  from atd_txdot_locations locations
+  left join crashes on (locations.location_id = crashes.location_id)
+  where true 
+    -- the criteria for being in the list at all, so we limit down the number of locations we're dealing with.
+    and locations.council_district > 0
```
roseeichelmann, line 57: much better than handling this in the VZE like before!

atd-vzd/views/locations_wiwth_crash_injury_counts.sql
frankhereford: Thanks for catching the filename typo!

atd-vzd/views/locations_wiwth_crash_injury_counts.sql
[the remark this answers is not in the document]
frankhereford: Thanks Rose! 

Sometimes I wonder when to put conditions in the definition of the view or if I want to put them in the javascript that uses it. I think it's often times a grey area type of thing. I think my rule of thumb might be that when the condition can change based on the state of the app or a users input, then it belongs in the app, but when it's a condition that is applied 100% of the time, then it belongs in the view. When it's in the view, it's defining the super-set of all possible results, and dynamic stuff belongs with all the dynamic code. It's def one of those things where there are lots of valid ways to do it.

patrickm02L (approved):
So nice to have a Location Page that actually loads. It does load slower like you mentioned on the first visit, but subsequent visits are much faster. 🐢

atd-vzd/views/locations_wiwth_crash_injury_counts.sql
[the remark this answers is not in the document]
mddilley: I agree that this works well in the view. I tripped over the config lines that you removed here in the VZE code the other day, and I was pretty surprised to find it there tbh. I think that focusing on the dynamic stuff being set in the front end makes total sense. 🙏

mddilley (approved):
Thanks for digging through all of this and coming up with a fresh view! ✨ I tested the steps, and I don't see any changes that jump out at me. I can load the locations table and use all of the existing features! 🙌

PR #1197, opened 2023-03-19
Tooling to steadily write to a VZ database (crashes + related records)

frankhereford:
## Problem Description & Intent

Long running queries on the VZ read replica are frequently terminated with an error that reads `canceling statement due to conflict with recovery / User query might have needed to see row versions that must be removed`. This problem degrades Xavier's analysis by killing his queries and often prevents the replication of production data onto local VZ databases. 

This program is tooling to load a database cluster with write operations on the primary database enabling the testing of long-running queries on a child read replica.

## Associated issues

No issues currently for this, please see [here](https://austininnovation.slack.com/archives/CM9MK950S/p1679060208951889) for background and [here](https://austininnovation.slack.com/archives/CM9MK950S/p1679079563267159) for discussion. 

This PR adds a toolbox program which can be used to write crash + unit + primaryperson + person transactions of `insert`s to a vision zero database. The records are populated with field type appropriate random data. 

## Testing

This program is tooling used to diagnose problems with the primary & read replica database instances. Please test it if you want, but I would consider it optional. The real acid test is if the read replica will service long running queries while the primary server is under write load.

* You can test the program by filling out a `.env` file and running it against your local database.
* If the solution found using this program has been installed on the RDS VZ servers, you can test the outcome by replicating from production successfully and reliably. 

---
#### Ship list
- [ ] Code reviewed


atd-toolbox/database_load_test/database_writer.py
```diff
@@ -0,0 +1,335 @@
...
+    return json
+
+
+def random_jsonb(field):
+    return random_json(field)
+
+
+def random_position():
+    latitude = 30.274722 + random.uniform(-0.1, 0.1)
+    longitude = -97.740556 + random.uniform(-0.1, 0.1)
```
frankhereford, line 320: Crash positions are set to the TX State Capitol ± 0.1 decimal degrees.

atd-toolbox/database_load_test/database_writer.py
```diff
@@ -0,0 +1,335 @@
...
+    position = f"ST_GeomFromEWKT('SRID=4326;POINT({longitude} {latitude})')"
+    return position
+    pass
+
+
+def get_next_crash_id():
+    primary = get_primary_connection()
+    cursor = primary.cursor(cursor_factory=psycopg2.extras.RealDictCursor)
+    cursor.execute("SELECT max(crash_id)+1 as next_id FROM atd_txdot_crashes;")
```
frankhereford, line 329: Because we get `crash_id`s from CRIS, we don't actually maintain a sequence for the crash table, and this is the only way to know you're not going to try to reuse a value in your insert.

atd-toolbox/database_load_test/database_writer.py
```diff
@@ -0,0 +1,335 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
patrickm02L: Welcome to my world! I've been using ChatGPT for a while to understand y'alls code.

PR #1200, opened 2023-03-23
Resolve AFP toggle naming confusion in crash index advanced filters

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/11819

## Testing
**URL to test:** 
[Netlify](https://deploy-preview-1200--atd-vze-staging.netlify.app/#/)

**Steps to test:**
1. Go to the crashes page advance filters section, see the new toggle label that says "Include Outside Of Austin Full Purpose".

---
#### Ship list
- [x] Code reviewed

### DIFF
--- a/atd-vze/src/views/Crashes/crashGridTableParameters.js
+++ b/atd-vze/src/views/Crashes/crashGridTableParameters.js
@@ -219,7 +219,7 @@ export const crashGridTableAdvancedFilters = {
       },
       {
         id: "geo_geocoded",
-        label: "Has been Geocoded",
+        label: "Has Been Geocoded",
         filter: {
           where: [
             {
--- a/atd-vze/src/views/Crashes/crashGridTableParameters.js
+++ b/atd-vze/src/views/Crashes/crashGridTableParameters.js
@@ -244,7 +244,7 @@ export const crashGridTableAdvancedFilters = {
       },
       {
         id: "geo_afd",
-        label: "Remove Austin Full Purpose",
+        label: "Include Outside Of Austin Full Purpose",
         invert_toggle_state: true,
         filter: {
           where: [


patrickm02L (approved):
Thanks for knocking this out! 🥊 

Screen Shot 2023-03-24 at 2 17 01 PM

atd-vze/src/views/Crashes/crashGridTableParameters.js
```diff
@@ -244,7 +244,7 @@ export const crashGridTableAdvancedFilters = {
       },
       {
         id: "geo_afd",
-        label: "Remove Austin Full Purpose",
+        label: "Include Outside Of Austin Full Purpose",
```
mddilley, line 247: I don't think that the VZ team will have any confusion here now! Super small detail, but I noticed that there is inconsistent capitalization on one toggle label - `Has been Geocoded`. Not a big deal but wondering if we should patch that one up while you are in here.

mddilley (approved):
Yay for clearer labels! 🚀 🚢 🙏

atd-vze/src/views/Crashes/crashGridTableParameters.js
frankhereford: I think this is a great idea. Polish to the UI is never a bad thing!

frankhereford (approved):
Big 👍 from me. Copy edit in the code looks good, and it continues to function as expected over different break points as I resize the window. This is a 🚀.

atd-vze/src/views/Crashes/crashGridTableParameters.js
roseeichelmann: thanks for catching that mike, just fixed that capitalization inconsistency!

chiaberry (approved):

PR #1202, opened 2023-03-28
Add Law Enforcement Number to the FTT Index

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/11805

For context the law enforcement number is APDs running count of fatal crashes for that year, similar to our "Year to date fatal crash" count. The VZ team will always be inputting it manually.

## Testing
**URL to test:** 
[Netlify](https://deploy-preview-1202--atd-vze-staging.netlify.app/#/fatalities)

**Steps to test:**
1. Click on a crash from the fatalities page
2. The law enforcement number can be found in the Fatalities card 
image
3. Edit the law enforcement num a few times, it is of type int. Check back at the Fatalities page to see it reflected there.

One thing I was thinking:
~Should the law enforcement number section only be seen on crashes that have fatalities? Or should it stay there for all crashes? Im guessing the VZ team wants it there for all crashes in case something changes but I'm not sure.~

---
#### Ship list
- [ ] Code reviewed


patrickm02L (approved):
Works well! 🚢 

I asked Joel if we wanted the Numbers for all crashes or just fatalities? 

He said, "just fatalities, thanks."

atd-vze/src/views/Crashes/crashDataMap.js
```diff
@@ -145,6 +145,11 @@ export const crashDataMap = [
         lookupOptions: "atd_txdot__y_n_lkp",
         lookupPrefix: "y_n",
       },
+      law_enforcement_num: {
+        label: "Law Enforcement Number",
+        editable: true,
+        uiType: "text",
```
mddilley, line 151: Looks like there is a mismatch in the data type between the UI config and the DB schema. I tested out `2023-01` like the example in the issue and get a GraphQL error.

mddilley (changes requested):
I tested this out, and I think this needs one small change to the DB schema. Also, going back to your question in dev sync about conditionally rendering on fatal crashes, it would be a good improvement to centralize that check somewhere in the code since we are checking it in recommendations, here, and other places too probably. That way we could update in one place only in the future if we wrangle these columns that can de-sync.

atd-vze/src/views/Crashes/crashDataMap.js
roseeichelmann: The values for law enforcement nums will always be plain integers, so the DB schema type is an int for this column. So `2023-01` wouldn't work because of the dash, I think Patrick just didn't realize that in the example. Sorry i should have specified that!

roseeichelmann: Also btw the "text" for the `uiType` just means that the user types in the value, the other option is "select" which would be a drop down. idk if you already know this but wanted to clarify bc it caught me off guard and i see how that could be confusing!

mddilley: Thanks for the context on this! I'm thinking in Moped-mode, and I completely forgot that this `dataMap` config was set up that way. It still makes me wonder if we should set the datatype to character varying since the UI will allow entry of non-integers. I'll give some time for others weigh in, and I'm ready to ✅ if we're good with this.

roseeichelmann: its weird, i noticed that when i test locally (using the staging database) i get the error you are talking about when typing a string into any of the fields that are supposed to be ints in the DB schema, including my law enforcement field but also the other fields like atd fatality count. but when i test on my netlify link or on the staging website i dont get an error, the non-int value just doesnt save. but i am totally open to addressing the possible mismatch in datatypes between what the DB expects and what the textbox allows you to input. if we decide to address it i think we would have to for all of the fields in the crash details page that are of type int with the "text" `uiType`, because they are all currently handled in the same way that i have the law enforcement field set up 🤔

mddilley: Gotcha - sorry, I didn't mean a large refactor. Just updating this one new `law_enforcement_num` column to from integer to character varying, but I recognize that this is a larger issue with the codebase that is way out of scope for this issue. The fact that you found other fields treated this way is a really good point and maybe this should be treated the same by leaving this as integer type in the DB then. 🙏

roseeichelmann: turns out i _will_ change the datatype for the law enforcement num after all because after taking a look into the spreadsheet for backfilling joel sent, there are a handful of values that are strings like "DPS" to represent crashes responded to by an entity other than APD.

mddilley: Cool cool - glad that we received more info that made this decision easier! 🚀

frankhereford (approved):
Reads great and works great. This is 💯. Thank you Rose!

PR #1203, opened 2023-03-28
Show level 5 polygons on Location view map

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/11735

This PR:
- adds some error handling so any null polygon shapes will not break the view in the future
- updates the location map to fit the polygon on load
- populates the `shape`, `latitude`, and `longitude` columns of level 5 location records in the `atd_txdot_locations` table  (more on this in the comments in the SQL below)
- sneaks in an update to the GraphQL abstract to skip `order_by` from applying to aggregate queries. I can't think of a case where this would be needed

## Testing
**URL to test:** 
https://deploy-preview-1203--atd-vze-staging.netlify.app

**Steps to test:**
1. Go to any of the following crashes: 
   - https://deploy-preview-1203--atd-vze-staging.netlify.app/#/locations/83405791792 
   - https://deploy-preview-1203--atd-vze-staging.netlify.app/#/locations/79DD65A6780 
   - https://deploy-preview-1203--atd-vze-staging.netlify.app/#/locations/9C95D829B2D
   - https://deploy-preview-1203--atd-vze-staging.netlify.app/#/locations/CAFDB95173F
      - you should see a yellow outline of the shape of this level 5 polygon
      - the map should be centered on the polygon
2. You can compare this to https://visionzero.austin.gov/editor/#/locations/83405791792 which breaks on load.
3. Go to https://deploy-preview-1203--atd-vze-staging.netlify.app/#/locations/C389A2CEFC1
   - I removed the `shape`, `latitude`, and `longitude` from this record
   - it should still load and the map will default to downtown

**Note: The following updates have been applied to the staging database. We will need to make the same changes to production when releasing.**

```sql
-- The UI shows the shape column on the map.
-- Level 1-4 locations have this same value in both shape and geometry columns
-- so we follow suit for level 5 records.
UPDATE
	atd_txdot_locations
SET
	shape = geometry
WHERE
	location_group = 2;
```

```sql
-- The latitude and longitude columns are used to center the map on the polygon
-- and query Nearmap API for a timestamp of the latest images captured for the location tiles.
UPDATE
	atd_txdot_locations
SET
	longitude = ST_X(ST_CENTROID(shape)), 
	latitude  = ST_Y(ST_CENTROID(shape))
WHERE
	location_group = 2;
```

---
#### Ship list
- [x] Code reviewed


patrickm02L (commented):
Step 1 works, but the polygon is not rendering on the map like polygons from other crashes. I expected to see the whole shape of the polygon in the map. It is cut off in this view.

### Normal view
normal

### Expected
expected

mddilley:
@patrickm02L Thanks for the feedback. 🙏 The existing map is set to always initialize at a fixed zoom (17) no matter how large or small the location polygon is. It would use Mapbox's [fitBounds](https://docs.mapbox.com/mapbox-gl-js/example/fitbounds/) method and would work like the Moped components map.

 I can update the map to fit to the bounds of the polygon it is, but this would affect how all locations look - level 1-4 and level 5. Does that sound okay to you?

frankhereford:
> Edit: I went ahead and updated this because it just looks better! 🗺️

Great change - I think fitting to bounds is totally the right way to go. 🙏

mddilley:
Sync'd with Patrick, and I'm going to update this to not show the transition when fitting the polygon when the map loads. That way it will just snap to the polygon.

atd-vze/src/views/Locations/LocationMap.js
```diff
@@ -53,10 +54,13 @@ const polygonDataLayer = {
...
     super(props);
+    this.polygon = this.props.data.atd_txdot_locations[0];
+    this.mapRef = React.createRef();
+
     this.state = {
       viewport: {
-        latitude: this.props.data.atd_txdot_locations[0].latitude || 30.2672,
-        longitude: this.props.data.atd_txdot_locations[0].longitude || -97.7431,
+        latitude: this.polygon?.latitude || 30.2672,
+        longitude: this.polygon?.longitude || -97.7431,
```
frankhereford, line 63: 🧼 🧽 Cleaner reading - this is helpful.

atd-vze/package.json
```diff
@@ -18,6 +18,7 @@
     "@coreui/icons": "0.3.0",
     "@coreui/react": "^2.5.1",
     "@handsontable/react": "^3.1.3",
+    "@turf/bbox": "^6.5.0",
     "@turf/difference": "6.0.1",
```
frankhereford, line 22: I wonder if you think it would be worth it to bring `@turf` up to `^6.5.0` so that we don't have two versions of the same library in play? It seems like we might be opening ourselves up to a really nasty bit of debugging if we're passing around GIS data with version changes unbeknownst to the methods we're using.

frankhereford (approved):
kermit map


This looks fantastic, esp with the suggestion from Patrick to keep the feature within the extent of the map. Thank you for doing this!

atd-vze/src/queries/gqlAbstract.js
```diff
@@ -373,6 +373,18 @@ gqlAbstractTableAggregateName (
...
+        let order_by = [];
+        for (let [key, value] of this.getEntries("order_by")) {
+          order_by.push(
+            this.isNestedKey(key)
+              ? this.sortifyNestedKey(key, value)
+              : `${key}: ${value}`
+          );
+        }
+        output.push(`order_by: {${order_by.join(", ")}}`);
+      }
```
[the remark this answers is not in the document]
mddilley: Thanks for bearing with the fast one that I pulled here! 😅 It was something that I noticed when looking at queries coming out of the gql abstract the other day, and it seemed like a small fix that was worth doing.

atd-vze/package.json
mddilley: That is a good catch - I didn't see that there was another turf method pulled in here. I'll take a look at this for sure. 🔍🙏

mddilley:
Currently working on a bug where the zoom of the map isn't updating in state after fitting to the polygon. You can replicate by zoom in or out after the map loads. There is a jump in zoom that isn't very fun to watch! 🙁

atd-vze/package.json
mddilley: @frankhereford On second thought, it doesn't look like this library is used anywhere in the app. I'm going to create a backlog issue because there is some clean up to be done beyond this library, and I'd like to be careful about removing things.

mddilley:
@patrickm02L This is ready for another look. 🙏 I refactored this to align with the way `react-map-gl` v5 works (less like the Mapbox GL JS way) because I was approaching it the way that `react-map-gl` v7 does which we use in Moped (which is more like the Mapbox GL JS way).

I added some more links above to level 5 locations too.

atd-vze/package.json
frankhereford: That sounds great, and I totally appreciate the idea of not yoinking it out without due care. One less dependency is one less thing dependabot can gripe at us about too. Thanks for looking into it @mddilley!

patrickm02L (approved):
Yes, very nice. Thanks for making that adjustment! 🚀

PR #1205, opened 2023-03-30
Improve secret exchange between VZ Hasura event triggers and CR3 API

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/11803

This PR:
- Updates the `HASURA_EVENT_API` secret name to `HASURA_TRIGGER_API_KEY`. This secret's role is to be passed in the headers of a Hasura event trigger request that is received by the CR3 API. In the API code, it is checked for a matching value that comes from `ZAPPA_SETTINGS`.  This secret will be moved to the Hasura environment variables that are set in the AWS ECS task definition so we don't leak it in the Hasura metadata accidentally.
- Updates the CR3 API code to get the environment name from env vars that come from `ZAPPA_SECRETS`
- Removes this env var from the SQS Lambdas code since it is not used in any of them

## Testing
**URL to test:** 
n/a

**Steps to test:**
1. Is the name of this secret any better? If anyone comes up with a stronger name, I'm 100% down to change it. 🙏
2. Do you agree with [this plan](https://github.com/cityofaustin/atd-data-tech/issues/11803#issuecomment-1489437919) to deploy these changes?
3. The rest of this will have to be tested after merge to `master`. From PRs, we deploy preview SQS + Lambda that handle the event triggers but not the CR3 API that processes and routes them there.

---
#### Ship list
- [x] Code reviewed

### DIFF
--- a/atd-cr3-api/server.py
+++ b/atd-cr3-api/server.py
@@ -45,7 +45,7 @@ AWS_S3_BUCKET = os.getenv("AWS_S3_BUCKET", "")
 # Hasura Config
 HASURA_ADMIN_SECRET = os.getenv("HASURA_ADMIN_SECRET", "")
 HASURA_ENDPOINT = os.getenv("HASURA_ENDPOINT", "")
-HASURA_EVENT_API = os.getenv("HASURA_EVENT_API", "")
+HASURA_TRIGGER_API_KEY = os.getenv("HASURA_TRIGGER_API_KEY", "")
 HASURA_EVENTS_SQS_URL = os.getenv("HASURA_EVENTS_SQS_URL", "")
 
 
@@ -428,10 +428,10 @@ def user_delete_user(id):
 @APP.route("/events/", methods=["PUT", "POST"])
 def associate_location():
     # Require matching token
-    incoming_token = request.headers.get("HASURA_EVENT_API")
+    incoming_token = request.headers.get("HASURA_TRIGGER_API_KEY")
     incoming_event_name = request.headers.get("HASURA_EVENT_NAME", "")
     hashed_events_api = hashlib.md5()
-    hashed_events_api.update(str(HASURA_EVENT_API).encode("utf-8"))
+    hashed_events_api.update(str(HASURA_TRIGGER_API_KEY).encode("utf-8"))
     hashed_incoming_token = hashlib.md5()
     hashed_incoming_token.update(str(incoming_token).encode("utf-8"))
 
--- a/atd-cr3-api/server.py
+++ b/atd-cr3-api/server.py
@@ -62,7 +62,9 @@ def get_api_token():
             "client_secret": API_CLIENT_SECRET,
             "audience": f"https://{AUTH0_DOMAIN}/api/v2/",
         },
-        {"content-type": "application/json", },
+        {
+            "content-type": "application/json",
+        },
     )
     return response.json().get("access_token", None)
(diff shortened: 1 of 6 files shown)


atd-cr3-api/server.py
```diff
@@ -452,6 +453,7 @@ def associate_location():
             HASURA_EVENTS_SQS_URL[0:48]  # This is the length of the url with the account number
             + "/atd-vz-data-events-"     # We're going to add a prefix pattern for our ATD VisionZero queues
             + incoming_event_name        # And append the name of the incoming event
+            + environment_for_queue_name # Append the environment name to target the correct queue
```
mddilley, line 456: Previously, `incoming_event_name` contained the name of the event and environment set in the Hasura console/metadata. This is where the crossed environments happened when production metadata was applied to the staging Hasura cluster. It now comes from the CR3 API's environment variables set in `ZAPPA_SETTINGS` (`STAGING` or `PRODUCTION`)

atd-cr3-api/server.py
```diff
@@ -428,10 +428,11 @@ def user_delete_user(id):
...
     # Require matching token
-    incoming_token = request.headers.get("HASURA_EVENT_API")
+    incoming_token = request.headers.get("HASURA_TRIGGER_API_KEY")
     incoming_event_name = request.headers.get("HASURA_EVENT_NAME", "")
+    environment_for_queue_name = API_ENVIRONMENT.lower()
     hashed_events_api = hashlib.md5()
-    hashed_events_api.update(str(HASURA_EVENT_API).encode("utf-8"))
+    hashed_events_api.update(str(HASURA_TRIGGER_API_KEY).encode("utf-8"))
     hashed_incoming_token = hashlib.md5()
     hashed_incoming_token.update(str(incoming_token).encode("utf-8"))
```
mddilley, line 437: Don't want to scope creep, but curious if anyone knows why we would md5 both values here to compare them? The value is a string. Could be overlooking something. 🤷

roseeichelmann (approved):
thanks so much for getting to the bottom of this mike! these changes look good to me, i think the updated name for the key is more descriptive and your next steps seem on point to me as well. this is a ✔️ for me!

atd-cr3-api/server.py
johnclary: this is a TIL for me—it appears to be an attempt to mitigate [short-circuiting](https://michaelhowardsecure.blog/2019/02/22/the-dangers-of-string-comparing-passwords/)

atd-cr3-api/server.py
johnclary: @mddilley i noticed we return a `403` if the event name is empty (`""`)—do we have a discrete list of event names we could validate against rather than accepting any string?

johnclary (approved):
thanks for tackling this @mddilley! it's so much better 😅

atd-cr3-api/server.py
mddilley: that sounds right to me, @johnclary - a TIL for me as well. 🙏

atd-cr3-api/server.py
mddilley: @johnclary - that is a great idea. We could add that to the code here for sure. There are three valid events at the moment.

patrickm02L (approved):
Thanks for showing this in Dev Sync today! 🚢

PR #1207, opened 2023-03-30
augment local storage key name

frankhereford:
This appends the code name, currently hardcoded into the app, to the local storage key name, so we can change that anytime we want to wipe out the currently used local storage.

As a follow on issue, I'd like to get this value pulled from the package.json file in some form or another. Perhaps a substring of a md5 sum of the version number.

#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vze/package.json
+++ b/atd-vze/package.json
@@ -1,6 +1,6 @@
 {
   "name": "atd-vz-data",
-  "version": "1.31.0",
+  "version": "1.31.1",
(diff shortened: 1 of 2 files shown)


mddilley (approved):
Thanks for making this update, and I like John's suggestion! 🚀🚀

patrickm02L (approved):
Looks good! 🚀

PR #1208, opened 2023-03-30
Patch VZE to invalidate local storage for GridTable

frankhereford:
This is a patch to invalidate local storage which contains the graphql query of the state of the user's table configuration. With changed table names, the stored state results in a graphql error. We intend to address this with a more elegant solution in how local storage is used to store table configuration state, and also to pull the `codeName` or similar from package.json, which is the authoritative source of that data.

---
#### Ship list
- [x] Code reviewed


patrickm02L (approved):
Good to goooooo!

PR #1213, opened 2023-04-11
Update local stack env var and replace Hasura metadata

mddilley:
## Associated issues
n/a

This PR addresses a warning in the Hasura console that happened after [updating the event triggers to use an environment variable](https://github.com/cityofaustin/atd-vz-data/pull/1205) called `HASURA_TRIGGER_API_KEY`. It also replaces the existing metadata that contains the leaked secret value with the updated metadata.

~_Edit: Also snuck in an update to gitignore the zappa_settings.json so we can prevent committing those secrets accidentally if we deploy the API manually._~ Nvm [Tilly handled this in another PR](https://github.com/cityofaustin/atd-vz-data/pull/1214) 🙏

## Testing
**URL to test:** 
Local only

**Steps to test:**
Spin up the local stack and go to the Hasura console. You should not be redirected to the settings page with the warning `You have been redirected because your GraphQL Engine metadata is in an inconsistent state`.

---
#### Ship list
- [x] Code reviewed


atd-vzd/hasura/hasura_metadata_2023_04_11_14_03_58_508.json
```diff
@@ -0,0 +1,5527 @@
+{
+  "version": 2,
```
mddilley, line 2: I noticed the existing metadata is version 3. Should export metadata from Hasura v1 which is in production or v2 which is the local stack? Happy to take this back to version 3 if that is what we've been doing.

roseeichelmann: this was my doing from exporting metadata from my local stack a few weeks ago, i didnt realize there was a diff in versions btwn local and prod. imo it makes sense to export the prod version though, so my bad!

mddilley: all good! just gathering opinions because I can see it both ways. 🙏

chiaberry (approved):
i looked at this last week but forgot to come back and give it an approval!

tillyw (approved):
merged this branch, restarted my stack and the console spun up without errors! thanks Mike!

mddilley:
Thanks, y'all - merging!

PR #1214, opened 2023-04-11
add pagination to user list

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/11940

this pr adds pagination to the users table, resolving an issue where only 50 users were visible due to a limitation in the number of records the auth0 management api returns in a single request. thanks @mddilley for getting me set up with the proper auth0 and aws credentials and pairing with me on updating the cr3-api.

## Testing
**URL to test:** 
https://deploy-preview-1214--atd-vze-staging.netlify.app/

**Steps to test:**
- go to the users page
- you should see a new pagination section at the top right showing the number of pages and users
- page through the table, you should see all records and be able to jump to the beginning and end of the table
- when you page through the table, you should see the loading spinner as vze queries the auth0 management api for the next set of users to display
- paging options should be disabled if they are not necessary (e.g., you get to the end of the table)
- if you shrink the screen, the 'add user' button and pagination menu should eventually take up their own rows

---
#### Ship list
- [ ] Code reviewed


atd-vze/src/containers/DefaultLayout/DefaultHeader.js
```diff
@@ -90,13 +90,13 @@ const DefaultHeader = props => {
                 href="https://atd.knack.com/dts#new-service-request/?view_249_vars=%7B%22field_398%22%3A%22Bug%20Report%20%E2%80%94%20Something%20is%20not%20working%22%2C%22field_399%22%3A%22Vision%20Zero%20(Editor)%22%7D"
                 target="_blank"
               >
-                <i class="fa fa-bug" /> Report a bug&nbsp;&nbsp; <i className="fa fa-external-link" />
+                <i className="fa fa-bug" /> Report a bug&nbsp;&nbsp; <i className="fa fa-external-link" />
```
tillyw, line 93: just fixing some console warnings

atd-vze/src/views/Users/Users.js
```diff
@@ -40,7 +43,11 @@ const UserRow = ({ user }) => {
       </td>
       <td>{format(parseISO(user.created_at), "MM/dd/yyyy")}</td>
       <td>{user.logins_count}</td>
-      <td>{format(parseISO(user.last_login), "MM/dd/yyyy")}</td>
+      <td>
+        {user.last_login
```
tillyw, line 47: prevents wsod bug when attempting to display users who have never logged in - thanks @mddilley for adding this!

patrickm02L (approved):
This looks great! Thanks for demoing in Dev Sync 🚀 

Screen Shot 2023-04-13 at 5 25 08 PM

atd-cr3-api/server.py
mddilley, line 1: Not change needed at all because we know exactly what this route needs to do in the UI, but, if this was less tightly bound to the view that it serves, I might consider handling the absence of the query parameters in the request.

atd-vze/src/views/Users/Users.js
mddilley: Thanks for adding this. I just noticed there is a WSOD when clicking through to the user details page for someone who has never signed in too. Could you add this fix there too? Another hidden bug revealed! 🐛

atd-vze/src/views/Users/Users.js
```diff
@@ -82,16 +101,56 @@ const Users = () => {
...
-                  <Row className="align-items-center mb-3">
-                    <Col col="6" sm="4" md="2" xl className="mb-xl-0">
-                      <Link to="/users/add" className="link">
-                        <Button color="primary">
-                          <i className="fa fa-user-plus"></i> Add User
-                        </Button>
-                      </Link>
+                  <Row className="align-items-center">
+                    <Col xs="12" md="6" className="mb-3">
+                      <Row className="d-flex justify-content-md-start justify-content-center ml-0">
```
mddilley, line 109: Not a huge deal, but I think that you can get away with removing the `Row` here and below where one is wrapping the `Pagination` component by moving the classes in them to the `Column`s that wrap them.

mddilley (approved):
Tested this locally and in the deploy preview, and it all works so well! I really like the pagination that you added here to keep it simple and elegant. 🎉 🚢  🎉  I know that the VZ team will be happy to see _all_ the users! 🚀

atd-vze/src/views/Users/Users.js
tillyw: fixed! thanks for catching that @mddilley

atd-vze/src/views/Users/Users.js
tillyw: thanks for pointing that out! updated the code to remove those rows and move the styling to the columns

frankhereford (commented):
This looks great Tilly - thank you!

PR #1215, opened 2023-04-12
Add Victim Names to the FTT Index

roseeichelmann:
## Testing
**URL to test:** 

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vze/src/views/Crashes/RelatedRecordsTable.js
+++ b/atd-vze/src/views/Crashes/RelatedRecordsTable.js
@@ -88,11 +88,17 @@ const RelatedRecordsTable = ({
       <Table responsive>
         <thead>
           <tr>
-            {Object.keys(fieldConfig.fields).map(field => (
-              <th key={`th_${fieldConfig.fields[field].label}`}>
-                {fieldConfig.fields[field].label}
-              </th>
-            ))}
+            {Object.keys(fieldConfig.fields).map(
+              field =>
+                // Dont render columns that should not be rendered (Victim Name if there are no fatalities)
+                ((fieldConfig.fields[field].render &&
+                  fieldConfig.fields[field].render(data)) ||
+                  !fieldConfig.fields[field].render) && (
+                  <th key={`th_${fieldConfig.fields[field].label}`}>
+                    {fieldConfig.fields[field].label}
+                  </th>
+                )
+            )}
           </tr>
         </thead>
         <tbody>
@@ -117,118 +123,129 @@ const RelatedRecordsTable = ({
 
                     const uiType = fieldConfig.fields[field].format;
 
-                    return (
-                      <td key={i}>
-                        {isEditing && (
-                          <form
-                            onSubmit={e => handleSubmit(e, mutationVariable)}
-                          >
-                            {uiType === "text" && (
-                              <Input
-                                type="text"
-                                defaultValue={formatValue(row, field)}
-                                onChange={e =>
-                                  handleInputChange(e, updateFieldKey)
-                                }
-                              />
-                            )}
-                            {uiType === "select" && (
-                              <Input
-                                name={field}
-                                id={field}
-                                onChange={e =>
-                                  handleInputChange(e, updateFieldKey)
-                                }
-                                defaultValue={
-                                  // Check for null values and display as blank
-                                  row[field] &&
-                                  row[field][`${fieldLookupPrefix}_id`] !== null
-                                    ? row[field][`${fieldLookupPrefix}_id`]
+                    // Only render victim name cell for fatalities
+                    if (
+                      field == "fatality" &&
+                      row.injury_severity.injry_sev_desc !== "KILLED"
+                    ) {
+                      return <td></td>;
+                    } else {
+                      return (
+                        <td key={i}>
+                          {isEditing && (
+                            <form
+                              onSubmit={e => handleSubmit(e, mutationVariable)}
+                            >
+                              {uiType === "text" && (
+                                <Input
+                                  type="text"
+                                  defaultValue={formatValue(row, field)}
+                                  onChange={e =>
+                                    handleInputChange(e, updateFieldKey)
+                                  }
+                                />
+                              )}
+                              {uiType === "select" && (
+                                <Input
+                                  name={field}
+                                  id={field}
+                                  onChange={e =>
+                                    handleInputChange(e, updateFieldKey)
+                                  }
+                                  defaultValue={
+                                    // Check for null values and display as blank
+                                    row[field] &&
+                                    row[field][`${fieldLookupPrefix}_id`] !==
+                                      null
+                                      ? row[field][`${fieldLookupPrefix}_id`]
+                                      : ""
+                                  }
+                                  type="select"
+                                >
+                                  {/* Show a NO DATA option only when formatValue is displayed. */}
+                                  {formatValue(row, field) === "NO DATA" && (
+                                    <option value={null}>NO DATA</option>
+                                  )}
+                                  {lookupOptions[
+                                    fieldConfig.fields[field].lookupOptions
+                                  ].map(option => {
+                                    return (
+                                      <option
+                                        value={
+                                          option[`${fieldLookupPrefix}_id`]
+                                        }
+                                        key={option[`${fieldLookupPrefix}_id`]}
+                                      >
+                                        {option[`${fieldLookupPrefix}_desc`]}
+                                      </option>
+                                    );
+                                  })}
+                                </Input>
+                              )}
+                              <div className="d-flex">
+                                <Button
+                                  type="submit"
+                                  block
+                                  color="primary"
+                                  size="sm"
+                                  style={{ minWidth: "50px" }}
+                                  className="btn-pill mt-2 mr-1"
+                                >
+                                  <i className="fa fa-check edit-toggle" />
+                                </Button>
+                                <Button
+                                  type="cancel"
+                                  block
+                                  color="danger"
+                                  size="sm"
+                                  className="btn-pill mt-2"
+                                  style={{ minWidth: "50px" }}
+                                  onClick={e => handleCancelClick(e)}
+                                >
+                                  <i className="fa fa-times edit-toggle"></i>
+                                </Button>
+                              </div>
+                            </form>
+                          )}
+
+                          {!isEditing &&
+                            (fieldConfig.fields[field].badge ? (
+                              <Badge
+                                color={fieldConfig.fields[field].badgeColor(
+                                  formatValue(row, field)
+                                )}
+                              >
+                                {formatValue(row, field)}
+                              </Badge>
+                            ) : (
+                              <span
+                                className={
+                                  formatValue(row, field) === "NO DATA"
+                                    ? "text-muted"
                                     : ""
                                 }
-                                type="select"
                               >
-                                {/* Show a NO DATA option only when formatValue is displayed. */}
-                                {formatValue(row, field) === "NO DATA" && (
-                                  <option value={null}>NO DATA</option>
-                                )}
-                                {lookupOptions[
-                                  fieldConfig.fields[field].lookupOptions
-                                ].map(option => {
-                                  return (
-                                    <option
-                                      value={option[`${fieldLookupPrefix}_id`]}
-                                      key={option[`${fieldLookupPrefix}_id`]}
-                                    >
-                                      {option[`${fieldLookupPrefix}_desc`]}
-                                    </option>
-                                  );
-                                })}
-                              </Input>
-                            )}
-                            <div className="d-flex">
+                                {formatValue(row, field)}
+                              </span>
+                            ))}
+
+                          {fieldConfig.fields[field].editable &&
+                            !isReadOnly(roles) &&
+                            !isEditing && (
                               <Button
-                                type="submit"
                                 block
-                                color="primary"
-                                size="sm"
-                                style={{ minWidth: "50px" }}
-                                className="btn-pill mt-2 mr-1"
-                              >
-                                <i className="fa fa-check edit-toggle" />
-                              </Button>
-                              <Button
-                                type="cancel"
-                                block
-                                color="danger"
+                                color="secondary"
                                 size="sm"
                                 className="btn-pill mt-2"
-                                style={{ minWidth: "50px" }}
-                                onClick={e => handleCancelClick(e)}
+                                style={{ width: "50px" }}
+                                onClick={e => handleEditClick(field, row)}
                               >
-                                <i className="fa fa-times edit-toggle"></i>
+                                <i className="fa fa-pencil edit-toggle" />
                               </Button>
-                            </div>
-                          </form>
-                        )}
-
-                        {!isEditing &&
-                          (fieldConfig.fields[field].badge ? (
-                            <Badge
-                              color={fieldConfig.fields[field].badgeColor(
-                                formatValue(row, field)
-                              )}
-                            >
-                              {formatValue(row, field)}
-                            </Badge>
-                          ) : (
-                            <span
-                              className={
-                                formatValue(row, field) === "NO DATA"
-                                  ? "text-muted"
-                                  : ""
-                              }
-                            >
-                              {formatValue(row, field)}
-                            </span>
-                          ))}
-
-                        {fieldConfig.fields[field].editable &&
-                          !isReadOnly(roles) &&
-                          !isEditing && (
-                            <Button
-                              block
-                              color="secondary"
-                              size="sm"
-                              className="btn-pill mt-2"
-                              style={{ width: "50px" }}
-                              onClick={e => handleEditClick(field, row)}
-                            >
-                              <i className="fa fa-pencil edit-toggle" />
-                            </Button>
-                          )}
-                      </td>
-                    );
+                            )}
+                        </td>
+                      );
+                    }
                   })}
                 </tr>
               );
--- a/atd-vze/src/queries/people.js
+++ b/atd-vze/src/queries/people.js
@@ -24,6 +24,9 @@ export const GET_PEOPLE = gql`
         ethnicity_id
         ethnicity_desc
       }
+      fatality {
+        victim_name
+      }
       unit_nbr
     }
     atd_txdot_person(where: { crash_id: { _eq: $crashId } }) {
@@ -45,6 +48,9 @@ export const GET_PEOPLE = gql`
       ethnicity {
         ethnicity_id
         ethnicity_desc
+      }
+      fatality {
+        victim_name
       }
       unit_nbr
     }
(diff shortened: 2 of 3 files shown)


atd-vzd/migrations/migration_fatalities_2023_04_27--1526.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE fatalities ADD COLUMN updated_by text;
```
roseeichelmann, line 1: keeps track of what email edited the victim name

atd-vze/src/Components/GridTable.js
```diff
@@ -32,7 +31,7 @@ import GridTableDoughnut from "./GridTableDoughnut";
 import GridTableHorizontalBar from "./GridTableHorizontalBar";
 import GridTableFilterBadges from "./GridTableFilterBadges";
 
-const codeName = 'jester';
+const codeName = "jester";
```
roseeichelmann, line 34: this was prettier formatting

atd-vze/src/views/Crashes/PeopleDetailsCard.js
```diff
@@ -83,6 +92,7 @@ const PeopleDetailsCard = ({ isExpanded, toggleAccordion, ...props }) => {
             keyField={"primaryperson_id"}
             lookupOptions={lookupSelectOptions}
             mutation={primaryPersonMutation}
+            secondaryMutation={fatalityMutation}
```
roseeichelmann, line 95: needed to add a secondary mutation that handles updates to `victim_name` bc this column exists in the fatalities table

atd-vze/src/views/Crashes/RelatedRecordsTable.js
```diff
@@ -101,135 +109,163 @@ const RelatedRecordsTable = ({
             .map(row => {
               return (
                 <tr key={`table-${tableName}-${row[keyField]}`}>
-                  {Object.keys(fieldConfig.fields).map((field, i) => {
```
roseeichelmann, line 104: this was all prettier formatting

atd-vze/src/views/Crashes/RelatedRecordsTable.js
```diff
@@ -101,135 +109,163 @@ const RelatedRecordsTable = ({
...
-                                  row[field][`${fieldLookupPrefix}_id`] !== null
-                                    ? row[field][`${fieldLookupPrefix}_id`]
-                                    : ""
+                  {Object.keys(fieldConfig.fields)
+                    // Filter out victim name column if there are no fatalities
+                    .filter(field =>
+                      fieldConfig.fields[field].render
+                        ? fieldConfig.fields[field].render(data)
+                        : true
+                    )
```
roseeichelmann, line 118: this `.filter` i added

atd-vze/src/views/Crashes/RelatedRecordsTable.js
```diff
@@ -101,135 +109,163 @@ const RelatedRecordsTable = ({
...
+
+                      const mutationVariable =
+                        fieldConfig.fields[field].mutationVariableKey;
+
+                      const uiType = fieldConfig.fields[field].format;
+
+                      // Render blank cell in victim name column if not a fatality
+                      if (
+                        field === "fatality" &&
+                        row.injury_severity.injry_sev_desc !== "KILLED"
```
roseeichelmann, line 138: this i added

atd-vze/src/views/Crashes/RelatedRecordsTable.js
```diff
@@ -88,11 +89,18 @@ const RelatedRecordsTable = ({
       <Table responsive>
```
johnclary, line 89: outside the scope of this PR, but it would be nice to see this return/render broken into smaller pieces—e.g. maybe a separate component for the `<form>`...

atd-vzd/migrations/migration_fatalities_2023_04_27--1526.sql
johnclary: probably out of scope, but if the user email is in the user's session token, it would be very cool to set this value via [Hasura permission rule](https://hasura.io/docs/latest/auth/authorization/permissions/column-presets/).

roseeichelmann: ohh cool, TIL! this is also a reminder to fix how we are granting users that have "Admin" access in the VZE actual admin privileges in hasura. bc from what i can tell thru the console you cant set column presets for the admin role

atd-vze/src/queries/fatalities.js
```diff
@@ -0,0 +1,32 @@
...
+    $crashId: Int
+    $changes: fatalities_set_input
+  ) {
+    update_fatalities(
+      where: {
+        crash_id: { _eq: $crashId }
+        _and: {
+          _or: [
+            { person_id: { _eq: $personId } }
+            { primaryperson_id: { _eq: $personId } }
```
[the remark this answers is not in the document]
roseeichelmann: hmm i think its super unlikely... these ids are default calculated as next val and the most recent `person_id` is 1473645 and `primaryperson_id` is 3955101 so thats a huge range (millions) before they start getting close enough to fear them overlapping. i also did a query of the database now and there are no crashes where the person/primaryperson ids are the same. looking back to the first crashes in our db it looks like the primary person and person values were started with a big gap between them to begin with.

mddilley (commented):
This looks so great! I added a few comments and also think it is worth thinking a bit more about handling the update mutation that could target either a primary person or person record.

atd-vze/src/views/Crashes/PeopleDetailsCard.js
roseeichelmann: oh yeah, i really like that idea as well. i forgot that was even a possibility, im going to implement that! and yeah +1 @mddilley, it felt weird writing a `secondaryMutation` and im laughing thinking about there being a `quaternaryMutation` 😁

atd-vze/src/queries/fatalities.js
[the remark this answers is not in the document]
frankhereford: I double checked, and the `person_id` and the `primaryperson_id` in their respective tables are surrogate keys which we generate on insert into the table. They do not come from CRIS, and they are only constrained such that they are unique in their respective table. All that to say, they are vanilla `id serial primary key` primary keys. They can, and do in some cases, have potential to overlap. 

The rule of thumb that I use is that the primary key column of every table be nothing except an integer that uniquely identifies the row, and equally importantly, has no imbued meaning encoded in the number. If you can tell something about the data in the row by looking at the ID -- anything at all -- there becomes an opportunity for snags to pop up.

I think that you need to pass some sort of indication along with the `$personId` value about which table you're intending to be querying from along with just the `$personId`. As this is written, if we had an ETL get away from us one day and consume a few million IDs, we'd have to be extra careful in how we cleaned that up because we might loose our method of distinguishing which table's ID field we're intending to match against. And there would be nothing except our own experience telling us that we needed to keep that in mind too, if that were to happen.

I agree with Rose that there is little potential for this to snag us up here, right now, because we just, by luck, have our sequences way out of pace with each other, but I would strongly support spinning off an issue to address this if we don't in this PR.

frankhereford (approved):
Rose, this works great. I think you've received some fantastic feedback from other reviewers as well; thank you everyone.

I did a little digging on the `primaryperson_id` and `person_id` fields. I'm not entirely comfortable with relying on just how out of sync those sequences are to indicate which table a certain integer id refers to, but I would be fine seeing fixing it spun off into another issue if you don't want to address it in this PR. 

Thank you again for this Rose, it's a great enhancement.

atd-vze/src/queries/fatalities.js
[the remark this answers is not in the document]
roseeichelmann: john and i synced yesterday on how to address the possibility of there being a crash record with a primary person and person with the same IDs so ill be pushing that change up today, thanks for the feedback yall!

roseeichelmann:
Also realized i need to take into account not showing the victim name column if the user doesnt have admin-level VZE permissions, so im also addressing that

tillyw:
tested and works and looks great!! it was really interesting reading the comments on this pr - i feel like i learned a lot!

the only question that came to mind that hasn't been discussed is whether it would be worth clearing the victim name column if a user changes a fatal crash back to a nonfatal crash after adding/editing a victim name. i'm sure this would rarely  happen (if ever) but it could result in a non-fatal crash having a victim name populated in the database. i'm not sure whether that would be a concern at all but just wanted to mention it!

tillyw (approved):

roseeichelmann:
im adding the WIP label while i work on adjusting this PR to reflect our decision to put the names column both in the primary person and person tables instead of it being in the fatalities table since we are going to implement importing names from the cr3s to the respective person tables. then, the names will be pulled into the view called `view_fatalities` for being displayed in the FTT index page.

another issue should be written to address importing names from the cr3s into their respective person tables, and this PR will just address the new column locations and being able to edit the name in the crash details page as mocked up in the issue. @patrickm02L 

one thing im wondering is if there should be separate fields for `first_name` `middle_name` and `last_name` bc i can see thats how it was done when cr3 names were imported for the `atd_txdot_persons` table. and if we do go this route instead of having just one `name` field for the whole name, how would this affect the UI interaction?

curious what you guys think!

frankhereford:
> one thing im wondering is if there should be separate fields for `first_name` `middle_name` and `last_name` bc i can see thats how it was done when cr3 names were imported for the `atd_txdot_persons` table. and if we do go this route instead of having just one `name` field for the whole name, how would this affect the UI interaction?

Hi Rose, this is a great question. I think your instinct to question this is right on. My gut says to mirror the columns in the that we get from cris, one for first, middle and last. As for the UI, that's a bit tricker, because you either have a <blank for first> <blank for middle> <blank for last> to make it easy to get the right part of the name into the right part of the record, or you can do something like this regex to split a single name field up into three strings, one for each name column.

`^([\w']+)(?:\s+([\w\.']+(?:\s+\w+)*))?\s+([\w']+)$`

This will give you 3 matches, where the first word ends up in the first one, optionally (0 or more) middle names in the 2nd and the last word will end up in the 3rd match. It handles abbreviated middle names and `'`s in names, so it's not a terrible start, but names are human, so it won't surprise me if it needs a little work. 

I'm not sure what's best, but maybe these will open up some options. Thanks for your continued work on this!!

roseeichelmann:
@frankhereford thanks for your comment, it was super helpful! have i misunderstood that we would actually get the name columns directly from the cris tables/csv's we are download? for some reason i was under the impression that the names are something we would have to scrape from the cr3s, but sounds like that is not the case.

frankhereford:
@roseeichelmann wrote:

> actually get the name columns directly from the cris tables/csv's we request?

That's a great question Rose. I think our recent conversations with the VZ team and internally have conflated the terms "pull from CR3s" and "receive from CRIS CSV exports." Here's the story as I understand it:

* CRIS does, for sure, provide the first, middle, and last names for both `persons` and `primaryperson`s in the CSVs we receive.
* At some point, some years back, it was decided that we shouldn't capture those fields because of concerns around PII or related reasons.
* We dropped the name columns from our `primaryperson`s table, so that our ETLs would not capture that data from the CSV.
* We failed/opted-not to drop the columns from our `person`s table, so our ETL did continue to capture that data. We don't use it anywhere though, to my knowleddge.

As a result, we happen to have name data for our `person`s, but do not have it for `primaryperson`s. By my read, I think we want to add those columns back to `primaryperson` so that we start to capture it. The LDM will include a full history backfill, so that will fill in the last ten years of names, as well, so I wouldn't sweat trying to backfill it beforehand. The VZ team can fill in those names by hand until the LDM, as I understand our current plan. 

Of course, I may be in a state of misunderstanding, so please anyone correct this if so, but I think this may be helpful to give you some additional background into where we are and how we got here. 

Thanks!!

mddilley:
@frankhereford Thanks for the background on that because I also started to think we were scraping the CR3 for names somehow. I'm just catching up on this convo, and I vote for treating first, middle, and last as separate fields to update in the UI. It keeps things simple and less error prone, and I think it fits with the pattern throughout the app of mapping one editable field to one DB column.

I am all ears for other opinions, though, because maybe I'm just scared of regex. 😅 It is tempting to find a way to make the UI look cleaner. ✨

frankhereford:
> because maybe I'm just scared of regex

Is there anything more ugly than regex code?

`^([\w']+)(?:\s+([\w\.']+(?:\s+\w+)*))?\s+([\w']+)$`

It's a insanely powerful construct though. One of those things in computer science that was discovered and laid down before there were microprocessors. I would be very happy to pair with you if it would be helpful to make that particularly gnarly one be more accessible to you. 

I played with it a fair bit, and I think it does a pretty stelar job of parsing names into their constituent parts, if that ends up being the route you want to go.

chiaberry:
It sounds like ultimately it's a design/UX question, but I'm for having different fields for first, middle and last as inputs in the UI. Just thinking about all the different variations names can have (i.e. two first names, two last names, etc)

roseeichelmann:
@frankhereford ahh thank you for the context on the names and where we get them, im glad to hear we dont have to do any scraping and they are just fields we already get from the CRIS CSVs. 

@mddilley i also vote in favor of having different first/middle/last name fields in the UI, like chia said names can be complicated and weird and this would be the least error prone way i think. but yeah, its defs a design question too

patrickm02L:
In Planning 5/3/23, we talked about creating the new columns that go into `primarypersons` that mirrors the columns found in` persons `table.

atd-vze/src/views/Crashes/RelatedRecordsTable.js
```diff
@@ -37,14 +44,18 @@ const RelatedRecordsTable = ({
   };
 
   const handleInputChange = (e, updateFieldKey) => {
+    // Change new value to null if it only contains whitespaces
+    // or save value as all uppercase
+    const newValue =
+      e.target.value.trim().length === 0 ? null : e.target.value.toUpperCase();
```
roseeichelmann, line 50: this handles the input for victim names but also works for the other text input types

roseeichelmann:
Finished updating this PR with the changed scope and UI design, its now ready for review!

frankhereford (approved):
🚢 🚢 Rose, this is ready to sail. I've tested it six ways to Sunday, and it's perfect.

I did want to ask - is there still a need for the `victim_name` field in the fatalities table? Maybe out of the scope of this issue, but is that going to get deprecated, or am I missing something from the bigger picture?

roseeichelmann:
@frankhereford you are correct in thinking that the `victim_name` column in the fatalities table is now deprecated and can be deleted. thats now having me thinking that a separate issue maybe should be created to address reflecting this change in the FTT Index Page bc currently that pulls from the old `victim_name` column and now needs to be updated to pull from the first, middle, and last name columns in the persons table and be concatenated for the UI.

image

patrickm02L:
> i guess that could be addressed in this issue instead but feels out of scope?

cc: @roseeichelmann @frankhereford 

Update Victim Name display in FTT Index Page [#12304](https://github.com/cityofaustin/atd-data-tech/issues/12304)

patrickm02L (approved):
## Review
This is so nice! It's very intuitive, and works like a dream 🚀 

## Testing

- [X] Go to the Fatalities tab and click on a Crash ID
- [X] On the crash details page click on the People tab in the Related Records table and test editing the first, middle, and last name of a person who had a fatal injury. (:03-22)
- [X] Test cancelling your edit and saving it. (:46)
- [X] The Victim Name column should only show up on a table if it has a fatality in it (Now called "Fatal Injury") 
- [X] Test changing the injury severity of another person from non-fatal to "Fatal Injury" back to non-fatal. Note how the victim name edit cell will appear and disappear based on the injury severity. (:28-:37)

https://github.com/cityofaustin/atd-vz-data/assets/5566727/8d199cf3-31b8-4375-82a6-5e12621eb6ae

mddilley (approved):
Awesome stuff! I went through all of the steps using the deploy preview, and they all worked just like you said. Thanks for sticking with this one through all the change. It is 🪄✨, and the VZ team is going to love this. 🚀

atd-vze/src/views/Crashes/RelatedRecordsTable.js
```diff
@@ -101,133 +121,185 @@ const RelatedRecordsTable = ({
...
+                        ? fieldConfig.fields[field].shouldRender(data) &&
+                          (isItSupervisor(roles) || isAdmin(roles))
+                        : true
+                    )
+                    .map((field, i) => {
+                      // Render victim name cell in victim name column if row is a fatality
+                      if (field === "victim_name") {
+                        if (row.prsn_injry_sev_id === 4) {
+                          return (
+                            <VictimNameField
```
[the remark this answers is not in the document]
roseeichelmann: ooh i like that idea!

atd-vze/src/views/Crashes/RelatedRecordsTable.js
```diff
@@ -101,133 +121,185 @@ const RelatedRecordsTable = ({
...
-                                  row[field][`${fieldLookupPrefix}_id`] !== null
-                                    ? row[field][`${fieldLookupPrefix}_id`]
-                                    : ""
+                  {Object.keys(fieldConfig.fields)
+                    // Filter out victim name field if there are no fatalities in the table
+                    // or if user doesn't have correct permissions
+                    .filter(field =>
+                      fieldConfig.fields[field].shouldRender
+                        ? fieldConfig.fields[field].shouldRender(data) &&
+                          (isItSupervisor(roles) || isAdmin(roles))
```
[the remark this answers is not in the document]
roseeichelmann: I just pushed some updates that moved the role check to the `shouldRender` function and generalized this section of the code and I can now see this easily being used in the future for other fields that should render conditionally. Lmk what you think and thanks for the suggestion :)

atd-vze/src/views/Crashes/VictimNameField.js
```diff
@@ -0,0 +1,133 @@
+import React from "react";
+import { Button, Input, Form, FormGroup, Row, Col, Label } from "reactstrap";
+import { useQuery } from "@apollo/react-hooks";
+
+import { GET_PERSON_NAMES } from "../../queries/people";
+
+const VictimNameField = ({
```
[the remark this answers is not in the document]
roseeichelmann: Done, good idea!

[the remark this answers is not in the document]
mddilley: Thanks, Rose! 🚀

atd-vze/src/views/Crashes/RelatedRecordsTable.js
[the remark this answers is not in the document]
mddilley: Looks amazing! 😎

PR #1217, opened 2023-04-18
Add an address search feature to the VZV map

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/11581

This PR updates the crash map with a geocoder search. It also updates the way the map ref is assigned (note in the code diff about this that points to some docs).

## Testing
**URL to test:** 
https://deploy-preview-1217--atd-vzv-staging.netlify.app/viewer/map

**Steps to test:**
1. Test the address search by entering a street, place, etc within the Austin area and select it from the dropdown
2. You should see the map fly to the location that you selected
3. Try searching for a place outside of the Austin area (Dallas, San Antonio, or wherever) and you'll see that results do not populate for locations outside of the navigable map area
4. To test the updates to features that could have been affected by the `ref` code change:
   - Test mobile view and you'll see that, when you click on a crash dot on the map, the popup info box opens towards the viewable space even if it is near the edge of the map (towards the horizontal center - the popup can still open off-screen vertically)
   - Toggle fatal and serious injury crash filters
   - Make sure the loading compass spinner shows when data is loading on the map

---
#### Ship list
- [x] Code reviewed

### DIFF
--- a/atd-vzv/src/views/map/Geocoder/Geocoder.js
+++ b/atd-vzv/src/views/map/Geocoder/Geocoder.js
@@ -4,10 +4,11 @@
 // TODO zoom to location
 // TODO clear search input
 import React from "react";
-import Geocoder from "react-mapbox-gl-geocoder";
+import Geocoder from "react-map-gl-geocoder";
+import "react-map-gl-geocoder/dist/mapbox-gl-geocoder.css";
 import { MAPBOX_TOKEN } from "../Map";
 
-const Geocoder = React.forwardRef(({ handleViewportChange }, ref) => {
+const MapGeocoder = React.forwardRef(({ handleViewportChange }, ref) => {
   return (
     <Geocoder
       mapRef={ref}
@@ -15,10 +16,11 @@ const Geocoder = React.forwardRef(({ handleViewportChange }, ref) => {
       mapboxApiAccessToken={MAPBOX_TOKEN}
       //   inputValue={geocoderAddress}
       options={{ flyTo: false }}
+      position="top-left"
       // Bounding box for auto-populated results in the search bar
       //   bbox={[-98.22464, 29.959694, -97.226257, 30.687526]}
     />
   );
 });
 
-export default Geocoder;
+export default MapGeocoder;
--- a/atd-vzv/src/views/map/Geocoder/Geocoder.js
+++ b/atd-vzv/src/views/map/Geocoder/Geocoder.js
@@ -1,24 +1,23 @@
-// TODO Geocoder in the top left
-// TODO Type street into search bar
-// TODO dropdown autocomplete of streets
-// TODO zoom to location
-// TODO clear search input
 import React from "react";
 import Geocoder from "react-map-gl-geocoder";
 import "react-map-gl-geocoder/dist/mapbox-gl-geocoder.css";
 import { MAPBOX_TOKEN } from "../Map";
+import { mapNavBbox } from "../mapData";
 
 const MapGeocoder = React.forwardRef(({ handleViewportChange }, ref) => {
+  // Apply the same map navigation bounding box to the geocoder search
+  const { latitude, longitude } = mapNavBbox;
+  const bbox = [longitude.min, latitude.min, longitude.max, latitude.max];
+
   return (
     <Geocoder
       mapRef={ref}
       onViewportChange={handleViewportChange}
       mapboxApiAccessToken={MAPBOX_TOKEN}
-      //   inputValue={geocoderAddress}
       options={{ flyTo: false }}
       position="top-left"
       // Bounding box for auto-populated results in the search bar
-      //   bbox={[-98.22464, 29.959694, -97.226257, 30.687526]}
+      bbox={bbox}
     />
   );
 });
--- a/atd-vzv/src/views/map/Geocoder/Geocoder.js
+++ b/atd-vzv/src/views/map/Geocoder/Geocoder.js
@@ -1,11 +1,24 @@
-// TODO: Geocoder in the top left
-// TODO: Type street into search bar
-// TODO: dropdown autocomplete of streets
-// TODO: zoom to location
-// TODO: clear search input
+// TODO Geocoder in the top left
+// TODO Type street into search bar
+// TODO dropdown autocomplete of streets
+// TODO zoom to location
+// TODO clear search input
+import React from "react";
+import Geocoder from "react-mapbox-gl-geocoder";
+import { MAPBOX_TOKEN } from "../Map";
 
-const Geocoder = () => {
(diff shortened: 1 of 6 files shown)


atd-vzv/src/views/map/Map.js
```diff
@@ -336,7 +335,7 @@ const Map = () => {
       getCursor={_getCursor}
       interactiveLayerIds={interactiveLayerIds}
       onClick={_onSelectCrashPoint}
-      ref={(ref) => (mapRef.current = ref && ref.getMap())}
+      ref={mapRef}
```
mddilley, line 338: This change updates the shape of the ref object. We were assigning the map instance to `ref.current` directly by calling [getMap()](https://github.com/visgl/react-map-gl/blob/6.1-release/docs/api-reference/static-map.md#getmap) on assignment. The geocoder added in this PR is[ expecting to call `getMap()` on the ref passed to it](https://github.com/SamSamskies/react-map-gl-geocoder/blob/2968c74e730c52f2ecece086e08ed9d518fd9bfd/src/index.js#L119-L123) and was breaking when trying to call map methods in conjunction with the custom way this ref was assigned.

**Before**
`ref.current` => the Mapbox map instance

**Now**
`ref.current.getMap()` => the Mapbox map instance

I updated the other uses of the map instance to get it with `getMap()` so they would continue to work.

atd-vzv/src/views/map/Map.js
```diff
@@ -280,9 +271,15 @@ const Map = () => {
...
-      b: selectedFeature.layer.paint[`${selectedFeature.layer.type}-color`].b * 255,
+      r:
+        selectedFeature.layer.paint[`${selectedFeature.layer.type}-color`].r *
+        255,
+      g:
+        selectedFeature.layer.paint[`${selectedFeature.layer.type}-color`].g *
+        255,
+      b:
+        selectedFeature.layer.paint[`${selectedFeature.layer.type}-color`].b *
+        255,
```
mddilley, line 282: This chunk is only Prettier formatting

atd-vzv/src/views/map/Map.js
```diff
@@ -27,20 +27,9 @@ import "mapbox-gl/dist/mapbox-gl.css";
 import "@mapbox/mapbox-gl-draw/dist/mapbox-gl-draw.css"; // Get out-of-the-box icons
 import MapInfoBox from "./InfoBox/MapInfoBox";
 import MapPolygonInfoBox from "./InfoBox/MapPolygonInfoBox";
+import MapGeocoder from "./Geocoder/Geocoder";
 
-const MAPBOX_TOKEN = `pk.eyJ1Ijoiam9obmNsYXJ5IiwiYSI6ImNrM29wNnB3dDAwcXEzY29zMTU5bWkzOWgifQ.KKvoz6s4NKNHkFVSnGZonw`;
-
-function useMapEventHandler(eventName, callback, mapRef) {
```
mddilley, line 33: Moved this to the helpers file and added a docstring because it was a quick change

atd-vzv/src/views/map/Map.js
```diff
@@ -27,20 +27,9 @@ import "mapbox-gl/dist/mapbox-gl.css";
...
-  useEffect(() => {
-    const currentMapRef = mapRef.current;
-    const mapDataListener = currentMapRef.on(eventName, function () {
-      callback();
-    });
-    return () => {
-      currentMapRef.off(eventName, mapDataListener);
-    };
-  }, [eventName, callback, mapRef]);
-}
```
mddilley, line 43: Moved this to the helpers file and added a docstring because it was a quick change

patrickm02L:
**Two recommendations**
- Zoom faster, similar to the speed in Moped
- Close the addresses drop down when zooming to location, keep it closed until clicking instead search input - see Moped behavior for reference

chiaberry (approved):
thanks for explaining the ref object change

mddilley:
@patrickm02L I updated this code to address the two recs that you had. Thanks for catching the issue with the dropdown staying open - I completely missed that. 🙏  The geocoder library that we need to use with this version of `react-map-gl` handles things a little different than what we do in Moped, but I tracked down a GH issue with a fix to make sure it closes on result select.

patrickm02L (approved):
¡Muy bien! 🚀 

https://user-images.githubusercontent.com/5566727/232942180-cb57d10a-05e0-4fb8-bcdb-f102c262506b.mp4

atd-vzv/src/views/map/Geocoder/Geocoder.js
```diff
@@ -0,0 +1,43 @@
...
+ * dropdown to close after selecting a result.
+ * @see https://github.com/SamSamskies/react-map-gl-geocoder/issues/96
+ * @param {Function} handleViewportChange - Callback function to update the map viewport
+ */
+const MapGeocoder = React.forwardRef(({ handleViewportChange }, ref) => {
+  const handleGeocoderViewportChange = useCallback(
+    (viewport) => {
+      // Speed up the flyTo transition
+      const geocoderDefaultOverrides = {
+        transitionDuration: 1500,
```
frankhereford, line 19: I think maybe this might want to be even faster? I'd be comfortable with it down at `500` milliseconds or even less.

frankhereford (approved):
This is great Mike, I think it's a really helpful addition to the VZV's map. Thank you!

atd-vzv/src/views/map/Map.js
```diff
@@ -27,20 +27,9 @@ import "mapbox-gl/dist/mapbox-gl.css";
...
+export const MAPBOX_TOKEN = `pk.eyJ1Ijoiam9obmNsYXJ5IiwiYSI6ImNrM29wNnB3dDAwcXEzY29zMTU5bWkzOWgifQ.KKvoz6s4NKNHkFVSnGZonw`;
```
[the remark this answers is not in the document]
mddilley: Thanks for dropping that link, Frank! I agree that an issue to review and secure these tokens is in order since we fully control them. It would be good to push them out of the code and rotate them.

atd-vzv/src/views/map/Geocoder/Geocoder.js
mddilley: 👀 I'll leave this up to @patrickm02L since we modeled this after the speed in Moped.

patrickm02L: let's keep it at this speed for now.

atd-vzv/src/views/map/Map.js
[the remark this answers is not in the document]
patrickm02L: Good looking out. I'll make an Issue for this.

PR #1220, opened 2023-04-28
Fix Person Badge Colors

frankhereford:
## Associated issues

This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/12135. 

This came out of [an observation in slack](https://austininnovation.slack.com/archives/CM9MK950S/p1682696549485749) by @roseeichelmann  that the updates today to many of our lookup tables had broken the color picking scheme for our person badges.

I'm also sneaking in a fix to the repo orchestration helper that updates it for our new multi-import-schema ETL mechanism.

Based on direction from @patrickm02L, should this patch be accepted, it will be deployed into production today.

## Testing
**URL to test:** 

https://deploy-preview-1220--atd-vze-staging.netlify.app/#/

**Steps to test:**

Check out various people on crashes pages and observe if their badges' colors match the injury they received. 

---
#### Ship list
- [ ] Code reviewed
- [ ] Product manager approved

PR #1221, opened 2023-04-28
Patch production with the badge color fix.

frankhereford:
This is a patch to production which incorporates the two commits related to fixing the bug that was fixed in PR #1220. It includes a third commit to bump the patch number for the app.


roseeichelmann (commented):
🚢🚢🚢

PR #1224, opened 2023-05-08
 Add New FTT fields to the crash export

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/12126

## Testing
**URL to test:** 
Local, check out this branch and you have to run the commands in the migrations file to update the recommendations column names

**Steps to test:**
1. Test as admin, IT supervisor, editor, and read only users if you really want to go above and beyond.
2. Go to the crashes table and click the export (save icon) button.
3. If you are admin or IT supervisor you will see 4 new fields at the end of the csv file for recommendation text, update, status, and partners. Note that if you export crashes that dont have data for these fields then the columns wont show up in the export, so make sure you are targeting crashes with fatalities that have recommendations.

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vzd/migrations/migration_recommendations_2023_05_03--0926.sql
+++ b/atd-vzd/migrations/migration_recommendations_2023_05_03--0926.sql
@@ -1,0 +1,11 @@
+ALTER TABLE recommendations
+RENAME COLUMN text TO rec_text;
+
+ALTER TABLE recommendations
+RENAME COLUMN update TO rec_update;
+
+ALTER TABLE atd__recommendation_status_lkp
+RENAME COLUMN description TO rec_status_desc;
+
+ALTER TABLE atd__coordination_partners_lkp
+RENAME COLUMN description TO coord_partner_desc;
--- a/atd-vze/src/Components/GridExportData.js
+++ b/atd-vze/src/Components/GridExportData.js
@@ -1,5 +1,6 @@
 import React, { useState } from "react";
 import { withApollo } from "react-apollo";
+import { useAuth0, isAdmin, isItSupervisor } from "../auth/authContext";
 import { useLazyQuery } from "@apollo/react-hooks";
 import { formatISO } from "date-fns";
 import { CSVLink } from "react-csv";
@@ -33,6 +34,25 @@ const StyledSaveLink = styled.i`
 
 const GridExportData = ({ query, columnsToExport, totalRecords }) => {
   const [isModalOpen, setIsModalOpen] = useState(false);
+
+  const { getRoles } = useAuth0();
+  const roles = getRoles();
+
+  // Columns that can only be exported by admins & it supervisors
+  const adminColumns = [
+    "recommendation { rec_text }",
+    "recommendation { rec_update }",
+    "recommendation { atd__recommendation_status_lkp { rec_status_desc } }",
+    "recommendation { recommendations_partners { atd__coordination_partners_lkp { coord_partner_desc }} }",
+  ];
+
+  // Remove admin-only columns from columnsToExport for users without admin/it supervisor permissions
+  if (!(isAdmin(roles) || isItSupervisor(roles))) {
+    console.log("not admin");
+    adminColumns.forEach(col => {
+      columnsToExport = columnsToExport.replace(col, "");
+    });
+  }
 
   // Use .queryCSV to insert columnsToExport prop into query
   let [getExport, { loading, data }] = useLazyQuery(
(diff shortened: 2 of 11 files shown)


atd-vze/src/Components/GridExportData.js
mddilley, line 1: Thanks for pointing out that we need to filter to crashes with fatalities to see the new columns. 🙏 It makes me wonder if we should spin off an issue to make sure columns export regardless of whether there is data retrieved or not to keep CSVs consistent between exports. Could get tricky since some of these fields are nested and the parent returns `null`. 🤷  @patrickm02L

mddilley (approved):
This looks and works perfectly! I tested all of the roles locally, and I saw the new columns in the export on with the admin and IT supervisor roles. 🙌 ✨

atd-vze/src/Components/GridExportData.js
```diff
@@ -34,6 +35,24 @@ const StyledSaveLink = styled.i`
...
+  const { getRoles } = useAuth0();
+  const roles = getRoles();
+
+  // Columns that can only be exported by admins & it supervisors
+  const adminColumns = [
+    "recommendation { rec_text }",
+    "recommendation { rec_update }",
+    "recommendation { atd__recommendation_status_lkp { rec_status_desc } }",
+    "recommendation { recommendations_partners { atd__coordination_partners_lkp { coord_partner_desc }} }",
+  ];
```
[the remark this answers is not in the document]
roseeichelmann: Yeah I was wondering if it would be better to go that route, thanks for the suggestion im gonna update this :)

atd-vze/src/Components/GridExportData.js
frankhereford: I think that having the columns only show up when a record happens to include non-null data for them is unexpected and potentially confusing behavior. I agree with Mike a lot here.

I can see a world where our VZ users are exporting data out of this app and are going to feed it into some program, as it's kind of what it's for, right? If we are not very stable in which columns we include in the CSV, then whatever ETL or analysis tool they run on it will have to handle the columns conditionally, and that's a headache.

If this is unreasonably tricky to implement in our front-end JS CSV building mechanism, I don't feel like it would be too out of line to create a flask app in ECS or a λ function to return it, as it's not the quickest system in the world, as is. This is some pretty serious scope creep though, so 🤷🏼.

frankhereford (approved):
This looks good Rose, thank you. I've read over the diff and tested it locally. This is ready to ship 📦!

For future testers, if you apply the migration file using a DB client, you'll need to make corollary changes to the metadata, so the `graphql-engine` middleware knows how to serve up that data. It makes me appreciate the moped system; might be a fun spike or discovery day one of these days to try to bring that back to VZ.

Thanks again Rose, this is sharp!

atd-vzd/migrations/migration_recommendations_2023_05_03--0926.sql
```diff
@@ -0,0 +1,11 @@
...
+RENAME COLUMN text TO rec_text;
+
+ALTER TABLE recommendations
+RENAME COLUMN update TO rec_update;
+
+ALTER TABLE atd__recommendation_status_lkp
+RENAME COLUMN description TO rec_status_desc;
+
+ALTER TABLE atd__coordination_partners_lkp
+RENAME COLUMN description TO coord_partner_desc;
```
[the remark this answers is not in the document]
roseeichelmann: I always forget 🤦

atd-vze/src/Components/GridExportData.js
```diff
@@ -31,9 +31,22 @@ const StyledSaveLink = styled.i`
   }
 `;
 
-const GridExportData = ({ query, columnsToExport, totalRecords }) => {
+const GridExportData = ({
+  query,
+  columnsToExport,
+  totalRecords,
+  roleSpecificColumns,
+  hasSpecificRole,
```
roseeichelmann, line 39: definitely open to other suggestions for these prop names `roleSpecificColumns` and `hasSpecificRole`

patrickm02L (approved):
Thanks for the demo. Approved. 🚢 Ship it!

atd-vze/src/Components/GridExportData.js
frankhereford: @patrickm02L I think we have agreement that the behavior where the application will add or remove columns from the export without a clear indication why is going to be difficult for the users, particularly when they are trying to export the data for use in fatality related work. Can you bring this to grooming next week with an issue? Thanks!

PR #1229, opened 2023-05-25
Implement FTT Index Advanced Search

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/11802

## Testing
**URL to test:** 
https://deploy-preview-1229--atd-vze-staging.netlify.app/#/fatalities

**Steps to test:**
1. Go to the Fatalities page. Use all of the different toggles. Check if the unit type toggle matches up with the fatalities that are showing up by clicking on the crash for the record, going to the Person tab in the Related Records table, and seeing if the type there is the same. You can also verify the other toggles for a few records
2. Turning on all of the toggles in the Roadway System or Status cards will give you all of the fatality records back. But if you turn on all the toggles in the Unit Type there will be some records missing out of the full amount because not all unit types are captured.
3. Use the export/save button to export the Fatalities data to a csv. Check out the csv, make sure the data looks right.

For merging to prod:

- [ ] Add person, primaryperson, and crash relationships to `view_fatalities`
- [ ] Add unit relationship to primary person and person tables

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vze/src/views/Crashes/RelatedRecordsTable.js
+++ b/atd-vze/src/views/Crashes/RelatedRecordsTable.js
@@ -85,7 +85,9 @@ const RelatedRecordsTable = ({
     const isFieldNull = fieldValue === null;
     const isFieldBlank = fieldValue === "";
 
-    return isFieldNull || isFieldBlank ? "NO DATA" : fieldValue;
+    return isFieldNull || isFieldBlank
+      ? "NO DATA"
+      : fieldValue.toString().toUpperCase();
   };
 
   return (
--- a/atd-vze/src/views/Fatalities/fatalityGridTableParameters.js
+++ b/atd-vze/src/views/Fatalities/fatalityGridTableParameters.js
@@ -94,7 +94,7 @@ export const fatalityGridTableAdvancedFilters = {
         filter: {
           where: [
             {
-              "_or: [ { _and: [ { person: { unit: { unit_desc_id: { _eq: 1 } } } }, { _or: [ { person: { unit: { veh_body_styl_id: { _eq: 71 } } } } , { person: { unit: { veh_body_styl_id: { _eq: 90 } } } } ] } ] }, { _and: [ { primaryperson: { unit: { unit_desc_id: { _eq: 1 } } } }, { _or: [ { primaryperson: { unit: { veh_body_styl_id: { _eq: 71 } } } } , { primaryperson: { unit: { veh_body_styl_id: { _eq: 90 } } } } ] } ] }]": null,
+              "_or: [ { _and: [ { person: { unit: { unit_desc_id: { _eq: 1 } } } }, { _or: [ { person: { unit: { veh_body_styl_id: { _eq: 71 } } } } , { person: { unit: { veh_body_styl_id: { _eq: 90 } } } } ] } ] }, { _and: [ { primaryperson: { unit: { unit_desc_id: { _eq: 1 } } } }, { _or: [ { primaryperson: { unit: { veh_body_styl_id: { _eq: 71 } } } } , { primaryperson: { unit: { veh_body_styl_id: { _eq: 90 } } } } ] } ] } ]": null,
             },
           ],
         },
@@ -135,7 +135,7 @@ export const fatalityGridTableAdvancedFilters = {
     ],
   },
   groupStatus: {
-    icon: "newspaper-o",
+    icon: "clipboard",
     label: "Status",
     filters: [
       {
@@ -263,7 +263,7 @@ case_id
 recommendation { rec_text }
 recommendation { rec_update }
 recommendation { atd__recommendation_status_lkp { rec_status_desc } }
-recommendation { recommendations_partners { atd__coordination_partners_lkp { coord_partner_desc }} }
+recommendation { recommendations_partners { atd__coordination_partners_lkp { coord_partner_desc } } }
 primaryperson { unit { unit_description { veh_unit_desc_desc } } }
 person { unit { unit_description { veh_unit_desc_desc } } }
 primaryperson { unit { body_style { veh_body_styl_desc } } }
--- a/atd-vze/src/views/Fatalities/fatalityGridTableParameters.js
+++ b/atd-vze/src/views/Fatalities/fatalityGridTableParameters.js
@@ -68,3 +68,139 @@ export const fatalityGridTableColumns = {
     type: "String",
   },
 };
+
+export const fatalityGridTableAdvancedFilters = {
+  groupStatus: {
+    icon: "map-marker",
+    label: "Status",
+    filters: [
+      {
+        id: "status_open_short_term",
+        label: "Open - Short Term",
+        filter: {
+          where: [
+            {
+              "recommendation: { recommendation_status_id: { _eq: 1 } } ": null,
+            },
+          ],
+        },
+      },
+      {
+        id: "status_open_long_term",
+        label: "Open - Long Term",
+        filter: {
+          where: [
+            {
+              "recommendation: { recommendation_status_id: { _eq: 2 } } ": null,
+            },
+          ],
+        },
+      },
+      {
+        id: "status_closed_no_action_rec",
+        label: "Closed - No Action Recommended",
+        filter: {
+          where: [
+            {
+              "recommendation: { recommendation_status_id: { _eq: 3 } } ": null,
+            },
+          ],
+        },
+      },
+      {
+        id: "status_closed_rec_implemented",
+        label: "Closed - Recommendation Implemented",
+        filter: {
+          where: [
+            {
+              "recommendation: { recommendation_status_id: { _eq: 4 } } ": null,
+            },
+          ],
+        },
+      },
+      {
+        id: "status_closed_no_action_taken",
+        label: "Closed - No Action Taken",
+        filter: {
+          where: [
+            {
+              "recommendation: { recommendation_status_id: { _eq: 5 } } ": null,
+            },
+          ],
+        },
+      },
+      {
+        id: "status_null",
+        label: "Null",
+        filter: {
+          where: [
+            {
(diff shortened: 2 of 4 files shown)


atd-vze/src/views/Fatalities/fatalityGridTableParameters.js
```diff
@@ -68,3 +71,202 @@ export const fatalityGridTableColumns = {
...
+    icon: "bicycle",
+    label: "Unit Type",
+    filters: [
+      {
+        id: "motor_vehicle",
+        label: "Motor Vehicle Fatality",
+        filter: {
+          where: [
+            {
+              "_or: [ { _and: [ { person: { unit: { unit_desc_id: { _eq: 1 } } } }, { _and: [ { person: { unit: { veh_body_styl_id: { _neq: 71 } } } }, { person: { unit: { veh_body_styl_id: { _neq: 90 } } } } ] } ] }, { _and: [ { primaryperson: { unit: { unit_desc_id: { _eq: 1 } } } }, { _and: [ { primaryperson: { unit: { veh_body_styl_id: { _neq: 71 } } } }, { primaryperson: { unit: { veh_body_styl_id: { _neq: 90 } } } } ] } ] } ]": null,
```
roseeichelmann, line 86: shoutout to @tillyw for some of these `where` filters from this [PR](https://github.com/cityofaustin/atd-vz-data/pull/1226) that I was able to adapt for the Fatalities page. the unit type filters look prettyyy crazy because I had to use an `or` operator since the record could be a person or a primary person.

atd-vze/src/views/Fatalities/fatalityGridTableParameters.js
```diff
@@ -68,3 +71,202 @@ export const fatalityGridTableColumns = {
...
+          ],
+        },
+      },
+      {
+        id: "status_null",
+        label: "Null",
+        filter: {
+          where: [
+            {
+              "_not: { recommendation: {recommendation_status_id: { _is_null: false } } }": null,
```
roseeichelmann, line 202: if i were to leave out the `not` and just check for records that have null recommendation statuses this filter wouldn't capture fatalities that dont have a recommendation associated with them at all yet.

atd-vze/src/views/Crashes/RelatedRecordsTable.js
```diff
@@ -85,7 +85,9 @@ const RelatedRecordsTable = ({
     const isFieldNull = fieldValue === null;
     const isFieldBlank = fieldValue === "";
 
-    return isFieldNull || isFieldBlank ? "NO DATA" : fieldValue;
+    return isFieldNull || isFieldBlank
+      ? "NO DATA"
+      : fieldValue.toString().toUpperCase();
```
roseeichelmann, line 90: i noticed that there were a few crash records that had lowercase text for some of the fields (saying "austin" instead of "AUSTIN") in the related records table which goes against the pattern and looked weird, so i added this to make sure all of the text is rendered in the UI in a consistent way.

atd-vze/src/views/Fatalities/fatalityGridTableParameters.js
```diff
@@ -68,3 +71,202 @@ export const fatalityGridTableColumns = {
...
+case_id
+recommendation { rec_text }
+recommendation { rec_update }
+recommendation { atd__recommendation_status_lkp { rec_status_desc } }
+recommendation { recommendations_partners { atd__coordination_partners_lkp { coord_partner_desc }} }
+primaryperson { unit { unit_description { veh_unit_desc_desc } } }
+person { unit { unit_description { veh_unit_desc_desc } } }
+primaryperson { unit { body_style { veh_body_styl_desc } } }
+person { unit { body_style { veh_body_styl_desc } } }
+crash { onsys_fl }
```
roseeichelmann, line 271: i chose these export fields based on what i assumed the vz team would expect to be able to export bc they didnt give an explicit list. since they wanted toggles for unit types i figured unit type and body style would be good fields to carry over into a csv export.

patrickm02L (approved):
## Review
Elegant implementation. It looks really nice, is intuitive, and works like a charm! ✨ 

## Testing

- [X] Go to the Fatalities page. Use all of the different toggles. Check if the unit type toggle matches up with the fatalities that are showing up by clicking on the crash for the record, going to the Person tab in the Related Records table, and seeing if the type there is the same. You can also verify the other toggles for a few records (0:00-1:35)
- [X] Turning on all of the toggles in the Roadway System or Status cards will give you all of the fatality records back. But if you turn on all the toggles in the Unit Type there will be some records missing out of the full amount because not all unit types are captured.
- [X] Use the export/save button to export the Fatalities data to a csv. Check out the csv, make sure the data looks right. (1:45-2:17)

## Notes

Tried a some additional advanced search functionality:
- [X] Search by victim name (2:33-2:49)
- [X] Search by date (2:50-3:19)

https://github.com/cityofaustin/atd-vz-data/assets/5566727/9e74c895-bc1d-42f9-8c18-96d642c30786

atd-vze/src/views/Fatalities/fatalityGridTableParameters.js
frankhereford, line 86: Agreed - big shoutout to @tillyw  for working out these `where` clauses. 

I'm a little concerned that they are hard to read in this form, which makes it tricky to debug them or to grok exactly what is going on. What would you think about reformatting these with whitespace to show how the parts work together?

Here's an example of how it might be done for this first one and I'm curious what you think.

```
    filters: [
      {
        id: "motor_vehicle",
        label: "Motor Vehicle Fatality",
        filter: {
          where: [
            {
              "_or: \
                [ \
                  { _and: \
                    [ \
                      { person: { unit: { unit_desc_id: { _eq: 1 } } } }, \
                      { _and: \
                        [ \
                          { person: { unit: { veh_body_styl_id: { _neq: 71 } } } }, \
                          { person: { unit: { veh_body_styl_id: { _neq: 90 } } } } \
                        ] \
                      } \
                    ] \
                  }, \
                  { _and: \
                    [ \
                      { primaryperson: { unit: { unit_desc_id: { _eq: 1 } } } }, \
                      { _and: \
                        [ \
                          { primaryperson: { unit: { veh_body_styl_id: { _neq: 71 } } } }, \
                          { primaryperson: { unit: { veh_body_styl_id: { _neq: 90 } } } } \
                        ] \
                      } \
                    ] \
                  } \
                ] \
              " : null,
            },
          ],
        },
      },
 ... ]
```

Because these clause like structures are acting as a key (not a value) in an object, we can't use a template literal there, which would look a little cleaner without the `\`s, but I still think this is more human readable.

frankhereford (changes requested):
Rose, this is great work. 

I think the `toUpperCase()` call is premium; it does bring things into accord under the existing pattern that we're used to for end users.

I've tested the application and it works great.

I'm putting this review as Request changes, because I think there are a couple of things that we might want to consider. I put a couple of suggestions for the labels that are currently "Null." I thin `null` makes a lot of sense in the context of the code, but I think that doesn't carry over to the UI as well. 

I also put up a possible way we can clarify the intent and presentation of the complex where clauses in `fatalityGridTableParameters.js`. I'm curious what you think about those.

If you don't agree with my requests, I'd love to chat about them, and I am very open to reconsidering and can turn this into an approval very quickly. This is great work - thank you!

atd-vze/src/views/Fatalities/fatalityGridTableParameters.js
roseeichelmann: Yeah this looks great, im going to implement this as well. Thanks Frank!

atd-vze/src/views/Fatalities/fatalityGridTableParameters.js
```diff
@@ -68,3 +71,202 @@ export const fatalityGridTableColumns = {
...
+          where: [
+            {
+              "recommendation: { recommendation_status_id: { _eq: 5 } } ": null,
+            },
+          ],
+        },
+      },
+      {
+        id: "status_null",
+        label: "Null",
```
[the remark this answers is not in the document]
mddilley: Just going to throw out another suggestion of "Not provided" because both null and undefined are both data types in JS. Not a strong feeling here because I think any word that Xavier and the VZ team can easily interpret as "no value provided by CRIS" is the most important part.

It gets confusing for me in the UI sometimes because CRIS lookup tables provide lots of values like "NO DATA", "NONE", "NOT REPORTED", etc that I confuse with words that we set in the code.

mddilley (approved):
This looks awesome, Rose! I went through your test steps, and this works so well. I added a comment about the wording for null filtering, but I think this is 100% shippable as is since Patrick is looking to get this in the release. We can always spin off an issue to work out the readability and UI feedback. 🚢 ✨

frankhereford (approved):
@mddilley wrote:

> I think this is 100% shippable as is since Patrick is looking to get this in the release

Good call Mike. I'm going to flip my review to approved in light of this observation and Rose saying she'd address the label and where clause formatting. Thank you @roseeichelmann!

atd-vze/src/views/Fatalities/fatalityGridTableParameters.js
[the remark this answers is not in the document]
roseeichelmann: Thanks for the input Mike! Also for context, the recommendation status is a value that the VZ team gives to crashes with fatalities (part of the Fatality Review Board feature) and its not something thats provided by CRIS. So if a fatality is missing a recommendation status its because the VZ team hasnt given it one yet. Im still not sure what would be the best toggle label to describe that, I personally like "None" but open to more opinions!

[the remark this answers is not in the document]
patrickm02L: Perhaps, "Missing Values" or "Not Available."

PR #1232, opened 2023-05-31
Refinements to Fatalities Advanced Search

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/12413

## Testing
**URL to test:** 
https://deploy-preview-1232--atd-vze-staging.netlify.app/#/fatalities

**Steps to test:**
1. Go to the Fatalities page, click on the advanced search button.
2. Do you like the new toggle names for null values in the on-system and recommendation status cards? The On-system flag is a value we get from CRIS, so "Unknown" as a label to represent null makes sense to me. Recommendation status is a value the VZ team gives to a fatal crash, so "None" makes sense to me bc its not really unknown, they just haven't given it a status yet so it doesn't exist. 
3. Can you think of better labels for these toggles? Leave your thoughts below!

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vze/src/views/Crashes/VictimNameField.js
+++ b/atd-vze/src/views/Crashes/VictimNameField.js
@@ -39,7 +39,7 @@ const VictimNameField = ({
   // Format name by concatenating first, middle, last or returning NO DATA
   const formatName = () => {
     var concatenatedName = "";
-    Object.keys(nameFieldConfig.subfields).map(field => {
+    Object.keys(nameFieldConfig.subfields).forEach(field => {
       if (personData?.[field] != null) {
         concatenatedName = concatenatedName.concat(" ", personData?.[field]);
       }
--- a/atd-vze/src/views/Fatalities/fatalityGridTableParameters.js
+++ b/atd-vze/src/views/Fatalities/fatalityGridTableParameters.js
@@ -83,7 +83,32 @@ export const fatalityGridTableAdvancedFilters = {
         filter: {
           where: [
             {
-              "_or: [ { _and: [ { person: { unit: { unit_desc_id: { _eq: 1 } } } }, { _and: [ { person: { unit: { veh_body_styl_id: { _neq: 71 } } } }, { person: { unit: { veh_body_styl_id: { _neq: 90 } } } } ] } ] }, { _and: [ { primaryperson: { unit: { unit_desc_id: { _eq: 1 } } } }, { _and: [ { primaryperson: { unit: { veh_body_styl_id: { _neq: 71 } } } }, { primaryperson: { unit: { veh_body_styl_id: { _neq: 90 } } } } ] } ] } ]": null,
+              "_or: \
+              [ \
+                { _and: \
+                  [ \
+                    { person: { unit: { unit_desc_id: { _eq: 1 } } } }, \
+                    { _and: \
+                      [ \
+                        { person: { unit: { veh_body_styl_id: { _neq: 71 } } } }, \
+                        { person: { unit: { veh_body_styl_id: { _neq: 90 } } } } \
+                      ] \
+                    } \
+                  ] \
+                }, \
+                { _and: \
+                  [ \
+                    { primaryperson: { unit: { unit_desc_id: { _eq: 1 } } } }, \
+                    { _and: \
+                      [ \
+                        { primaryperson: { unit: { veh_body_styl_id: { _neq: 71 } } } }, \
+                        { primaryperson: { unit: { veh_body_styl_id: { _neq: 90 } } } } \
+                      ] \
+                    } \
+                  ] \
+                } \
+              ] \
+              ": null,
             },
           ],
         },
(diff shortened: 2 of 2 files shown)


atd-vze/src/views/Fatalities/fatalityGridTableParameters.js
```diff
@@ -83,7 +83,32 @@ export const fatalityGridTableAdvancedFilters = {
...
+                    { _and: \
+                      [ \
+                        { primaryperson: { unit: { veh_body_styl_id: { _neq: 71 } } } }, \
+                        { primaryperson: { unit: { veh_body_styl_id: { _neq: 90 } } } } \
+                      ] \
+                    } \
+                  ] \
+                } \
+              ] \
+              ": null,
```
roseeichelmann, line 111: much more readable, thanks for the idea frank!

atd-vze/src/views/Fatalities/fatalityGridTableParameters.js
```diff
@@ -94,7 +119,32 @@ export const fatalityGridTableAdvancedFilters = {
...
+                    { _or: \
+                      [ \
+                        { primaryperson: { unit: { veh_body_styl_id: { _eq: 71 } } } } , \
+                        { primaryperson: { unit: { veh_body_styl_id: { _eq: 90 } } } } \
+                      ] \
+                    } \
+                  ] \
+                } \
+              ] \
+              ": null,
```
johnclary, line 147: 😵 TIL you can't use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) as object keys.  Oof this is a weird pattern. might be good to document in here somewhere what these magic body numbers mean?

frankhereford: I know, right!! That was exactly where my mind went first too to get multiline strings, but I guess it would require some bonkers just-in-time compiler magic to support keys that could shift at run-time. 

I totally agree that sticking some comments in there to denote what the magic numbers mean would be helpful. FWIW, Rose is getting these from Tilly who got them originally without magic-number explanations from Xavier, so I can totally get how they got here like this too - maybe even just a link back to the appropriate issue at the root of these where-like clauses might be helpful.

mddilley: Thank y'all for discussing this some more here. I looked a little closer and noticed that we get some console warnings with these multiline strings. I started thinking about how we could assign that string to a variable and use a template literal there along with referencing it as the key - which works. But, I'm still questioning why this string need to be a key with value of `null`. Gotta dig a little deeper. 🕵️‍♂️

Screenshot 2023-05-31 at 12 06 08 PM

Maybe something like this:
```js
const whereUnitIsMotorVehicle = `_or: 
[ 
  { _and: 
    [ 
      { person: { unit: { unit_desc_id: { _eq: 1 } } } }, 
      { _and: 
        [ 
          { person: { unit: { veh_body_styl_id: { _neq: 71 } } } }, 
          { person: { unit: { veh_body_styl_id: { _neq: 90 } } } } 
        ] 
      } 
    ] 
  }, 
  { _and: 
    [ 
      { primaryperson: { unit: { unit_desc_id: { _eq: 1 } } } }, 
      { _and: 
        [ 
          { primaryperson: { unit: { veh_body_styl_id: { _neq: 71 } } } }, 
          { primaryperson: { unit: { veh_body_styl_id: { _neq: 90 } } } } 
        ] 
      } 
    ] 
  } 
] 
`;

Comment (mddilley) [reply atd-vze/src/views/Fatalities/fatalityGridTableParameters.js:147]: Ok, it looks like the `{ whereFilter: null }` is a GridTable/gqlAbstract thing [handled here](https://github.com/cityofaustin/atd-vz-data/blob/a2e254942ce7fdc967cc0d5d07f41fcd63652e0a/atd-vze/src/queries/gqlAbstract.js#L389-L395). 100% outside the scope of this PR but I overlooked that aspect in my last review so I wanted to track it down.

Comment (johnclary) [reply atd-vze/src/views/Fatalities/fatalityGridTableParameters.js:147]: cool suggestion, Mike. going on a tangent now: we use this same gqlabstract code in Moped, so i went looking for an example of this config object taking a query as a key but couldn't find one. i think that's because the Moped search builds up the configs dynamically from the advanced search UI, so the same issue might be happening but it's hidden from us 😬

File: atd-vze/src/views/Crashes/VictimNameField.js [hunk 3]
```diff
@@ -39,7 +39,7 @@ const VictimNameField = ({
   // Format name by concatenating first, middle, last or returning NO DATA
   const formatName = () => {
     var concatenatedName = "";
-    Object.keys(nameFieldConfig.subfields).map(field => {
+    Object.keys(nameFieldConfig.subfields).forEach(field => {
```
roseeichelmann, line 42: just getting rid of a console warning here, not trying to return an array so `forEach` is better

atd-vze/src/views/Fatalities/fatalityGridTableParameters.js
roseeichelmann: i found out that if i wrap the template literal as an element in an array i can use it as an object key :0
i just pushed those changes up, lmk what yall think. definitely down to approach it with mikes suggestion instead if that makes more sense to yall!

johnclary: oooh slick! i don't have an opinion on this one 🚢

mddilley: That is a cool find that keeps the relevant bits in place so I say keep it. 🚀 In the end, I think you are backed in a corner here by the existing code - which happens! Just part of our journey of finding ways to build queries dynamically. I like adding the comment for future context as well. 🙏

mddilley (approved):
🛳️ it!

patrickm02L (approved):
## Review

Approved! This is good to go 💯 

## Testing
- [X] Go to the Fatalities page, click on the advanced search button.
- [ ] Do you like the new toggle names for null values in the on-system and recommendation status cards? The On-system flag is a value we get from CRIS, so "Unknown" as a label to represent null makes sense to me. Recommendation status is a value the VZ team gives to a fatal crash, so "None" makes sense to me bc its not really unknown, they just haven't given it a status yet so it doesn't exist.
- [X] Can you think of better labels for these toggles? Leave your thoughts below!

## Notes
"Unknown" and "None" make a lot of sense. Clear and simple.

Screen Shot 2023-06-05 at 8 02 51 AM

patrickm02L:
@roseeichelmann let's patch this in the latest release.

PR #1233, opened 2023-06-01
Updates crash mode trigger, socrata python script

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/12065

## Testing
**URL to test:** 
you can test this on [staging](https://visionzero-staging.austinmobility.io/editor/#/crashes)! i have run all the scripts in this pr on staging, with the exception of the python script (to update the socrata open data portal dataset that powers the viewer), which should deploy when this pr is merged to main.

this pr adds new mode categories and includes a one-time script to update the crashes table's mode category metadata column to reflect changes made to units records (i ran this on staging as well, and we will need to run it on prod before the next release), and a trigger which will automate this process going forward. this should result in more accurate mode data in the vision zero viewer and will make it possible to represent micromobility devices and e-scooters in the viewer

we will need a separate issue (if there isn't one already) to make sure socrata is updating properly and make sure vzv can represent the new categories. thanks y'all!

**Steps to test:**
- open a record in the crashes table on staging and the [hasura staging console](https://vzd-staging.austinmobility.io/console/api-explorer) in a separate tab
- change the 'type' on one of them to be 'micromobility device' and the body style to 'unknown'
- query the record in the hasura console's graphiql interface by running the following query and replacing the crash id with the current record
```
query MyQuery {
  atd_txdot_crashes(where: {crash_id: {_eq: 18900579}}) {
    atd_mode_category_metadata
    units {
      atd_mode_category
      body_style {
        veh_body_styl_desc
        veh_body_styl_id
      }
      unit_desc_id
      unit_description {
        veh_unit_desc_desc
        veh_unit_desc_id
      }
      unit_id
    }
  }
}
```
- the `atd_txdot_crashes` `atd_mode_category_metadata` column for the updated record should show 'micromobility device' (this should allow the vz team to account for micromobility devices that are not scooters)
- change the body style to 'e-scooter' and rerun the query
- the metadata column should show 'e-scooter'
- change the type to 'motor vehicle' and body style to 'police motorcycle' and rerun the query
- the metadata column should show 'motorcycle' (these were previously not being counted as motorcycles)
- change the type to 'motorized conveyance' and body style to 'unknown' and rerun the query
- the metadata column should show 'motorized conveyance'
- try out other combinations and rerun the query to confirm the metadata matches the unit record
---
#### Ship list
- [ ] Code reviewed


atd-vzd/triggers/get_crash_modes.sql
```diff
@@ -19,24 +19,32 @@ BEGIN
         FROM (
             SELECT unit_id, (
                        CASE
-                           -- PEDALCYCLIST / BYCICLE
+                           -- PEDALCYCLIST / BICYCLE (5)
```
roseeichelmann, line 22: i like how you added the mode id into the comment!

roseeichelmann (approved):
this looks awesome tilly, all the test steps worked! ✨
my only question is do we also need a trigger that happens on insert to the `atd_txdot_crashes` table to get the `atd_mode_category_metadata` for new records we get from cris? bc currently that column is null for new records, not sure if thats how its supposed to work though!
thanks for knocking this out and untangling everything that was involved in this!! 🚀🚀

atd-vzd/migrations/migration_atd_txdot_units_2023_06_29--1509.sql
```diff
@@ -0,0 +1,6 @@
+-- trigger when atd_mode_category is changed
+CREATE OR REPLACE TRIGGER atd_txdot_units_mode_category_metadata_update
+    AFTER UPDATE ON atd_txdot_units
+    FOR EACH ROW
+    WHEN (OLD.atd_mode_category IS DISTINCT FROM NEW.atd_mode_category)
+    EXECUTE FUNCTION atd_txdot_units_mode_category_metadata_update();
```
roseeichelmann, line 6: wondering if we need another trigger like this on inserts to the `atd_txdot_crashes` table?

mddilley: Rose - I could use some confirmation on whether I'm looking at this wrong but it appears to occur on insert and update in the `atd_txdot_crashes_updates_audit_log` function. It is buried in there and reminds me of some of the conversation about improvements we all had in one of the recent meetings. 🙏

mddilley (commented):
Going to pick back up on this after playing with the triggers more locally, but I left a few comments for now. All of the test steps check out beautifully! ✨

atd-vzd/triggers/atd_txdot_units_create_update.sql
mddilley, line 1: Beyond the scope of this issue, but it would be cool to find a way to reuse the CASE statement in this function that is also in `atd_txdot_units_create_update`

atd-vzd/migrations/migration_atd_txdot_units_2023_06_29--1509.sql
tillyw: thanks, y'all! as far as i know it is buried [here](https://github.com/cityofaustin/atd-vz-data/blob/2a1236084f647d2c074bbf320a693cf432a73794/atd-vzd/triggers/atd_txdot_crashes_updates_audit_log.sql#L152) in a catch-all trigger that runs on insert or update to the `atd_txdot_crashes` crashes table. it was difficult for me to find initially because the trigger is labeled as an audit log trigger but does much more than that. definitely agree that making these triggers more discrete would make them much more manageable!

tillyw: @roseeichelmann i just reread your comment and it seems like maybe the trigger isn't firing when a new record is added - if that's the case i'll definitely need to do some digging to figure out why!

atd-vzd/triggers/atd_txdot_units_mode_category_metadata_update.sql
```diff
@@ -0,0 +1,11 @@
+CREATE OR REPLACE FUNCTION public.atd_txdot_units_mode_category_metadata_update()
+ RETURNS trigger
+ LANGUAGE plpgsql
+AS $function$
+    BEGIN
+        UPDATE atd_txdot_crashes SET atd_mode_category_metadata = get_crash_modes(crash_id)
+        WHERE (atd_txdot_crashes.crash_id = NEW.crash_id);
```
mddilley, line 7: I was reading about how `NEW` and `OLD` are available automagically. Do you know if this `NEW.crash_id` is referring to the unit's crash ID from where this is called [here](https://github.com/cityofaustin/atd-vz-data/pull/1233/files#diff-fbbc4d9a8a4ef8f9129aad0c174750571255c4d8a25a7476cebdd31d937f4278)?

I'm just curious to learn more, and I'm getting hung up on how the trigger knows what crash record to update.

mddilley (approved):
This looks great! I made all the necessary updates to my local instance to test it out a little while using `EXPLAIN` with some updates so I could learn more, and I'm seeing everything working as described. Thanks for wading into all of this and arriving at a solution that considers the unit updates. 🙏 🚢 🚀 🙌

atd-vzd/migrations/migration_atd_txdot_units_2023_06_29--1509.sql
roseeichelmann: @mddilley ahh yes it does look like there is a trigger on the `atd_txdot_crashes` table that executes the function `atd_txdot_crashes_updates_audit_log` where the metadata column is updated like @tillyw linked to. that `atd_txdot_crashes_updates_audit_log` function is so involved and updates so many diff things its hard to keep track of, it seems like a really good candidate to downsize/move away from using it 😅 but for now if its correctly updating the metadata column on inserts to the crashes table then ignore my comment, but looks like maybe its not?

atd-vzd/triggers/atd_txdot_units_mode_category_metadata_update.sql
roseeichelmann: Yeah thats right, the functions called by triggers use the `NEW` and `OLD` values from the table the trigger was called on. So it would use the unit's `crash_id`. I think it works bc units share the crash id with their associated crash, so a trigger can happen on updates to the unit table and cause changes to the correct crash record by using the crash id to identify the crash record (the `WHERE` clause in the `UPDATE` statement here).

mddilley: Thanks and that's pretty rad - learning a lot in this PR! 😎

atd-vzd/migrations/migration_atd_txdot_units_2023_06_29--1509.sql
roseeichelmann: nvm, on a second glance at new crash records it looks like they do have the `atd_mode_category_metadata` filled out. so i think that value is being calculated on inserts like we expect, my bad!

frankhereford (approved):
I went over the code and tested this. Great instructions; I was able to test the various combinations easily. This looks 🚀 to me! 

Thank you Tilly for this work and to Rose & Mike for the good discussion and observations in here.

PR #1234, opened 2023-06-05
Add Engineering Areas to the VZ database

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/12444

All the credit goes to @johnclary for the [Moped council district](https://github.com/cityofaustin/atd-moped/tree/main/moped-toolbox/council_districts_load) loading scripts, I just modified them to pull the engineering areas. 

## Testing
**URL to test:** 
None

**Steps to test:**
1. If you want you can test the scripts on your local VZ DB, but you dont have to
2. You can see a preview of the engineering areas in data/engineering_areas and compare to https://austin.maps.arcgis.com/apps/mapviewer/index.html?layers=28c369f7ab534c64a26a646fca3fd6ec

For merge to prod:

- [x] Add `engineering_areas` table to the db

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vzd/schema/engineering_areas.sql
+++ b/atd-vzd/schema/engineering_areas.sql
@@ -1,14 +1,12 @@
 CREATE TABLE public.engineering_areas(
     area_id integer,
-    label text NOT NULL,
+    label text NOT NULL unique,
     geometry geometry(MultiPolygon, 4326) NOT NULL
 );
 
 ALTER TABLE ONLY public.engineering_areas
     ADD CONSTRAINT engineering_areas_pkey PRIMARY KEY (area_id);
 
-CREATE UNIQUE INDEX engineering_areas_label_key ON public.engineering_areas USING btree (label);
-
 CREATE INDEX engineering_areas_gix
-ON engineering_areas
+ON public.engineering_areas
 USING GIST (geometry);


atd-vzd/schema/engineering_areas.sql
frankhereford, line 1: I think it may be helpful here to add a spatial index (bounding box) GIST index on the geometry column. I think it'd be done with something like:

```
CREATE INDEX engineering_areas_gix
ON engineering_areas
USING GIST (eng_area_geometry);
```

atd-vzd/schema/engineering_areas.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE public.engineering_areas(
+    id integer NOT NULL,
+    eng_area_label text,
+    eng_area_geometry geometry,
```
frankhereford, line 4: What do you think about calling these columns something like "label" and "geometry" instead of repeating the table name in an abbreviated form? Since the table only holds engineering areas, I think you could get away with less specific column names.

atd-toolbox/engineering_areas_load/process_areas.js
frankhereford, line 1: This is so premium to have a script to do this. Thank you to @johnclary for this original work and thank you to @roseeichelmann for demonstrating what a reusable and useful ability this is! This is just so great.

atd-toolbox/engineering_areas_load/data/engineering_areas.geojson
frankhereford, line 1: I think we can drop this out of the repo. I would rather not have another copy around that could confuse people who are looking for the latest data. I would, however, find a place to include a link to our AGOL resource which should remain up to date.

atd-toolbox/engineering_areas_load/.nvmrc
frankhereford, line 1: Little details that save the day - thank you!

frankhereford (changes requested):
Rose, I think this is just super work. Thank you!

Getting a viz on my local DB showing this data - it's perfect. 

image

I'm requesting changes with regard to my comment about the `CREATE` statement not executing and my question about an index on the geometry. Really though, after the `CREATE` fix to get it running, I'm eager to turn this into approval. Thank you again!

atd-toolbox/engineering_areas_load/package-lock.json
frankhereford, line 1: Should have added this to my review, but I think this might be something that can be dropped too, unless you're reliant on very specific versions of tooling. Since it's a toolbox app, I think we want to specify as little as we can get away with. I'd like other's input here though; node apps are far from my forte.

mddilley: I think that this is ok to leave in here since John did and we have done that in all other JS scripts or apps that I'm aware of. I can also see your point about only leaving only what is 100% necessary.

mddilley (approved):
I ran this SQL and the scripts locally, and I can see all three of the engineering areas in my DB. +1 to Frank's comment about being happy to see John's code reused here (and another Node script 🚀 for that matter). I'm throwing an approval on here even though I ran into the same snag in the `CREATE` statement. I know that it will be updated soon enough. Thanks - this is awesome! 🚢

atd-vzd/schema/engineering_areas.sql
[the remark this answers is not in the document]
roseeichelmann: oops i noticed that when i ran it on my local db but forgot to fix it here, thanks for catching that!

atd-toolbox/engineering_areas_load/data/engineering_areas.geojson
roseeichelmann: i think maybe it would be helpful to leave here bc it gives a reference to what we are currently using in our database/app and can also help hold us accountable to make sure we keep it up to date? but i see what you mean about confusing people, maybe i could make a note in the README linking to the AGOL resource and saying thats the best place to get the most up to date data?

atd-vzd/schema/engineering_areas.sql
roseeichelmann: yeah i can definitely make these less specific, i never know whats a better route for column names but that makes sense to me!

atd-vzd/schema/engineering_areas.sql
roseeichelmann: oh thats a cool idea, ill do that! thanks :-)

atd-toolbox/engineering_areas_load/package-lock.json
roseeichelmann: yeah i was following the pattern of johns work and other apps in our repos, but truly i dont have much of an opinion on what is most helpful.. i have referred to this [comment](https://stackoverflow.com/a/44210813) though in the past describing why package-lock is meant to be source controlled so i have always just went with that logic bc it makes sense to me.

frankhereford: That works for me Rose, thanks!

atd-toolbox/engineering_areas_load/data/engineering_areas.geojson
frankhereford: I like your solution a ton. I think a readme in the same directory as the geojson is awesome. Maybe it could read:

```
This geoJSON file is an intermediate product from loading the engineering areas into the Vision Zero database. The upstream, and up to date, source of this data is found here: https://austin.maps.arcgis.com/home/item.html?id=28c369f7ab534c64a26a646fca3fd6ec. This file is not intended to be used as a source of the TPW engineering areas, but rather to show the form of the data when it was loaded.
```

atd-vzd/schema/engineering_areas.sql
frankhereford: Awesome - I think our queries will be more readable in the future because we'll be doing stuff like `engineering_areas.label` which captures what's in the column really well. Thanks!

frankhereford (approved):
Thanks Rose!

atd-toolbox/engineering_areas_load/.gitignore
frankhereford, line 1: @roseeichelmann - sorry for another comment outside of a review, but I've already approved, and I don't want to revoke that for this. 

Do you mind adding a second line to this file that reads:
```
node_modules
```

This will prevent git from wanting to pick up the output of `npm install` when it's run in here.

Thanks!!

atd-vzd/schema/engineering_areas.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE public.engineering_areas(
+    id integer NOT NULL,
+    eng_area_label text,
```
johnclary, line 3: it's not obvious from the toolbox script, but in Moped we have a unique constraint on the council district number in order to prevent dupe inserts. in the same way, i think you probably want to apply a unique index to this eng area label.

johnclary: just took a quick look at that layer. are y'all sure we don't need any of these other columns?

Screenshot 2023-06-06 at 10 58 28 AM

johnclary (commented):
glad you were able to repurpose some Moped code—this looks great! i have not tested but did flag a few things for consideration 😄

atd-vzd/schema/engineering_areas.sql
roseeichelmann: im not sure about the other columns, afaik this feature is just so that the the vz team can search for fatalities by engineering area/get lists of fatalities for each engineering area.

patrickm02L: Teams Mary on 6/6/23 about other columns, she said she wanted Council Districts, but not the way they were being aggregated in to the Engineering Areas in the map. Made a separate Issue [#12503](https://github.com/cityofaustin/atd-data-tech/issues/12503) to capture them in the crash export.

roseeichelmann:
Just pushed up a few last changes addressing yalls comments! Ran into a really weird situation after changing the column names that was preventing the insert mutation from working in the `insert_areas` script that I think had to do with metadata being corrupted, but John and I paired for awhile and I finally got it to work on my local db! Just had to explicitly make the `geometry` column a MultiPolygon type.

atd-vzd/schema/engineering_areas.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE public.engineering_areas(
+    area_id integer,
```
roseeichelmann, line 2: idk if this is repetitive again  and if it should just be `id` but i feel like `area_id` helps make it more clear that this isnt a random id we are creating but that we pull it from the data itself? idek

johnclary: since there's a decent chance this data will be refreshed from the same source down the road, i would lean toward preserving the column names as they are in the source data, but i draw the line at capitalization. so this column would be `engineering_area_id`.

johnclary: derp, i see Frank already made a comment with a different approach. sorry for the back n forth. ignore me...

PR #1235, opened 2023-06-07
CRIS import in a docker image

frankhereford:
## Introduction

This PR is big at 1,500+ lines, but it's a lot of old code from https://github.com/cityofaustin/atd-prefect/pull/28.  This ETL is jumping out of the `atd-prefect` repo into here, and that makes for new files instead of edited ones. I'm going to point out where meaningful changes have occurred. 

## Associated issues and PRs

This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/11732.
This PR is a sibling of https://github.com/cityofaustin/atd-airflow/pull/108.

## Testing

I think this PR is best tested via reading the code, focusing on the parts that I've called out as new to this port. Everything in it is fair game of course, but I don't want to force y'all to wade through 1500 lines when we're talking about couple hundred or so. Most of this edit came in the form of the delete button because we're translating this from a ETL-system-integrated DAG into a vanilla python script.

In addition to reading this code, please see the [sibling PR](https://github.com/cityofaustin/atd-airflow/pull/108) in the airflow repo. By testing that PR, you'll be executing the contents of this one.

## Continuous Integration

I think it would be a neat spin-off issue from this work to create a GitHub workflow action that is pointed at a certain folder in a repo and will rebuild a docker container from the Dockerfile in that folder when any files in or below that folder change. For this new docker image, it would be pointed at `atd-etl/cris_import`. This would run on PR merges. What do y'all think?

---
#### Ship list
- [ ] Code reviewed


atd-etl/cris_import/Dockerfile
frankhereford, line 1: Here's the definition of the docker image the script runs in. I believe it is sufficiently unique to live in its own container and not try to make it share one with any other DAGs. That wouldn't be too hard to do, but this seems different enough to me. 🤷🏻‍♂️ I'm curious what y'all think about that.

atd-etl/cris_import/docker-compose.yml
frankhereford, line 1: This is a barebones docker compose file which is helpful when developing the script locally. It really just serves to remember all the arguments that you'd otherwise have to put in a `docker run .......` command, and let's you just `docker compose run cris-import bash` to get a shell and start doing dev stuff.

atd-etl/cris_import/entrypoint.sh
frankhereford, line 1: This is the script that is run by default on the container. The `sshd` line starts the local SSH service so that the container can use itself as a bastion. The keys needed are setup in the Dockerfile.

atd-etl/cris_import/lib/graphql.py
frankhereford, line 1: Unchanged file from a previous version of this script running on prefect.

atd-etl/cris_import/lib/helpers_import.py
frankhereford, line 1: This function used to be made available to the previous generation of this script via a git submodule, which brought in the VZ repo into the prefect one. I could copy in the file out of the source in the ETL directory above, but I think this is cleaner to just borrow the function I need and stick it in the context of this program. It's nice to freeze the state of the function too, so we don't have an outside library dependency.

atd-etl/cris_import/lib/mappings.py
frankhereford, line 1: Unchanged file from a previous version of this script running on prefect.

atd-etl/cris_import/lib/sql.py
frankhereford, line 1: Unchanged file from a previous version of this script running on prefect.

atd-toolbox/materials_from_ldm_conversion/.gitignore
frankhereford, line 1: This was an error that I didn't intend to commit and push, but I would like to ask if I can keep it. I have a environment in this folder, and anytime I switch away from my LDM work, it leaves the untracked `env` file in there hanging out in my local checkout. It then becomes so easy to commit and push, having this file will make my dev life a ton easier for a while.

atd-etl/cris_import/cris_import.py
```diff
@@ -0,0 +1,801 @@
...
+            "opitem": "SFTP Endpoint Key",
+            "opfield": ".private key",
+            "opvault": VAULT_ID,
+            },
+    }
+
+    # instantiate a 1Password client
+    client: Client = new_client(ONEPASSWORD_CONNECT_HOST, ONEPASSWORD_CONNECT_TOKEN)
+    # get the requested secrets from 1Password
+    return onepasswordconnectsdk.load_dict(client, REQUIRED_SECRETS)
```
frankhereford, line 190: These lines, 1-190 are all worth giving a close look at. It's a whole new `main()` function and a `get_secrets()` function.

atd-etl/cris_import/cris_import.py
```diff
@@ -0,0 +1,801 @@
...
+def download_extract_archives():
+    """
+    Connect to the SFTP endpoint which receives archives from CRIS and
+    download them into a temporary directory.
+
+    Returns path of temporary directory as a string
+    """
+
+    with SshKeyTempDir() as key_directory:
+        write_key_to_file(key_directory + "/id_ed25519", SFTP_ENDPOINT_SSH_PRIVATE_KEY + "\n")
```
frankhereford, line 208: You can see here I'm now using a class, defined below, to give me a way to get a directory that automatically is removed when it goes out of scope. A bit overkill now that this is in an ephemeral docker container, but I had the code all written, so it was easier to just do this fancy style.

atd-etl/cris_import/cris_import.py
```diff
@@ -0,0 +1,801 @@
...
+    Delete the archives which have been processed from the SFTP endpoint
+
+    Arguments:
+        zip_location: Stringing containing path of a directory containing the zip files downloaded from SFTP endpoint
+
+    Returns: None
+    """
+    with SshKeyTempDir() as key_directory:
+        write_key_to_file(key_directory + "/id_ed25519", SFTP_ENDPOINT_SSH_PRIVATE_KEY + "\n") 
+
```
frankhereford, line 328: We write out that SSH key again we got from 1password. I found it quicker to write it out again than pass the first location around in a variable down to here.

atd-etl/cris_import/cris_import.py
```diff
@@ -0,0 +1,821 @@
...
+    """
+    Connect to the SFTP endpoint which receives archives from CRIS and
+    download them into a temporary directory.
+
+    Returns path of temporary directory as a string
+    """
+
+    with SshKeyTempDir() as key_directory:
+        write_key_to_file(key_directory + "/id_ed25519", SFTP_ENDPOINT_SSH_PRIVATE_KEY + "\n") 
+
```
frankhereford, line 216: We're starting a long stretch of this file which is essentially unchanged from previous production versions.

atd-etl/cris_import/cris_import.py
```diff
@@ -0,0 +1,821 @@
...
+        cursor = pg.cursor()
+        sql = f"DROP SCHEMA IF EXISTS {map_state['import_schema']} CASCADE "
+        cursor.execute(sql)
+
+        pg.commit()
+        cursor.close()
+        pg.close()
+
+    return map_state
+
```
frankhereford, line 795: Everything from here on out is new.

atd-etl/cris_import/cris_import.py
```diff
@@ -0,0 +1,821 @@
...
+def write_key_to_file(path, content):
+    # Open the file with write permissions and create it if it doesn't exist
+    fd = os.open(path, os.O_WRONLY | os.O_CREAT, 0o600)
+
+    # Write the content to the file
+    os.write(fd, content.encode())
+
+    # Close the file
+    os.close(fd)
```
frankhereford, line 817: Both of these functions are used to provide a place to briefly store SSH keys to disk because SSH insists on getting that way. This gives a place to store them and writes them out with the correct file permissions.

atd-etl/cris_import/Dockerfile
frankhereford, line 1: Here's the definition of the docker image the script runs in. I believe it is sufficiently unique to live in its own container and not try to make it share one with any other DAGs. That wouldn't be too hard to do, but this seems different enough to me. 🤷🏻‍♂️ I'm curious what y'all think about that.

atd-etl/cris_import/docker-compose.yml
frankhereford, line 1: This is a barebones docker compose file which is helpful when developing the script locally. It really just serves to remember all the arguments that you'd otherwise have to put in a docker run ....... command, and let's you just docker compose run cris-import bash to get a shell and start doing dev stuff.

atd-etl/cris_import/entrypoint.sh
frankhereford, line 1: This is the script that is run by default on the container. The sshd line starts the local SSH service so that the container can use itself as a bastion. The keys needed are setup in the Dockerfile.

atd-etl/cris_import/lib/graphql.py
frankhereford, line 1: Unchanged file from a previous version of this script running on prefect.

atd-etl/cris_import/lib/helpers_import.py
frankhereford, line 1: This function used to be made available to the previous generation of this script via a git submodule, which brought in the VZ repo into the prefect one. This script would import the function out of that checkout's python files. Since we're actually back in the VZ repo, we could just import something like ../app/process/helpers_import.py, but I think this is cleaner to just borrow the function I need and stick it in the context of this program. It's nice to freeze the state of the function too, so we don't have an outside library dependency.

atd-etl/cris_import/lib/mappings.py
frankhereford, line 1: Unchanged file from a previous version of this script running on prefect.

atd-etl/cris_import/lib/sql.py
frankhereford, line 1: Unchanged file from a previous version of this script running on prefect.

atd-toolbox/materials_from_ldm_conversion/.gitignore
frankhereford, line 1: This was an error that I didn't intend to commit and push, but I would like to ask if I can keep it. I have a environment in this folder, and anytime I switch away from my LDM work, it leaves the untracked env file in there hanging out in my local checkout. It then becomes so easy to commit and push, having this file will make my dev life a ton easier for a while.

atd-etl/cris_import/cris_import.py
```diff
@@ -0,0 +1,821 @@
...
+            "opitem": "RDS Bastion Key",
+            "opfield": ".private key",
+            "opvault": VAULT_ID,
+            },
+    }
+
+    # instantiate a 1Password client
+    client: Client = new_client(ONEPASSWORD_CONNECT_HOST, ONEPASSWORD_CONNECT_TOKEN)
+    # get the requested secrets from 1Password
+    return onepasswordconnectsdk.load_dict(client, REQUIRED_SECRETS)
```
frankhereford, line 197: Everything north of line 197 is worth giving a close look at. It's a whole new main() function and a get_secrets() function.

frankhereford (commented):
I meant to leave these comments as individual comments, but I started a review by accident.

frankhereford:
@Charlie-Henry wrote:

> I don't want to get in the way of this getting approved but I am having a hard time following what's going on. Just the one comment about the dockerignore. Also maybe an excalidraw diagram or something to put in the docs would help.

I think that is a fantastic idea Charlie. I'll try to cook one up and get it in a README.md for this docker container. Thank you for the suggestion, and thanks again for catching the `env` getting bundled earlier. 🙏

johnclary:
@frankhereford without asking you to write too much, what are are the hurdles to being able run this ETL against a local VZ instance?

frankhereford:
@johnclary wrote:

> @frankhereford without asking you to write too much, what are are the hurdles to being able to run this ETL against a local VZ instance?

Hi John - thanks for the question. I think there are no substantive hurdles to doing this. This setup is how it was developed and tested. 

I think the directions in the [sibling PR](https://github.com/cityofaustin/atd-airflow/pull/108) in the airflow repo may be helpful to get this going. 

Let me know if I can provide anything else that would make this easier.

Thanks!

atd-etl/cris_import/cris_import.py
```diff
@@ -0,0 +1,821 @@
...
+class SshKeyTempDir:
+    def __init__(self):
+        self.path = None
+
+    def __enter__(self):
+        self.path = tempfile.mkdtemp(dir='/tmp')
+        return self.path
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        shutil.rmtree(self.path)
```
johnclary, line 807: TIL `__enter__` and `__exit__`: i've always wondered how you set up a context manager like this! v cool.

atd-etl/cris_import/.dockerignore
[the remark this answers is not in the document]
frankhereford: @Charlie-Henry Thanks again for your idea here and for reviewing the diagram. 

It's been added to a new README.md file and can be seen here:

https://github.com/cityofaustin/atd-vz-data/tree/docker-cris-import/atd-etl/cris_import

frankhereford:
@Charlie-Henry wrote: 

> Also maybe an excalidraw diagram or something to put in the docs would help.

Thanks for this idea Charlie. I've put a diagram in the README.md for this file, and it can be seen here: https://github.com/cityofaustin/atd-vz-data/tree/docker-cris-import/atd-etl/cris_import

chiaberry:
I mentioned this in our 1:1 last week, here's an example of where we have a github action that runs when a specific folder is updated https://github.com/cityofaustin/atd-moped/blob/main/.github/workflows/atd_moped_api.yml#L19

chiaberry (approved):
I compared this to the existing code in the prefect repo, and I find this one easier to understand what's going on. 

I didn't look too closely at the code that you said was copy/paste from the prefect version of this. 

Thanks for bringing this over to airflow! Big relief

atd-etl/cris_import/cris_import.py
```diff
@@ -0,0 +1,821 @@
...
+        for logical_group in logical_groups_of_csvs:
+            desired_schema_name = create_import_schema_name(logical_group)
+            schema_name = create_target_import_schema(desired_schema_name)
+            pgloader_command_files = pgloader_csvs_into_database(schema_name)
+            trimmed_token = remove_trailing_carriage_returns(pgloader_command_files)
+            typed_token = align_db_typing(trimmed_token)
+            align_records_token = align_records(typed_token)
+            clean_up_import_schema(align_records_token)
+    remove_archives_from_sftp_endpoint(zip_location)
+    upload_csv_files_to_s3(archive)
```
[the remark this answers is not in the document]
roseeichelmann: +1!

roseeichelmann (approved):
this is massive, thanks for handling this frank! i skimmed through all the code and focused especially on the stuff you called out as being new. i like the excalidraw diagram you drew up!

atd-etl/cris_import/cris_import.py
mddilley: +1 - this is cool pattern!

mddilley (approved):
Looks great - I did not test but read through the new parts since I've reviewed some of this before. Thanks for adding the Excalidraw! It is a great refresher and pair nicely along with the new loop in `main()`. Thanks for taking this on! 🚀

frankhereford:
@mddilley wrote:

> This global keyword also caught my eye 👀 Is it possible to avoid using globals by declaring secrets at the top level and then assigning it here on line 34? Not sure if there is a limitation with that route.

I know, it's crazy you can't `global variable_name = value` to me. But i've found that either you `global` them while not in the global scope and then define them, or you define them in the global scope as you suggested. I'm trying to get in the habit of doing as little code as possible in my python outside of functions because I feel like it's good style - such that one's python programs can be python libraries without modification just by importing them. 

I waffled back and forth on this real hard, but I decided I wanted to go `global` definitions in the `main()` function so that they are entirely scoped & defined in single section of the file. I realize that because `main()` sits at the top of the file, this isn't too far away from where you'd scope them in practice, but "outside vs. inside a function" seems like a significant separation of context. Imagine that `main()` were defined as the last function of the file, which is equally valid, then maybe this choice makes sense? I know I'm like beyond splitting hairs, so please don't take me too seriously here!

Either way - it's not pretty, and I'd be so happy to conform to the pattern you suggest if you think that's better after reading this. What feels best to you?

atd-etl/cris_import/cris_import.py
```diff
@@ -0,0 +1,821 @@
...
+)  # our current environment from ['production', 'development']
+ONEPASSWORD_CONNECT_TOKEN = os.getenv("OP_API_TOKEN")  # our secret to get secrets 🤐
+ONEPASSWORD_CONNECT_HOST = os.getenv("OP_CONNECT")  # where we get our secrets
+VAULT_ID = os.getenv("OP_VAULT_ID")
+
+
+def main():
+    secrets = get_secrets()
+
+    # 😢 why not `global variable = value`??
```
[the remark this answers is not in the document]
frankhereford: @mddilley wrote:

> This global keyword also caught my eye 👀 Is it possible to avoid using globals by declaring secrets at the top level and then assigning it here on line 34? Not sure if there is a limitation with that route.

I know, it's crazy you can't `global variable_name = value` to me. But i've found that either you `global` them while not in the global scope and then define them, or you define them in the global scope as you suggested. I'm trying to get in the habit of doing as little code as possible in my python outside of functions because I feel like it's good style - such that one's python programs can be python libraries without modification just by importing them. 

I waffled back and forth on this real hard, but I decided I wanted to go `global` definitions in the `main()` function so that they are entirely scoped & defined in single section of the file. I realize that because `main()` sits at the top of the file, this isn't too far away from where you'd scope them in practice, but "outside vs. inside a function" seems like a significant separation of context. Imagine that `main()` were defined as the last function of the file, which is equally valid, then maybe this choice makes sense? I know I'm like beyond splitting hairs, so please don't take me too seriously here!

Either way - it's not pretty, and I'd be so happy to conform to the pattern you suggest if you think that's better after reading this. What feels best to you?

patrickm02L (approved):
Adding my rubber stamp based on the Dev reviews!

PR #1238, opened 2023-06-09
Make Onsys Flag editable

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/12078

~We decided to make this flag editable for all crashes, not just temps.~
After updates from Xavier we are making the on sys flag only editable for temp crashes.

## Testing
**URL to test:** 
[Netlify](https://deploy-preview-1238--atd-vze-staging.netlify.app/#/crashes)

**Steps to test:**
1. Click on a temp crash, scroll down to the "Flags" section, test editing the "On TxDOT Highway System Flag".
2. Click on a non-temp crash, the flag should not be editable.

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vze/src/Components/DataTable.js
+++ b/atd-vze/src/Components/DataTable.js
@@ -29,7 +29,6 @@ const DataTable = ({
   handleFieldUpdate,
   handleButtonClick,
   downloadGlobal,
-  isTempRecord,
 }) => {
   // Disable edit features if only role is "readonly"
   const { getRoles } = useAuth0();
@@ -76,13 +75,8 @@ const DataTable = ({
                       const fieldLabel = fieldConfigObject.label;
 
                       // Disable editing if user is only "readonly"
-                      if (
-                        (fieldConfigObject.editable ||
-                          fieldConfigObject.isEditable) &&
-                        isReadOnly(roles)
-                      ) {
+                      if (fieldConfigObject.editable && isReadOnly(roles)) {
                         fieldConfigObject.editable = false;
-                        fieldConfigObject.isEditable = false;
                       }
 
                       // Set data table (alternate if defined in data map)
@@ -208,15 +202,12 @@ const DataTable = ({
                             )}
                           </td>
                           <td>
-                            {(fieldConfigObject.editable ||
-                              (fieldConfigObject.isEditable &&
-                                fieldConfigObject.isEditable(isTempRecord))) &&
-                              !isEditing && (
-                                <i
-                                  className="fa fa-pencil edit-toggle"
-                                  onClick={() => setEditField(field)}
-                                />
-                              )}
+                            {fieldConfigObject.editable && !isEditing && (
+                              <i
+                                className="fa fa-pencil edit-toggle"
+                                onClick={() => setEditField(field)}
+                              />
+                            )}
                           </td>
                         </tr>
                       );
--- a/atd-vze/src/Components/DataTable.js
+++ b/atd-vze/src/Components/DataTable.js
@@ -166,7 +166,7 @@ const DataTable = ({
                                   <Input
                                     name={field}
                                     id={field}
-                                    onClick={e => handleInputChange(e)}
+                                    onChange={e => handleInputChange(e)}
                                     defaultValue={fieldValue}
                                     type="select"
                                   >
--- a/atd-vze/src/Components/DataTable.js
+++ b/atd-vze/src/Components/DataTable.js
@@ -29,6 +29,7 @@ const DataTable = ({
   handleFieldUpdate,
   handleButtonClick,
   downloadGlobal,
+  isTempRecord,
 }) => {
   // Disable edit features if only role is "readonly"
   const { getRoles } = useAuth0();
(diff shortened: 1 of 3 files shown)


mddilley (approved):
Tested this on the deploy preview and 🚢 🚢 🚢

patrickm02L (approved):
## Testing
Functionality works as expected. 🥏 

Screen Shot 2023-06-12 at 3 37 19 PM

## Note
Before we ship, we're checking with the VZ team on whether or not this field should be protected.

patrickm02L:
We talked with Xavier, and he was adamant that CRIS is the source of truth for the Onsys flag, and it is "NOT ok to edit the on-sys_fl id."

He will however edit the flag if it's a temporary records, then manually removes the temporary records when it becomes a duplicate. 

With this clarity we have rescoped the solution to ONLY make the flag available for temporary records. Notes and estimate have been adjusted in the original Issue.

atd-vze/src/views/Crashes/crashDataMap.js
```diff
@@ -1,3 +1,7 @@
+const isTempCrash = isTempRecord => {
```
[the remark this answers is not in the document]
roseeichelmann: thanks mike for the suggestion, i just pushed up changes going that route instead ✨

atd-vze/src/Components/DataTable.js
```diff
@@ -166,7 +166,7 @@ const DataTable = ({
                                   <Input
                                     name={field}
                                     id={field}
-                                    onChange={e => handleInputChange(e)}
+                                    onClick={e => handleInputChange(e)}
```
roseeichelmann, line 169: I noticed that when you go to a temp record without an onsys value yet and click edit and select "Yes" and save, it wasnt saving. You had to first click No and then Yes for it to log the value and save. Changing this to `onClick` allows your first click to be logged as the new value. Its still somewhat of a confusing behavior that you have to click at all when Yes already looks to be selected in the UI so im trying to think through other solutions..

atd-vze/src/views/Crashes/crashDataMap.js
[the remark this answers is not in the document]
mddilley: Awesome! Getting into nit-picky territory here, but I think the function should have a verb in it. I like `make` or `create` in cases like this. Tried to find an article real quick and [this is the first useful thing that came up](https://dmitripavlutin.com/coding-like-shakespeare-practical-function-naming-conventions/#2-prefer-explanatory-names) - lots of opinions out there! All part of finding your code style. 🍄

atd-vze/src/Components/DataTable.js
mddilley: Good catch! I took a small dive into this, and I think you have uncovered another example where users cannot clear out a value. They can only select `Y` or `N` in this case and are unable to clear that out once chosen. It seems like the input is taking the `null` value and and just going with `Y` which is why the `onChange` handler isn't firing (since it appears as no change has happened when the value goes `Y` => `Y`). 

Could be a spin-off issue, but I think that adding another option to the select like `<option value={null}></option>` or something would patch this up.

atd-vze/src/views/Crashes/crashDataMap.js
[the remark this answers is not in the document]
roseeichelmann: yeah that totally makes sense! i usually put action verbs in functions, it just didnt cross my mind to rename `crashDataMap` when converting it into a function, but it should have. thanks for keeping me thinking about good code practices :-)

atd-vze/src/Components/DataTable.js
roseeichelmann: yeahh i agree that it would be good to add a null/clear option. @patrickm02L another issue or address that here?

atd-vze/src/views/Crashes/crashDataMap.js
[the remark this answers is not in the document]
mddilley: awesome, i'll keep thinking about them too! 🙏

patrickm02L (approved):
Everything looks and works perf! 🥏

atd-vze/src/Components/DataTable.js
mddilley: @roseeichelmann Could be a good issue to write up and bring to backlog refinement tomorrow!

mddilley (approved):
Tested this out locally and with the deploy preview, and it works great! Thanks for hanging in there with this one while it evolved. 🙏 🚀 🎉

atd-vze/src/views/Crashes/crashDataMap.js
```diff
@@ -1,460 +1,463 @@
...
-        editable: true,
-        uiType: "select",
-        lookupOptions: "atd_txdot__road_type_lkp",
-      },
-      traffic_cntl_id: {
-        label: "Traffic Control",
-        editable: true,
-        uiType: "select",
-        lookupOptions: "atd_txdot__traffic_cntl_lkp",
+export const createCrashDataMap = isTempRecord => {
```
frankhereford, line 1: This is a lot of diff for a pretty small change! 

I like how you handled this - very elegant.

chiaberry (approved):
🚢

frankhereford (approved):

PR #1239, opened 2023-06-09
 Add Engineering Areas to the FTT Index

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/12125

## Testing
**URL to test:** 
[Netlify](https://deploy-preview-1239--atd-vze-staging.netlify.app/#/fatalities)

**Steps to test:**
1. Go to the fatalities page, see the engineering areas in the last column now.
2. Note the engineering area of one of the fatalities. Click on the crash id for that fatality, edit the coordinates, move the coordinate to south/north/central austin. Go back to the fatalities page, search by that crash id, see how the engineering area automatically updates after you edit the coordinates. (This is kinda guess work unless you know exactly where the engineering areas are, but i just moved coordinates to super south or super north to test. You can refer to [the arcgis layer](https://austin.maps.arcgis.com/apps/mapviewer/index.html?layers=28c369f7ab534c64a26a646fca3fd6ec) to see where the engineering areas are).
3. Click the save button on the fatalities page to export as csv. The engineering area has been added as a column to be exported.

On merge to prod:

- [ ] Update DB with new `view_fatalities`. Manually add back in the object relationships it has with other tables.

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vzd/views/view_fatalities.sql
+++ b/atd-vzd/views/view_fatalities.sql
@@ -37,11 +37,13 @@ CREATE OR REPLACE VIEW view_fatalities AS (
             AS ytd_fatal_crash,
         crashes.case_id,
         crashes.law_enforcement_num
+        engineering_areas.label as engineering_area,
     FROM
         fatalities f
     INNER JOIN atd_txdot_crashes crashes ON f.crash_id = crashes.crash_id
     LEFT JOIN atd_txdot_primaryperson primaryperson ON f.primaryperson_id = primaryperson.primaryperson_id
     LEFT JOIN atd_txdot_person person ON f.person_id = person.person_id
+    LEFT JOIN engineering_areas ON ST_CONTAINS(engineering_areas.geometry, crashes.position)
     WHERE
         crashes.austin_full_purpose = 'Y'
     AND 
--- a/atd-vzd/views/view_fatalities.sql
+++ b/atd-vzd/views/view_fatalities.sql
@@ -43,7 +43,7 @@ CREATE OR REPLACE VIEW view_fatalities AS (
     INNER JOIN atd_txdot_crashes crashes ON f.crash_id = crashes.crash_id
     LEFT JOIN atd_txdot_primaryperson primaryperson ON f.primaryperson_id = primaryperson.primaryperson_id
     LEFT JOIN atd_txdot_person person ON f.person_id = person.person_id
-    LEFT JOIN engineering_areas ON ST_CONTAINS(engineering_areas.geometry, crashes.position)
+    LEFT JOIN engineering_areas ON engineering_areas.geometry && crashes.position AND ST_CONTAINS(engineering_areas.geometry, crashes.position)
     WHERE
         crashes.austin_full_purpose = 'Y'
     AND 
--- a/atd-vzd/views/view_fatalities.sql
+++ b/atd-vzd/views/view_fatalities.sql
@@ -36,8 +36,8 @@ CREATE OR REPLACE VIEW view_fatalities AS (
             ORDER BY crashes.crash_date ASC, crashes.crash_time ASC, crashes.crash_id) 
             AS ytd_fatal_crash,
         crashes.case_id,
-        crashes.law_enforcement_num
-        engineering_areas.label as engineering_area,
+        crashes.law_enforcement_num,
+        engineering_areas.label as engineering_area
     FROM
         fatalities f
     INNER JOIN atd_txdot_crashes crashes ON f.crash_id = crashes.crash_id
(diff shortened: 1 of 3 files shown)


atd-vze/.env
```diff
@@ -1,5 +1,5 @@
 REACT_APP_HASURA_ENDPOINT="https://vzd-staging.austinmobility.io/v1/graphql"
-#REACT_APP_HASURA_ENDPOINT="https://localhost:8080/v1/graphql"
+# REACT_APP_HASURA_ENDPOINT="http://localhost:8080/v1/graphql"
```
roseeichelmann, line 2: i get a network error when i run locally and uncomment out this line and sometimes it takes me awhile to remember why, but its because this `https` needs to be changed to `http`. does this happen to anyone else? if not, i can change this back!

atd-vzd/views/view_fatalities.sql
```diff
@@ -36,12 +36,14 @@ CREATE OR REPLACE VIEW view_fatalities AS (
...
         crashes.case_id,
-        crashes.law_enforcement_num
+        crashes.law_enforcement_num,
+        engineering_areas.label as engineering_area
     FROM
         fatalities f
     INNER JOIN atd_txdot_crashes crashes ON f.crash_id = crashes.crash_id
     LEFT JOIN atd_txdot_primaryperson primaryperson ON f.primaryperson_id = primaryperson.primaryperson_id
     LEFT JOIN atd_txdot_person person ON f.person_id = person.person_id
+    LEFT JOIN engineering_areas ON ST_CONTAINS(engineering_areas.geometry, crashes.position)
```
roseeichelmann, line 46: Because this is a view, the `engineering_area` will automatically update on edits to the crash position through the edit coordinates functionality in the VZE. I toyed with some other solutions but I personally like this one. Another way would have been to have this be a static column in the `atd_txdot_crashes`  table or `fatalities` table (which includes fatalities outside of AFP that are not included in the FTT Index) that is updated by database triggers when the crash coordinates change. But since engineering areas are only relevant to AFP fatalities and the whole point of having them in the DB is bc mary uses them when sending out messages for the Fatality Review Board, i thought it would be most clean and simple to have it be calculated on the fly when the view for the Fatalities page is created. If people have other opinions i am open to hearing them!

roseeichelmann:
Also -
Im pretty certain this is working as expected but this is a really important feature/something that would be bad to get wrong, so im nervous about that and wondering if there is a fool proof way to make sure the `ST_CONTAINS` is for sure correctly identifying the correct engineering area based on the crash position? Maybe im being paranoid but wondering if anyone has ideas on how to do a thorough test that this is working correctly?

atd-vzd/views/view_fatalities.sql
johnclary: I definitely like calculating on-the-fly so long as you don't run into performance issues. because the table has relatively view rows this seems like a great solution.

roseeichelmann: Yeah, ive tested loading all of the fatalities dating back to 2013 in the VZE fatalities page and it doesnt seem to be loading any slower than before (at least from a human perspective).

roseeichelmann: mhmm yeah i know, which frank has mentioned that `st_contains` can be an expensive calculation, so thats definitely something i was considering performance-wise and thats why i want to open up discussion incase anyone thinks doing it this way is too performance-heavy

johnclary: all good! i'd say leave it as-is until it becomes a problem 🚢

atd-vze/.env
chiaberry: it happens to me too, im good with this change

frankhereford (approved):
This looks awesome Rose. 🚢🚢🚢

I found a bug unrelated to this work while testing, which I've put in the "needs scoping" queue, https://github.com/cityofaustin/atd-data-tech/issues/12541. This shouldn't hold up merging this at all though.

Thanks again! Code reviewed and tested on netlify.

atd-vze/.env
mddilley: All good with this change too! 🙏 I could even see us updating this so the staging endpoint is commented out so we default to local.

mddilley (approved):
Tested this on the preview, and I was able to see the updated engineering area after updating crash locations. I like adding this to the existing DB view and seeing if we need to optimize down the road. 🚢 🚢 🚢

atd-vzd/views/view_fatalities.sql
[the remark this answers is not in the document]
roseeichelmann: Ooh thank you frank, thats super cool! i just pushed up changes implementing that :))

PR #1241, opened 2023-06-15
Dockerized EMS & AFD import scripts for Airflow 2

frankhereford:
## Testing
**URL to test:** 

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed


atd-etl/afd_ems_import/afd_incident_upload.py
```diff
@@ -0,0 +1,384 @@
...
+)  # our current environment from ['production', 'development']
+ONEPASSWORD_CONNECT_TOKEN = os.getenv("OP_API_TOKEN")  # our secret to get secrets 🤐
+ONEPASSWORD_CONNECT_HOST = os.getenv("OP_CONNECT")  # where we get our secrets
+VAULT_ID = os.getenv("OP_VAULT_ID")
+
+
+def main():
+    secrets = get_secrets()
+
+    global DB_USERNAME
```
mddilley, line 41: I know that I [left you hanging](https://github.com/cityofaustin/atd-vz-data/pull/1235/files#r1245923996) that last time I brought this up, but I'm starting to think that this ties into the 1PW in the DAG vs 1PW in the container discussion. I'm starting to feel like the Docker containers should not be concerned with how the secrets are obtained but only with receiving them. It would help avoid a refactor if we ever shifted from the 1PW server (which is unlikely and I surely hope we never have to!).

Like you mentioned today, it is a lot to think about, and I'll keep thinking about it over the weekend.

mddilley, line 41: I know that I [left you hanging](https://github.com/cityofaustin/atd-vz-data/pull/1235/files#r1245923996) that last time I brought this up, but I'm starting to think that these global variables tie into the 1PW in the DAG vs 1PW in the container discussion. I'm starting to feel like the Docker containers should not be concerned with how the secrets are obtained but only with receiving them. It would help avoid a refactor if we ever shifted away from the 1PW server (which is unlikely and I surely hope we never have to!).

Like you mentioned today, it is a lot to think about, and I'll keep thinking about it over the weekend.

mddilley (commented):
I've read through most of this, and it makes sense. 💯  I'll pick back up when I get the error that I was running into locally sorted out. I'm sure it is my local setup or something!

chiaberry (approved):
did not run, but reviewed code. Frank and I talked about secrets handling here IRL and I like the idea of being able to use 1password but also having the option that a local var could override it. (The Charlie way)

atd-etl/afd_ems_import/requirements.txt
```diff
@@ -0,0 +1,11 @@
...
+psycopg2-binary==2.*
+boto3==1.*
+sshtunnel==0.*
+openpyxl==3.*
+
+# The following two specific versions are needed. 
+# They are borrowed from the dockerized agent vz setup.
+# numpy broke its API in 1.24 wrt to type definitions.
+numpy==1.23.5
+pandas==1.4.4
```
johnclary, line 11: eek—can you link to an an issue related to this and/or where this impacts the script? how will we know if/when we can upgrade these packages?

atd-etl/afd_ems_import/afd_incident_upload.py
johnclary: in either case, i think you could slash all of these `global` declarations, `get_secrets()` outside of `main`, and declare these variables once in the global scope.

atd-etl/afd_ems_import/ems_incident_upload.py
```diff
@@ -0,0 +1,531 @@
+#!/usr/bin/env python
+
+"""
+Name: EMS Incident Uploads
+Description: This flow uploads EMS Incident Response CSVs (EMS Contact: Lynn C).
```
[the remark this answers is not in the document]
frankhereford: Great idea, 100% will do.

atd-etl/afd_ems_import/afd_incident_upload.py
```diff
@@ -0,0 +1,384 @@
+#!/usr/bin/env python
+
+"""
+Name: AFD Incident Uploads
+Description: This flow uploads AFD Incident Response CSVs
```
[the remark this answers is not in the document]
frankhereford: 100%, yes. This is completed.

atd-etl/afd_ems_import/README.md
```diff
@@ -0,0 +1 @@
+## Import scripts for AFD and EMS Daily Data Exports
```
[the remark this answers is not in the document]
johnclary: @frankhereford  in the corresponding [Airflow DAG](https://github.com/cityofaustin/atd-airflow/pull/112/files) you reference a pullable docker image:

>  `docker_image = "atddocker/vz-afd-ems-import:latest"`

do we need to setup CI for this?

johnclary (approved):
I asked a few questions. I did not test the code. Thank you for getting this ported @frankhereford!!!!!

atd-etl/afd_ems_import/Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM ubuntu:22.04
```
johnclary, line 1: this is another one of those cases where it would be nice to start from a python-based image, but i understand that might be a non-trivial change to make at this stage and i know we need to ship 🚢 🚢 🚢 🚢

atd-etl/afd_ems_import/README.md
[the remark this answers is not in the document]
frankhereford: Chia - thank you for this suggestion. I've added some basic invocation information to the readme and into the scripts themselves for local dev using docker compose.

John - I am going to pull your comment out into a top-level comment so I can resolve this issue from Chia. I think the CI would be helpful for sure.

atd-etl/afd_ems_import/README.md
frankhereford, line 1: @johnclary wrote in a previous thread:

> @frankhereford in the corresponding [Airflow DAG](https://github.com/cityofaustin/atd-airflow/pull/112/files) you reference a pullable docker image:
> 
> `docker_image = "atddocker/vz-afd-ems-import:latest"`
> 
> do we need to setup CI for this?

atd-etl/afd_ems_import/Dockerfile
frankhereford: I think it can go either way on this one. If the python image is based on debian, then I'm 100% on board because it'd get you the fixed version of python, like we've talked about in the past and I'd be able to reuse the code below that installs the SSH daemon. 

Quick look at docker hub, and this is 100% the case: [3.9.17-bullseye](https://hub.docker.com/layers/library/python/3.9.17-bullseye/images/sha256-9c50b04e7ebc4c5a1afbf73786863d9e317414d5f2850d37cbb78fdc9742239f?context=explore). ✅

I'll work on porting this over this afternoon to a python image.

atd-etl/afd_ems_import/README.md
frankhereford: @johnclary Totally 100% yes. I [wrote a bit more](https://github.com/cityofaustin/atd-airflow/pull/112#discussion_r1258613010) in the sibling PR, but short story is that I totally agree and am working on it this afternoon.

atd-etl/afd_ems_import/requirements.txt
frankhereford: @johnclary I'm baffled on this one. I remember working on why these scripts would not run for a non-trivial amount of time, and only after pinning them back to the versions that had been in the docker image they previous ran on fixing it. 

I've unpinned them.

The best I can figure, one of the following is true:

* I'm totally out to lunch and was confused.
* There was something anomalous in the data for EMS or AFD during the time I developed this PR.
* There has been a more recent version of the libraries which addressed the problem that I was encountering.

But it's working now, and I'm utterly unable to recreate this problem. I wish that I had written up an issue at the time so that I knew what had happened.

Thanks for asking for more details!

atd-etl/afd_ems_import/afd_incident_upload.py
frankhereford: John, thank you for your suggestion!

I had done this ugly thing with the `global` declarations inside `main()` because I was telling myself that it's good to get just as much out of the global execution as possible because that prevents it from being executed if someone were to `import` the script as a library. 

As I write this, I also think that the way it looks so 🤮 that it can't be right. I am going to ask about it today in dev sync, but I am increasingly feeling that I'm doing a lot of jumping-through-hoops for an esoteric reason that isn't really actually possible.

mddilley: In response to your question in dev sync, the thing that stood out to me was that it looks like variables are being declared and defined in one function and then being pushed up into global scope so that another function can read the values. I could be misunderstanding what is going on, but I would expect globally scoped variables to be declared in the global scope if multiple functions in that scope are reading them.

johnclary: i think that's a great point about not polluting the global scope for code that gets imported elsewhere. for scripts like this, i kind of prefer to just cram everything global up front at the top level so it's clear what all the dependencies are. 

either way (declaring globals in `main()` or in the global scope), maybe the real anti-pattern is not using `os.getenv()` and `os.setenv()` to pass around environment settings? 

i do not feel strongly about this at all, and think what you currently have is free of any red flags that would make maintaining the code difficult!

atd-etl/afd_ems_import/requirements.txt
johnclary: That's great news—thanks for retesting that! ✨ ⬆️ 🍄

.github/workflows/build_docker_images.yml
frankhereford, line 1: Here's some CI based on @johnclary's feedback. Thank you for that! This should conditionally build either the CRIS import image or the AFD import image with both ARM64 and AMD64 support. I'm pretty sure I have the `on:` stuff right, but it's not impossible that will need some tweaks once this is in `master` and `production`.

atd-etl/afd_ems_import/afd_incident_upload.py
frankhereford: Thank you both a ton for these thoughts! 🙏🏻 I really like how Mike put it about expecting global variables to be declared in the global scope -- he's identifying why just seeing `global` in the code looks strange. 

I cleaned up both scripts to make this look more normal. Because the `get_secets()` is still in a function, that needs to get declared first, but that's totally fine.

atd-etl/afd_ems_import/Dockerfile
frankhereford: This worked out super well - thanks for this awesome suggestion. I'm adopting this method going forward. If there's a hint of python on the image's task, start with a python image. You can still get 100% of debian based OS stuff right in there.

johnclary: niiice! thanks for getting this change in. really good stuff!

mddilley (approved):
This is awesome, Frank! This was a fun one that opened up some really good discussion. Thanks for bearing with the time that it took to collect my thoughts on some of this. 🚀🚀🚀

.github/workflows/build_docker_images.yml
```diff
@@ -0,0 +1,67 @@
...
+        if: steps.changes.outputs.afd_ems == 'true'
+        uses: docker/build-push-action@v2
+        with:
+          platforms: linux/amd64,linux/arm64
+          context: atd-etl/afd_ems_import
+          push: true
+          tags: atddocker/vz-afd-ems-import:${{ github.ref == 'refs/heads/production' && 'production' || 'latest' }}
+
+      - name: Build and push CRIS import image
+        if: steps.changes.outputs.cris == 'true'
```
johnclary, line 61: whoa—awesome 🤯

johnclary:
i wish i could bookmark this new github action: it's next-level stuff. thank you @frankhereford!! 🚢 🚢 🚢

PR #1247, opened 2023-06-27
Replace lambda function to make sure austin_full_purpose (AFP) is flagged correctly

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/10126

First of all, I apologize for the length of this!

This PR addresses pulling the functionality for updating the `austin_full_purpose` flag out of AWS lambda which looks like it has not been servicing its purpose lately (I tested updating crash coordinates to be outside of Austin/vice versa through the VZE UI and the flag wasn't updating). We instead want the flag to be handled directly in the DB which will be easier to maintain. The route I ended up taking for this was creating a stored generated column called `in_austin_full_purpose` that calls a function `is_in_austin_full_purpose` whenever a record is added to the `atd_txdot_crashes` table or when a record is updated. The function returns true/false for whether the crash is in AFP and then that value is stored in the column.

_Cool thing about stored generated columns from the postgres docs:
"A stored generated column is computed when it is written (inserted or updated) and occupies storage as if it were a normal column"_

The determination for whether a crash is in AFP is based off this message that Xavier sent us in the past:
["The austin_full_purpose flag should only be "Y" when it is in jurisdiction = 5 OR when the crash does not have a latitude/longitude and the city_id = 22"](https://austininnovation.slack.com/archives/CM9MK950S/p1678398208097029)

The AFP flag being correct will ensure that crashes are showing up in the Crashes page correctly and the same goes for the Fatalities page since the view that powers that page uses the AFP flag to determine what fatalities make it into the view. This will also help with making sure fatality records have an engineering area, because previously we had some fatalities that were showing up in the Fatalities page that didn't have engineering areas because they werent actually in Austin. With that being said, there are still about 7 fatalities where the engineering area is NULL because it seems like the engineering areas dont perfectlyyy line up with the AFP jurisdiction, and I'm talking like meters off, so its definitely an edge case thing and I'm not sure if thats a problem or how to handle it.

Heres an example of that:

The crash id 13511720 has a fatality and is in the AFP jurisdiction but it just barely falls outside of what would be its fatality engineering area, so the record is showing up in the Fatality view as it should, but it has a null engineering area.

AFP Jurisdiction layer (includes the right lane of Dessau rd):
Screenshot from 2023-06-27 16-07-21

Engineering area layer (does not include the right lane of Dessau rd):
Screenshot from 2023-06-27 16-07-38

Where the crash is located (on the right lane of Dessau rd):
Screenshot from 2023-06-28 17-44-55

## Testing
**URL to test:** 
https://deploy-preview-1247--atd-vze-staging.netlify.app/#/fatalities

**Steps to test:**
1. Go to the Fatalities page, observe that its rendering without errors and the engineering areas are there for mostly all the fatalities.
2. Go to the Crashes page, toggle on and off the AFP toggle.
3. Mostly though look at the code and let me know what you think of this solution!

Steps moving forward:

- [ ] Retire the `austin_full_purpose` column?
- [x] Need to address the other part of what the lambda function did, which was updating the `city_id`. That should be its own issue. And this probably shouldn't be merged until that is done and we can shut down the lambda function all at once?
- [ ] Run all these commands in prod before merge.
- [ ] Make sure all views and the socrata etl is using the new `in_austin_full_purpose` column
- [ ] Delete the `notify_hasura_jurisdiction_from_latlon_INSERT` - AFTER INSERT
`notify_hasura_jurisdiction_from_latlon_UPDATE` - AFTER UPDATE triggers
- [ ] Retire the `jurisdiction_from_latlon` hasura event
- [ ] Shut down the lambda functions

---
#### Ship list
- [ ] Code reviewed


atd-vzd/triggers/atd_txdot_crashes_updates_audit_log.sql
```diff
@@ -152,15 +152,6 @@ BEGIN
...
 
-    ------------------------------------------------------------------------------------------
-    -- AUSTIN FULL PURPOSE
-    ------------------------------------------------------------------------------------------
-    -- Set Austin Full Purpose to Y (TRUE) when it has Austin City ID and no coordinates.
-    IF (NEW.position IS NULL and NEW.city_id = 22) THEN
-        NEW.austin_full_purpose = 'Y';
-    END IF;    
-    --- END OF AUSTIN FULL PURPOSE ---
-
```
roseeichelmann, line 163: dont need this anymore, will be handled in new `in_austin_full_purpose` generated column

atd-vzd/triggers/is_crash_in_austin_full_purpose.sql
```diff
@@ -0,0 +1,13 @@
+-- This function determines whether crash coordinates fall within the Austin Full Purpose jurisdiction polygon.
+-- If the crash has no coordinates, it returns whether the crash is in Austin (city_id = 5)
+CREATE OR REPLACE FUNCTION is_crash_in_austin_full_purpose(location geometry(Geometry,4326), city_id int) RETURNS boolean AS $$
+        BEGIN
+            -- If we dont have coordinates for the crash and the city_id is Austin (22), return true
+            IF (location IS NULL) THEN --TODO: Do we need to handle if city_id is null? If so should that make the afp flag default to false?
```
roseeichelmann, line 6: do crashes ever have `city_id = null`? if so, how should we handle that here? no position and no city id means not in austin full purpose? i can see in the DB `city_id` is nullable

atd-vzd/views/view_crashes_inconsistent_numbers.sql
```diff
@@ -37,7 +37,7 @@ FROM atd_txdot_crashes atc
          LEFT JOIN primarypersons atpp ON atc.crash_id = atpp.crash_id
          LEFT JOIN units atu ON atc.crash_id = atu.crash_id
 WHERE 1 = 1
-  AND (austin_full_purpose = 'Y' OR (atc.city_id = 22 and atc.position IS NULL))
+  AND (atc.in_austin_full_purpose = true)
```
roseeichelmann, line 40: i basically searched for everywhere we use the old austin_full_purpose flag in the DB and changed all those instances to using the new column. all these views will need to be recreated in staging and prod.

atd-vze/src/views/Fatalities/fatalityGridTableParameters.js
```diff
@@ -73,9 +73,9 @@ export const fatalityGridTableColumns = {
   engineering_area: {
     searchable: true,
     sortable: true,
-    label_table: "Engineering Area",
+    label_table: "Current Engineering Area",
```
roseeichelmann, line 76: based on Xaviers comments in release party today

atd-vzd/migrations/migration_atd_txdot_crashes_2023_06_26--1622.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE atd_txdot_crashes
+ADD COLUMN in_austin_full_purpose boolean GENERATED ALWAYS AS (is_crash_in_austin_full_purpose(position, city_id)) STORED NOT NULL;
```
roseeichelmann, line 2: when we close out this PR ill add this column and its meaning/how its calculated to our gitbook data dictionary

atd-etl/app/process/socrata_queries.py
roseeichelmann, line 1: I'm not sure where this etl actually lives or anything about our socrata process, so i may need some help/direction in how i can get these changes with the new `in_austin_full_purpose` reflected in the actual flow that is run. is this an airflow thing?

roseeichelmann:
Also it looks like these aren't source controlled on github but i can see in dbeaver that we have these materialized views that all currently use the `austin_full_purpose` flag and i guess i would need to update those as well?

materialized views that use austin_full_purpose:
all_atd_txdot_crashes
all_cr3_crashes_off_mainlane
five_year_all_crashes_off_mainlane
five_year_all_crashes_off_mainlane_outside_surface_polygons
five_year_all_crashes_outside_any_polygons
five_year_all_crashes_outside_surface_polygons
five_year_atd_apd_blueform
five_year_atd_txdot_crashes
five_year_cr3_crashes_off_mainlane
five_year_cr3_crashes_outside_surface_polygons

atd-etl/app/process/socrata_queries.py
```diff
@@ -26,7 +26,7 @@
                       { crash_time: { _is_null: false }}
                     ]
                     _or: [
-                        {austin_full_purpose: {_eq: "Y"}},
+                        {in_austin_full_purpose: {_eq: true}},
                         {
                             _and: [
                                 {city_id: {_eq: 22}},
```
mddilley, line 32: I really like the new column name! Looking at the lines below this update, we probably need to update this query since `in_austin_full_purpose` now includes `city_id = 22` and lat/lon is not null.

There may be other places where this logic is present in queries and could be a separate issue to sweep through the code for other places where there could be some overlap before this and the other PR for the `city_id` Lambda are merged. _Edit: Just saw the rest of the PR so it looks like you covered most of these already 🙏_

atd-vzd/triggers/is_crash_in_austin_full_purpose.sql
mddilley: I don't see any nulls in the prod DB - just some marked as `9999` which doesn't correspond to a value in the lookup table. Maybe the PR for the other part of this Lambda will uncover some more info?

mddilley (commented):
Test steps check out, and I read through all of the code. Thanks for finding a new and better way to handle this value, and this diff shows that we can now encapsulate some of the AFP logic that was added over time in this generated column. That will make maintaining this much easier. 🙌 🚀 I added one comment about the ETL queries that may need to be addressed.

atd-etl/app/process/socrata_queries.py
roseeichelmann: ooh thank you for catching that! ill defs update this query and remove the part about `city_id = 22` since that logic is already captured in the `in_austin_full_purpose` column

atd-etl/app/process/socrata_queries.py
```diff
@@ -94,15 +86,7 @@
...
-                    _or: [
-                        {crash: {austin_full_purpose: {_eq: "Y"}}},
-                        {
-                            _and: [
-                                {crash: {city_id: {_eq: 22}}},
-                                {crash: {position: {_is_null: true}}} 
-                            ]
-                        }
-                    ]
+                    crash: {in_austin_full_purpose: { _eq: true}}
```
roseeichelmann, line 89: these queries look rly confusing to me, i cant tell if this is right

mddilley: These trip me up too! I'm reading this update as crash date between the two date vars AND crash in AFP. Previously, it was crash date between the two date vars AND [in AFP (jurisdiction = 5) OR (city_id = 22 AND position not null)]. 

The new definition of `in_austin_full_purpose` covers everything in the brackets so that looks good to me. 👍

mddilley (approved):
Thanks for updating the queries to reflect the updated AFP boolean. Looks great to me! 🚀

atd-vzd/views/view_fatalities.sql
```diff
@@ -45,7 +45,7 @@ CREATE OR REPLACE VIEW view_fatalities AS (
     LEFT JOIN atd_txdot_person person ON f.person_id = person.person_id
     LEFT JOIN engineering_areas ON engineering_areas.geometry && crashes.position AND ST_CONTAINS(engineering_areas.geometry, crashes.position)
     WHERE
-        crashes.austin_full_purpose = 'Y'
+        crashes.in_austin_full_purpose = true
```
frankhereford, line 48: Thank you for this! A boolean as a boolean. 💯💯💯

atd-vze/src/views/Fatalities/fatalityGridTableParameters.js
frankhereford: I wish i recalled exactly what the context was around this, because adding the "current" makes it feel like sometimes we operate on "previous" or "future" Engineering Areas, which doesn't make a ton of sense to me. But it's totally all good - not worth digging into.

atd-etl/app/process/socrata_queries.py
frankhereford: Yea, this is a really good question. We'll need to make sure that the docker image ([Dockerfile](https://github.com/cityofaustin/atd-vz-data/blob/master/atd-etl/Dockerfile) & [dockerhub](https://hub.docker.com/repository/docker/atddocker/atd-vz-etl/general)) that is built from `atd-etl/app` is updated when this is merged into `master`, and then when that is done, we'll want to get on `atd-data02`'s airflow and make sure to pull down the image if it's not already done so by a cron job.

Please feel free to assign working this stuff out to me during the launch if you want, as you already have a lot of moving parts, and I'd be super happy to help.

frankhereford (approved):
This looks great Rose, thank you for working this out, and extra thank you for starting down the path of simplifying the AFD tracking mechanism by bringing it down into the database and out of a lambda function.

🚀🚀🚀

atd-etl/app/process/socrata_queries.py
roseeichelmann: Thanks for the context frank and that would be amazing if you could help out with that part when it comes to it!

atd-vze/src/views/Fatalities/fatalityGridTableParameters.js
roseeichelmann: I thinkk Xavier wanted to capture the fact that the engineering areas could change at some point and that the VZE is operating off of the "current" areas rather than past ones?

patrickm02L (approved):
## Review

Checked out the UI parts of the testing, everything worked as expected. 🥏 

1. Go to the Fatalities page, observe that its rendering without errors and the engineering areas are there for mostly all the fatalities. ✅ 
2. Go to the Crashes page, toggle on and off the AFP toggle.✅ 
3. Mostly though look at the code and let me know what you think of this solution!

atd-vzd/triggers/is_crash_in_austin_full_purpose.sql
```diff
@@ -0,0 +1,13 @@
+-- This function determines whether crash coordinates fall within the Austin Full Purpose jurisdiction polygon.
+-- If the crash has no coordinates, it returns whether the crash is in Austin (city_id = 5)
```
johnclary, line 2: it might be good to add a note here that this is invoked by a generated column function, in case that migration kind of gets buried over time.

johnclary:
This looks very great! A few questions for y'all:

- what happens to the old `austin_full_purpose` column now? should it be dropped since it will no longer reflect reality? 
- should we give Xavier a heads up that we're deprecating this column? i assume he has lots of queries that use it.
- do we render this column in the Editor UI anywhere? the only field i see on the crash detals page is **Outside City Limit Flag**
- can y'all clarify how **Outside City Limit Flag** (`rpt_outside_city_limit_fl`) relates to this column? I see that it is editable.

roseeichelmann:
@johnclary Personally i think we should drop the old `austin_full_purpose` column since it will be deprecated, and yeah i also think we should give Xavier a heads up about the new column. As for the old column being rendered anywhere, i searched for everywhere we use the old `austin_full_purpose` column in the codebase and i think i covered everything in this PR :crossed_fingers: 

I don't think the Outside City Limit Flag (`rpt_outside_city_limit_fl`) column actually relates to this column. I think thats a value we get from CRIS that the vz team can change if they decide its wrong, and `in_austin_full_purpose` is never something we get from CRIS but a value we calculate based on the `position` of the crash and whether that point falls within the AFP jurisdiction polygon we have in the DB. I'm not sure much more about the context of `rpt_outside_city_limit_fl`

atd-vzd/views/view_crashes_inconsistent_numbers.sql
roseeichelmann, line 40: just released that in the staging/prod DBs, this view doesnt actually even use `austin_full_purpose` in the where clause, its only here in the code that it does :sweat_smile: it instead just filteres by city_id = 22

image

these are the problems that come along with not having a great way to source control our db changes. does anyone have any context on this view/if its important, and should i just leave the view as is then?

chiaberry: I have zero context on this view, but this is really frustrating -- i wonder how many other places the db doesnt match what we have in github

roseeichelmann: The other views that i updated to use `in_austin_full_purpose` actually do use the austin full purpose flag, its just this one that doesnt match. but yeah curious where else this happens in github :face_with_spiral_eyes:

mddilley: this is frustrating! I think it is a combination of the criteria of AFP changing over time and being set in so many individual places that it probably wasn't consistently updated. This really highlights the upside of centralizing that logic in the new column that you've added in this PR! 🚀

PR #1248, opened 2023-07-06
Escape double quotes and ensure properly formatted CSV

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/12541

The `rec_text` and `rec_update` columns were formatting weird when opening the exported csv from the Crashes or Fatalities pages in a spreadsheet app. This is because VZE users often use double quotes within their inputs for these columns. We weren't properly escaping these double quotes so that apps like Excel can consume the csv without formatting issues.

https://datatracker.ietf.org/doc/html/rfc4180 Lays out the format for CSVs and how to properly escape double quotes

## Testing
**URL to test:** 
https://deploy-preview-1248--atd-vze-staging.netlify.app/#/fatalities

**Steps to test:**
1. Go to the Fatalities tab, click on the save/export csv option.
2. Open the csv in your spreadsheet app of choice (Excel, I use LibreOffice Calc, etc)
3. Are the columns formatted correctly? Is any weird behavior happening? Specifically `rec_text` and `rec_update` were the ones causing issues before, so check that those values look right.

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vze/src/Components/GridExportData.js
+++ b/atd-vze/src/Components/GridExportData.js
@@ -112,36 +112,12 @@ const GridExportData = ({
     });
 
     const cleanedAndFlattenedData = flattenedData.map(item => {
-      // Some fields are presenting issues when they are exported as .csv and
-      // imported into MSFT Excel, Acces & Power BI. For now, we're just going to
-      // modify the data for export to solve issues related to returns and double-quotes
-      // In the future, we may choose to clean these values at the database level.
-
-      const columnsToClean = [
-        "rpt_street_name",
-        "rpt_street_desc",
-        "rpt_sec_street_desc",
-        "rpt_sec_street_name",
-      ];
-
-      // These columns contain user input which may contain double
-      // quotes that need to be escaped for proper csv formatting
-      const columnsToEscapeDoubleQuotes = ["rec_text", "rec_update"];
-
-      columnsToClean.forEach(col => {
-        if (item[col]) {
-          // remove return carriage and replace with space
-          item[col] = item[col].replace(/(\r\n|\n|\r|")/gm, " ");
-          // remove double-quotes
-          item[col] = item[col].replace(/"/g, "");
-        }
-      });
-
-      columnsToEscapeDoubleQuotes.forEach(col => {
-        if (item[col]) {
-          // Replace all double quotes with two double quotes
-          // This escapes double quotes in a csv file
-          item[col] = item[col].replace(/"/g, '""');
+      // We want to escape any instances of double quotes by preceding
+      // them with another double quote. This will put us in compliance
+      // with CSV formatting guidelines https://www.rfc-editor.org/rfc/rfc4180
+      Object.keys(item).forEach(key => {
+        if (typeof item[key] === "string" && item[key].includes('"')) {
+          item[key] = item[key].replace(/"/g, '""');
         }
       });
 
--- a/atd-vze/src/Components/GridExportData.js
+++ b/atd-vze/src/Components/GridExportData.js
@@ -111,19 +111,6 @@ const GridExportData = ({
       return flattenedRow;
     });
 
-    const cleanedAndFlattenedData = flattenedData.map(item => {
-      // We want to escape any instances of double quotes by preceding
-      // them with another double quote. This will put us in compliance
-      // with CSV formatting guidelines https://www.rfc-editor.org/rfc/rfc4180
-      Object.keys(item).forEach(key => {
-        if (typeof item[key] === "string" && item[key].includes('"')) {
-          item[key] = item[key].replace(/"/g, '""');
-        }
-      });
-
-      return item;
-    });
-
     // Create array of columns that should be displayed in the table
     let columnsToExportParsed = [];
 
@@ -142,7 +129,7 @@ const GridExportData = ({
       }
     });
 
-    const cleanedFlattenedAndParsedData = cleanedAndFlattenedData.map(item => {
+    const flattenedAndParsedData = flattenedData.map(item => {
       // Parse out unnecessary columns before exporting table, rename one column
       Object.keys(item).forEach(col => {
         if (!columnsToExportParsed.includes(col)) {
@@ -154,6 +141,19 @@ const GridExportData = ({
           delete item["death_cnt"];
         }
       });
+      return item;
+    });
+
+    const cleanedFlattenedAndParsedData = flattenedAndParsedData.map(item => {
+      // We want to escape any instances of double quotes by preceding
+      // them with another double quote. This will put us in compliance
+      // with CSV formatting guidelines https://www.rfc-editor.org/rfc/rfc4180
+      Object.keys(item).forEach(key => {
+        if (typeof item[key] === "string" && item[key].includes('"')) {
+          item[key] = item[key].replace(/"/g, '""');
+        }
+      });
+
       return item;
     });
 
--- a/atd-vze/src/Components/GridExportData.js
+++ b/atd-vze/src/Components/GridExportData.js
@@ -124,6 +124,8 @@ const GridExportData = ({
         "rpt_sec_street_name",
       ];
 
+      // These columns contain user input which may contain double
+      // quotes that need to be escaped for proper csv formatting
       const columnsToEscapeDoubleQuotes = ["rec_text", "rec_update"];
 
       columnsToClean.forEach(col => {
@@ -137,6 +139,8 @@ const GridExportData = ({
 
       columnsToEscapeDoubleQuotes.forEach(col => {
         if (item[col]) {
+          // Replace all double quotes with two double quotes
+          // This escapes double quotes in a csv file
           item[col] = item[col].replace(/"/g, '""');
         }
       });
@@ -176,8 +180,6 @@ const GridExportData = ({
       });
       return item;
     });
-
-    console.log("data", cleanedFlattenedAndParsedData);
 
     return cleanedFlattenedAndParsedData;
   };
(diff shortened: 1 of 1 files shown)


atd-vze/src/Components/GridExportData.js
```diff
@@ -133,6 +137,14 @@ const GridExportData = ({
         }
       });
 
+      columnsToEscapeDoubleQuotes.forEach(col => {
```
johnclary, line 140: this approach looks sound to me, but isn't it possible we end up with quotes in other fields? might be better to just escape any field that has a string value? it will be a performance hit for sure, although you could also test if a value contains a quote before escaping it (iirc regex is quite expensive but a string-contains check not so much).

in any case, i was surprised to see we were already using a CSV library and that it was not handling quote escaping. fwiw there are [open PRs](https://github.com/react-csv/react-csv/issues/199) 🤷 

i was also wondering about the need to escape commas as well, but apparently the default behavior is to wrap all values [in double quotes](https://github.com/react-csv/react-csv#--enclosingcharacter-props), which takes care of commas.

@roseeichelmann is there

frankhereford (approved):
Thank you Rose! This works great, and it totally solves the issue at hand.


### suggestion
There is a loop just above where you added your work that looks like this:

```
      columnsToClean.forEach(col => {
        if (item[col]) {
          // remove return carriage and replace with space
          item[col] = item[col].replace(/(\r\n|\n|\r|")/gm, " ");
          // remove double-quotes
          item[col] = item[col].replace(/"/g, "");
        }
      });
```
What would you think about combining the new loop you've added and this one together and make one that does 2 things:
* Replace any combination of `\r`, `\n` and `\r\n` with ` ` as written
* Adopt your new regex to properly escape double quotes instead of blast them away?

You could get this down to one list of fields and that list of fields would have some columns found in one export and some columns found in another, which appears to be fine in this context. 


### outside of scope
FWIW, I wanted to see if there were any other rules we might not be following, and I think everything else is ✅, except we are only using new-lines instead of carriage-return-new-lines at the end of every record.

Here's the line of `od` output where the header row ends and the first record begins:

`$ od -c view_fatalities2023-07-05T16_50_41-05_00.csv`

```
0000440    i   n   e   e   r   i   n   g   _   a   r   e   a   "  \n   "
```

 [RFC 4180](https://datatracker.ietf.org/doc/html/rfc4180) defines the EOL format. This is squarely outside of the scope of this issue, and everything we care about being able to parse this is already doing just fine with just new lines.

atd-vze/src/Components/GridExportData.js
johnclary: one more idea on the topic of just quote-escaping every field... you could generate the list of string-type columns dynamically up front by check the `typeof` for the first not-null value of every column in the dataset 🤔

roseeichelmann: yeah i was wondering about escaping all string columns too, i was looking into the other columns we have and came to the conclusion that the only other columns that might have double quotes are in the `columnsToClean`list and those cases were already handled by removing the double quotes in the loop frank commented. im not sure whats the best approach but it does make sense to me to quote escape every string field to cover all our bases and make sure no fields w/ double quotes are flying under our radar and throwing off the formatting. i was thinking of the performance hit but maybe its negligible!

roseeichelmann:
@frankhereford i really like your idea to combine the two loops! i was thinking of doing that at one point but wasnt sure if the decision to just completely remove double quotes from the `columnsToClean` list was purposeful/should stay that way.

Im curious what you think about @johnclary thoughts on checking every string field for double quotes and escaping them instead of just doing that for the columns we expect to have double quotes. If we go that route, should we also do the `/r` `/n` replacement check on every string field?

frankhereford:
@roseeichelmann wrote:

> If we go that route, should we also do the `/r` `/n` replacement check on every string field?

This regex that you're talking about looks at field values and removes any `\r` and `\n` characters it finds, which are the carriage return character and the line feed character respectively. 

It's an interesting question, because it's really asking: "Do we have linefeeds in any of our data that we are exporting that is meaningful and should be included?" I think that in the current set of data that we're exporting, the only fields that have potential for including meaningful linefeeds are actually the ones that you've recently added as part of the fatality tool. Rose, do you think it's important to preserve them? My gut says yes, and really, this whole regex bothers me a little because it's throwing away data that made it in there for a good reason. I wouldn't be sad if you tried to remove that line.

I think a subtle thing to observe is that in a previous comment, I was talking about how the CSV as a whole is breaking the CSV format rules laid out in RFC 4180 by having linefeeds that are wrong, but that's talking about the linefeed at the end of the CSV record, not about linefeeds that are contained in values, which is what Rose is asking about here.

roseeichelmann:
@frankhereford yeah thats a great point and it makes a lot of sense to me that we should be preserving the linefeeds in the csv export, especially for the fatality columns such as `rec_update`, bc the vz team added linebreaks to their inputs for a reason. i wonder why the decision was made to remove them for those few columns in `columnsToClean` in the first place b/c its not like leaving them throws off the csv formatting. I am comfy with throwing out that regex and not removing any linefeeds or carriage returns at all anymore, or just leaving that line as is to just filter them out for the columns in columnsToClean since that was a decision that was made at some point? ill see if i can find the issue for doing that in the first place which may help our decision.

atd-vze/src/Components/GridExportData.js
```diff
@@ -112,24 +112,12 @@ const GridExportData = ({
...
-        "rpt_street_name",
-        "rpt_street_desc",
-        "rpt_sec_street_desc",
-        "rpt_sec_street_name",
-      ];
-
-      columnsToClean.forEach(col => {
-        if (item[col]) {
-          // remove return carriage and replace with space
-          item[col] = item[col].replace(/(\r\n|\n|\r|")/gm, " ");
```
roseeichelmann, line 130: frank and i discussed that removing all newlines and carriage returns for these columns "rpt_street_name", "rpt_street_desc", "rpt_sec_street_desc", and "rpt_sec_street_name" is altering the data somewhat and maybe it would be best to not do that. newlines are allowed in CSV format guidelines and they dont cause apps like excel to freak out like the double quotes do (the rec_updates column often has newlines and they format fine in excel). this can be opened up for discussion though if anybody has other thoughts!

mddilley: @roseeichelmann I thought that I dropped the link to the original PR https://github.com/cityofaustin/atd-vz-data/pull/649 for these changes in response to [this comment](https://github.com/cityofaustin/atd-vz-data/pull/1248#issuecomment-1631474955) the other day, but I guess that I forgot!  The only thing that concerns me about taking this out is that it seems like this solved an issue that Xavier was facing. 

There is no GH issue attached to that PR so not super helpful, but maybe there is risk of regression for the VZ team? There are a couple example crash record IDs listed in that PR.

roseeichelmann: ooh yeah thats a good point! i tested exporting a csv containing that crash record (17272765) with the code changes i made in this pr and none of the rpt fields for it seem to have newlines? i wonder if maybe the data has changed.

atd-vze/src/Components/GridExportData.js
```diff
@@ -169,6 +144,19 @@ const GridExportData = ({
       return item;
     });
 
+    const cleanedFlattenedAndParsedData = flattenedAndParsedData.map(item => {
```
roseeichelmann, line 147: moved this to take place after we parse out the columns we dont need to export so we arent cleaning columns that arent even going into the csv

atd-vze/src/Components/GridExportData.js
frankhereford: If we have new lines in data that should not have new lines in it, like for example, a street name, then I think we should address that data in the DB and get it corrected. LDM will make this super easy too, in the future.

There are fields where newlines are valid and useful data, like the notes fields that are now being generated by the VZ team. I am going to make a note to look at this first thing in the morning and see if i can learn more about what was causing the VZ team headaches on it, and I will also possible reach out to Xavier for more background.

I guess put another way, I am keen to not drop new lines, but, of course, I obviously don't want regressions either.

It's a little late in the day to jump into an investigation, but I would be very eager to explore some solutions that let us not strip out newlines, but also ensure that there are none where they do not belong. Don't let me hold this up, of course, and I will see if i can offer something more helpful than this opinion tomorrow.

Thank you Rose & Mike!!

mddilley: I was curious about this code block so I looked into these columns.

- `rpt_street_name`
    - newline
        - none
    - double quotes
        - none
    - carriage return
        - none

- `rpt_street_desc`
    - newline
        - none
    - double quotes
        - 11243056, 11415653, 11459597, 11895097, 11988003, 14648733, 16565337, 16795451, 17069323, 17569955, 18385294, 18582777, 18627255, 19263271
    - carriage return
         - none

- `rpt_sec_street_desc`
    - newline
        - none
    - double quotes
        - 13308293, 15437978, 16202128, 16388510, 16452903, 16504241, 16529694, 16795451, 17471777, 18385294
    - carriage return
        - none

- `rpt_sec_street_name`
    - newline
        - none
    - double quotes
        - none
    - carriage return
        - none

- `All crash table columns`
    - newline
        - 18004825 has `...lso \northbound on manor...` in `investigator_narrative_ocr` column (probably a typo)
    - double quotes
        - couldn’t narrow down this search enough but we have examples above
    - carriage return
        - none

atd-vze/src/Components/GridExportData.js
mddilley: Great call!

mddilley (approved):
I did a dive into the crash columns in the legacy code `columnsToClean` and added that to the discussion. It does seem like an issue that was more prevalent in the past, and I think the risk of regression is a lot lower than I was thinking. This was a tough one! Ship it! 🚀

patrickm02L (approved):
Good to go! 💎

PR #1253, opened 2023-07-14
Replace lambda function to update city_id

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/12744

This PR addresses the other part of what the lambda function does (first part was addressed here https://github.com/cityofaustin/atd-vz-data/pull/1247) which is updating the `city_id` to be in Austin when a crashes position is updated to be inside of Austin. I tested the current set up and the lambda function is not doing this as it should be.

To determine whether city_id = 22 (Austin) im using the same jurisdictions that the OG lambda functions use, which is (5, 3, 7, 8, 10). 

image

## Testing
**URL to test:** 
Test [staging](https://visionzero-staging.austinmobility.io/editor/#/crashes)

**Steps to test:**
1. Go to a crash, change the city to be outside of Austin on this Details card.
image
2. Now edit the crash coordinates to be somewhere inside of Austin.
4. Refresh the page, the City should now say Austin in the place you originally edited (aka city_id was updated to 22).

Steps for merge:

- [ ] Recreate these triggers/functions in prod
- [ ] Merge at the same time as https://github.com/cityofaustin/atd-vz-data/pull/1247
- [ ] Delete the `notify_hasura_jurisdiction_from_latlon_INSERT` - AFTER INSERT
`notify_hasura_jurisdiction_from_latlon_UPDATE` - AFTER UPDATE triggers
- [ ] Retire the `jurisdiction_from_latlon` hasura event
- [ ] Shut down the lambda functions

---
#### Ship list
- [ ] Code reviewed


frankhereford (approved):
This is powerful work Rose! You're bringing VZ lightyears ahead step by step with these optimizations and restructuring how the data is managed.

Interestingly, following your directions, I don't even have to refresh the page to get the reassigned city name after the trigger runs. It's just 💯 🎯.

patrickm02L (approved):
## Testing

- [X] 1. Go to a crash, change the city to be outside of Austin on this Details card.
image
- [X] 2. Now edit the crash coordinates to be somewhere inside of Austin.
- [X] 3. Refresh the page, the City should now say Austin in the place you originally edited (aka city_id was updated to 22).

## Notes
This is really cool! 

Same observation as Frank, I didn't need to refresh it happened automatically.

No sound on vid
https://github.com/cityofaustin/atd-vz-data/assets/5566727/2c53b554-4f8b-4ff7-bbe3-fa99074671d2

atd-vzd/triggers/update_crash_city_id.sql
```diff
@@ -0,0 +1,29 @@
...
+        -- loop through jurisdictions that contain the new position
+        -- if jurisdiction is in the list of valid jurisdictions then
+        -- update city id to be in Austin (22)
+        FOR temprow IN 
+        (
+            SELECT jurisdictions.id
+        FROM atd_txdot_crashes AS crashes
+            INNER JOIN atd_jurisdictions jurisdictions
+            ON (jurisdictions.geometry && NEW.position) AND ST_Contains(jurisdictions.geometry, NEW.position)
+        WHERE crashes.crash_id = NEW.crash_id
```
johnclary, line 18: 👋 just eyeballing this sql, @roseeichelmann can you help me understand why it is necessary to join to `atd_txdot_crashes`? it looks to me like you only neext to select from `atd_jurisdictions` to get the info you need.

johnclary:
Do we need to tear down the lambda function in concert with this change? Or is the plan to just disable the hasura event trigger?

atd-vzd/triggers/update_crash_city_id.sql
roseeichelmann: no you are totally right, its not necessary! brain fart :p 

just pushed up changes

roseeichelmann:
@johnclary i was thinking we would disable the hasura event trigger and then tear down the lambda as well, but curious what others think

mddilley:
@roseeichelmann @johnclary I think that tearing down the related lambdas and sqs AWS resources is a great idea for a spin-off issue unless someone really thinks that it should happen now.

I quickly looked at the code/CI/AWS resources, and I'm reading that [the CI will continue to deploy lambdas and SQS](https://github.com/cityofaustin/atd-vz-data/blob/c5000cbc2086adf717fbcbd126b74e8175a3a4a9/.github/workflows/aws-lambda-helper.sh#L166-L184) for each folder in the `atd-vz-data/atd-events`. So, they will keep coming back until that code is removed too.

roseeichelmann:
@mddilley okay that sounds like a good plan to me, ill bring this up in planning today as a potential spin-off issue to create

PR #1254, opened 2023-07-18
Update CR3 download script

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/12849

Most of this is formatting with a small change to how the cookie is obtained in the script. No more prompt, and this script is now split out from the other CR3 tasks bundled in `atd-etl/app`. The script is now built and pushed to its own Docker image that is used by Airflow v2. The companion Airflow PR is https://github.com/cityofaustin/atd-airflow/pull/133.

## Testing
**URL to test:** 
None

**Steps to test:**
Trigger the DAG in the other Airflow PR or run locally by following the steps in the readme added in this diff. If you decide to run locally, the secrets for `.env` come out of the 1PW entries called **CRIS CR3 Download** and ~**Vision Zero CRIS Import**~ **Vision Zero graphql-engine Endpoints**. The cookie will also need to be added to the env file from CRIS.

You should see the script find no new pdfs to download and store in S3 unless you run it in the morning before the designated CRIS downloader runs the current process. 

If you want to go the extra mile, you can: 
- find an older crash record without serious injuries or fatalities (we don't want to disrupt VZ team efforts)
- carefully, [SSH tunnel and connect to the production database](https://app.gitbook.com/o/-LzDQOVGhTudbKRDGpUA/s/-MIQvl_rKnZ_-wHRdp4J/dev-guides/how-tos/how-to-connect-to-the-rds-instance#2.-open-the-tunnel) and update the one crash record's `cr3_stored_flag` column from `Y` to `N` (i tested this with crash id `16302089`)
- run the script and it will log one pdf download from CRIS and upload to S3

---
#### Ship list
- [x] Code reviewed


atd-etl/app/process_cris_cr3.py
```diff
@@ -55,21 +48,28 @@ def wait(int):
...
 except Exception as e:
     crashes_list_without_skips = []
     print("Error, could not run CR3 processing: " + str(e))
 
 
 for crash_record in crashes_list_without_skips:
-    process_crash_cr3(crash_record, CRIS_BROWSER_COOKIES, skipped_uploads_and_updates)
+    process_crash_cr3(
+        crash_record,
+        ATD_ETL_CONFIG["CRIS_CR3_DOWNLOAD_COOKIE"],
```
mddilley, line 70: The only change in this file. The cookie now comes from a environment variable.

atd-etl/cr3_download/process/helpers_cr3.py
```diff
@@ -56,26 +46,29 @@ def download_cr3(crash_id, cookies):
     for key, morsel in cookie.items():
         baked_cookies[key] = morsel.value
 
-    crash_id_encoded = base64.b64encode(str("CrashId=" + crash_id).encode("utf-8")).decode("utf-8")
+    crash_id_encoded = base64.b64encode(
+        str("CrashId=" + crash_id).encode("utf-8")
+    ).decode("utf-8")
     url = ATD_ETL_CONFIG["ATD_CRIS_CR3_URL"] + crash_id_encoded
-    download_path = ATD_ETL_CONFIG["AWS_CRIS_CR3_DOWNLOAD_PATH"] + "%s.pdf" % crash_id
```
mddilley, line 61: This path was coming from the env sometimes and hardcoded in other places. I decided to hardcode it to `/tmp/` since it is unlikely to need change from the outside of the script.

atd-etl/cr3_download/cr3_download.py
```diff
@@ -15,11 +15,6 @@
 from process.config import ATD_ETL_CONFIG
 from process.helpers_cr3 import *
 
-def wait(int):
```
mddilley, line 18: This method was not used in this script. It is removed in `helpers_cr3.py` for the same reason.

atd-etl/cr3_download/cr3_download.py
```diff
@@ -29,8 +24,6 @@ def wait(int):
 # the CR3 pdf, upload to S3
 #
 
-# ask user for a set of valid cookies for requests to the CRIS website
-CRIS_BROWSER_COOKIES = input('Please login to CRIS and extract the contents of the Cookie: header and please paste it here:')
```
mddilley, line 33: We no longer need to prompt for the cookie. 🤖

atd-etl/cr3_download/process/config.py
mddilley, line 1: We don't have to keep this pattern around, but I left it since it gives some fallback values for the Hasura requests made in `request.py`. Maybe better to just hardcode those? 🤷

atd-etl/cr3_download/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# This image was used over python:3.10-slim to support python-magic library
```
mddilley, line 1: I am inexperienced on choosing base images so I went with the one that Frank used for the afd/ems one. Open to other thoughts here. I kept getting and error with `python:3.10-slim` that `libmagic` wasn't found even though it logged the pip install.

.github/workflows/build_docker_images.yml
```diff
@@ -79,3 +83,12 @@ jobs:
...
           tags: atddocker/atd-vz-cr3-extract:${{ github.ref == 'refs/heads/master' && 'production' || 'latest' }}
+
+      - name: Build and push CR3 download image
+        if: steps.changes.outputs.cr3_download == 'true'
+        uses: docker/build-push-action@v4
+        with:
+          platforms: linux/amd64,linux/arm64
+          context: atd-etl/cr3_download
+          push: true
+          tags: atddocker/vz-cr3-download:${{ github.ref == 'refs/heads/master' && 'production' || 'latest' }}
```
johnclary, line 94: i don't think this is the place to change this pattern, but my two cents is that it would be more clear/explicit to default to a tag name of `staging` instead of `latest` for these images.

atd-etl/cr3_download/Dockerfile
johnclary: it would be nice to use the same base image to save a little on image storage, but i think this is totally fine and definitely not worth debugging any further!

atd-etl/cr3_download/cr3_download.py
```diff
@@ -55,21 +48,28 @@ def wait(int):
...
     crashes_list_without_skips = [
         x for x in crashes_list if x["crash_id"] not in known_skips
     ]
-    print("\nList of crashes needing CR3 download: %s" % json.dumps(crashes_list_without_skips))
- 
+    print(
+        "\nList of crashes needing CR3 download: %s"
+        % json.dumps(crashes_list_without_skips)
+    )
+
```
johnclary, line 60: might just be my personal preference, but i think it'd be nice if this instead (or additionally) logs the **number** of crashes needing download.

atd-etl/requirements.txt
```diff
@@ -7,4 +7,4 @@ web-pdb==1.*
 sodapy==2.*
 boto3==1.*
 mail-parser==3.*
-python-magic==0.*
\ No newline at end of file
+python-magic==0.*
```
johnclary, line 10: to be sure, the cr3 download script no longer uses this requirements file, right? do other scripts that use this requirements file need python-magic?

johnclary (approved):
i have not tested this, but it looks very great. thank you @mddilley! i think DAG-ifying this is a major improvement 🚢🚢🚢🚢

atd-etl/requirements.txt
mddilley: The cr3 download script does still use it to validate that files are pdfs before storing them in S3. But, you are right! This requirement should be removed here since it is covered in requirements for the script that is split out here. The only other place it is used in this repo is a toolbox script that has its own requirements.txt. 🙏 I'll remove it here.

.github/workflows/build_docker_images.yml
chiaberry: If someone opens a PR with changes in these folders, it will build the image tagged with the other tag, so latest/staging. I was thinking we shouldn't overwrite images tagged `staging` with code that hasn't been reviewed or approved, but I don't think we have a staging data ETL setup, so it wouldn't affect anything.

atd-etl/cr3_download/Dockerfile
chiaberry: I used 3.10 because one of the packages the cr3 extract script uses didn't list python 3.11 as compatible.

frankhereford (approved):
Thanks for packaging this up and making it available for use in the [airflow stack](https://github.com/cityofaustin/atd-airflow/pull/133#pullrequestreview-1543528987). This is so creative and well executed. Awesome work!

patrickm02L (approved):
Adding my concurrence based on Dev reviews 💎

atd-etl/cr3_download/cr3_download.py
mddilley: I agree that the count is useful. I'm adding that 🙏

chiaberry (approved):
👍🏼

PR #1260, opened 2023-07-24
Update local Hasura port from 8080 to 8084

mddilley:
## Associated issues
**_This is not urgent_**

This is an update to make it easier to run the VZ local stack and the local Airflow stack at the same time. See https://austininnovation.slack.com/archives/CHZE6BC6L/p1690229808701409. Hasura [defaults to serving up the GraphQL endpoint at 8080](https://hasura.io/docs/latest/deployment/graphql-engine-flags/reference/#server-port) which conflicts with the local Airflow dashboard webserver on port 8080 so this update to the Docker compose file maps 8080 in the container to 8084 locally.

## Testing
**URL to test:** 

**Steps to test:**
1. Update `.env` by commenting out line 1 and uncommenting line 2
2. Start the local VZ stack and make sure that the app is sending requests to `http://localhost:8084/v1/graphql`
3. You agree with `8084` and see that [this 1Password entry](https://start.1password.com/open/i?a=RPNXYTT5QNABRL3IE5XBV3362I&v=dgumte7pqvqkpsm7ihg4ni63ke&i=tzlu67ntmpb4gdnpkfdpnmrmrm&h=austintransportation.1password.com) shows `8084` in the **DEVELOPMENT** section endpoints.

---
#### Ship list
- [x] Make sure that 1Password entries that contain this local endpoint all have their ports updated
- [x] Code reviewed

### DIFF
--- a/atd-vze/.env
+++ b/atd-vze/.env
@@ -1,5 +1,5 @@
 REACT_APP_HASURA_ENDPOINT="https://vzd-staging.austinmobility.io/v1/graphql"
-# REACT_APP_HASURA_ENDPOINT="http://localhost:8080/v1/graphql"
+# REACT_APP_HASURA_ENDPOINT="http://localhost:8084/v1/graphql"
 REACT_APP_AUTH0_DOMAIN=atd-datatech.auth0.com
 REACT_APP_AUTH0_CLIENT_ID=2qbqz2sf8L9idBOwn0d5YA9efNgQbL7c
 REACT_APP_MAPBOX_TOKEN="pk.eyJ1Ijoiam9obmNsYXJ5IiwiYSI6ImNqemU4OGQzaTBlZjczYm1rZWg1c2lidTcifQ.2eooq4Q7dv_1Le5la9D96Q"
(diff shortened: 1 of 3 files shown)


atd-vze/.env
```diff
@@ -1,5 +1,5 @@
 REACT_APP_HASURA_ENDPOINT="https://vzd-staging.austinmobility.io/v1/graphql"
-# REACT_APP_HASURA_ENDPOINT="http://localhost:8080/v1/graphql"
+# REACT_APP_HASURA_ENDPOINT="http://localhost:8084/v1/graphql"
```
mddilley, line 2: Should switch this to default to the local endpoint? I've been using local more and more by default but not sure about everyone else.

johnclary (approved):
👍 i did not test but I support this change 🚢

atd-vze/.env
chiaberry: but in the meantime, sure, change it to default to local endpoint!

frankhereford (approved):
This looks great - thanks for bringing this order to the port-dance we had going on.

atd-vze/.env
roseeichelmann: defaulting to local sounds good, i have found myself using it more often as well

roseeichelmann (approved):
thanks for doing this mike! ill update my council districts PR to default to 8084 as well

atd-vze/.env
mddilley: thanks y'all! I'll go ahead and switch the commented lines here.

PR #1263, opened 2023-07-27
Update Socrata ETL GitHub Action

mddilley:
## Associated issues
Since this ETL queries the production VZ database, we need the production Airflow DAG to run production code only. Otherwise, a schema change in staging (like adding a new column, etc) could break the ETL. This happened when the GraphQL query to get crash data had `in_austin_full_purpose` in the filters in staging while the production database still uses `austin_full_purpose`.

See https://austininnovation.slack.com/archives/C013G4RNJ01/p1690432638099059?thread_ts=1690431847.716579&cid=C013G4RNJ01

## Testing
**URL to test:** 
n/a

**Steps to test:**
1. Give me a sanity check on the logic in this change please 🙏

I read the action as:
- Runs when code is pushed to `master` or `production` branch
- if there is a code change within `atd-etl/socrata_export/**`:
    - if the push was from `production` then update Docker image tagged with `production`
    - otherwise, update Docker image tagged with `development` (for Airflow local testing - I realize that I am breaking the convention we started and can change this back to `latest` if anyone feels strongly)

---
#### Ship list
- [x] Code reviewed

### DIFF
--- a/.github/workflows/build_docker_images.yml
+++ b/.github/workflows/build_docker_images.yml
@@ -91,4 +91,4 @@ jobs:
           platforms: linux/amd64,linux/arm64
           context: atd-etl/socrata_export
           push: true
-          tags: atddocker/vz-socrata-export:${{ github.ref == 'refs/heads/master' && 'production' || 'latest' }}
+          tags: atddocker/vz-socrata-export:${{ github.ref == 'refs/heads/production' && 'production' || 'development' }}


johnclary (approved):
Looks good to me. I agree that using `latest` isn't ideal. Fwiw `staging` might be a more explicit tag name? In any case take my approval with a grain of salt since i don't do development here :)

frankhereford (approved):
Thanks Mike!

mddilley:
@johnclary yep, i also am not sure about the tag. I picked `development` only because this DAG uses the environment variable to decided which image tag to run (`development` locally or `production`). Not sure how useful having the `development` tag with code from `master` is but we shall see.

roseeichelmann (approved):
yay thanks for doing this mike, looks good to me!

johnclary:
matching the env var seems like a good enough reason to me 🙏

PR #1267, opened 2023-07-28
Add council districts to crash & fatality export

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/12503

I decided to add another [stored generated column](https://www.postgresql.org/docs/current/ddl-generated-columns.html) called `council_district` that calls the function `find_council_district()` to calculate the council district for each crash. Lmk what yall think!

## Testing
**URL to test:** 
https://deploy-preview-1267--atd-vze-staging.netlify.app/#/crashes

**Steps to test:**
1. Export a csv from the crashes page.
2. Do you see the council_districts column?
3. Repeat for fatalities page.
4. You can view the new generated `council_district` column in the `atd_txdot_crashes` table as well

---
#### Ship list
- [ ] Code reviewed


roseeichelmann:
Wondering if it would also be helpful to add the new in_austin_full_purpose column to the crash export?

atd-vzd/migrations/migration_atd_txdot_crashes_2023_07_14--1325.sql
roseeichelmann, line 1: In the release yesterday Xavier told us it wasn't necessary to have this trigger and he actually prefers that the `city_id` always stays untouched from what we get from CRIS. I already dropped this trigger in staging/prod, now removing this code from Github so we aren't confused in the future. Would it be better tho if I leave the code and commit a "migration" to gh that drops the trigger? Idk, the manual migration-esque files here are hard to keep track of :shrug:

mddilley: I'm going to be asking myself the same question soon in a PR! I'm thinking that these migration-like files aren't all that useful besides showing what SQL needs to run at launch so removing this is ok. It seems like, when we get around to implementing migrations here in VZ, our first migration will include the current state of things [like in Moped](https://github.com/cityofaustin/atd-moped/tree/main/moped-database/migrations/1599856186244_init). I could be way off base, though! Eager to hear other opinions.

mddilley (approved):
Tested this with the deploy preview, and it works great! I see council districts in both the crash and fatality exports. 🙏

PR #1269, opened 2023-08-03
Replace Lambda function blueform_location_from_latlon with trigger

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/13057

This PR updates `location_id` as a stored generated column. I'm already thinking about deployment strategy because some of these DB operations take a really long time locally (probably around 1 hour total). We can brainstorm some ideas here in this PR. I also created https://github.com/cityofaustin/atd-data-tech/issues/13275 to handle the AWS and Github Actions cleanup after all these triggers are replaced.

Questions:
- ~There are a lot of materialized views affected by this change. Do we think it is worth ~inventorying~ _assessing_ them in case some of them are no longer used? Do we know if Xavier uses any of them? Could be a backlog issue but it could also save deployment time if `five_year_surface_polygons_with_crash_data` isn't used anymore.~ _Update: See below for discussion and decision to drop these materialized views since they are not used by the VZ team or us._
- Are we ok with SQS build failures until we clean everything up in the issue linked above?

## Testing
**URL to test:** 
Local

**Steps to test:**
To test this, start up your local VZDB instance by replicating the latest DB dump and then follow the steps below. ~Heads up that some of the steps will take a long time locally to test!~

### Test steps
1. On your local VZ instance, follow the deployment steps below except for the last step of turning off the ETLs.
2. Afterwards, you can see the insert trigger working by running the sql statements below on your local DB
```sql
-- insert a blueform crash with coordinates that will associate to location_id 8337406736
INSERT INTO atd_apd_blueform(latitude, longitude, date, case_id)
VALUES (30.16841600, -97.78449000, '2023-08-14', '00000001');

-- insert a blueform crash with coordinates that will not associate to a location
INSERT INTO atd_apd_blueform(latitude, longitude, date, case_id)
VALUES (30.386136, -97.67309, '2023-08-14', '00000002');

-- insert a blueform crash without coordinates that will not associate to a location
INSERT INTO atd_apd_blueform(latitude, longitude, date, case_id)
VALUES (NULL, NULL, '2023-08-14', '00000003');
```
3. Update an existing record with a `location_id` of `NULL` to have the latitude and longitude of a record with a location. After update, you should see the record associate with the same location as the record you took the lat/lon from.
4. Update an existing record with a `location_id` that is not null and update it with coordinates from an existing record that has latitude and longitude and no `location_id` associated.
5. Update an existing record with a `location_id` that is not null and update it with `NULL` for both latitude and longitude. You should see the record `location_id` update to `NULL`.

### Deployment steps
The DB changes were applied through the Hasura CLI with the migrations in the this diff.
~- [ ] Remove Hasura event trigger called `blueform_location_from_latlon`~
~- [ ] Run SQL in [update_noncr3_location.sql](https://github.com/cityofaustin/atd-vz-data/blob/md-13057-blueform-lat-lon-trigger/atd-vzd/triggers/update_noncr3_location.sql)~
~- [ ] Run SQL in [migration_noncr3_location_2023_08_02--1648.sql](https://github.com/cityofaustin/atd-vz-data/blob/md-13057-blueform-lat-lon-trigger/atd-vzd/migrations/migration_noncr3_location_2023_08_02--1648.sql)~
~- [ ] Switch off non-cr3 crash location housekeeping Airflow ETLs~

---
#### Ship list
- [x] Code reviewed


johnclary:
> I'm already thinking about deployment strategy because some of these DB operations take a really long time locally (probably around 1 hour total).

I don't want to re-hash old convo if this has already been discussed nauseam, but i do wonder if a static column + traditional DB trigger might be the safer way to go. The main limitation of the generated column that i'm worried about is that it cannot subscribe to updates based on which columns were edited, so the postGIS operations will execute whenever any column in the crash record is edited. I can see situations where we are really hampered when we need to batch update (or reimport?) many rows of crashes, because the update operation will be very slow. In contrast, the benefit of a trigger is that we can (1) listen to specific column changes and (2) switch it off if we need to. 

I might be mischaracterizing that limitation of the generated column, but I wanted to throw this out there before y'all go further down the road 🤔

mddilley:
@johnclary Totally open to more discussion on this and very valid concerns. Going the trigger direction would not be a big lift at all now that we have a better handle on these. We can also do some benchmarking. Going to think about this some more. Thanks for bringing that up!

It also makes me wonder if we have overlooked a role of the ETLs. Do we expect the shape of locations to change over time? If so, it may be necessary to periodically update associations (via recalc of the generated column or retrigger if we go that route).

johnclary:
>  if we expect the shape of locations to change over time

Yep I think 100% we're expecting regular updates to the location records so there will need to be some process to efficiently update crashes affected by those changes. It might be ok to leave that as a separate concern, because I think there are ways to do that for either generated columns or trigger-based cols?

roseeichelmann:
I think its definitely worth having a conversation about the views and seeing if any of them aren't used/necessary anymore esp if they do expensive calculations, I've also noticed we maintain lots of views that aren't used in our codebase. Maybe Xavier is using them somewhere, but if hes not then it would be good to cut down on them I think.

atd-vzd/triggers/update_noncr3_location.sql
```diff
@@ -0,0 +1,40 @@
...
+                AND ST_Contains(
+                        ST_Transform(
+                                ST_Buffer(
+                                    ST_Transform(ncr3m.geometry, 2277),
+                                    1,
+                                    'endcap=flat join=round'
+                                ),
+                                4326
+                        ), /* transform into 2277 to buffer by a foot, not a degree */
+                        atc.position
```
roseeichelmann, line 23: i think if you pass in the position of the crash as a parameter to this function, you could just make this a select without having to do a join. like
`IF EXISTS (SELECT * FROM non_cr3_mainlanes WHERE (position && non_cr3_mainlanes.geometry) AND ....`

atd-vzd/triggers/update_noncr3_location.sql
```diff
@@ -0,0 +1,40 @@
...
+    ELSE 
+        -- If it isn't main-lane and is of concern to Vision Zero, try to find a location_id for it
+        RETURN (SELECT aab.location_id FROM atd_apd_blueform AS aab
+                INNER JOIN atd_txdot_locations AS atl
+                ON ( atl.location_group = 1
+                    AND (atl.geometry && aab.position)
+                    AND ST_Contains(atl.geometry, aab.position)
+                    )
+                WHERE aab.case_id = blueform_case_id);
+    END IF;
```
roseeichelmann, line 38: same thing here, i think you could remove the join if you pass `position` as a parameter to `update_noncr3_location()`

atd-vzd/migrations/migration_noncr3_location_2023_08_02--1648.sql
```diff
@@ -0,0 +1,495 @@
...
+
+-- 5. Then, drop current atd_apd_blueform location_id column
+ALTER TABLE atd_apd_blueform DROP COLUMN location_id;
+
+-- 6. Rename generated_location_id to location_id to replace the previous column
+ALTER TABLE atd_apd_blueform 
+RENAME COLUMN generated_location_id TO location_id;
+
+-- 7. Add DB views that use location_id (need to be tracked in Hasura console)
+CREATE OR REPLACE VIEW locations_with_crash_injury_counts AS
```
roseeichelmann, line 39: just checkin - did you copy the code from these views from the read replica or from the codebase? just thinking about how sometimes the gh version controlled db stuff isnt accurate anymore :face_with_spiral_eyes:

atd-vzd/migrations/migration_noncr3_location_2023_08_02--1648.sql
roseeichelmann, line 1: for the views that also touch the main `atd_txdot_crashes` table `location_id`, we should probably coordinate this part so we only have to drop and recreate them once? i will compile the list of views affected by the main crashes `location_id` and we can go from there :thinking:

frankhereford:
Hi @mddilley  - you wrote:

> There are a lot of materialized views affected by this change. Do we think it is worth inventorying assessing them in case some of them are no longer used? Do we know if Xavier uses any of them? Could be a backlog issue but it could also save deployment time if five_year_surface_polygons_with_crash_data isn't used anymore.

These are materialized views that I proposed ages ago, and I don't think they are widely used, or perhaps even at all. I think it would be worthwhile to ask @xavierapostol if he uses any of the following, but my hunch is that they can be dropped without bothering anyone. Currently we have the following:

image

I suspect, at the very least, that the `five_year_...` ones can go. The others, I'd want to run through the codebase to double check on.

xavierapostol:
> Hi @mddilley - you wrote:
> 
> > There are a lot of materialized views affected by this change. Do we think it is worth inventorying assessing them in case some of them are no longer used? Do we know if Xavier uses any of them? Could be a backlog issue but it could also save deployment time if five_year_surface_polygons_with_crash_data isn't used anymore.
> 
> These are materialized views that I proposed ages ago, and I don't think they are widely used, or perhaps even at all. I think it would be worthwhile to ask @xavierapostol if he uses any of the following, but my hunch is that they can be dropped without bothering anyone. Currently we have the following:
> 
> image
> 
> I suspect, at the very least, that the `five_year_...` ones can go. The others, I'd want to run through the codebase to double check on.

I don't have any use for those Materialized Views.

mddilley:
@frankhereford thanks for raising this up to @xavierapostol! I'll poke around the codebase and update this PR to remove any that aren't seeing any use. Thank y'all!

atd-vzd/migrations/migration_noncr3_location_2023_08_02--1648.sql
mddilley: great thinking - these did come from the production replica. just double-checked to be sure 🙏

atd-vzd/triggers/update_noncr3_location.sql
mddilley: oh nice - i definitely just copied/pasted without looking at these queries too much 😝 I'll do a pass over these. i like the suggestions - thanks! 🙏

mddilley:
This PR is now updated to use a trigger to associate location IDs again. It is ready for review once more! 🙏

atd-vzd/triggers/update_noncr3_location.sql
```diff
@@ -0,0 +1,51 @@
...
+  RETURNS TRIGGER 
+  LANGUAGE PLPGSQL
+  AS
+$$
+DECLARE 
+    found_location_id varchar;
+
+BEGIN
+    -- Check if crash is on a major road and of concern to TxDOT.
+    -- NEW.position is recalculated in BEFORE trigger to ensure it is up to date
```
johnclary, line 12: i'm not familiar with the distinction between `position` and the lat/lon columns, but fwiw triggers on the same event fire in alphabetical order, in case you did want to chain them together ([docs](https://www.postgresql.org/docs/9.5/trigger-definition.html)):

> If more than one trigger is defined for the same event on the same relation, the triggers will be fired in alphabetical order by trigger name

johnclary (commented):
@mddilley i've offered some feedback. i hope it is helpful. it is so awesome to see all the red!

atd-vzd/triggers/update_noncr3_location.sql
```diff
@@ -0,0 +1,51 @@
...
+    -- when this runs AFTER.
+    IF EXISTS (SELECT ncr3m.*
+                FROM non_cr3_mainlanes AS ncr3m WHERE (
+                            (NEW.position && ncr3m.geometry)
```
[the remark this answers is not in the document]
mddilley: I looked at the DB just now, and these geometries are not indexed. I just added one and will push that up. Thanks for looking more closely at the logic in these functions. I was mostly focused on porting the existing functions, but we might as well try to think carefully about what is going on here while we have the chance. 🙏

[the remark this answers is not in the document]
johnclary: ah, right, this is existing code being ported. i get this may be best left out of scope then!

atd-vzd/triggers/update_noncr3_location.sql
```diff
@@ -0,0 +1,51 @@
...
+                                        4326
+                                ), /* transform into 2277 to buffer by a foot, not a degree */
+                                NEW.position
+                            )
+                    )) THEN
+        -- If it is, then set the location_id to None
+		UPDATE atd_apd_blueform SET location_id = NULL 
+        WHERE case_id = NEW.case_id;
+    ELSE 
+        -- If it isn't on a major road and is of concern to Vision Zero, try to find a location_id for it.
```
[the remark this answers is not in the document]
mddilley: Same on this one. I trusted the existing logic that used to be in a separate function, but I totally get what you are saying here. We briefly touched on how mainlane/non-mainlane (or TxDOT concern/VZ concern) are determined and agreed to keep this existing logic but re-address it. It has been really hard to wrangle the legacy terminology and logic to create a path forward! 😵‍💫

Since we decided to focus on improving how to differentiate TxDOT concern from VZ concern for CR3 crashes, this was punted on to keep things status quo and keep the method of calculating of the non-CR3 crash figures as they are now. 

Also worth noting that the improved method of associating CR3 crashes with major roads (or what we were calling level 5s) relies on columns that only exist for CR3 crashes and not for non-CR3 crashes. This is part of the challenge of figuring out how to improve this part of the code. Maybe we should figure that out now while we are making major changes, though? I need to think about that. I really like your breakdown! We need fresh takes because it has been tough.

atd-vzd/triggers/update_noncr3_location.sql
mddilley: TIL - I think that could be a better option here so there isn't the extra dependency of order of `AFTER` triggers. I'll make this change. 🙏

atd-vzd/triggers/update_noncr3_location.sql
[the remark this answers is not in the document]
mddilley: All good! The diff definitely suggests that this is new code. I created a spin-off issue that I can bring to the next refinement. Thanks, John! https://github.com/cityofaustin/atd-data-tech/issues/13425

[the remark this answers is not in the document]
mddilley: Forgot that I was going to tag @patrickm02L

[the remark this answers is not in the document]
patrickm02L: Thanks @mddilley I tagged it for next Sprint's refinement

atd-vzd/triggers/update_noncr3_location.sql
mddilley: Totally left this comment in the dust - sorry! You were spot on that this could be refactored to exclude the join. 🙏

atd-vzd/migrations/migration_noncr3_location_2023_08_02--1648.sql
```diff
@@ -0,0 +1,37 @@
+-- 1. Add triggers to update location_id on insert and update
+CREATE TRIGGER update_noncr3_location_on_insert
+AFTER INSERT ON atd_apd_blueform
+FOR EACH ROW
+WHEN (NEW.latitude IS NOT NULL AND NEW.longitude IS NOT NULL)
```
roseeichelmann, line 5: im wondering whether `position` is what should be checked here since thats whats used later in the trigger function?

mddilley: this is a good point. It also makes me wonder if using [ST_Equals](https://postgis.net/docs/en/ST_Equals.html) for the update trigger would be the right way to go about comparing `NEW` to `OLD`. I'm going to look into this. 🙏

mddilley: @johnclary @frankhereford this is what I was talking about in planning - any experience with this?

johnclary: i'm having trouble finding the docs that support this, but my understanding is that for point geometries you can test with an old fashion equality operator. e.g., `NEW.position != OLD.position`. 

for sure `ST_Equals` will get the job done for you though!

as an aside, i suspect that testing on the two numeric columns would be the most performant (by probably a very thin margin). e.g. `NEW.x != OLD.x AND NEW.y != OLD.y`

mddilley: finally coming back to this! I'm going to use a bit of both of y'all's advice and stick with using lat and lon with `IS DISTINCT FROM`. Thanks!

roseeichelmann (approved):
thanks mike, all the test steps worked! im wondering whats our plan for merge with these non LDM lambda function PRs now, should we be turning them into up/down migrations?

atd-vzd/migrations/migration_noncr3_location_2023_08_02--1648.sql
frankhereford: @mddilley wrote:

> @johnclary @frankhereford this is what I was talking about in planning - any experience with this?

Generally, I think sticking with the postGIS functions are the way to go when operating on GIS data. There are just so many little gotchas, none-the-least-of-which is that you're comparing equality of floating point numbers if you go at it via the latitude and longitude. `0.999… == 1`, `sqrt(2)**2 != 2`, etc. I always want to let the gis libraries deal with all that minutia if I can.

That said, postGIS is going to have overloaded operators based on the input type, so != will become the GIS functions that test for non-equality if the fields you're `!=`ing are GIS data. I think you're good with something like `NEW.position != OLD.position`.

frankhereford (approved):
I am approving based on my testing of https://github.com/cityofaustin/atd-vz-data/pull/1286. Thank you Mike!

patrickm02L (approved):
✅ based on Dev reviews!

mddilley:
Just merged `master` branch, ran through the steps in https://github.com/cityofaustin/atd-vz-data/blob/master/atd-vzd/README.md, and applied the migrations and metadata in this branch to the staging instance. Merging the code now. 🚀

PR #1272, opened 2023-08-16
VZE | Download CR3 "NoSuchKey The specified key does not exist" error

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/13021'

This is a script for our toolbox for fixing the above error in the VZE which is caused by crashes having the cr3_stored_flag set to Y when we dont actually have a CR3 for them in our AWS S3 bucket. This script updates that flag for those crashes to be N which allows the CR3 downloader to download the missing CR3s when it next runs. Follow up issue is to figure out why this desyncing happened in the first place.

I already ran the script for staging & prod which fixed the issue, this PR is just to review the code to put in our toolbox for potential future needs if this ever happens again.

## Testing
**URL to test:** 
You can test this script locally if youd like!

**Steps to test:**
1. Follow steps in the readme

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-toolbox/fix_cr3_stored_flag/fix_cr3_stored_flag.py
+++ b/atd-toolbox/fix_cr3_stored_flag/fix_cr3_stored_flag.py
@@ -31,105 +31,119 @@ CRASHES_UPDATE = """
   }
 """
 
+
 def get_s3_cr3_files():
-  """Connect to the aws s3 bucket and get all of the names of the cr3 pdfs we have
-  and pipe them into a new file in this directory.
-  """
-  # Creates a new txt file in the current directory to store output from the following commands
-  subprocess.run(["touch", "s3_cr3_file_names.txt"])
-  s3_cr3_file_names = open("s3_cr3_file_names.txt", "w")
-  print("Writing names of cr3 pdfs we have in s3 to a new file in this directory called 's3_cr3_file_names.txt', this can take a few minutes...")
-  aws = subprocess.Popen(["aws", "s3", "ls", "s3://atd-vision-zero-editor/production/cris-cr3-files/"], stdout=subprocess.PIPE)
-  output = subprocess.Popen(["awk", "{print $4}"], stdin=aws.stdout, stdout=s3_cr3_file_names)
-  output.wait()
+    """Connect to the aws s3 bucket and get all of the names of the cr3 pdfs we have
+    and pipe them into a new file in this directory.
+    """
+    # Creates a new txt file in the current directory to store output from the following commands
+    subprocess.run(["touch", "s3_cr3_file_names.txt"])
+    s3_cr3_file_names = open("s3_cr3_file_names.txt", "w")
+    print(
+        "Writing names of cr3 pdfs we have in s3 to a new file in this directory called 's3_cr3_file_names.txt', this can take a few minutes..."
+    )
+    aws = subprocess.Popen(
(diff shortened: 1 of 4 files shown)


atd-toolbox/fix_cr3_stored_flag/fix_cr3_stored_flag.py
```diff
@@ -0,0 +1,149 @@
...
+        crashes_with_cr3s = [line.rstrip().removesuffix(".pdf") for line in cr3s_file]
+    return crashes_with_cr3s
+
+
+def find_crashes_with_wrong_flag(crashes_with_cr3_flag, crashes_with_cr3s):
+    """Finds crash ids that we don't have cr3s for but have cr3_stored_flag = 'Y'."""
+    crashes_with_cr3_flag_list = []
+    for crash in crashes_with_cr3_flag:
+        crashes_with_cr3_flag_list.append(str(crash["crash_id"]))
+    crashes_to_update = list(set(crashes_with_cr3_flag_list) - set(crashes_with_cr3s))
```
mddilley, line 70: oooo nice that you used sets here 🚀

atd-toolbox/fix_cr3_stored_flag/fix_cr3_stored_flag.py
```diff
@@ -0,0 +1,149 @@
...
+    and pipe them into a new file in this directory.
+    """
+    # Creates a new txt file in the current directory to store output from the following commands
+    subprocess.run(["touch", "s3_cr3_file_names.txt"])
+    s3_cr3_file_names = open("s3_cr3_file_names.txt", "w")
+    print(
+        "Writing names of cr3 pdfs we have in s3 to a new file in this directory called 's3_cr3_file_names.txt', this can take a few minutes..."
+    )
+    aws = subprocess.Popen(
+        ["aws", "s3", "ls", "s3://atd-vision-zero-editor/production/cris-cr3-files/"],
```
mddilley, line 46: cool to see an alternative to using boto3. Using that library would probably have produced more code too.

mddilley (approved):
I tested this on production, and I saw 0 crashes needing updates - just as expected! 🎉 Thanks for adding this to the toolbox. It will be nice to have if we ever need to run it again or adapt it into an ETL (which i really hope not). 🙏 🚀

atd-toolbox/fix_cr3_stored_flag/fix_cr3_stored_flag.py
roseeichelmann: lol i didnt even think to use that library/didnt know about it!

mddilley: Oh cool - totally valid either way! 😎

patrickm02L (approved):
### Review

Approved. Thanks for running the script in Prod & Staging. Glad we have a new 🔧  for the 🧰 

### Testing

Testing the example crashes from Issue #13021 in Production worked
https://visionzero.austin.gov/editor/#/crashes/19662002
https://visionzero.austin.gov/editor/#/crashes/19659622

I don't think those crashes are in Staging. I put the crash ID in the URL and also did a search and couldn't find them.

I did not test locally

chiaberry (approved):
✨

PR #1274, opened 2023-08-22
LDM Blueform crash location trigger

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/12644

This PR adds the non-CR3 crash location association trigger to the LDM schema. It also extends the `vision-zero` program for local testing to be able to restore from a specific .sql file and also pg_dump the current state of the DB. I'm hoping that the restore and dump commands makes starting/stopping LDM dev easier when jumping between work on the current schema and on the LDM schemas.

**FYI: We are still feeling out how to test LDM work so open to any feedback! Going to discuss how this goes in the next LDM standup, and the target branch is for the updates to the vision-zero program. Removing the .sql files before merge is definitely on the table.**

## Testing
**URL to test:** 
Local only

**Steps to test:**
1. To get a LDM data dump, go to the 1Password Developers vault and look for the entry called **Vision Zero LDM Staging pg_dump command**. You will need to update the command with the path to your snapshots folder.
2. If you can access the VZ read replica, then you should be able run this command. It is restricted to the same IPs as the replica through the security group called `visionzero-rds-read-replica`. We can add anyone who is looking to test this if needed.
4. Once you have the file in your snapshots folder, run `./vision-zero replicate-db -f <your filename>.sql` to restore from the LDM dump that you just made. Totally open to suggestions on changing up the command arguments. 🙏 I just went with `-f` to get it working.
5. Connect to your local VZ database and you will know that it worked when you see the `cris` and `vision_zero` schemas.
6. Run the SQL in `atd-vzd/ldm_triggers/update_non_cr3_location/update_noncr3_location.sql`
7. Run the SQL in `atd-vzd/ldm_triggers/update_non_cr3_location/up.sql`
8. You can now run the test steps 1-4 in https://github.com/cityofaustin/atd-vz-data/pull/1269
9. You can also test a data dump of the current state of your DB by running `./vision-zero dump-local-db`

---
#### Ship list
- [ ] Code reviewed


mddilley:
Closing this since it is now split into:
- LDM part https://github.com/cityofaustin/atd-vz-data/pull/1281 
- Local stack updates https://github.com/cityofaustin/atd-vz-data/pull/1280

PR #1275, opened 2023-08-25
Bring local `graphql-engine` version into sync with production / latest version

frankhereford:
## Associated issues

[This issue](https://github.com/cityofaustin/atd-data-tech/issues/13392) is the root cause of this change, but does not directly address it. 

This is a change that we would normally do during a release party directly into `master`, but due to our recent upgrade to the `graqphql-engine` middleware, this is needed to spin up the local stack with a replicated or copied database from production.

## Testing

**Steps to test:**

* Spin down your `graphql-engine` endpoint locally
* Replicate the production database into your local database via a backup fround on S3 or pulled directly from the read replica. `./vision-zero replicate-db`
* Check to see if your local `graphql-engine` starts up

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,14 +1,7 @@
 version: '3.7'
 services:
   graphql-engine:
-    image: hasura/graphql-engine:v2.20.1  # added to patch critical security vulnerability
-
-    # I'm keeping these next two notes around until we're fully up to date with 2.20.1+ in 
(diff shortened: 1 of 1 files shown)


mddilley (approved):
This lines up with the version that I used to init the migrations so I'm approving based on that testing. Thanks for the cleaning up the comments too! ✨

patrickm02L (approved):
Good to go 🚢

chiaberry (approved):

PR #1280, opened 2023-08-30
Extend the local stack script to pg_dump and restore the local DB

mddilley:
## Associated issues

This splits out the local tooling updates from https://github.com/cityofaustin/atd-vz-data/pull/1274 so that we can get this merged into master sooner.

## Testing
**URL to test:** 
Local

**Steps to test:**
1. Start up the local stack by activating the virtual env and then running `./vision-zero replicate-db`
2. You should now have a new snapshot in `/atd-vzd/snapshots/` with today's date
3. Now, run `./vision-zero dump-local-db` to pg_dump the local DB into `/atd-vzd/dumps/`
4. You should now have a new data dump
5. Now you can move that data dump file into the snapshots folder and run `./vision-zero replicate-db -f <your data dump filename>.sql` to restore into the local DB

---
#### Ship list
- [ ] Code reviewed


roseeichelmann (approved):
Thanks for adding these new commands and functionality to the local stack, they really make switching between diff local environments & ldm work much easier! :ship: :sparkles:

chiaberry (approved):
ran through the steps and 👍🏼

vision-zero
```diff
@@ -360,10 +396,12 @@ def get_main_parser():
             "psql",
             "tools-shell",
             "replicate-db",
+            "dump-local-db",
             "remove-snapshots",
             "hasura-console"
         ],
     )
+    parser.add_argument("-f", "--filename", required=False)
```
frankhereford, line 404: At first, I was going to suggest something more specific that describes what the file needs to be, but then I realized that you can only use the `-f` flag meaningfully after the `replicate-db` command, so 👍.

vision-zero
```diff
@@ -94,11 +96,16 @@ def replicateDb(args):
...
-    )
+    snapshotFilename = ""
+    if(args.filename):
+        snapshotFilename = args.filename
+    else:
+        today = date.today().strftime("%Y-%m-%d")
+        snapshotFilename = (
+            "visionzero_" + today + "_" + (
+                "with-change-log" if args.include_change_log_data else "without-change-log") + ".sql"
+        )
```
frankhereford, line 107: I reallllly like that you can specify a filename to be used when pulling data down from the production machine.

frankhereford (approved):
This looks and works great Mike.

I have one question though: You've added support to dump the local DB into the `dumps` directory, and you've added support to specify the filename that you're going to use to replicate into from the live DB and to restore from. Is the intent that if you want to load a local dump file you have handy into your DB for you to copy it into the snapshots folder and then `replicate-db` with the `-f` arg pointing at it?

I guess put another way, is the intent to make it a manual operation to promote a local dump file into a snapshot that you can use by `mv`ing it from `atd-vzd/dumps` to `atd-vzd/snapshots`? 

If so, that's totally cool, I'm just curious if I am getting the whole picture. 

Thanks!! 🚢 🚢 🚢

mddilley:
@frankhereford Yeah, it does involve manually moving files between the `/dumps/` and `/snapshots/` folder. I've thought about how those folders serve the same purpose and went with a separate folder since they come from different sources. Not sure if that is the best call, but we can update it if it gets annoying!

frankhereford:
@mddilley wrote:

> Not sure if that is the best call, but we can update it if it gets annoying!

I think it's great just the way it is. I just wanted to make sure I wasn't missing anything. This is top shelf stuff.

PR #1282, opened 2023-08-30
LDM lookup table support for the CRIS import & queries to transform existing data

frankhereford:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/12577

### `cris_import.py`
The updated script which will now use the LDM lookup tables
### `data/`
Where you can put your CRIS schema spreadsheet named `cris_spec.xlsx`
### `parse_cris_schema.py`
One-time use program to parse the CRIS schema XLSX document and populate the lookup tables and create the materialized views of them.

## Testing

NB: I've done my best to make this as streamlined as possible, but even doing that has been a challenge. Please let me know what information you need to better test this work. I'll be prompt in supplying answers here in the PR.

* With your local VZ DB container running, we're going to replicate the "HEAD" version, if you will, of the LDM database schema as found on the `ldm-staging` RDS cluster. Heads up: this will wipe out anything you have locally in `atd_vz_data`. 
  * This takes 15-20 minutes.
  * The snipped part of the hostname and password can be found in 1PW under the entry: `RDS Instance: LDM staging`
```bash
docker exec -i \
-e PGHOST=localhost \
-e PGDATABASE=postgres \
-e PGPASSWORD=visionzero \
-e PGUSER=visionzero \
visionzero-postgis \
psql -c "drop database atd_vz_data with (force)";

docker exec -i \
-e PGHOST=localhost \
-e PGDATABASE=postgres \
-e PGPASSWORD=visionzero \
-e PGUSER=visionzero \
visionzero-postgis \
psql -c "create database atd_vz_data";

docker exec -i \
-e PGHOST=ldm-staging.<snip>.us-east-1.rds.amazonaws.com \
-e PGDATABASE=atd_vz_data \
-e PGUSER=postgres \
-e PGPASSWORD=<snip> \
visionzero-postgis \
pg_dump --no-owner| \
perl -pe 's/^(GRANT ALL ON .* TO) vze/$1 visionzero/g' | \
perl -pe 's/^(REVOKE.*rdsadmin)/-- $1/g' | \
docker exec -i \
-e PGHOST=localhost \
-e PGDATABASE=atd_vz_data \
-e PGPASSWORD=visionzero \
-e PGUSER=visionzero \
visionzero-postgis \
psql -v ON_ERROR_STOP=1
```
* Next, please delete the old lookup schemas created prior to this new set which are based on the official CRIS schema specification.
```
docker exec -i \
-e PGHOST=localhost \
-e PGDATABASE=atd_vz_data \
-e PGPASSWORD=visionzero \
-e PGUSER=visionzero \
visionzero-postgis \
psql -c "drop schema lookup cascade"

docker exec -i \
-e PGHOST=localhost \
-e PGDATABASE=atd_vz_data \
-e PGPASSWORD=visionzero \
-e PGUSER=visionzero \
visionzero-postgis \
psql -c "drop schema cris_lookup cascade"

docker exec -i \
-e PGHOST=localhost \
-e PGDATABASE=atd_vz_data \
-e PGPASSWORD=visionzero \
-e PGUSER=visionzero \
visionzero-postgis \
psql -c "drop schema vz_lookup cascade"
```

* Edit: Since we're about to swap in LDM FK values in our fact tables, we need to drop the one constraint we have that links them to the old CRIS lookup table. See [this comment below](https://github.com/cityofaustin/atd-vz-data/pull/1282#issuecomment-1714187261).

```
docker exec -i \
-e PGHOST=localhost \
-e PGDATABASE=atd_vz_data \
-e PGPASSWORD=visionzero \
-e PGUSER=visionzero \
visionzero-postgis \
psql --set ON_ERROR_STOP=on \
-c "alter table atd_txdot_person drop constraint if exists atd_txdot_person_prsn_injry_sev_id_fkey;"
```

* Next, please apply the queries found in [this file](https://gist.githubusercontent.com/frankhereford/b35034518a14a63042c9c256748db547/raw/417f630b78c27cf13f33277b8e678e903fbf399c/dump_of_lookup_schemata.sql) to add the lookup table schemata and contents. BTW, I have not added that file to the repo because it's one time use and I don't want to bloat the repo with a 17 meg file.
  * Please note, these were generated programmatically via this script using a spreadsheet that is provided by TXDOT here, however, the spreadsheet is imperfect, and you have to clean up various things here and there to make it work. I believe it's created by a human, not a program, and you get these sorts of things. 
  * Please review `parse_cris_schema.py` in this PR if you'd like to see the details of how the `lookup`, `cris_lookup`, and `vz_lookup` are populated. I'd be happy to share the polished up XLSX file to anyone who would like it, or you can create your own starting from the file [standardextractfilespecification.xlsx](https://ftp.txdot.gov/pub/txdot-info/trf/crash_statistics/automated/standardextractfilespecification.xlsx) available here: https://ftp.txdot.gov/pub/txdot-info/trf/crash_statistics/automated/.
  * This program is intended to run inside the same docker image/container that the main CRIS import script runs in. I like to `docker compose exec -it cris-import bash` to get a shell on the container and then run it from there.
  * This program takes about 10-15 minutes to run.

```bash
curl https://gist.githubusercontent.com/frankhereford/b35034518a14a63042c9c256748db547/raw/417f630b78c27cf13f33277b8e678e903fbf399c/dump_of_lookup_schemata.sql  | \
docker exec -i \
-e PGHOST=localhost \
-e PGDATABASE=atd_vz_data \
-e PGPASSWORD=visionzero \
-e PGUSER=visionzero \
visionzero-postgis \
psql  -v ON_ERROR_STOP=1
```

* Now that you have the lookup tables in place, it's time to transform the existing fact tables to use them. This is done with a series of SQL queries to be run against the database. Connect to it in DBeaver or your client of choice and run the queries found in `atd-etl/cris_import/ldm-materials/queries_to_transform_existing_data_to_ldm_lookup_tables.sql`
  * This takes 5-10 minutes.

* Next you'll need to get a CRIS extract in place in `/home/txdot` on the SFTP endpoint. You can do this by requesting one from CRIS or using some that we have on the server. As root on the machine, you can copy any of the files in `/root/schema_inspection_extracts` to `/home/txdot` and `chown` them to be own by txdot.
  * Note, in the following example, you're loading a zip file which is primarily or entirely made of 2023 CRIS schema crashes. This process should be executed for all CRIS schema versions. You can find other extracts created for this in that same directory.

```
# cp /root/schema_inspection_extracts/extract_2023_20230829092129000_86885_20230812_HAYSTRAVISWILLIAMSON.zip /home/txdot; 
chown txdot:txdot /home/txdot/*zip
```

* Finally, you're good to run the actual import and test how it works inserting new data into your database.
  * From the `atd-etl/cris_import` directory, you should be able to `docker compose run cris-import`

---
#### Ship list
- [ ] Code reviewed


atd-etl/cris_import/cris_import.py
```diff
@@ -517,6 +502,88 @@ def align_db_typing(map_state):
     return map_state
 
 
+def convert_to_ldm_lookup_ids(state):
```
mddilley, line 505: I think this function could use a bit of doc about what it is doing. I could see the context of what is happening here getting away from us over time. This is tricky and heady stuff!

atd-etl/cris_import/cris_import.py
```diff
@@ -517,6 +502,88 @@ def align_db_typing(map_state):
...
+                        from lookup.{field["lookup_table"]}
+                        where true 
+                            and source = 'cris'
+                            and cris_id = {state['import_schema']}.{table["imported_table"]}.{field["field_name"]}::integer
+                        )""")
+        if len(assignments) > 0:
+            sql += ", ".join(assignments) 
+
+            # debugging functionality
+            # sql = remove_newlines_and_collapse_spaces(sql)
```
mddilley, line 569: Should this stick around?

atd-etl/cris_import/cris_import.py
```diff
@@ -588,12 +652,14 @@ def align_records(map_state):
...
                 # generate some record specific SQL fragments to identify the record in larger queries
                 record_key_sql, import_key_sql = util.get_key_clauses(table_keys, output_map, table, source, map_state["import_schema"])
+                if not record_key_sql or not import_key_sql:
+                    continue
 
                 # To decide to UPDATE, we need to find a matching target record in the output table.
                 # This function returns that record as a token of existence or false if none is available
                 if util.fetch_target_record(pg, output_map, table, record_key_sql):
                     # Build 2 sets of 3 arrays of SQL fragments, one element per column which can be `join`ed together in subsequent queries.
-                    column_assignments, column_comparisons, column_aggregators, important_column_assignments, important_column_comparisons, important_column_aggregators = util.get_column_operators(target_columns, no_override_columns, source, table, output_map, map_state["import_schema"])
```
mddilley, line 596: cool so this stuff can be removed now that we are not concerned with "important columns" since the VZ layer let's us separate CRIS data from edits. As I'm getting into these details, it reminded me to ask about https://github.com/cityofaustin/atd-vz-data/pull/1240. Does that PR still fit into the code that we are trying to push into the LDM collector branch?

atd-etl/cris_import/ldm-materials/queries_to_transform_existing_data_to_ldm_lookup_tables.sql
mddilley, line 1: this looks similar to what is happening further up in the `convert_to_ldm_lookup_ids` helper. Are these related? I'm also wondering if this is code that will be maintained over time unlike the one-time gist that was part of the test instructions.

atd-etl/cris_import/lib/sql.py
mddilley, line 1: okay, looks like the important changes here are updating to target the `cris` schema. 👍

atd-etl/cris_import/parse_cris_schema.py
mddilley, line 1: Also thinking about if this script will need maintenance or if this is one and done. I guess a better question is what happens when we have a new schema in the future?

If we would need to run this again, some docstrings would go a long way in helping the next person out.

atd-etl/cris_import/parse_cris_schema.py
```diff
@@ -0,0 +1,688 @@
...
+            cursor = pg.cursor(cursor_factory=psycopg2.extras.DictCursor)
+            cursor.execute(sql)
+            rows = cursor.fetchall()
+            cursor.close()
+            if len(rows) > 0:
+                print(rows)
+                print("ERROR: Duplicate values in lookup table")
+                exit(1)
+
+def create_materialized_views(file_path, pg):
```
mddilley, line 157: I inspected the DB and see the the `lookup` schema has these materialized view, and the `cris_lookup` and `vz_lookup` have tables. What are the relationships between these three schema? I'm also curious about the materialized views out of curiosity. I just don't have a good enough feel about postgres to connect the dots.

mddilley (commented):
The test instructions worked out great, and the import script is a lot less intimidating to run locally now that we have it separated out like the other ETLs! I left comments and questions as I read through the code. I have a mental block on some of this, but I'm getting there! Maybe just talking through some of this early next week could help me out too. Thanks for tackling this because it is a lot to think through! 🙏

atd-etl/cris_import/cris_import.py
```diff
@@ -76,22 +73,21 @@ def main():
...
         logical_groups_of_csvs = group_csvs_into_logical_groups(archive, dry_run=False)
         for logical_group in logical_groups_of_csvs:
             desired_schema_name = create_import_schema_name(logical_group)
+            print("desired_schema_name", desired_schema_name)
             schema_name = create_target_import_schema(desired_schema_name)
             pgloader_command_files = pgloader_csvs_into_database(schema_name)
             trimmed_token = remove_trailing_carriage_returns(pgloader_command_files)
-            typed_token = align_db_typing(trimmed_token)
-            align_records_token = align_records(typed_token)
+            typed_token = align_db_typing(trimmed_token) # reminder, these "tokens" are from when this was a prefect script, and they now contain execution context
```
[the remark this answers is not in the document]
frankhereford: This definitely deserves some explanation and some inline comments. Great catch!

Just like you're suggesting, this wonky pattern initially came out of the `flow` definition code when this was a prefect script. I /think/ this is how it played out.

* Passing something from one function to another was the mechanism that prefect used to derive the flow's task plan. Sometimes the "something" being passed is just the signal, not any particular data, so these variables got the "token" name. I think of it as each function passing the next the one permission to run.
* Also, from the prefect era, the script's task execution took advantage of the `.map()` function, allowing portion of the tasks to operate on multiple sets of data in "parallel" execution.
* When this was ported back into a stand-alone package, the `.map()` functionality become the two `for` loops [here](https://github.com/cityofaustin/atd-vz-data/blob/ldm-cris-import-with-lookup/atd-etl/cris_import/cris_import.py#L80-L82).
  * During that port process, the tasks / functions were written with the prefect-like expectation that any context you would be getting out of a prefect object and not some enclosing scope. So, outside of prefect, it was natural to build up an object containing the state and parameters of the series of tasks, and then pass it into and then back out of the tasks/functions. 
  * And due to all that, a `dict` is created early in the job's iterations over each logical group of files to process, [here](https://github.com/cityofaustin/atd-vz-data/blob/ldm-cris-import-with-lookup/atd-etl/cris_import/cris_import.py#L715-L725), and it's passed down one-by-one through all the functions.

I'll stick a couple of sentences in there and push it up and maybe a link back to this comment. Thanks for asking about this.

atd-etl/cris_import/cris_import.py
frankhereford: I agree, totally. I stuck some line-by-line comments of what is going on, and put this out in front of the function:

```
# The following function is used to change, in place, the foreign key values
# from those used by CRIS to those used by the LDM.  This function gets the "import"
# schema name in the argument, and then will use some hard-coded maps between columns 
# and tables to build a SQL command which updates all the appropriate foreign key values
# in the import schema's tables.
```

atd-etl/cris_import/cris_import.py
frankhereford: Good call 🧹

atd-etl/cris_import/cris_import.py
frankhereford: Won't fib that I had a moment of fear while I tried to think about how all the branches related, but it's all good. This PR's branch actually came off of the one in 1240, and merging this would include that work entirely. I think I'll close that PR.

atd-etl/cris_import/parse_cris_schema.py
frankhereford: Great question. 

I'd say it's a one-and-done situation. I think that the spreadsheet that CRIS produces, worked really well getting at the map of columns to tables, I don't think it will ever be something that we'd want to try to automate or use again. I think we should pursue using the "lookup" CSV that comes in every export to maintain the CRIS lookup schema or at least alert us when it changes.

I'm thinking that we should get it into `ldm-collector` pre-LDM launch and maybe remove it after the LDM in place?

atd-etl/cris_import/parse_cris_schema.py
frankhereford: Sure thing, here's the three schema and their relationships:

### `cris_lookup`
This is a schema holding the CRIS provided lookup tables and their values. They are populated from parsing a slightly-modified version of a XLSX file that CRIS makes available [here](https://ftp.txdot.gov/pub/txdot-info/trf/crash_statistics/automated/standardextractfilespecification.xlsx). 

In the future, I think it'd be neat to be checking and possibly maintaining the contents of these tables by examining the `lookup` CSVs provided in our daily exports. 

These tables match up, one-to-one, to a table in the `vz_lookup` schema, and entries in both tables in each pair, share a common ID sequence so they can define a always-unique column between each other.

Most of the tables have a similar shape, but there are some outliers, and you can see them being handled with special handler functions in this program.

### `vz_lookup`
This is a schema holding the VZ provided lookup table values. By and large, the shape of these tables is predictable based on the columns that CRIS gave us. Also, there's nothing preventing us from adding VZ specific associated data which is a feature of the LDM. 

These tables are not populated by this script, because it's interpreting CRIS provided data. However, we will need to add some specific VZ provided values that have been added into our production database. 

When we search for orphaned FK values before we launch🎗️, it'll reveal which ones VZ has requested and used.

### `lookup` 
* This contains materialized views that unify the `vz_lookup` and `cris_lookup` pairs of tables. The query to `create` each materialized view is constructed programmatically, so they may be easier to review via a DB client and inspecting the materialized view.
* Materialized views are used here because lookup tables change their values very slowly, and it's well worth having to `refresh materialized view ...` when we make the very occasional change and pick up the ability to index these tables.
* This feels like it should get a better name, but I'm not sure ... `ldm_lookup`? It feels redundant and not like adding more information by adding the `ldm_`, but I'm not sure.

atd-etl/cris_import/ldm-materials/queries_to_transform_existing_data_to_ldm_lookup_tables.sql
frankhereford: Good call. It's very similar. These queries are used during the testing instructions (and ultimate LDM launch) to update all the data that is in place in the database being converted. The place that you referenced earlier in the execution process is where the data that has been imported from the CRIS CSV files into an `import_...` schema is converted.

Here it's a one-time operation during LDM launch, and above it's every time we pull in new data during our daily ETL run.

atd-etl/cris_import/lib/sql.py
frankhereford: 👍 You got it. These changes were very much a search and replace type operation.

frankhereford:
While working on another PR, I've observed something that adds a wrinkle to this PR. Of the four primary fact tables which we use, there one which contains a single FK relationship to a lookup table which is thrown a curve ball by this PR because this PR is switching up the foreign table that is being referred to. The constraint in question is shown here, and I'm going to add to the instructions to remove the constraint near or at the beginning of the process. 

Because it's a highly arbitrary column to be enforcing this data, I think it's OK to remove it here, and further, I think we should consider the implications of adding it back, along with /all/ other FK relationships (all 100 or so of them), back in as part of the LDM refactor. I'll ask tomorrow in the standup. 

```sql
--
-- Name: atd_txdot_person atd_txdot_person_prsn_injry_sev_id_fkey; Type: FK CONSTRAINT; Schema: public; Owner: -
--
    
ALTER TABLE ONLY cris_fact_tables.atd_txdot_person
    ADD CONSTRAINT atd_txdot_person_prsn_injry_sev_id_fkey FOREIGN KEY (prsn_injry_sev_id) REFERENCES public.atd_txdot__injry_sev_lkp(injry_sev_id) ON UPDATE RESTRICT ON DELETE RESTRICT;

```

atd-etl/cris_import/cris_import.py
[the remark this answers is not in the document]
mddilley: awesome, yes, this code (and you) have been through a lot with the different ways that we have run this! Maybe some day in the future we can clean it up some more since we are now settled in Airflow land! 🏡

atd-etl/cris_import/cris_import.py
mddilley: awesome - i think that is very helpful. 🙏

atd-etl/cris_import/parse_cris_schema.py
mddilley: cool, thanks, yeah, or maybe we have a ldm toolbox or something? But, now that I've seen the end result of the LDM build up process, it is all making more sense to me. 🙏

atd-etl/cris_import/parse_cris_schema.py
mddilley: awesome, never commented here that this helped me explore the DB a lot after running through the LDM transformation steps. We should copy/paste your explanation here into some docs!

mddilley (approved):

roseeichelmann (approved):
approving based on my review of https://github.com/cityofaustin/atd-vz-data/pull/1288

patrickm02L (approved):
Great job to everyone involved! 🎉

PR #1290, opened 2023-09-12
Migrations for cr3 location triggers

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/13055

## Testing
**URL to test:** 
Local

**Steps to test:**

1. Use the readme in the /atd-vzd/ folder to get your local stack up and running with production data and the latest migrations and metadata applied
2. Use the Hasura CLI instructions in that readme to apply the migrations and metadata in this branch. From /atd-vzd/:
```
hasura migrate apply --envfile .env.local
hasura metadata apply --envfile .env.local
```
3. Follow the test steps 2-4 (the first step of those instructions is to manually make the changes that these migrations do for us) in https://github.com/cityofaustin/atd-vz-data/pull/1270
5. You can also run ./vision-zero hasura-console to open your local Hasura console and make sure that the [Event Trigger](http://localhost:9695/console/events/data/manage) called location_from_latlon is no longer present in Hasura.
6. If you want to test the down migration, run:

`hasura migrate apply --down 1 --envfile .env.local`

---
#### Ship list
- [x] Run migration steps in readme in /atd-vzd/ if needed
- [x] Code reviewed


atd-vzd/metadata/databases/default/functions/functions.yaml
```diff
@@ -1,8 +1,8 @@
 - "!include public_find_cr3_collisions_for_location.yaml"
-- "!include public_find_cr3_mainlane_crash.yaml"
 - "!include public_find_crash_in_jurisdiction.yaml"
 - "!include public_find_crash_jurisdictions.yaml"
 - "!include public_find_crashes_in_jurisdiction.yaml"
+- "!include public_find_location_for_noncr3_collision.yaml"
 - "!include public_find_location_for_cr3_collision.yaml"
```
mddilley, line 6: I ran into an inconsistent metadata warning when I applied the Hasura metadata locally. The first line added back here is a function that was dropped in https://github.com/cityofaustin/atd-vz-data/pull/1286 and the second one is being dropped in this PR. These two lines can go since Hasura doesn't need to worry about tracking or not tracking them anymore.

atd-vzd/migrations/default/1694468261303_cr3_location_update_insert_triggers/up.sql
```diff
@@ -0,0 +1,19 @@
...
+FOR EACH ROW
+WHEN (OLD.position IS DISTINCT FROM NEW.position)
+EXECUTE PROCEDURE update_cr3_location();
+
+COMMENT ON FUNCTION update_cr3_location IS 'This trigger function is used to update the location_id on insert and update of a record in atd_txdot_crashes.';
+
+-- Drop functions that are no longer needed
+DROP FUNCTION find_cr3_mainlane_crash;
+DROP FUNCTION find_location_for_cr3_collision;
+DROP FUNCTION find_location_id_for_cr3_collision;
```
mddilley, line 19: Just to make sure that we can migrate down and then back up again, these drop statements could use `IF EXISTS` since they aren't added back in the down migrations. I ran into this in my PR too.

atd-vzd/triggers/update_cr3_location.sql
mddilley, line 1: This works great! I also tested nulling out `latitude`, `longitude`, `latitude_primary`, and `longitude_primary` because you caught that on mine, and I saw the location id clear out. 🚀

mddilley (changes requested):
All of the test steps work perfectly! Just a couple small fixes and this will be 🚢 🚢 🚢  🙌

atd-vzd/migrations/default/1694468261303_cr3_location_update_insert_triggers/up.sql
roseeichelmann: thanks for catching this!

roseeichelmann:
@mddilley thanks for catching those things, just pushed up those updates!

mddilley:
@roseeichelmann awesome, just tested and there is a small bug with the syntax `DROP IF EXISTS FUNCTION` vs. `DROP FUNCTION IF EXISTS`. Ready to approve when that is patched up 🚀

roseeichelmann:
@mddilley ahh sorry i should have tested the migration again before pushing up lol but i just fixed that and tested the down/up and metadata, shoulddd be good to go now

mddilley (approved):
So happy to see all your work pay off! 🙌 🥳 ✅

atd-vzd/triggers/update_cr3_location.sql
frankhereford, line 1: Rose, I'm curious what you think about if we need to keep these migration-like files in here with proper `graphql-engine` migration support in place. Like, I think now if I wanted to find what the latest and greatest version of this function is, I would probably get into the root of the repo and do:

```bash
find atd-vzd/migrations/default -name 'up.sql' -exec grep -l 'update_cr3_location' {} + | sort -r
```

The top line in the response is the most recent `up.sql` that mentions the string and going backwards is its history. 

I totally think I may be missing something, and I'm super curious to what your reaction is. 

CC/ @mddilley too - I'm curious if you have a take on this as well. Thanks to you both!!!

frankhereford (approved):
So much delete!! I sure love seeing the events folder clearing out. 😅 

This is superb, and it works great. So nice to see the fruits of the migration system already manifesting in such an organized and well presented change to the database. Thank you Rose!

atd-vzd/triggers/update_cr3_location.sql
chiaberry: I may be misunderstanding what your comment is Frank, are you saying you don't see the need for this file because you could execute that command to find what the latest version is?

I do not work much in VZ, but I do find it easier that in moped we have the views in their own files to be able to easily see changes between migrations, and in moped the top of the file has what up.sql file is the latest aka
`-- latest version 1694809058492_list_view_optimize`

chiaberry (approved):

patrickm02L (approved):
Send it 🥏

atd-vzd/triggers/update_cr3_location.sql
chiaberry: Ah yes, I understood that it would tell you what file to look in, but not the contents. In moped we manually add that line to the top of the git tracked file so devs can see what the latest migration is, and having the file is to help at a glance see changes in the PR from one migration to another.

roseeichelmann: Yeah im cool with leaving this file out of the repo, I honestly was kinda just following Mikes lead and saw that he left his trigger function in his similar [PR](https://github.com/cityofaustin/atd-vz-data/pull/1286/files#diff-7f96b3b42cf4b8e4346e5865e828db0c8f11322e41df45add370860a4f0b8bf7) for the blueform crashes so figured I would do the same thing, but I agree that it's not necessary and maybe will lead to confusion now? What do you think @mddilley

mddilley: hey y'all - late to the party, but I think that we can leave it in for now and decide later so this can 🚢. I agree with Chia that the [Moped views folder](https://github.com/cityofaustin/atd-moped/tree/main/moped-database/views) is great for seeing a diff during reviews, and we've succeeded at keeping files up to date over time there. I created https://github.com/cityofaustin/atd-data-tech/issues/13950 to bring to a future refinement meeting.

PR #1291, opened 2023-09-13
Implement custom 404 page

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/12995

Even though the issue only specifies to do this for crash ids that dont exist, i also implemented it for location ids.

I know the mockup doesn't include the navbar on the left side, but I thought that leaving the navbar would make it easier for users to navigate to any other tab if they reach the 404 page rather than only having the option to go back to the Home page. Let me know what y'all think.

## Testing
**URL to test:** 
[Netlify](https://deploy-preview-1291--atd-vze-staging.netlify.app/#/dashboard)

**Steps to test:**
1. Go to the crashes page and then type in an int after /crashes/ in the URL that isn't a crash id we have, like https://deploy-preview-1291--atd-vze-staging.netlify.app/#/crashes/22. The 404 page should be rendered.
2. Click on the Back to Homepage button to make sure it works.
3. Go to the Locations page and do the same thing, for example go to https://deploy-preview-1291--atd-vze-staging.netlify.app/#/locations/h750. You should once again get the 404 page if that location id doesn't exist.
4. Typing in a random url like https://deploy-preview-1291--atd-vze-staging.netlify.app/#/hello should just reroute you back to the dashboard which I feel like is acceptable behavior.

---
#### Ship list
- [ ] Code reviewed


atd-vze/src/routes.js
```diff
@@ -147,7 +148,12 @@ const routes = roles => [
   { path: "/dev/widgets", name: "Widgets", component: Widgets },
   { path: "/dev/charts", name: "Charts", component: Charts },
   { path: "/profile", name: "Profile", component: Profile },
-  { path: "/dashboard", name: "Dashboard", component: VZDashboard },
+  {
+    path: "/dashboard",
+    exact: true,
```
roseeichelmann, line 153: This made it so typing in something like /#/dashboard/hello would reroute you back to the dashboard.

atd-vze/src/views/Crashes/Crash.js
```diff
@@ -163,19 +164,21 @@ function Crash(props) {
     geocode_method: geocodeMethod,
     cr3_file_metadata: cr3FileMetadata,
     investigator_narrative_ocr: investigatorNarrative,
-  } = data.atd_txdot_crashes[0];
+  } = !!data?.atd_txdot_crashes[0] ? data?.atd_txdot_crashes[0] : 0;
```
roseeichelmann, line 167: If i leave this unhandled we get an error when going to a crash id that doesn't exist because  `data.atd_txdot_crashes[0]` is undefined. So I handled this by setting these values to 0 if the data doesn't exist, which feels kinda weird to do so lmk if there's a better way.

atd-vze/src/views/Widgets/VZNoticeWidget.js
```diff
@@ -63,7 +63,14 @@ class VZNoticeWidget extends Component {
...
 
-    const card = { style: "clearfix", color: color, icon: icon, raster_icon: raster_icon, raster_icon_alt: raster_icon_alt, classes: "" };
+    const card = {
+      style: "clearfix",
+      color: color,
+      icon: icon,
+      raster_icon: raster_icon,
+      raster_icon_alt: raster_icon_alt,
+      classes: "",
+    };
```
roseeichelmann, line 73: this is all formatting w prettier :nail_care:

patrickm02L:
Thanks for leaving in the navbar - that's perfect.

### Quick observation from preliminary testing
The links in the Crashes and Location tabs are going directly to the 404 page.

https://github.com/cityofaustin/atd-vz-data/assets/5566727/5178402d-5724-4682-8f4b-e463f2be6ab1

roseeichelmann:
@patrickm02L Whoops thanks for catching that, just pushed up the fix!

atd-vze/src/routes.js
```diff
@@ -63,6 +63,7 @@ const ToolsUploadNonCR3 = React.lazy(() =>
 const CreateCrashRecord = React.lazy(() =>
   import("./views/Tools/CreateCrashRecord")
 );
+const Page404 = React.lazy(() => import("./views/Pages/Page404/Page404"));
```
chiaberry, line 66: I'm getting this warning
`  Line 66:7:  'Page404' is assigned a value but never used`

atd-vze/src/views/Crashes/Crash.js
chiaberry: I'm not sure if theres a better way to do it, but instead of a 0, what about setting the object to `{}`

chiaberry: Since if `data?.atd_txdot_crashes[0]` exists, you get an object and not a number

atd-vze/src/routes.js
roseeichelmann: Thanks yeah i dont need that line anymore and forgot to take it out!

atd-vze/src/views/Crashes/Crash.js
roseeichelmann: Oo yeah that makes more sense

chiaberry (approved):
Tested and i like the 404 page within the side menus. I left a note about how to handle the undefined object

patrickm02L (approved):
Works like a charm! 💎

atd-vze/src/routes.js
mddilley: This works great! I wanted to offer up the idea of also adding a wildcard route to the very end of this array for any other routes not defined in this file that would redirect to the 404 page. Not sure if we would want that behavior, though. It would use a `*` path kind of like [in this article](https://ui.dev/react-router-v5-handling-404-pages).

mddilley (approved):
This looks great! Awesome to see this design implemented and to have unavailable records handled much more gracefully. 🙏 🚢

atd-vze/src/routes.js
mddilley: @roseeichelmann this works great as-is! For some reason, I thought there were routes that weren't captured by the redirect behavior so I was thinking that both the wildcard and the redirect behavior would work together to catch all routes. It doesn't seem to be needed though! 🚢🚢🚢

mddilley:
@roseeichelmann Forgot to mention this one earlier in dev sync. I think that this one can merge to make the release too, right?

frankhereford (approved):
This is great, thank you Rose! 🚢 🚢 🚢

PR #1293, opened 2023-09-21
LDM cr3 crash location trigger

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/12646

So originally I thought the `cris_facts.atd_txdot_crashes` table shouldn't have a `location_id` column since that's not a value we get from CRIS but that proved to be a reallyyyyy difficult and complicated pattern to carry out in these triggers. It involved a lot of extra steps and complex logic. In the end I'm deciding this solution is best, to have the `location_id` column in both schemas and in the combined view it will either be the `location_id` from the `cris_facts` or `vz_facts` if there is a non-null value for it there. Now with some thought this solution makes more sense to me in general, it would have been weird imo for there to be `location_id` in the `vz_facts.atd_txdot_crashes` that really represented the location of the coordinates in the `cris_facts` schema.

A follow up to this issue is to address the stuff in the TODO section, the UI now needs to match up to these changes and stop using the `latitude_primary` and `longitude_primary` columns.

## Testing
**URL to test:** 
Local

**Steps to test:**

1. Pull this branch locally.
2. Replicate from prod locally by running `./vision-zero -l replicate-db`
3. From the atd-vzd folder, run `hasura migrate apply --envfile .env.local` and then `hasura metadata apply --envfile .env.local`
4. Test inserts to `cris_facts.atd_txdot_crashes`:
```plpgsql
-- insert a cr3 crash with coordinates that will associate to location_id 8337406736
INSERT INTO cris_facts.atd_txdot_crashes(latitude, longitude, crash_id)
VALUES (30.16841600, -97.78449000, 9999);

-- insert a cr3 crash with coordinates that will not associate to a location
INSERT INTO cris_facts.atd_txdot_crashes(latitude, longitude, crash_id)
VALUES (30.386136, -97.67309, 8888);

-- insert a cr3 crash without coordinates that will not associate to a location
INSERT INTO cris_facts.atd_txdot_crashes(latitude, longitude, crash_id)
VALUES (NULL, NULL, 7777);
```
5. Update an existing record with a `location_id` of NULL to have the `latitude` and `longitude` of a record with a location. After update, you should see the record associate with the same location as the record you took the lat/lon from.
6. Update an existing record with a `location_id` that is not null and update it with `latitude` and `longitude` coordinates from an existing record that has latitude and longitude and no location_id associated.
7. Now test editing records in `vz_facts.atd_txdot_crashes`.
```plpgsql
-- edit one of the records in vz_facts to associate to location_id 5B5D5E72E4
UPDATE vz_facts.atd_txdot_crashes
SET latitude = 30.298207784614654,
    longitude = -97.74852940415873
WHERE crash_id = 7777;
```
8. Check in the unified view `public.atd_txdot_crashes` that crash_id 7777 has the location_id from the vz_facts schema, 5B5D5E72E4.

---

### TODO:

- [ ] I think we should only have `latitude` and `longitude` columns, no more `latitude_primary` or `latitude_geocoded`. The `latitude_primary`/`longitude_primary` columns were used to keep track of lats from CRIS vs lats from VZ edits. We need to deprecate these other columns and adjust the code so that when users update coordinates through the UI map feature, they update the `latitude` `longitude` columns instead of the primary columns
- [ ] I also think we should deprecate the `position` column because once again it is repetitive data that opens up room for de normalization.

---
#### Ship list
- [ ] Code reviewed


atd-vzd/migrations/default/1698277990422_create_location_id_triggers/up.sql
```diff
@@ -0,0 +1,21 @@
...
+
+CREATE OR REPLACE TRIGGER update_cr3_location_on_cris_insert
+BEFORE INSERT ON cris_facts.atd_txdot_crashes
+FOR EACH ROW
+EXECUTE PROCEDURE public.update_cr3_location();
+
+CREATE OR REPLACE TRIGGER update_cr3_location_on_vz_insert
+BEFORE INSERT ON vz_facts.atd_txdot_crashes
+FOR EACH ROW
+EXECUTE PROCEDURE public.update_cr3_location();
```
roseeichelmann, line 21: I'm actually realizing this trigger isn't necessary because new inserts to vz_facts.atd_txdot_crashes should always have null values, correct? Have we decided how we are going to handle temp crash records in the LDM? Because maybe that would be a situation in which we would need this trigger? I'll leave this in for now depending on what y'all think but I can update the up/down migrations and delete this trigger.

mddilley: I don't think that we need it either, and I tested removing this trigger and see `NULL` in `vz_facts`. I thought that the order of the triggers would be the opposite, but using `EXPLAIN ANALYZE` shows that the trigger that copies the record from `cris_facts` to `vz_facts` happens before the trigger that populates `location_id` in `cris_facts`.

Screenshot 2023-10-26 at 5 08 58 PM

atd-vzd/migrations/default/1698277990422_create_location_id_triggers/up.sql
mddilley, line 1: I like your naming here, and I think we should lean into better names across the board. My units ones have `units` in all the names which gets a bit redundant when exploring the DB with a client. I'll update in my PR.

mddilley (commented):
I tested this and all of the steps work out as advertised. ✅ I left some comments but holding off on approval until we agree what to do about storing derived values in the CRIS layer or not. I went in the opposite direction, but I'm open to refactoring when we see what we learn from the person tables.

frankhereford (approved):
## Approval & Intent of review
Rose, this is good work - thank you for this. In terms of the content of the PR, it's a solid approve. I do want to respond to some of the comments you made in your PR's comment as well. 

The following comments are in response to your PR comment's only, and they discuss the past technical decisions of the VZ database work and not the choices you've made in this PR. This PR does just what it should given those past choices, but I think this discussion is helpful in the context of your PR and your comments. 

## Form of fact tables being identical between VZ and CRIS or not

One of the issues your raise is the connection between our position 
In terms of keeping the form of the fact tables in the VZ and CRIS schemata the same so that the columns line up exactly -- if that is going to make writing triggers easier, then I'm all for it. 

##  Storing location_id in the CRIS fact table

I think that your decision to put the location_id from the CRIS latitude and longitude values in the CRIS fact table is a reasonable choice. I don't love having data derived from VZ parameters (the locations table) in the CRIS schema, but I get the benefits of doing it, and I think it can be argued either way.

## Floats vs. Geometry

Our real problem is that we have columns that store coordinates at all, like `latitude`, `longitude`, `latitude_primary`, `longitude_primary`, etc. Storing geometry simultaneously in floats and geometry columns is bad practice for, at least, the following reasons:

* It's denormalized
  * It's unclear which is the field of record and which is derived
  * It becomes up to us to keep these in synchronicity
* IEEE floats have variable precision based on the scale of the number. There are a lot more possible values near `0` than there are say near `1000000`. In reality, a measuring tape is the same measuring tape if you're standing on the equator and if you're standing at the north pole -- illustrating why floats are problematic here.
  * Precision in measurements and records of geometry are equally precise no matter how far you are from the datum (origin) of the plane, which implies that geometry needs to be processed with arbitrary precision math.
  * This is just the tip of this iceberg, and but it does show the importance of letting postGIS handle geometry and not try to reinvent that.
 
In your comment, you write:

> I think we should only have latitude and longitude columns, no more latitude_primary or latitude_geocoded. 

I would prefer that we drop all the floating point representations of parts of geometry entirely and settle on using the geometry type to store geometry exclusively. If we need to get at those components, we can use `ST_Y()` and `ST_X()`. This would let the nasty math of storing and manipulating geometry fall to postGIS where we want it, re-normalize the database both. It would also make storing the whole of the crash location in the correct schema as easy as how we might store the road name. 

When you think of "geometry" as a native type, which it is with postGIS, then it becomes pretty plain -- no more than you'd try to encode `a string` into integer columns (even though you can!), you store them in a character field. Let the DB do that encoding. Geometry goes in geometry columns, strings into characters, integers into ints, etc.

## `geometry` vs. `position` columns

>  I also think we should deprecate the position column because once again it is repetitive data that opens up room for de normalization.

A thousand times yes. I delivered this data back when with a `geometry` column only. That was copied into a `position` column but the source was left in there. 

The general practice in geographicly-enabled databases used to be to use `the_geom` which I detest because of the article in the name. That evolved into people tending to use `geom` which feels like a unnecessary abbreviation, which I don't care too much for, so I like to use `geometry`. That's just my two cents.

Your point about not having a vestigial column hanging around to trip us up -- let's drop it yesterday! @patrickm02L do you think we can get this hopefully snackoo like issue queued up?

PR #1294, opened 2023-09-22
12516 add user email export button

tillyw:
## Associated issues

Closes [#12516](https://github.com/cityofaustin/atd-data-tech/issues/12516) which allows one to download all the emails in the VZE Users tab.

## Testing
**URL to test:** 
https://deploy-preview-1294--atd-vze-staging.netlify.app/#/users

**Steps to test:**
- go to the users page and click the 'copy user emails' button at the top right of the table
- you should see a loading spinner while fetching the user list from the the auth0 api
- this may take a few seconds because we need to make multiple asynchronous api requests
- once that is done, the button should reappear and a success popover should display for two seconds and then disappear
- open a text doc and paste the user list from your clipboard!
- note: if you click the button again the previously fetched data will be copied to your clipboard. since fetching the data is somewhat resource intensive, it will only refetch the data if the number of users somehow changes

---
#### Ship list
- [ ] Code reviewed


patrickm02L (approved):
Super nice! Really like the loading action and hover notification when it finishes. Was able to the paste the emails into my notepad perfectly. 🚢

atd-vze/src/views/Users/Users.js
```diff
@@ -83,8 +87,57 @@ const Users = () => {
...
+    // then query all users
+    !!copyUserEmailsClicked &&
+      !!totalUsers &&
+      // only do this if the user hasn't already fetched data. if they have, skip this step
+      // and copy what is in state to the clipboard
+      allUsers?.length !== totalUsers &&
+      getAllUsers();
+    // once we have all users, copy the emails to the clipboard
+    !!allUsers.length && !!copyUserEmailsClicked && getUserEmails();
+  });
```
mddilley, line 101: I was expecting to see a dependency array in this `useEffect` but I see that your boolean conditions isolate how often this code fires so I think this is a-okay. 👍

atd-vze/src/views/Users/Users.js
```diff
@@ -83,8 +87,57 @@ const Users = () => {
       });
   }, [token, page, perPage]);
 
+  useEffect(() => {
```
mddilley, line 90: It would be cool to explore wrapping the new state and useEffect in a custom hook to keep the new functionality in a logical chunk, but your documentation about what is going on here makes it straightforward to see what is happening! 🙏

mddilley (approved):
Test steps check out, and I love the feedback after clicking the button - makes it easy to know when the addresses are ready to paste somewhere else. 📋 🚀 🚢

atd-vze/src/views/Users/Users.js
chiaberry: will netlify still build this without a dependency array?

mddilley: yes, i think the only warning that we would see that might affect the build is if there was a dependency array that was missing dependencies. Without the array, React just treats this as "run this on every render" and the boolean checks prevent any code beyond them from running.

chiaberry (approved):
Tested and got a list of emails! thank you for the clear code comments along each step

atd-vze/src/views/Users/Users.js
johnclary: I feel a bit more strongly that there needs to have a dependency array. if absolutely necessary, i would prefer to see the array and an eslint ignore rather than no array at all. with no array, the hook runs every render and so there's not really any point to having it at all. 

if putting the api call in the useEffect hook is feeling awkward, it might be worth trying to refactor so that the API call function is fired directly by the button click event. 

it also seems like the two chunks of code that call the `list_users` API could be combined if they do the same thing but for paginating.

roseeichelmann (approved):
I really like the loading wheel and the pop up message once the emails are copied. Also, your comments made the code easy to follow :sparkles:

roseeichelmann:
Does anyone else get a network error when they click on any user?

patrickm02L:
> Does anyone else get a network error when they click on any user?

I clicked on the User Name and the email links for multiple records and was able to get to the details pages. I did not get a network error.

frankhereford (approved):
This does work great, and I think that with a little refinement on the useEffect() hook, this is ready to go. Take me with a grain of salt, my react is rusty.

roseeichelmann:
Just pushed up some changes adding a dependency array to the useEffect hook and also refactored the list_users API call into its own function to combine some of the repetitive code.

patrickm02L (approved):
Worked like a charm! 💎  Thanks for making the adjustments!

atd-vze/src/views/Users/Users.js
```diff
@@ -66,28 +68,82 @@ const Users = () => {
...
   useEffect(() => {
+    getCurrentPageUsers(currentPage, perPage).then(res => {
+      setUserList(res.data.users);
+      setTotalUsers(res.data.total);
+    });
+    // eslint-disable-next-line
+  }, [token, currentPage, perPage]);
+
+  // Handles the copy user emails button click
+  useEffect(() => {
```
mddilley, line 86: I was looking at this `useEffect` with fresh eyes, and I'm seeing what @johnclary was suggesting earlier about tying these events to the click handler of the copy button. What do you think about something like below? I think that `getAllUsers` could also be moved outside of the component definition too.

The `isFetchingAllUsers` state is total pseudocode so I'm not sure if that is an addition or maybe `copyUserEmailsClicked` could fit in here to tracking loading.

```js
 const handleCopyUserEmails = async () => {
   if(allUsers === null) {
       setIsFetchingAllUsers(true);
       const users = await getAllUsers(perPage, pageCount);
       setAllUsers(users);
    }
    getUserEmails(users);
    setIsFetchingAllUsers(false);
  };
```
The hard part of this is that the button click event depends on the initial data fetch for the total number of user records to iterate in `getAllUsers`. I think that could be addressed by disabling this button until the first page of users is returned.

mddilley (commented):
Looks great - love the refactor 🙌 and dropped a comment to maybe take it a little further. Been thinking a lot of `useEffect`, and this is a great one to think about when to use it.

atd-vze/src/views/Users/Users.js
johnclary: nice—i really like this example! letting the button click trigger the fetching allows the state logic to be a lot tighter. this looks like the way to go to me.

frankhereford (approved):
This works well - good UX and it's ✅. Thank you Tilly & Rose.

I do also like the idea that John proposed and that Mike prototyped for the benefits they bring forward in terms of maintainability and readability.

mddilley:
@roseeichelmann Thanks for pushing more code! I have a couple suggestions to get this over the line.
1. I think the show/hide toggle could use its own state apart from `isFetchingAllUsers` which could then focus solely on the spinner
2. Moving `handleCopyUserEmails` back into the component could decrease the # of params fed to functions that it calls

Thanks for being patient with me being a bit scattered on this today - i know that point 2 is backtracking on my part. Was going to drop comments but I think chat or pairing tomorrow would get us there faster.

roseeichelmann:
@mddilley thanks for those suggestions, i just pushed up some updates! im pretty sure by show/hide toggle you meant the popover? If so I added a new state separate state `isPopoverOpen` to keep track of that so now `isFetchingAllUsers` just keeps track of the spinner.

chiaberry (approved):
Thanks for working through the refactor requests -- tested, got my list of users.

patrickm02L (approved):
Worked like a charm! 💎  Thanks for making the adjustments!

atd-vze/src/views/Users/Users.js
```diff
@@ -11,15 +11,62 @@ import {
...
+// Fetch an array of all the users
+async function getAllUsers(pageCount, perPage, totalUsers, setAllUsers, token) {
+  let page = 0;
+  let users = [];
+  // query every page of users until we have fetched them all
+  while (page <= pageCount) {
+    const res = await getCurrentPageUsers(page, perPage, token);
+    res.data.users.forEach(user => {
+      // make sure the user is not blocked/inactive, then push to an array
+      if (user.status === false || user.status === undefined) {
```
johnclary, line 45: just to double-check: `status === false || status ==== undefined` means the user is active?

atd-vze/src/views/Users/Users.js
```diff
@@ -11,15 +11,62 @@ import {
...
+        users.push(user);
+      }
+    });
+    page++;
+  }
+  // once we have fetched all users, update the users array in state
+  if (users.length === totalUsers) {
+    setAllUsers(users);
+    return users;
+  }
```
johnclary, line 55: and if `users.length !== totalUsers` that would mean something weird happened with the Auth0 data fetching? from a UI perspective, would that just cause the loading spinner keep spinning forever? if this is an extreme edge case i could live with that, but at least log a console error?

johnclary (approved):
i have not tested this code! 

but it looks very great ✨

i have asked a few questions and a made a minor suggestion.

thank you Rose and Tilly both for standing up this feature!!!

atd-vze/src/views/Users/Users.js
```diff
@@ -66,30 +113,52 @@ const Users = () => {
...
-      .then(res => {
-        setUserList(res.data.users);
-        setTotalUsers(res.data.total);
-      });
-  }, [token, page, perPage]);
+    getCurrentPageUsers(currentPage, perPage, token).then(res => {
+      setUserList(res.data.users);
+      setTotalUsers(res.data.total);
+    });
+    // eslint-disable-next-line
```
frankhereford, line 128: Neither here nor there, but I really like development experience of having a unforgiving linter and then disabling it on a per-line basis as needed. This is great to see in here.

frankhereford (approved):
Works great, and looks great - thank you!

atd-vze/src/views/Users/Users.js
chiaberry: I was concerned about this, because if inactive users are in the list that is returned (and counted as part of total users) but then not added to the email list, then this wont work out. 

When I was testing I checked to see if any inactive users were returned in the users request, and they are not. every user is returned with an undefined status.

chiaberry: Now that I'm thinking more about it though, I can see how even though this works now, all it would take would be someone to be in the list of users that is inactive and this feature breaks.

atd-vze/src/views/Users/Users.js
roseeichelmann: Yeah it sounds counterintuitive, but this definitely seems to be working as expected in terms of copying all 107 active users to clipboard. I just tested some things and changed this to an if else statement to capture users that would fall under "blocked/inactive" and we dont have any. Fwiw i did not write this part of the code

atd-vze/src/views/Users/Users.js
roseeichelmann: good catch y'all. there are multiple places in the code that check if allUsers = totalUsers, which like chia said is gonna cause issues if there is ever an inactive user in totalUsers. thinking thru a solution now

roseeichelmann: Okay im gonna just filter out inactive users in the `getUserEmails` function instead of filtering them out of `allUsers` state in `getAllUsers`

roseeichelmann:
I removed mentions of `user.status` in the feature and also added a `console.error` in the rare case that `users.length` and `totalUsers` doesn't match up. If that happens the spinner won't spin forever, it will stop but "user emails copied" popover won't happen and we get an error message in the console.

[Screencast from 11-02-2023 11:18:07 AM.webm](https://github.com/cityofaustin/atd-vz-data/assets/67130963/7f499a9c-4073-4aea-9745-31bc88bf4bf5)

chiaberry:
I like that solution! no success message, but no eternal spinner either

PR #1296, opened 2023-09-25
Update VZ branding for the VZV

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/13678

## Testing
**URL to test:** 

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed


chiaberry (approved):
tested and resized the screen a few times. thanks for the update

atd-vzv/src/views/nav/Header.js
```diff
@@ -15,74 +15,73 @@ const Header = () => {
   const isSummaryView = currentPath === "/";
 
   const StyledNavbar = styled.div`
-  .header-navbar {
```
roseeichelmann, line 18: most of this is prettier formatting

atd-vzv/src/views/nav/Header.js
```diff
@@ -119,10 +118,11 @@ const Header = () => {
                 Vision Zero -- Help Austin reach zero traffic deaths
               </h1>
               <img
-                className="vz-logo pl-lg-3"
+                className="vz-logo"
                 // Need to adjust location of public folder to account for /viewer/ basepath
                 src={process.env.PUBLIC_URL + "/vz_logo.svg"}
                 alt="Vision Zero Austin Logo"
+                width={300}
```
mddilley, line 125: I noticed a small regression on mobile. I think that the logo could get a bit smaller on mobile to avoid overlap with the hamburger on mobile screens. I used iPhone 12 size in Chrome for the screenshots.

Before 
Screenshot 2023-09-25 at 4 05 59 PM

After
Screenshot 2023-09-25 at 4 06 14 PM

mddilley (commented):
Yay! New branding. I added a comment about the mobile width of the new logo and wanted to get your thoughts.

atd-vzv/src/views/nav/Header.js
roseeichelmann: How does it look now? Had to futz around and push up a few diff changes to test the deploy preview on mobile :p

atd-vzv/src/views/nav/SideDrawerContent.js
```diff
@@ -21,9 +21,12 @@ const SideDrawerContent = ({ type }) => {
     @media only screen and (max-width: ${responsive.bootstrapMedium}px) {
       display: none;
     }
-
     .vz-logo {
-      height: ${responsive.headerButtonHeight + responsive.headerLogoOffset}px;
```
roseeichelmann, line 26: because of the diff logo dimensions these parameters dont work anymore as a good measure for height

atd-vzv/src/views/nav/Header.js
mddilley: I think that it looks great! ✨

atd-vzv/src/views/nav/SideDrawerContent.js
mddilley: cool - those old values look specific to the dimensions of the old logo so I think adding some that are specific to the new one makes sense!

mddilley (approved):
Looks awesome - thanks for updating to the new branding! 🙏 🚢

frankhereford (commented):
Thanks for updating this branding Rose. I think your changes to the code look great.

Do you think that the antialiasing issues with the new logo are worth working on? See below. It's not ideal, but passable, on desktop, but I think that it's degraded on mobile. I'm surprised by this, but perhaps SVG isn't the best format for the image, because it's giving up all control to the renderer about how to handle the crisp edges of the logo. This may be a case where a lossless PNG that targets the displayed size may give some more control over it. 

I'm curious what you think - thanks!

image

frankhereford:
@roseeichelmann, it's so much better!! I don't think I can put an approval on this from mobile, but it's good to go. Thanks!

patrickm02L (approved):
Good to go!

PR #1299, opened 2023-09-27
CI workflow for applying migrations and metadata

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/13294

This update adds a GH action to apply Hasura migrations and metadata when merging to `master` or `production` branches. Mostly copied from the [Moped workflow](https://github.com/cityofaustin/atd-moped/blob/main/.github/workflows/atd_moped_database.yml) but reworked since we need to get the Hasura secrets out of the `ZAPPA_SETTINGS` secret. I went with [jq](https://jqlang.github.io/jq/) to handle that since we've used it in the past.

Having the general shape of `ZAPPA_SETTINGS` could be helpful for reviewing what the bash script is doing.
```json
{
    "production": {
        "aws_environment_variables": {
            "HASURA_ENDPOINT": "https://vzd.austinmobility.io/v1/graphql",
            "HASURA_ADMIN_SECRET": "thisisaprodsecret"
        },
    },
    "staging": {
        "aws_environment_variables": {
            "HASURA_ENDPOINT": "https://vzd-staging.austinmobility.io/v1/graphql",
            "HASURA_ADMIN_SECRET": "thisisastagingsecret"
        },
    }
}
```

## Testing
**URL to test:** 

**Steps to test:**
1. Read the code and updated readme to see if it makes sense. 🙏
2. Double check the flags on the CLI commands that intend to keep any user interactions from needing to happen?
- [hasura migrate apply](https://hasura.io/docs/latest/hasura-cli/commands/hasura_migrate_apply/)
- [hasura metadata apply](https://hasura.io/docs/latest/hasura-cli/commands/hasura_metadata_apply/)
3. Other than that, this will be another one that we can merge, test, and patch if needed.
---
#### Ship list
- [ ] ~Run migration steps in readme in /atd-vzd/ if needed~
- [x] Code reviewed


.github/pull_request_template.md
```diff
@@ -8,6 +8,6 @@
 
 ---
 #### Ship list
-- [ ] Run migration steps in readme in /atd-vzd/ if needed
+- [ ] Check migrations to make sure they work with latest in master branch
```
mddilley, line 11: Wording? Basically, I'm trying to say that we need to check that the migrations are in the correct order. Just like we do in Moped.

.github/workflows/migration-helper.sh
```diff
@@ -0,0 +1,51 @@
...
+function export_hasura_env_vars() {
+    # Use jq to get Hasura endpoint and admin secret from ZAPPA_SETTINGS
+    export HASURA_GRAPHQL_ENDPOINT=$(echo ${ZAPPA_SETTINGS} | \
+    jq -r ".\"${WORKING_STAGE}\".aws_environment_variables.HASURA_ENDPOINT")
+    
+    export HASURA_GRAPHQL_ADMIN_SECRET=$(echo ${ZAPPA_SETTINGS} | \
+    jq -r ".\"${WORKING_STAGE}\".aws_environment_variables.HASURA_ADMIN_SECRET")
+    
+    # Remove /v1/graphql from the end of the graphql endpoint for Hasura CLI commands
+    export HASURA_GRAPHQL_ENDPOINT=$(basename "${HASURA_ENDPOINT%/v1/graphql}")
```
mddilley, line 27: Is this gross? I was hesitant to add yet another secret into `ZAPPA_SETTINGS` or GitHub Actions secrets. The problem comes down to Hasura CLI not needing the full graphql endpoint since it doesn't interface with GraphQL. I could see this catching us off guard if we ever upgraded to a v2 endpoint or something.

.github/pull_request_template.md
chiaberry: this works but how about
Check migrations for any conflicts with latest migrations in master branch

mddilley: 🙏 Yes, i like that a lot better! I'm updating.

.github/workflows/migration-helper.sh
chiaberry: I don't dislike this, but I can see your point about it maybe catching us off guard at some point. I do agree with you about not wanting to add another secret to the zappa settings. I'm fine with this!

chiaberry (approved):
test as we go!

.github/workflows/migration-helper.sh
roseeichelmann: This doesn't really bother me!

patrickm02L (approved):
🚢

.github/workflows/apply_db_migrations_and_metadata.yml
```diff
@@ -0,0 +1,44 @@
...
+      - production
+    paths:
+      - "atd-vzd/**"
+      - ".github/workflows/apply_db_migrations_and_metadata.yml" 
+      - ".github/workflows/migration-helper.sh"
+
+jobs:
+  apply:
+    name: Apply Migrations
+    runs-on: ubuntu-20.04
```
frankhereford, line 19: Do you think it would work to run this on 22.04? It's the newer LTS and will stay updated until '27, a couple of years longer than 20.04.

.github/workflows/migration-helper.sh
frankhereford: +1, I think this good. Great question though.

frankhereford (approved):
I think updating the runner would be cool, but this looks ready to ship either way. Thanks Mike!

.github/workflows/apply_db_migrations_and_metadata.yml
mddilley: yeah, let's go for it! Makes me think that we should look into updating this for other workflows running 20.04 too.

.github/pull_request_template.md
```diff
@@ -8,6 +8,6 @@
 
 ---
 #### Ship list
-- [ ] Run migration steps in readme in /atd-vzd/ if needed
+- [ ] Check migrations for any conflicts with latest migrations in master branch
```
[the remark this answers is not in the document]
mddilley: To me, that means merging master branch and testing the sequence of migrations to make sure that they don't conflict with each other. Then, updating the new migration's version (timestamp) if needed. I appreciate y'all helping with this wording! 🙏 Definitely open to changing it more

PR #1302, opened 2023-10-02
HAProxy for VZ Read Replica

frankhereford:
## Context

This PR is in pursuit of completing https://github.com/cityofaustin/atd-data-tech/issues/11895.

This PR includes support for setting the backend server via an environment variable instead of hard coding our RDS instance hostname into the `haproxy.cfg` file.

## Testing

⚠️ This is already running as of Oct 2, 2023 @ 9 PM central. If you're able to to connect to the read replica, then 👍 

From the RDS bastion:
* Checkout this branch and get into `atd-toolbox/read_replica_proxy/`
* `docker compose build`
* `docker compose up`
* You should then be able to connect to the read replica by connecting to the RDS bastion on `5432`. 

---
#### Ship list
- [ ] Code reviewed


atd-toolbox/read_replica_proxy/entrypoint.sh
```diff
@@ -0,0 +1,2 @@
+envsubst < /usr/local/etc/haproxy/haproxy.cfg.template > /usr/local/etc/haproxy/haproxy.cfg
```
frankhereford, line 1: This takes the environment variables of the docker container once it starts up and substitutes them into the `haproxy.cfg.template` file. The result is what's used when we fire up `haproxy` on the next line.

atd-toolbox/read_replica_proxy/haproxy.cfg.template
```diff
@@ -0,0 +1,16 @@
+global
+  log stdout format raw local0
+
+defaults
+  log global
+  mode tcp
+  timeout connect 5000ms
+  timeout client 50000ms
+  timeout server 50000ms
+
```
frankhereford, line 10: Boilerplate, setting up logging and timeouts.

atd-toolbox/read_replica_proxy/haproxy.cfg.template
```diff
@@ -0,0 +1,16 @@
...
+frontend postgresql_front
+  bind 0.0.0.0:5432
+  default_backend postgresql_back
```
frankhereford, line 13: This defines what we listen for, which is incoming connections on port `5432` on all addresses. All addresses are used because this is happening inside a docker container which only has `5432` forwarded to it from all NICs on the host.

atd-toolbox/read_replica_proxy/haproxy.cfg.template
```diff
@@ -0,0 +1,16 @@
...
+
+backend postgresql_back
+  server postgresql_server ${DATABASE_SERVER}:5432
```
frankhereford, line 16: This defines, once substituted with the environment variable, what backend service is used to fulfill requests that come in on the front end.

atd-toolbox/read_replica_proxy/docker-compose.yaml
```diff
@@ -0,0 +1,11 @@
+version: '3'
+services:
+  haproxy:
+    build: .
+    restart: always
```
frankhereford, line 5: This is what makes this service persist after reboot. It's our system's job to get the docker service running, which then takes care of this for us.

frankhereford:
This is running in production proxying incoming requests to the read-replica DNS entry back to the subordinate, read-only Aurora instance as of Oct 2.

chiaberry (approved):
thank you for the documentation! i can connect to the read replica

atd-toolbox/read_replica_proxy/env_template
```diff
@@ -0,0 +1,2 @@
+# Copy this file to `env` and fill in the value
+DATABASE_SERVER=<hostname of reader instance>
```
mddilley, line 2: I was doing that googling that I owed this work this morning, and I tripped over a SO comment that cautioned against using the instance endpoints and instead using the cluster endpoints. Do you think that we should adjust this? I'm still learning about this but wanted to share this. https://stackoverflow.com/a/52996681

mddilley (commented):
I checked out the code, and it makes sense to me. I also logged on the bastion host to check out the config and see how this is working. I had one question after reading a bit about proxies and Aurora.

atd-toolbox/read_replica_proxy/env_template
frankhereford: This worked great. Thank you again Mike! 🙌

frankhereford:
@mddilley wrote:

> Can't remember and not familiar with haproxy logging but did we have to set up log rotation for the haproxy logs coming out of the Airflow stack?

I'm not sure why I can't inline reply to this, but no big. I am not actually making any attempt to capture the logs for this, except in a ephemeral way. They are trapped while the container is running by the docker service, but as soon as the container shuts down and gets pruned, they are gone. 

I don't see any real value in it, but I can if we want. I'm confident enough in the combination of strong passwords on our user accounts + the network ACL provided by the security group that I don't feel the need to keep these. I'm open to reconsidering this and if we're going to get more into log analyzation, we should talk about some sort of log aggregation server.

atd-toolbox/read_replica_proxy/env_template
mddilley: Awesome, give all credit to the person who replied to that question on SO. They are the real MVP!!! 🏆

patrickm02L (approved):
Send it 🥏

PR #1303, opened 2023-10-05
Start a fresh migration and metadata for the LDM

frankhereford:
## Associated issues

Roughly associated to https://github.com/cityofaustin/atd-data-tech/issues/13902. 

## Testing

This is hard to test, as it's really the representation of the `graphql-engine` configuration after starting everything fresh. 

To test, start up a `hasura console` pointing at the LDM staging instance and you can review the configuration of the `graphql-engine` endpoint. Please let me know if I can offer you an environment file for your `atd-vzd` directory.

## Next Steps

@mddilley noted below that the `vision-zero` orchestration can be improved to allow us to replicate the LDM staging DB down into our local stack. I'm going to spin off [an issue](https://github.com/cityofaustin/atd-data-tech/issues/14181) for that.

---
#### Ship list
- [ ] Code reviewed


frankhereford:
Mike, I think this is the PR we need to record having setup a fresh graphql-engine migration history and to have tracked `public`, `vz_facts` and `cris_facts`. Am I on the right path? If you approve, I'll open this up to the wider crew. Thanks!

atd-vzd/metadata/databases/default/tables/tables.yaml
mddilley, line 1: I noticed that none of the tables are tracked in the layers. Should we do this right from the start? I think we should but am open to anything that I'm missing. Then we can adjust as needed and see the changes track in the metadata over time.

Screenshot 2023-10-05 at 2 23 41 PM

atd-vzd/config.yaml
```diff
@@ -1,6 +1,5 @@
 version: 3
-endpoint: http://localhost:8084
-admin_secret: thisisasecret
```
mddilley, line 3: Thanks for removing this! We really need to remove this secret that I added to the docker compose file a while back. It prevents VZE from connecting to the Hasura instance in the non-LDM branches.

mddilley (commented):
Thanks for doing this Frank! This is looking great. 🙌 I dropped a couple of questions, and I also wanted see if you think making some small tweaks to make the `./vision-zero replicate-db` command target the LDM staging instance. Maybe renaming the env vars in the LDM branches so the we can bounce between LDM and non-LDM branches. Something like `LDM_RR_USERNAME`?

atd-vzd/config.yaml
frankhereford: Oh cool, that was all Hasura's work during the `init` command process.

atd-vzd/config.yaml
```diff
@@ -1,6 +1,5 @@
...
+endpoint: http://localhost:8080
```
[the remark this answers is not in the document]
mddilley: Just saw your comment below so this was also probably Hasura init doing it thing. 👍

atd-vzd/metadata/databases/default/tables/tables.yaml
frankhereford: Great question Mike. I thought about this for a little bit when I was starting to track tables. Like you, I'm also 100% open to discussion and don't have any monstrously strong feelings about what our table/view tracking policy should be at this time. 

That said, my rational here was to only track and expose tables which are going to be queried by the VZE to make it work. The great bulk of data will be read out of `public` and written into `vz_facts`. `cris_facts` would be useful to enable showing both the current CRIS value vs the current VZ value, which is a feature we've spoken about a few times.

Do you like this approach, or is there another approach you'd like to pursue?

Either way is 💯 good with me, and I am eager to hear your thoughts and the group's as well. Thanks!

atd-vzd/config.yaml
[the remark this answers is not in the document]
frankhereford: Thanks for catching this. This does need to move back to `8084` like you suggested so our local environments keep working without changing the env file and also to not clash with local moped dev. 🙏

frankhereford:
Mike, thank you for giving this a first pass. I think I addressed your other questions, so I'm going to open this up to the rest of the crew.

@mddilley also asked:

> Thanks for doing this Frank! This is looking great. 🙌 I dropped a couple of questions, and I also wanted see if you think making some small tweaks to make the ./vision-zero replicate-db command target the LDM staging instance. Maybe renaming the env vars in the LDM branches so the we can bounce between LDM and non-LDM branches. Something like LDM_RR_USERNAME?

Great question and it's an awesome idea. I'm 100% on board. I would like to spin it off into its own issue and PR, and I'll write that up now. Thank you Mike - cool idea. 🚀

atd-vzd/metadata/databases/default/tables/tables.yaml
mddilley: Yeah, I agree. I think that it makes sense to track any table that could have columns added to it in the future. That way we could generate diffs in the metadata by editing in the Hasura console like we do in Moped. Definitely open to discussion, and I'm sure that we will touch on it in the ldm standup tomorrow. 🙏

frankhereford:
This is now tracking all trackable relations in the LDM staging database.

roseeichelmann (approved):
yay im excited to be able to start writing migrations to this ldm staging instance! really think this well help speed up development

patrickm02L (approved):
So fresh and so clean 🍰

frankhereford:
Thanks everyone!

mddilley:
@frankhereford Maybe I missed something but I pulled the latest commits and was browsing, and I'm not seeing anything tracked in the Hasura console.

mddilley (commented):
Finally worked out the issues on my machine, and this works great! ✨ Thanks for bearing with my discussion about Hasura tracking! Looks amazing. 🚢

PR #1308, opened 2023-10-12
Add ems__incidents table schema and trigger to LDM

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/13707
https://github.com/cityofaustin/atd-data-tech/issues/14141

This PR adds migrations for the `ems__incidents` table schema (no data) and the one trigger set on it. I'm using the same patterns that Frank started in https://github.com/cityofaustin/atd-vz-data/pull/1305, and the idea is to apply these migrations and backfill the data on LDM staging when this code merges. We'll keep building that instance up while building up the LDM transformation steps (transform script + migrations + backfills).

## Testing
**URL to test:** 
Local

**Steps to test:**
1. Make sure that your `.env` is updated with the `LDM_` variables shown in the `env_template`. Check the 1Pass secrets called `RDS Instance: LDM staging` for the values. `LDM_DATABASE` is `atd_vz_data`.
2. Replicate the LDM staging database locally by running:
```bash
$ ./vision-zero -l replicate-db
```
3. From the `atd-vzd` folder run:
```bash
$ hasura migrate apply --envfile .env.local
$ hasura metadata apply --envfile .env.local
```
4. To backfill the EMS incidents data from the VZ read replica to your local LDM database, run:
```bash
export RR_PASSWORD=<your VZ read replica password>
export RR_USER=<your VZ read replica username>

docker exec -i \
-e PGHOST=db-rr.vision-zero.austinmobility.io \
-e PGDATABASE=atd_vz_data \
-e PGPASSWORD=$RR_PASSWORD \
-e PGUSER=$RR_USER \
visionzero-postgis \
pg_dump -t ems__incidents --data-only | \
docker exec -i \
-e PGHOST=localhost \
-e PGDATABASE=atd_vz_data \
-e PGPASSWORD=visionzero \
-e PGUSER=visionzero \
visionzero-postgis \
psql --set ON_ERROR_STOP=on;
``` 
5. Through your database explorer or the local Hasura console, look for the `public.ems__incidents` table
6. You should see the columns set in the `public.ems_incidents_trigger()` function populated with data (if there are values to generate from) which happened on insert.

---
#### Ship list
- [ ] Code reviewed


atd-vzd/migrations/default/1697144772669_add_ems_trigger_function/up.sql
```diff
@@ -0,0 +1,26 @@
...
+    austin_full_purpose = (
+      select ST_Contains(jurisdiction.geometry, incidents.geometry)
+      from ems__incidents incidents
+      left join atd_jurisdictions jurisdiction on (jurisdiction.jurisdiction_label = 'AUSTIN FULL PURPOSE')
+      where incidents.id = new.id),
+    location_id = (
+      select locations.location_id
+      from ems__incidents incidents
+      join atd_txdot_locations locations on (locations.location_group = 1 and incidents.geometry && locations.shape and ST_Contains(locations.shape, incidents.geometry))
+      where incidents.id = new.id),
```
johnclary, line 16: you might want to add `LIMIT 1` in case there were some scenario (crufty polygons?) where the incident intersected with multiple location polygons?

atd-vzd/migrations/default/1697144772669_add_ems_trigger_function/up.sql
```diff
@@ -0,0 +1,26 @@
...
+    latitude = ST_Y(ems__incidents.geometry),
+    longitude = ST_X(ems__incidents.geometry),
```
johnclary, line 18: i'm sure this merely is carried forward from current production—but i'm curious why it is necessary to have these separate lat/lon columns in addition to what's stored in the geometry

atd-vzd/migrations/default/1697144772669_add_ems_trigger_function/up.sql
```diff
@@ -0,0 +1,26 @@
...
+    apd_incident_number_1 = ems__incidents.apd_incident_numbers[1],
+    apd_incident_number_2 = ems__incidents.apd_incident_numbers[2],
+    mvc_form_date = date(ems__incidents.mvc_form_extrication_datetime),
+    mvc_form_time = ems__incidents.mvc_form_extrication_datetime::time
```
johnclary, line 22: also wondering if these columns could be managed by generated columns—or perhaps even better: as part of the ingest ETL for these records?

atd-vzd/migrations/default/1697145115428_add_ems_table/up.sql
```diff
@@ -0,0 +1,430 @@
...
+--
+
+COMMENT ON COLUMN public.ems__incidents.pcr_patient_acuity_final IS 'Final patient acuity determined by provider';
+
+
+--
+-- Name: COLUMN ems__incidents.apd_incident_numbers; Type: COMMENT; Schema: public; Owner: vze
+--
+
+COMMENT ON COLUMN public.ems__incidents.apd_incident_numbers IS 'A comma delimitted list of incident numbers for APD incidents that are linked to the EMS incident. This field can be used to determin if there is an associated APD incident.';
```
johnclary, line 379: this is such good metadata 😭

atd-vzd/migrations/default/1697144772669_add_ems_trigger_function/up.sql
```diff
@@ -0,0 +1,26 @@
+CREATE OR REPLACE FUNCTION public.ems_incidents_trigger()
+ RETURNS trigger
+ LANGUAGE plpgsql
+AS $function$
+BEGIN
+  update ems__incidents set
+    austin_full_purpose = (
+      select ST_Contains(jurisdiction.geometry, incidents.geometry)
```
johnclary, line 8: is there a concern that geometry would ever be null/missing?

johnclary (commented):
@mddilley i threw some comments on here to try to be helpful. i will not approve or request changes because i'm not close enough to this work. 🚢🚢🚢🚢🚢🚢

atd-vzd/migrations/default/1697145115428_add_ems_table/up.sql
mddilley: agreed! i want to thank whoever wrote this 🙏

atd-vzd/migrations/default/1697144772669_add_ems_trigger_function/up.sql
mddilley: that is a good check. I think that since we are limiting to `location_group = 1` here takes care of any overlap, but it is hard to tell from the code what is going on. Group 1 (level 1-4) can overlap with group 2 (level 5 or what we are calling major roads now).

atd-vzd/migrations/default/1697144772669_add_ems_trigger_function/up.sql
mddilley: You are right on about this coming straight from prod. I'll have to look at the data source for this, but I also think that it feels denormalized!

atd-vzd/migrations/default/1697144772669_add_ems_trigger_function/up.sql
mddilley: I checked this table in production and there are a handful of nulls for geometry. Ends up with a nulled `austin_full_purpose` column. That should be an easy fix.

atd-vzd/migrations/default/1697144772669_add_ems_trigger_function/up.sql
mddilley: Edit from LDM discussion:
- spinoff?

atd-vzd/migrations/default/1697144772669_add_ems_trigger_function/up.sql
mddilley: On second thought, I noticed that this is comes from using `ST_SetSRID` and `ST_Point` in the ETL. Maybe we should have some upstream changes to null these out there. Calling @frankhereford for phone a friend ☎️

atd-vzd/migrations/default/1697144772669_add_ems_trigger_function/up.sql
mddilley: @patrickm02L I made the update to include limit, but I think that we can spin off an issue to address the idea of making this always return the same location polygon to associate with the EMS incident. Frank had the idea of using the centroid or something similar.

atd-vzd/migrations/default/1697144772669_add_ems_trigger_function/up.sql
mddilley: @patrickm02L I created https://github.com/cityofaustin/atd-data-tech/issues/14316 to capture this and the comment below.

atd-vzd/migrations/default/1697144772669_add_ems_trigger_function/up.sql
frankhereford: I think this function as written will work fine, because if `ST_Contains()` gets a `null` argument the result is null. Here's a test fixture:

```
SELECT ST_Contains(
    ST_GeomFromText('POLYGON((0 0, 0 10, 10 10, 10 0, 0 0))'), 
    ST_GeomFromText('POINT(5 5)')
) AS result;

SELECT ST_Contains(
    ST_GeomFromText('POLYGON((0 0, 0 10, 10 10, 10 0, 0 0))'), 
    null
) AS result;

SELECT ST_Contains(
    null, 
    ST_GeomFromText('POINT(5 5)')
) AS result;
```

To Mike's point about how these bogus `0,0` points have snuck into the DB - I do think we should fix that. `0,0` is a valid point, which is why it's slipping through the GIS functions..., but we should trap it as invalid because it's just not one we're going to run into here in Austin when using SRID 4326, 3857, or 2277. I think that a new issue though, not one we should try to pick off in this work.

frankhereford (approved):
It's another winner here. This worked great. 

Thanks to @johnclary for asking great questions about the fundamentals of the table's structure, and I would be eager to see us address those normalization issues. 

If I didn't fully answer the question you were asking Mike, please let me know and I'll give it another whirl. Thank you!

roseeichelmann (approved):
Worked great, thanks Mike!

PR #1311, opened 2023-10-18
LDM Persons Table Triggers

frankhereford:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/13703

## Backfill commands

This command populates the fatalities table from the production read replica.

This won't work until we have a complete set of crashes instead of the subset we have now in the ldm staging database.

```

docker exec -i \
-e PGHOST=db-rr.vision-zero.austinmobility.io \
-e PGDATABASE=atd_vz_data \
-e PGPASSWORD=$RR_PASSWORD \
-e PGUSER=$RR_USER \
visionzero-postgis \
pg_dump -t public.fatalities --data-only | \
docker exec -i \
-e PGHOST=localhost \
-e PGDATABASE=atd_vz_data \
-e PGPASSWORD=visionzero \
-e PGUSER=visionzero \
visionzero-postgis \
psql --set ON_ERROR_STOP=on;
```

## Testing

Testing is most easily done in your local database. 
* Get yourself the latest and greatest LDM data via `./vision_zero --ldm replicate-db`
* Step through the migrations as they are applied to see what you have here:

```bash
hasura migrate apply --database-name default  --up 1; hasura migrate --database-name default status;
```

  * `1697646059131_changing-keying-on-ldm-view`
    * Adjust the `public.atd_txdot_person` to only use the actual trio of fields that comprise the primary key of persons records. (`crash_id`, `unit_nbr`, & `prsn_nbr` and not `prsn_type_id` & `prsn_occpnt_pos_id`.)
  * `1697646059132_fatalities_table`
    * Create the fatalities table. We'll back fill this at merge. 
  * `1697646059133_fataility_insert_function`
    * Create the `fatality_insert_function`. Note that this doesn't need a if/else construct because it's going to be used in an on-insert trigger, and those only ever will happen that we care about when we insert into the `cris_facts` schema.
  * `1697646059134_persons_fatality_insert_trigger`
    * Bind the above function to insert on `cris_facts.atd_txdot_persons` table.
  * `1697646059135_rename-existing-trigger`
    * Rename an existing trigger to simply understanding and establishing the order in which they run.
  * `1697646059136_modify-ldm-maint-trigger`
    * Define trigger that is responsible for creating or removing the record in `vz_facts.atd_txdot_persons` when a record is created or removed in the `cris_facts` schema.
  * there is no `1697646059136` migration. I had something here and removed it.
  * `1697646059138_update_fatality_soft_delete_function`
    * Add the soft_delete hander for the fatalities table. This has a branch for when this is called from an edit on the VZ side and on the CRIS side.
  * `1697646059139_update_fatality_soft_delete_trigger`
    * Bind soft_delete handler to both the `vz_facts` and `cris_facts` `atd_txdot_persons` tables.
  * `1697646059140_add_person_audit_log_function`
    * Create the `audit_log` function for storing a backup of a person record when a change is made. This is the old change log tracking system.
  * `1697646059141_add_person_audit_log_trigger`
    * Bind the `audit_log` function to updates to the persons tables in the `vz_facts` and `cris_facts` schemata.

## Helpful commands
Step through migrations, backwards and forwards
```
hasura migrate apply --database-name default  --up 1; hasura migrate --database-name default status
hasura migrate apply --database-name default  --down 1; hasura migrate --database-name default status
```

---
#### Ship list
- [ ] Code reviewed


atd-vzd/migrations/default/1697646059132_fatalities_table/up.sql
```diff
@@ -0,0 +1,134 @@
...
+ALTER TABLE ONLY public.fatalities
+    ADD CONSTRAINT fatalities_primaryperson_id_key UNIQUE (primaryperson_id);
+
+
+--
+-- Name: fatalities fatalities_crash_id_fkey; Type: FK CONSTRAINT; Schema: public; Owner: vze
+--
+
+ALTER TABLE ONLY public.fatalities
+    ADD CONSTRAINT fatalities_crash_id_fkey FOREIGN KEY (crash_id) REFERENCES vz_facts.atd_txdot_crashes(crash_id) ON UPDATE CASCADE ON DELETE CASCADE;
```
frankhereford, line 98: This references the VZ half of this logical table group.

atd-vzd/migrations/default/1697646059131_changing-keying-on-ldm-view/up.sql
```diff
@@ -0,0 +1,54 @@
...
+    --COALESCE(vz_atd_txdot_person.person_id, cris_atd_txdot_person.person_id) AS person_id,
+    COALESCE(vz_atd_txdot_person.is_retired, cris_atd_txdot_person.is_retired) AS is_retired,
+    COALESCE(vz_atd_txdot_person.years_of_life_lost, cris_atd_txdot_person.years_of_life_lost) AS years_of_life_lost,
+    vz_atd_txdot_person.person_id as vz_person_id,
+    cris_atd_txdot_person.person_id as cris_person_id
+   FROM vz_facts.atd_txdot_person vz_atd_txdot_person
+     JOIN cris_facts.atd_txdot_person cris_atd_txdot_person ON 
+       vz_atd_txdot_person.crash_id = cris_atd_txdot_person.crash_id AND 
+       vz_atd_txdot_person.unit_nbr = cris_atd_txdot_person.unit_nbr AND 
+       vz_atd_txdot_person.prsn_nbr = cris_atd_txdot_person.prsn_nbr;
```
frankhereford, line 54: Join the tables on just what we know will be available in both tables, always. These are the actual fields that are used to uniquely identify a person.

mddilley (commented):
Thanks for working out these triggers! I dropped some notes and questions. I also read your comments on Rose's PR and agree that having consistent naming for these triggers would be a big level up!

atd-vzd/migrations/default/1697646059138_update_fatality_soft_delete_function/up.sql
```diff
@@ -0,0 +1,60 @@
+CREATE OR REPLACE FUNCTION public.update_fatality_soft_delete()
+ RETURNS trigger
+ LANGUAGE plpgsql
+AS $function$
+DECLARE
+    cris_id integer; -- used to hold vz id when triggered from cris_facts schema
+    injury_severity_id integer;
+    BEGIN
+    RAISE NOTICE 'Here we are.';
```
roseeichelmann, line 9: is this like a stray console log :smile: also TIL, i love console logging for debugging so this is a helpful tip!!

atd-vzd/migrations/default/1697646059138_update_fatality_soft_delete_function/up.sql
```diff
@@ -0,0 +1,60 @@
...
+        -- if updated table was primary person
+    	IF (TG_TABLE_NAME = 'atd_txdot_primaryperson') THEN
```
roseeichelmann, line 11: is the plan to at a later date add in the `TG_TABLE_SCHEMA = 'cris_facts'` or `TG_TABLE_SCHEMA = 'vz_facts'` parts of the if else, like you did with the persons table?

atd-vzd/migrations/default/1697646059138_update_fatality_soft_delete_function/up.sql
```diff
@@ -0,0 +1,60 @@
...
+            SELECT cris_person_id, prsn_injry_sev_id 
+            INTO cris_id, injury_severity_id 
+            FROM public.atd_txdot_person 
+            WHERE vz_person_id = NEW.person_id;
+            --RAISE NOTICE 'The value of cris_id is %', cris_id;
+            --RAISE NOTICE 'The value of prsn_injry_sev_id is %', injury_severity_id;
+
+            IF (NEW.prsn_injry_sev_id = 4) THEN
+                INSERT INTO fatalities (crash_id, person_id)
+                VALUES (NEW.crash_id, NEW.person_id)
```
roseeichelmann, line 48: would there ever be a situation that we have a person in the vz schema that doesnt exist in the cris schema? im thinking about for example when the vz team creates temp records? if so, it would be possible for there to be fatalities that don't have a `person_id` connecting to the cris schema, but only the vz schema.

the `cris_person_id` vs `vz_person_id` distinction made my head hurt, i kinda forgot that the two tables in the diff schemas would have diff ids :face_with_spiral_eyes:

roseeichelmann:
This trigger work is complicated, thanks for untangling all of these! I left some comments and here are some other thoughts I have:

- We talked in a [slack discussion](https://austininnovation.slack.com/archives/C05F2052S67/p1697664864221439?thread_ts=1697662816.360719&cid=C05F2052S67) recently about the fatalities table just being a view, instead of having both a table and a view. In the LDM, it would be a view that pulls from the views in the public schema (crash, persons, any other tables that make up the fatalities view). I think this would clear up some of the issues that come with the complexity of having the fatalities table being comprised of insert/update/delete triggers from both the vz and cris schemas?
- It seems like we need to have a discussion about what we want to happen when we get updates from cris that conflict with vz edits - this has been coming up a lot
- Also I've been noticing the role that temp records play coming up a lot causing questions/problems with the LDM, do we plan on handling temp records any differently in an LDM implementation than we currently do?

atd-vzd/migrations/default/1697646059139_update_fatality_soft_delete_trigger/up.sql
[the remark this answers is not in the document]
mddilley: Yup, my bad. I was grouping them together and this PR is for people only (not primary).

PR #1316, opened 2023-11-07
Invalidate CR3 metadata to cause them to be redownloaded when updated by CRIS

frankhereford:
## Associated issues

This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/14400

## Tag-along back-port

I've back ported a feature from the LDM that allows you to put a CRIS extract in a local directory and have it use that instead of pulling one from the SFTP server. It's super helpful when working on this script, so it felt like a good time to bring it over.

## Testing

* Spin up the local DB and get an export from CRIS. Let me know, and I can wormhole you one. Any old one will work. 
* Replicate production into your local DB
* Put the export in `/atd-vz-data/atd-etl/cris_import/development_extracts/`
* Comment in the `mess_with_incoming_records ...` line around 109 in `cris_import.py`. This will ensure every record is always an update from CRIS. You can try it without that line too.
* From `/atd-vz-data/atd-etl/cris_import` do `docker compose run cris-import`
* While this runs, capture a crash ID that is being updated. This doesn't have to be from the crashes table. For example, it can be the crash_id of a unit that gets an update. Look for something in the STDOUT stream that looks like `Executing update in atd_txdot_crashes for where public.atd_txdot_crashes.crash_id = 19842483` or anything that gives you a crash_id of an imported record from any table.
* Confirm that that record in the `atd_txdot_crashes` table has these fields set this way:
```
cr3_stored_flag = 'N'
cr3_file_metadata = null
cr3_ocr_extraction_date = null
```
* Here's some sql to make that query easier:
```
select crash_id, cr3_stored_flag, cr3_file_metadata, cr3_ocr_extraction_date
          from public.atd_txdot_crashes
          where crash_id = <crash_id>;
```

---

#### Ship list
- [ ] Code reviewed


atd-etl/cris_import/cris_import.py
```diff
@@ -81,7 +82,21 @@ def main():
...
+
+    zip_location = None
+    if not local_mode: # Production
+        zip_location = download_archives()
+    else: # Development. Put a zip in the development_extracts directory to use it.
+        zip_location = specify_extract_location()
+
+    if not zip_location:
+        return
+
```
frankhereford, line 99: Backport this functionality from LDM, it's super helpful for testing this sort of thing.

atd-etl/cris_import/cris_import.py
```diff
@@ -91,10 +106,12 @@ def main():
             pgloader_command_files = pgloader_csvs_into_database(schema_name)
             trimmed_token = remove_trailing_carriage_returns(pgloader_command_files)
             typed_token = align_db_typing(trimmed_token)
+            #typed_token = mess_with_incoming_records_to_ensure_updates(typed_token) # for testing PR 1316
```
frankhereford, line 109: This is to help test this PR. I'll remove this one line before merging.

This function will add a random string to a field in each of the tables we're importing. This ensures that every line that comes in from CRIS is an update.

atd-etl/cris_import/cris_import.py
```diff
@@ -197,6 +214,63 @@ def get_secrets():
     return onepasswordconnectsdk.load_dict(client, REQUIRED_SECRETS)
 
 
+def mess_with_incoming_records_to_ensure_updates(map_state):
```
frankhereford, line 217: This function is what ensures that every incoming record from CRIS is messed with a little to ensure that it's an update.

atd-etl/cris_import/cris_import.py
```diff
@@ -242,6 +316,59 @@ def download_extract_archives():
         return zip_tmpdir
 
 
+def specify_extract_location():
```
frankhereford, line 319: Backport functionality from LDM.

atd-etl/cris_import/cris_import.py
```diff
@@ -592,6 +719,10 @@ def align_records(map_state):
                 # To decide to UPDATE, we need to find a matching target record in the output table.
                 # This function returns that record as a token of existence or false if none is available
                 if util.fetch_target_record(pg, output_map, table, record_key_sql):
+                    
+                    # Cause the CR3s to be re-downloaded next time that ETL is run
+                    util.invalidate_cr3(pg, source["crash_id"])
```
frankhereford, line 724: If we're in this execution branch, we're in the "update" branch, so invalidate the CR3.

atd-etl/cris_import/lib/sql.py
```diff
@@ -21,6 +21,14 @@ def get_pgfutter_path():
...
     return None
 
+def invalidate_cr3(pg, crash_id):
+    invalidate_cr3_sql = f"""UPDATE atd_txdot_crashes 
+    SET cr3_stored_flag = 'N', cr3_file_metadata = null, cr3_ocr_extraction_date = null
+    WHERE crash_id = {crash_id}"""
+    cursor = pg.cursor(cursor_factory=psycopg2.extras.RealDictCursor)
+    cursor.execute(invalidate_cr3_sql)
+    pg.commit()
+
```
frankhereford, line 31: This snippet of SQL is what we've been using for a good while to cause CR3s to be re-downloaded.

frankhereford, line 31: This SQL snippet has been used to invalidate CR3s manually for a long time -- tried and true.

vision-zero
```diff
@@ -177,7 +177,7 @@ def replicateDb(args):
 
     drop_database_command = [
         "psql",
-        "--command=DROP DATABASE " + os.environ["POSTGRES_DB"] + "",
+        "--command=DROP DATABASE " + os.environ["POSTGRES_DB"] + " WITH(force)",
```
frankhereford, line 180: A final DX backport from the LDM.

atd-etl/cris_import/lib/sql.py
```diff
@@ -21,6 +21,14 @@ def get_pgfutter_path():
...
     return None
 
+def invalidate_cr3(pg, crash_id):
+    invalidate_cr3_sql = f"""UPDATE public.atd_txdot_crashes 
+    SET cr3_stored_flag = 'N', cr3_file_metadata = null, cr3_ocr_extraction_date = null
+    WHERE crash_id = {crash_id}"""
+    cursor = pg.cursor(cursor_factory=psycopg2.extras.RealDictCursor)
+    cursor.execute(invalidate_cr3_sql)
+    pg.commit()
+
```
frankhereford, line 31: Tried and true -- this is the snippet we've used to invalidate CR3s for ages when doing it manually.

atd-etl/cris_import/cris_import.py
mddilley: Cool to see this getting added here. I could see a tiny blurb about this in the readme being helpful.

atd-etl/cris_import/cris_import.py
mddilley: This was really helpful for testing this out. 🙏 What do you think about moving the function called here to another module? I can see it coming in handy again the future, but I initially thought it was part of the fix so I'm not sure if it should live with the rest of the import code.

atd-etl/cris_import/cris_import.py
mddilley, line 319: I think that there is some duplication here with code just before this line. `specify_extract_location` is defined twice, and I think that there is overlap in `download_extract_archives` and `download_archives`. Might be another duplicate?

mddilley (changes requested):
I tested this out with an extract that I have on hand, and I see that all the expected fields are reset. 🙌 I see some duplicate functions that need some addressing before ship, and I left a few comments too.

atd-etl/cris_import/cris_import.py
```diff
@@ -81,7 +82,21 @@ def main():
     SFTP_ENDPOINT_SSH_PRIVATE_KEY = secrets["sftp_endpoint_private_key"]
 
     # 🥩 & 🥔
-    zip_location = download_extract_archives()
+    #zip_location = download_extract_archives()
```
chiaberry, line 85: this line should be removed, right? and as Mike pointed out, download_extract_archives and download_archives overlap, and this is the only place I could find that referenced download_extract_archives, so I'm guessing thats the one that can be deleted?

atd-etl/cris_import/cris_import.py
chiaberry: I was going to ask if you were removing this line, if you would also remove the function, but I like Mike's idea about moving the function to another module for testing purposes.

atd-etl/cris_import/cris_import.py
```diff
@@ -242,6 +316,59 @@ def download_extract_archives():
...
+
+    zip_tmpdir = tempfile.mkdtemp()
+
+    for file_to_copy in zip_files:
+        shutil.copy(file_to_copy, zip_tmpdir)
+
+    return zip_tmpdir
+
+
+
```
chiaberry, line 332: my linter is saying there are two many lines between some of these functions

atd-etl/cris_import/cris_import.py
chiaberry: could you add this as a docstring or some comments on the function?

atd-etl/cris_import/cris_import.py
frankhereford: Yea, totally good call about adding a blurb. I put one in the readme in ee872c02d62e43329747fd7fa00d32bd16564d06.

atd-etl/cris_import/cris_import.py
frankhereford: Thank you Mike for catching this. These dups are the result of me keeping a slightly less refined version of this "pull zips from local directory" mechanism that I had floating in my checkout -- and getting some of those names jumbled as I brought it into a real branch. The dups are removed in 92efb2cc74dffe9c262e8f51e51937c59b50f864.

atd-etl/cris_import/cris_import.py
frankhereford: Thanks y'all for the great suggestions on this. I have pulled the testing function down into `lib/testing.py` with a little blurb in there about usage. This also required pulling a class I had tucked in the bottom of the main script out into a file, so that's what i'm up with with lib/sshkeytempdir.py.

atd-etl/cris_import/cris_import.py
frankhereford: I did add a comment with usage instructions in the file containing this function now -- thank you for that suggestion!

atd-etl/cris_import/cris_import.py
frankhereford: Great catch. I've linted the entire application with `black` now in 39cf8557641193901f79c75bf6bd34df3528e288.

mddilley (approved):
This works great! Thanks for adding the new modules and docs. I just used them to test the CR3 columns resets on record update again. 🚢 🚀 ✨

patrickm02L (approved):
Good to go! 💯

PR #1318, opened 2023-11-14
Create a new field peh_fl in the database for a Person Experiencing Homelessness

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/13561

This issue didn't have specs for what the UI interaction should be, but there really is no other place in the editor for editing the fields of the persons table that makes sense other than the Related Records section, so I put the functionality there.

## Testing
**URL to test:** 
Local!

**Steps to test:**
1. Checkout this branch, start up your local db and run the new migrations in this branch:
```bash
hasura migrate apply --envfile .env.local
hasura metadata apply --envfile .env.local
``` 
3. Load up the VZE pointing to your local db
4. Go to a crash, you should now see the "Experiencing Homelessness" field in the People section of the Related Records table.
5. Test editing the field to be True, False, and back to No Data (null in the db).

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vzd/migrations/default/1699990226463_alter_table_public_atd_txdot_person_add_column_peh_fl/up.sql
+++ b/atd-vzd/migrations/default/1699990226463_alter_table_public_atd_txdot_person_add_column_peh_fl/up.sql
@@ -1,0 +1,4 @@
+alter table "public"."atd_txdot_person" add column "peh_fl" boolean
+ null;
+
+COMMENT ON COLUMN public.atd_txdot_person.peh_fl IS 'Boolean flag that indicates whether the person was experiencing homelessness at the time of the crash.';
--- a/atd-vze/src/views/Crashes/personDataMap.js
+++ b/atd-vze/src/views/Crashes/personDataMap.js
@@ -108,7 +108,7 @@ export const primaryPersonDataMap = [
         editable: false,
       },
       peh_fl: {
-        label: "Experiencing Homelessness",
+        label: "Unhoused",
         editable: true,
         format: "boolean",
         mutationVariableKey: "personId",
@@ -189,7 +189,7 @@ export const personDataMap = [
         mutationVariableKey: "personId",
       },
       peh_fl: {
-        label: "Experiencing Homelessness",
+        label: "Unhoused",
         editable: true,
         format: "boolean",
         mutationVariableKey: "personId",
--- a/atd-vze/src/views/Crashes/personDataMap.js
+++ b/atd-vze/src/views/Crashes/personDataMap.js
@@ -21,12 +21,12 @@ const getInjurySeverityColor = personRecord => {
     case 4: // FATAL INJURY: red
       return "danger";
     case 99: // KILLED: red
-      return "secondary"
-    default: // Other cases: Not reported, Reported invalid, Not injured
+      return "secondary";
+    default:
+      // Other cases: Not reported, Reported invalid, Not injured
       return "secondary";
   }
 };
-
 
 export const primaryPersonDataMap = [
   {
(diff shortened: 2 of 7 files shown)


atd-vze/src/views/Crashes/personDataMap.js
```diff
@@ -107,6 +107,12 @@ export const primaryPersonDataMap = [
         label: "ZIP",
         editable: false,
       },
+      peh_fl: {
+        label: "Experiencing Homelessness",
+        editable: true,
+        format: "boolean",
```
roseeichelmann, line 113: created this new "boolean" format because the "select" format uses lookup tables to determine what choices are in the drop down.. this new column doesn't have a lookup table, its not something we get from CRIS.

atd-vzd/migrations/default/1699990226463_alter_table_public_atd_txdot_person_add_column_peh_fl/up.sql
```diff
@@ -0,0 +1,4 @@
+alter table "public"."atd_txdot_person" add column "peh_fl" boolean
+ null;
+
+COMMENT ON COLUMN public.atd_txdot_person.peh_fl IS 'Boolean flag that indicates whether the person was experiencing homelessness at the time of the crash.';
```
roseeichelmann, line 4: i think using postgres comments as much as we can especially for columns is great practice!! ive been noticing we have been taking advantage of this more as a team lately

atd-vze/src/views/Crashes/personDataMap.js
frankhereford: No, I think you've made the perfect choice. It's exactly what it is - true/false to the person experiencing homelessness.

atd-vzd/migrations/default/1699990226463_alter_table_public_atd_txdot_person_add_column_peh_fl/up.sql
frankhereford: I need to do much better about this - thank you for mentioning it. I agree with you, it's nothing but good. 🚀

frankhereford (approved):
I think you've done this perfectly. I've spun it up and watched the records change while editing them in the UI. I think where you put the UI elements make perfect sense. too. This is 🚀🚀🔥.

mddilley (approved):
Looks great, and I agree that the related records area is a fine place to add this to the UI. 🙏 Testing checks out - thanks for adding this so the VZ team can start capturing this data. 🚢 🚀

atd-vze/src/views/Crashes/personDataMap.js
roseeichelmann: Hmm thats a good point! I think since it's not in the section with the rest of the "Flags" it doesn't stand out as being abnormal, but I think changing to `YES` and `NO` does make sense from a UI perspective!

johnclary (commented):
👋 just chiming in to say that this looks good to me! it's cool to look at how y'all manage these re-usable inputs in VZ 🚀

chiaberry (approved):
I like where you put the field editing and the labels you chose

patrickm02L:
Back-channeled with Rose about the UI title "Experiencing Homelessness" and requested the title be changed to "Unhoused" so as to take up less real estate on the People card.

image

roseeichelmann:
I changed the label to say "Unhoused" and also thanks to @mddilley suggestion I changed true/false to be yes/no instead which I think is nicer from a UI standpoint!

patrickm02L (approved):
Good to go! 💯

mddilley:
@roseeichelmann I tested this again locally and looked at the new commits, and it is 🚢✨!

PR #1320, opened 2023-11-20
Remove vestigial docker image for VZ ETLs

frankhereford:
## Associated issues

This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/11858. In particular, it is responding to the revised intent of the issue found here: https://github.com/cityofaustin/atd-data-tech/issues/11858#issuecomment-1819737032.

This issue was written when we were still using a unified docker image for our VZ ETL processes. When we transitioned to Airflow 2, we reimplemented each ETL in its own docker image. This left the old setup unused, and this PR is cleaning that out.

## A note on the `atd-etl/README.md`

This was full of documentation of the old ETL system which we are no longer using. I have struck all that documentation from the file and replaced with with a meager two sentences. Do you think we should spin off an issue to write documentation for these to replace what has been removed?

## Testing

This one is all removal of content, so testing is pretty much the following:

* Look at what's removed. Can you think of any impacts to our current operation with the removal of these programs and supporting infra?
* Look at what remains. Is there anything else that you think should go?

---

#### Ship list


mddilley (approved):
These ETLs have come a long way! I looked through the removed files and scripts, and nothing jumps out at me as something not covered by the individual Docker images. Thanks for cleaning out this legacy code! 🙏 🧹 🙌 🚀

atd-etl/README.md
mddilley, line 1: The only thing that I can think to add here is the GH action that updates the images with new code in these folders. I think that we are covered by the READMEs in each folder. I did notice that we don't have one in `process_missing_cr3_pdfs` though.

chiaberry: +1 to Mike's comment!

chiaberry (approved):
thanks for cleaning this up

roseeichelmann (approved):
Looks good, thanks for cleaning this out Frank!

patrickm02L:
Confirmed it's good to merge in Refinement 11/22

PR #1323, opened 2023-11-28
Replace moment.js with date-fns for VZV

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/12007

I've refactored the entire codebase to use `date-fns` instead of `moment.js` except `SideMapControlDateRange` which uses a component called `DateRangePicker` from this [react-dates](https://github.com/react-dates/react-dates) library that expects/requires a moment object as props, so we need to find something else to use that isn't dependent on moment. Finding a new date range component to use or making our own will take some time so I'm gonna create another issue to address that since this is getting pretty big already.

## Testing
**URL to test:** 
https://deploy-preview-1323--atd-vzv-staging.netlify.app/

**Steps to test:**
1. I think the best way to test is to pull up the [current VZV](https://visionzero.austin.gov/viewer/) and compare everything to this deploy preview to make sure the values, dates, numbers, etc are the same and the functionality is the same.
2. On the dashboard make sure the numbers and graphs are the same.
3. On the map view adjust the dates and test other functionality, make sure its all working.

---
#### Ship list
- [ ] Code reviewed


atd-vzv/src/constants/nav.js
```diff
@@ -5,12 +5,6 @@ import React from "react";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faChartBar, faMap } from "@fortawesome/free-solid-svg-icons";
 
-// Detect the environment
-export const isDevelopment =
-  !process.env.NODE_ENV || process.env.NODE_ENV === "development";
-
-export const isPreview = process.env.REACT_APP_VZV_ENVIRONMENT === "PREVIEW";
-
```
roseeichelmann, line 13: If I understood correctly in dev sync, I dont think we need this distinction anymore and we are just gonna pull production socrata data going forward whether or not we are in development?

atd-vzv/src/views/nav/SideMapControlDateRange.js
```diff
@@ -32,8 +34,8 @@ import {
 
 const SideMapControlDateRange = ({ type }) => {
   const [focused, setFocused] = useState(null);
-  const [start, setStart] = useState(mapStartDate);
-  const [end, setEnd] = useState(mapEndDate);
+  const [start, setStart] = useState(moment(mapStartDate));
+  const [end, setEnd] = useState(moment(mapEndDate));
```
roseeichelmann, line 38: need to reconvert this into a moment object here for the `DateRangePicker` component since the rest of the changes i did turned this into a normal date object.

atd-vzv/src/constants/nav.js
mddilley: This makes sense to me so we don't trip over this when running locally going forward. I think that we can decide how to point the app to staging when/if https://github.com/cityofaustin/atd-data-tech/issues/13087 is prioritized.

atd-vzv/src/views/nav/SideMapControlDateRange.js
mddilley: Not sure exactly where this is coming from but I see that the UI is showing the "restore" icon when first visiting the map. iirc that icon should only show after the date start, end, or both are updated. When I click the icon in the preview, it breaks. Not my favorite situation that this picker relies on moment.js 😔 - thanks for updating around this limitation!

Default in staging/production
Screenshot 2023-12-21 at 4 02 39 PM

Default in preview
Screenshot 2023-12-21 at 4 02 33 PM

mddilley (changes requested):
This is looking great! I tested the dashboard, and it looks 100% like production. Thanks for getting us into the modern date library world. ⏳ I found one regression in the map date picker, and it looks like a small fix.

atd-vzv/src/views/nav/SideMapControlDateRange.js
roseeichelmann: Thanks for catching that! Just pushed up the fix

mddilley (approved):
Checked this side by side against the production site, and I see everything matching up. Thanks for updating! 🙏 🚀 🙌

atd-vzv/src/constants/nav.js
chiaberry: agreed.

chiaberry (approved):

patrickm02L (approved):
Good to go! 🥏

PR #1326, opened 2023-12-07
Patch bugs from missing hasura permissions

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/14786

I've already tested the Editor and Read-only users with these updated permissions so im pretty confident this is good to go :ok_hand: 

## Testing
**URL to test:** 
Local

**Steps to test:**
1. Checkout this branch, apply the metadata to your local database.
2. Test clicking on crash pages while using the Editor user and Read only user.
3. Also test updating the coordinates of a crash as an Editor user, make sure it works and saves the new coordinates.

---
#### Ship list
- [ ] Code reviewed


atd-vzd/metadata/databases/default/tables/public_atd_txdot_crashes.yaml
```diff
@@ -116,6 +116,233 @@ array_relationships:
         remote_table:
           name: atd_txdot_units
           schema: public
+insert_permissions:
```
roseeichelmann, line 119: I went thru all of these columns in the crashes table to make sure none of them are PII or sensitive information we would only want admin level users to insert/edit, but I couldn't find any. So I toggled on all columns for insert/update on Editor permissions, let me know if yall see something that should be blocked for Editors.

mddilley:
Looking at this right now to check against [the permissions from 4/11/2023](https://github.com/cityofaustin/atd-vz-data/blob/635a8a5a4698310b19b11bd859d71eae50e7572c/atd-vzd/hasura/hasura_metadata_2023_04_11_14_03_58_508.json#L1626-L2770)

chiaberry:
Is this going to be a patch to prod? Should the base branch be production and bump the version number?

mddilley (approved):
I compared the old and new metadata for crash table permissions and started [this thread](https://austininnovation.slack.com/archives/CM9MK950S/p1701984494483269). Thank you for catching the additional issues with editor permissions. 🚀 🙏

PR #1330, opened 2023-12-20
Remove part of VZDB readme that uses the Hasura CLI

mddilley:
## Associated issues

There are still remnants of the old steps we used to manage migrations with the Hasura CLI before we moved to the GH action CI. Since we are trying to avoid any manual steps against our Hasura instances, I tried to describe the checks we have in place in Moped to make sure we don't have conflicting migrations. We can do this without our staging Hasura admin secret which these current steps depend on.

## Testing
**URL to test:** 
n/a

**Steps to test:**
1. Can this be worded better?

---
#### Ship list
- [x] Code reviewed

### DIFF
--- a/atd-vzd/README.md
+++ b/atd-vzd/README.md
@@ -26,14 +26,20 @@ hasura migrate apply --envfile .env.local
 hasura metadata apply --envfile .env.local
 ``` 
 - Start the local Hasura console and make any changes needed which will then reflect in your project folder
+
 ### Merging an approved feature branch
 
-We need to check the order of migrations against those in the `master` **before merging a feature branch** so that we can make updates to the migration version order if needed. The version refers to the timestamp in migration folder name.
+We need to check the order of migrations against those in the `master` branch **before merging a feature branch** so that we can make updates to the migration version order if needed. The version refers to the timestamp in the migration folder name.
 
-To check migrations for any conflicts with latest migrations in the `master`` branch:
+To check migrations for any conflicts with the latest migrations in the `master` branch:
 - Make sure that your branch is up to date with `master`
 - Check to make sure no one else is actively merging their work and coordinate if needed
-- Update the migration version in your project so it is the newest migration
-- Test locally using the steps in the  [Generating migrations and metadata changes section](#generating-migrations-and-metadata-changes)
+- Update the migration versions in your branch so they are the newest migrations if needed
+- Start up the local database and Hasura engine and replicate using the latest production data dump
+- Then, run:
+```bash
+hasura migrate apply --envfile .env.local
+hasura metadata apply --envfile .env.local
+``` 
 
-Once everything looks good, we can merge and the CI will apply the new migrations and metadata to the staging database.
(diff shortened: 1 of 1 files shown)


atd-vzd/README.md
```diff
@@ -29,16 +28,9 @@ hasura metadata apply --envfile .env.local
...
 
-We need to check the status of new migrations against the staging Hasura engine **before merging a feature branch** so that we can make updates to the migration version order if needed. The version refers to the timestamp in migration folder name.
-
-To check the status from your local project folder, run:
-```bash
-hasura migrate status --envfile .env.staging
-```
-
-Check that your new migration shows `NOT PRESENT` in the `DATABASE` column and that it is also the last row in the table shown. We need to make sure that migrations are applied in the same order.
+We need to check the order of migrations against those in the `master` **before merging a feature branch** so that we can make updates to the migration version order if needed. The version refers to the timestamp in migration folder name.
```
roseeichelmann, line 31: typo here :see_no_evil: , update to "those in master" or "those in the master branch" ?

atd-vzd/README.md
```diff
@@ -29,16 +28,9 @@ hasura metadata apply --envfile .env.local
...
 
-If the new migration is not the last migration to apply in the sequence:
+To check migrations for any conflicts with latest migrations in the `master`` branch:
 - Make sure that your branch is up to date with `master`
 - Check to make sure no one else is actively merging their work and coordinate if needed
 - Update the migration version in your project so it is the newest migration
```
roseeichelmann, line 36: the line after this `- Test locally using the steps in the  [Generating migrations and metadata changes section](#generating-migrations-and-metadata-changes)` confuses me bc is this rly how ur supposed to test that there are no conflicts? to me this just seems to be instructions on how to set up your env before creating any migrations

atd-vzd/README.md
mddilley: 🙏 fixing this!

atd-vzd/README.md
mddilley: Thanks for looking at this! This was really hard for me to capture in words, and I think that I was mostly trying to recycle a section instead of writing out more stuff. I pushed an edit, but I'm happy to continue making changes. I don't want this to be confusing. 🙏

roseeichelmann (approved):
Yay this looks great, thanks for bearing with me on clarifying that last part :smile:

atd-vzd/README.md
roseeichelmann: Yeah its hard to capture the migration process in words for sure :face_with_spiral_eyes:  I see now how recycling that section without writing out more stuff could make total sense for some ppl but I think the edit you just pushed makes it crystal clear! Esp if we ever get new ppl on the team, im always trying to think of newbies being able to understand the docs and im like well if im getting confused they might too :stuck_out_tongue_closed_eyes:

patrickm02L (approved):
Thanks for cleaning this up 🧹

chiaberry (approved):
makes sense!

mddilley:
Merging this y'all - thanks!!!

PR #1331, opened 2023-12-27
Updates that came out of backfilling EMS incidents

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/15077

Most of these changes were made by Black formatter. I added some notes, but this update revolves around making the number of days we backfill into a parameter for the AFD and EMS scripts.

Also, since we had a gap in data received from EMS that was over a month long, **do y'all think that we should increase the default # of days?** I ran 120 days to backfill, but I could see bumping it to 30 or 60 days. Not sure if there is more info on why we picked 15 days before but I'm guessing it was because we never had a lapse in source data.

## Testing
**URL to test:** 

**Steps to test:**
1. Start up VZ locally by replicating from production
2. Create a `.env` from the new template in `atd-vz-data/atd-etl/afd_ems_import`
3. Follow the readme in that same folder to run the AFD and EMS scripts on your local DB
4. You can update the number of days in the development section of the 1PW entry called `Vision Zero AFD and EMS Import` to test out different numbers of days back the script goes

---
#### Ship list
- [x] Code reviewed


atd-etl/afd_ems_import/afd_incident_upload.py
```diff
@@ -111,7 +112,12 @@ def get_secrets():
             "opitem": "RDS Bastion Key",
             "opfield": ".private key",
             "opvault": VAULT_ID,
-            },
+        },
+        "record_age_maximum_in_days": {
```
mddilley, line 116: I think it would be great to be able to update the number of days that the script looks back from 1PW but let me know if anyone sees a reason not to add this parameter.

atd-etl/afd_ems_import/docker-compose.yml
```diff
@@ -6,6 +6,4 @@ services:
     volumes:
       - .:/app
     env_file:
-      - ./env
-    environment:
-      - ENVIRONMENT=development
+      - ./.env
```
mddilley, line 9: The readme said to create a `.env` so I updated it here.

atd-etl/afd_ems_import/docker-compose.yml
```diff
@@ -6,6 +6,4 @@ services:
     volumes:
       - .:/app
     env_file:
-      - ./env
-    environment:
-      - ENVIRONMENT=development
```
mddilley, line 11: I removed this since we already default to `development` in the script, and we have an `.env` file. I missed this and kept running on development after changing my `.env` to production to do the backfill. 🙃

atd-etl/afd_ems_import/afd_incident_upload.py
chiaberry: I think this is a great idea too. I dont remember how long we were getting empty csvs, and being able to change the maximum without having to change the code is great.

chiaberry (approved):
thanks for the cleanup on the code and more importantly, thanks for jumping on the request so quickly.

atd-etl/afd_ems_import/afd_incident_upload.py
roseeichelmann: I like this idea!

frankhereford: 👍 from me, I think it makes a lot of sense.

frankhereford (approved):
This is awesome Mike, thanks for adding this flexibility and access to the scripts parameters. I spun it up locally and went through the code -- thank you for the cleanup with `black` too! ✨🍰 ✨

patrickm02L (approved):
🚀 Ship

PR #1332, opened 2023-12-28
Update Unhoused description in People card

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/15043

Idk if Xavier realized that the data type for this field already was boolean in the database and we just made the front end say "Yes/No" from a UI/UX point of view, but I updated the wording anyways and also updated the label.

## Testing
**URL to test:** 
[Netlify](https://deploy-preview-1332--atd-vze-staging.netlify.app/#/)

**Steps to test:**
1. Go to a crash with both primary and other people
2. Update their "Suspected Unhoused" columns. Make sure the update persists. Options should be No Data, True, or False.

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vze/src/views/Crashes/personDataMap.js
+++ b/atd-vze/src/views/Crashes/personDataMap.js
@@ -108,7 +108,7 @@ export const primaryPersonDataMap = [
         editable: false,
       },
       peh_fl: {
-        label: "Unhoused",
+        label: "Suspected Unhoused",
         editable: true,
         format: "boolean",
         mutationVariableKey: "personId",
@@ -189,7 +189,7 @@ export const personDataMap = [
         mutationVariableKey: "personId",
       },
       peh_fl: {
-        label: "Unhoused",
+        label: "Suspected Unhoused",
         editable: true,
         format: "boolean",
         mutationVariableKey: "personId",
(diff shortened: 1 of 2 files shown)


atd-vze/src/views/Crashes/personDataMap.js
```diff
@@ -108,7 +108,7 @@ export const primaryPersonDataMap = [
         editable: false,
       },
       peh_fl: {
-        label: "Unhoused",
+        label: "Suspected Unhoused",
```
chiaberry, line 111: is there a preference on sentence case vs title case @patrickm02L ?

roseeichelmann: All the other labels use title case so i went with that

chiaberry: i just noticed that when i was testing it and was going to remove my comment!

chiaberry (approved):
thanks for making their updates

patrickm02L (approved):
Send it 🥏

mddilley (approved):
Tested on the deploy preview, and this works great! Thanks for updating. 🙏 🚢

PR #1337, opened 2024-01-05
Process pending Dependabot updates for VZE

frankhereford:
## Associated issues

This PR is part of the work to close https://github.com/cityofaustin/atd-data-tech/issues/14729. 

## Note

After a discussion in dev sync, we agreed that disabling this geocoder was an acceptable route, given that we're going to take up a spin-off issue to work out the real issues at hand with it.

## Testing

https://deploy-preview-1337--atd-vze-staging.netlify.app/#/crashes/19325058

* Visit all parts of the application which have a slippy map on them:
  * Crash details page view location
  * Crash details page edit location
  * Location details page view location

---
#### Ship list
- [ ] Code reviewed


atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
frankhereford, line 1: I'm leaving the geocoder stuff in the file to assist whomever is assigned the follow on task of restoring that functionality.

atd-vze/src/scss/_custom.scss
frankhereford, line 1: These icons are being added because this appears to be the path of least resistance in restoring the UI and UI's functionality for the map after the 2-major-verions upgrade included in this PR. Not doing so would have been a regression stemming from this PR's work, whereas the GeoCoder regression was pre-existing to this PR.

atd-vze/src/scss/_custom.scss
```diff
@@ -1 +1,22 @@
 // Here you can add other styles
+
+div.fullscreen>div>button {
```
mddilley, line 3: I need someone else to verify, but I was able to see the default Mapbox buttons without these styles added here. If that can't be replicated, I think that this selector should be a little more specific because there are two different icons for the expand and collapse (depending on whether in full screen mode or not).

atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
mddilley: This issue felt really familiar, and [it looks like we ran into it in VZV too](https://github.com/cityofaustin/atd-vz-data/blob/1db22a57ee1dd575cce7f9e93d45cdd746bb1163/atd-vzv/src/views/map/Geocoder/Geocoder.js#L7-L13). This workaround would be useful info for the spin-off. cc/ @patrickm02L 

Versions 6+ of this react-map-gl switched over to the real deal Mapbox GL geocoder plugin instead of this library, which is great, but it leaves behind a lot of quirks like this! 😵‍💫

mddilley (commented):
Thanks for getting us up to date! We are very close version-wise to VZV now. 🙌 I left one comment about the Mapbox icons.

atd-vze/src/scss/_custom.scss
frankhereford: Mike, I have been able to confirm that I am able to see the icons. I have no idea why they were so consistently not showing up for me before, but this time, I've tested by entirely wiping out my `node_modules` folder and my `packge-log.json` file and building from as scratch as I can conceive of doing. They seem to show up fine, so I am going to strike those changes from this PR.

Additionally, I have had to add two checks for null objects that end up containing data from the graphql-engine endpoint but cause a WSOD. I mention this here because I had had to add them when I first started working this issue, and then removed them because I found them not to be needed in testing, which feels a lot like this icon thing.

There's something going on, most likely on my computer or between my keyboard and chair that is preventing me from feeling like I have a 100% handle on what's going on here. That said, I will feel comfortable with shipping this PR with approval from you and others on the team.

atd-vze/public/assets/img/expand.svg
mddilley, line 1: @frankhereford Could you remove this and the other two svgs? I don't think we need them anymore (could be wrong).

frankhereford: Shoot - i had that sitting staged in my repo but didn't commit it - thank you! They are out in 35251ab.

atd-vze/src/scss/_custom.scss
mddilley: Hmmm I did initially see no icons before I npm installed so maybe that is a piece of it. Glad you sorted it out on your end. It looks great in the preview!

atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
mddilley, line 1: This question is for @patrickm02L - are we going to try to get the geocoder fix in before the next release? Just curious, and I didn't think to ask yesterday in dev sync. 🙃

mddilley (approved):
I tested the maps every which way that I could think of, and I see no regressions with the updated versions. 😎 🙌

atd-vze/src/Components/DataTable.js
```diff
@@ -118,8 +118,11 @@ const DataTable = ({
 
                       // If there is no lookup options, we can assume the field value can be displayed as is.
                       // If there is a lookup option, then the value is an ID to be referenced in a lookup table.
-                      const selectOptions =
-                        lookupSelectOptions[fieldConfigObject.lookupOptions];
+                      let selectOptions;
+                      if (lookupSelectOptions) {
+                        selectOptions =
+                          lookupSelectOptions[fieldConfigObject.lookupOptions];
+                      }
```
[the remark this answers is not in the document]
roseeichelmann: i am an enjoyer of optional chaining ever since i learned about it :raising_hand:

roseeichelmann:
Im getting this weird behavior that i showed in this screen recording :0

where i click edit and the latitude and longitude boxes just have 0 in them instead of the current coords and if you click save you get an error and sometimes get caught in a weird loop.

[simplescreenrecorder-2024-01-12_14.24.36.webm](https://github.com/cityofaustin/atd-vz-data/assets/67130963/ff3dc16f-2730-4577-a593-0275ec429055)

frankhereford:
> Im getting this weird behavior that i showed in this screen recording :0
> 
> where i click edit and the latitude and longitude boxes just have 0 in them instead of the current coords and if you click save you get an error and sometimes get caught in a weird loop. curious if anyone else can replicate, im testing on firefox

Rose, thank you so much for this!! I am going to see what I can do to run this down, and I certainly won't merge until I'm sure it's under control. I may reach out to you if I am unable to replicate. Again, thank you for tagging this!!

frankhereford:
Here's the video of the geocoder behavior that prompted me to talk to the group where we decided to disable the geocoder and spin off an issue to fix it back up.

https://github.com/cityofaustin/atd-vz-data/assets/3653206/eeefb3b8-11d3-4237-8131-ddb36beaffe9

It's broken in a similar way in production, except it doesn't drop down the geocoder suggestions, which makes it easy to overlook the regression. It's the new version of the library that causes the existing regression to be disruptive and significantly degrade the usability of the tool.

mddilley:
@frankhereford @patrickm02L Thanks for hearing me out about the geocoder in planning! Only wanted to make sure we didn't forget that it would disappear if we merged this into Tuesday's release. I'm good either way since I don't think it will be missed in its broken state. All of this looks excellent. ✨

mddilley (changes requested):
Requesting changes on Frank's behalf. See https://austininnovation.slack.com/archives/CM9MK950S/p1706285501154119

frankhereford:
closing this because it's been long enough that we'd better off crafting a new one with the other dependabot prs that have come in

PR #1338, opened 2024-01-05
Rework requirements.txt and remove `_request_ctx_stack` use from CR3 API service

frankhereford:
## Associated issues

This PR is part of closing https://github.com/cityofaustin/atd-data-tech/issues/14729.

## Testing

Testing should be local for this one. Here are some steps to test:

1) create a `.env` in the `atd-cr3-api` folder if you don't have one already. It may be easiest to just ask me to wormhole you one, or you can piece it together out of our 1PW entries. If you happen to have an existing env file, please make sure it has `API_CLIENT_ID` and `API_CLIENT_SECRET`.
1) Spin up the flask app after creating a venv for it and installing the requirements set forth in the `requirements.txt`. This should setup an API endpoint for you at `localhost:3010`, which you can visit to get a status message at the `/` route.
1) Now, head over to your VZE, and edit the environment file for it. You'll want to set the following value as: `REACT_APP_CR3_API_DOMAIN="http://localhost:3010"`
1) Now you can use the VZE to interact with this endpoint. You can read the [flask app's routes](https://github.com/cityofaustin/atd-vz-data/blob/cr3-api-dependabot/atd-cr3-api/server.py) to see what you can test, but here are some highlights:

  * Can you download a CR3?
  * List users on the users page? 
 
---
#### Ship list
- [ ] Code reviewed


atd-cr3-api/.env.example
```diff
@@ -1,8 +1,10 @@
...
+API_ENVIRONMENT=STAGING
+AUTH0_DOMAIN=atd-datatech.auth0.com
+CLIENT_ID=2qbqz2sf8L9idBOwn0d5YA9efNgQbL7c
+AWS_DEFALUT_REGION=us-east-1
+AWS_S3_CR3_LOCATION=staging/path_to_files
+AWS_S3_BUCKET=your_bucket_name_here
+AWS_S3_KEY=your_s3_key_here
+AWS_S3_SECRET=your_s3_secret_here
+API_CLIENT_ID=secret_here
+API_CLIENT_SECRET=secret_here
```
frankhereford, line 10: NB: these two are new environment variables

atd-cr3-api/server.py
```diff
@@ -328,7 +329,13 @@ def user_list_users():
...
     per_page = request.args.get("per_page")
     if isValidUser(user_dict) and hasUserRole("admin", user_dict):
-        endpoint = f"https://{AUTH0_DOMAIN}/api/v2/users?page=" + page + "&per_page=" + per_page + "&include_totals=true"
+        endpoint = (
+            f"https://{AUTH0_DOMAIN}/api/v2/users?page="
+            + page
+            + "&per_page="
+            + per_page
+            + "&include_totals=true"
+        )
```
frankhereford, line 338: Linting

atd-cr3-api/.env.example
mddilley: Thanks for adding these! 🙏 I happened to notice that the `API_ENVIRONMENT` is not used in the server code. We can probably get away with removing it here if you agree we should.

mddilley (approved):
Thanks for the thorough test steps - it's been a while since I've run this Flask app locally. I tested downloading a CR3 and also CRUDing a user. All works just like before. 🚀

roseeichelmann (approved):
works like a charm, i was able to download CR3s and CRUD users as well. thanks for taking care of this Frank, love to see all that red in the requirements file!! :ship: :rocket:

patrickm02L (approved):
🚢

atd-cr3-api/.env.example
frankhereford: Great idea; it's outta there. 🏟️ ⚾

PR #1339, opened 2024-01-05
Add confirmation to delete notes on the Location page and Crash page

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/14771

This PR adds an existing `ConfirmModal` component to the `Notes` component so a confirm modal appears when deleting a crash or location note.

## Testing
**URL to test:** 
https://deploy-preview-1339--atd-vze-staging.netlify.app/

**Steps to test:**
1. Log in to the deploy preview with either editor, admin, or IT supervisor permissions
2. Go to a crash page like [this one](https://deploy-preview-1339--atd-vze-staging.netlify.app/#/crashes/19329553) and add a note to the crash
3. Click the trash can icon and you should see a confirm modal pop up
4. Click **Cancel** or the **X icon** and the modal should close without deleting the note
5. Try to delete again and, this time, click **Confirm**, and the note should delete
6. Repeat the previous steps on a location page like [this one](https://deploy-preview-1339--atd-vze-staging.netlify.app/editor/#/locations/05186FFEDB)

---
#### Ship list
- [x] Code reviewed

### DIFF
--- a/atd-vze/src/containers/DefaultLayout/DefaultHeader.js
+++ b/atd-vze/src/containers/DefaultLayout/DefaultHeader.js
@@ -105,7 +105,7 @@ const DefaultHeader = props => {
                 target="_blank"
                 rel="noreferrer"
               >
-                <i class="fa fa-briefcase" /> CR3 code sheet&nbsp;&nbsp; <i className="fa fa-external-link" />
+                <i className="fa fa-briefcase" /> CR3 code sheet&nbsp;&nbsp; <i className="fa fa-external-link" />
               </DropdownItem>
             </DropdownMenu>
         </UncontrolledDropdown>
--- a/atd-vze/src/Components/ConfirmDeleteButton.js
+++ b/atd-vze/src/Components/ConfirmDeleteButton.js
@@ -1,15 +1,14 @@
 import React, { useState } from "react";
 import { Button } from "reactstrap";
-import ConfirmModal from "../ConfirmModal.js";
+import ConfirmModal from "./ConfirmModal.js";
 
-const ConfirmDeleteButton = ({ onConfirmClick }) => {
+const ConfirmDeleteButton = ({ onConfirmClick, modalHeader, modalBody }) => {
   const [showModal, setShowModal] = useState(false);
 
   const toggleModal = () => {
     setShowModal(!showModal);
   };
 
-  // render notes card and table
   return (
     <>
       <Button
@@ -23,8 +22,8 @@ const ConfirmDeleteButton = ({ onConfirmClick }) => {
         <i className="fa fa-trash" />
       </Button>
       <ConfirmModal
-        modalHeader={"Delete Confirmation"}
-        modalBody={`Are you sure you want to delete this note?`}
+        modalHeader={modalHeader}
+        modalBody={modalBody}
         confirmClick={onConfirmClick}
         toggleModal={toggleModal}
         showModal={showModal}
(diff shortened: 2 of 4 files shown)


patrickm02L (approved):
I went back and forth thinking about whether the CTA should say "Delete" or "Confirm" I think confirm makes sense to keep it as "Confirm" here because you're asking them if they're sure they want to delete. 

But I think we should change the `modalHeader` to Delete Confirmation versus Delete Note.

Otherwise, everything's go to go!
Screen Shot 2024-01-05 at 5 02 23 PM

mddilley:
@patrickm02L Thanks! I just pushed the update to the header text.

frankhereford (changes requested):
Mike, thanks for adding this helpful feature. I think the modal looks great, and it's cool you were able to reuse one we had in the code base.

I was able to get it to do some unexpected behavior, which you can see here:

https://github.com/cityofaustin/atd-vz-data/assets/3653206/d10e404b-b15f-4b60-b9b7-a8f74ef7d578

I think you might need to do something like set the record ID in a state variable at the time toggleModal is called and use that later if/when handleDeleteClick is invoked. I think right now handleDeleteClick is going to operate on the last row that was rendered, which ends up being the last row on the table. Take me with a grain of salt of course!

mddilley:
@frankhereford Thanks for catching that! I was just able to reproduce that bug locally, and I'll take a look at it today.

atd-vze/src/containers/DefaultLayout/DefaultHeader.js
```diff
@@ -105,7 +105,7 @@ const DefaultHeader = props => {
                 target="_blank"
                 rel="noreferrer"
               >
-                <i class="fa fa-briefcase" /> CR3 code sheet&nbsp;&nbsp; <i className="fa fa-external-link" />
+                <i className="fa fa-briefcase" /> CR3 code sheet&nbsp;&nbsp; <i className="fa fa-external-link" />
```
mddilley, line 108: Fixes a console warning

mddilley:
@frankhereford I just pushed a fix, and I think that should do the trick. The bug came down to me overlooking that every modal rendered was sharing the same `showModal` state. Thanks again!

frankhereford (approved):
Thanks Mike!! 🚢 🚢 🚢

chiaberry (approved):
tested and ✔️ 

A future enhancement could be to include the text of the note the person is deleting, or somehow denote what specific note a person is deleting. but thats for the future!

mddilley:
@chiaberry I like that idea a lot! It would be a small change and small effort to fit that in this PR. What you think @patrickm02L? Also happy to make a spin-off issue

patrickm02L:
> @chiaberry I like that idea a lot! It would be a small change and small effort to fit that in this PR. What you think @patrickm02L? Also happy to make a spin-off issue

Let's include it!

mddilley:
Alright, I updated the modal to show text of the note being deleted. To test:
1. add several notes and one with really long text (my go-to is [pizza ipsum](https://pizzaipsum.com/?paras=5&type=all-custom&start-with-lorem=1))
2. try deleting any of them and see that the note or truncated note is shown in the modal

patrickm02L (approved):
Thanks for making the addition of adding the text being deleted in the confirmation pop-up, and making sure numbers show up. Send it 🍕 

Screen Shot 2024-01-11 at 8 48 07 AM

Screen Shot 2024-01-11 at 8 49 42 AM

PR #1340, opened 2024-01-09
Display annual totals for all, fatalities, and serious injuries for the year

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/15047

## Testing
**URL to test:** 
https://deploy-preview-1340--atd-vzv-staging.netlify.app/

**Steps to test:**
1. Go to the summary page, make sure you are seeing the new "Totals" row in the By Travel Mode section.
2. Make sure it shows up for all of the crash types (All, Fatality, Serious Injury)
3. Are they adding up correctly?

Lmk if yall have any design thoughts like if this row should look ~ different ~ from the others somehow.

---
#### Ship list
- [x] Code reviewed

### DIFF
--- a/atd-vzv/src/views/summary/CrashesByMode.js
+++ b/atd-vzv/src/views/summary/CrashesByMode.js
@@ -334,10 +334,6 @@ const CrashesByMode = () => {
                             <div>
                               <hr className="my-0"></hr>
                               <p className="h6 text-center my-0 py-1">
-                                <FontAwesomeIcon
-                                  aria-hidden="true"
-                                  className="block-icon"
-                                />
                                 <span className="sr-only">Total</span>
                                 <span className="mode-label-text"> Total</span>
                               </p>
--- a/atd-vzv/src/views/summary/CrashesByMode.js
+++ b/atd-vzv/src/views/summary/CrashesByMode.js
@@ -174,16 +174,17 @@ const CrashesByMode = () => {
     datasets: !!chartData && createTypeDatasets(),
   };
 
-  const yearTotalsArray = data.datasets
-    ? yearsArray().map((year, index) => {
-        let currentYearTotal = 0;
-        data.datasets.forEach((mode) => {
-          currentYearTotal += mode.data[index];
-        });
-        console.log(currentYearTotal, "current year total");
-        return currentYearTotal;
-      })
-    : null;
+  // Get an array of totals for the selected crash type (all, fatal, injury) for each year
+  const getYearTotalsArray = () => {
+    const yearTotalsArray = yearsArray().map((year, index) => {
+      let currentYearTotal = 0;
+      data.datasets.forEach((mode) => {
+        currentYearTotal += mode.data[index];
+      });
+      return currentYearTotal;
+    });
+    return yearTotalsArray;
+  };
 
   const StyledDiv = styled.div`
     .year-total-div {
@@ -336,8 +337,6 @@ const CrashesByMode = () => {
                                 <FontAwesomeIcon
                                   aria-hidden="true"
                                   className="block-icon"
-                                  // icon={dataset.icon}
-                                  // color={legendColors[i]}
                                 />
                                 <span className="sr-only">Total</span>
                                 <span className="mode-label-text"> Total</span>
@@ -377,8 +376,9 @@ const CrashesByMode = () => {
                                     }
                                   )}
                                   <hr className="my-0"></hr>
-                                  <p className={`h6 text-center my-1 pb-1`}>
-                                    {yearTotalsArray[yearIterator]}
+                                  <p className={`h6 text-center my-1`}>
+                                    {data.datasets &&
+                                      getYearTotalsArray()[yearIterator]}
                                   </p>
                                 </div>
                               </StyledDiv>
--- a/atd-vzv/src/views/summary/CrashesByMode.js
+++ b/atd-vzv/src/views/summary/CrashesByMode.js
@@ -174,6 +174,17 @@ const CrashesByMode = () => {
     datasets: !!chartData && createTypeDatasets(),
   };
 
+  const yearTotalsArray = data.datasets
+    ? yearsArray().map((year, index) => {
+        let currentYearTotal = 0;
+        data.datasets.forEach((mode) => {
+          currentYearTotal += mode.data[index];
+        });
+        console.log(currentYearTotal, "current year total");
+        return currentYearTotal;
+      })
+    : null;
+
   const StyledDiv = styled.div`
     .year-total-div {
       color: ${colors.dark};
@@ -319,6 +330,19 @@ const CrashesByMode = () => {
                                 </div>
                               );
                             })}
+                            <div>
+                              <hr className="my-0"></hr>
+                              <p className="h6 text-center my-0 py-1">
+                                <FontAwesomeIcon
+                                  aria-hidden="true"
+                                  className="block-icon"
+                                  // icon={dataset.icon}
+                                  // color={legendColors[i]}
+                                />
+                                <span className="sr-only">Total</span>
+                                <span className="mode-label-text"> Total</span>
+                              </p>
+                            </div>
                           </StyledDiv>
                         </Col>
                         {chart.data.labels.map((year, yearIterator) => {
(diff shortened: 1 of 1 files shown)


frankhereford (approved):
Rose, this is great. I've looked at your code and tested it on the deploy preview, and it's ship shape.

I did notice a regression which is not the result of your work here, but I did want to mention it in case you want to tack it on to this PR or if @patrickm02L wants to spin it off into an issue:

image

There's something about our "other" and "total rows" that make them 4 pixels taller than the rows above them, and you can see that by looking for misalignment between the label column and the leftmost data column.

But again, this is existing in production, and this PR is ready to roll IMHO. Thanks again!

atd-vzv/src/views/summary/CrashesByMode.js
```diff
@@ -352,6 +371,11 @@ const CrashesByMode = () => {
                                       );
                                     }
                                   )}
+                                  <hr className="my-0"></hr>
+                                  <p className={`h6 text-center my-1`}>
```
mddilley, line 375: I saw in the git history that you had a `pb-1` in here, and I think that it should come back if that looks right to you. ✨ It adds some grey space below the final row of cells that I think the regression that Frank mentioned was meant to address. The final row hard coded in the code is not the final row anymore with the addition of recent rows and this total row.

Also, super small detail but the braces and backticks can be replaced with double quotes. Purely stylistic and feel free to ignore!

atd-vzv/src/views/summary/CrashesByMode.js
```diff
@@ -174,6 +174,18 @@ const CrashesByMode = () => {
...
   };
 
+  // Get an array of annual totals for the selected crash type
+  const getYearTotalsArray = () => {
+    const yearTotalsArray = yearsArray().map((year, index) => {
+      let currentYearTotal = 0;
+      data.datasets.forEach((mode) => {
+        currentYearTotal += mode.data[index];
+      });
+      return currentYearTotal;
```
mddilley, line 184: [reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) could be fun alternative here. Not a change request as this is great too!

I once had a code block like this with declaration + work + return described to me as a "code sandwich". I've never been able to unsee it. 🥪

mddilley (commented):
This all looks great! I think that fixing the padding issue should happen in this PR if possible. Glad that Frank caught that because it slipped by us before!

atd-vzv/src/views/summary/CrashesByMode.js
frankhereford: ## 🥪 🤯❗

What a great name for that. I think it just got burnt into my mind too,

mddilley: Haha glad you enjoyed - felt like sharing that funny story. 😄 Yep, reduce can be hard to follow sometimes too - so many ways to do things! I think the code here is great. 🚀

atd-vzv/src/views/summary/CrashesByMode.js
roseeichelmann: lol good catch, i'm throwing it back in! unfortunately it doesn't address the regression frank mentioned but it does add some grey space to the last row that makes it look a little more even compared to the others :sparkles:

patrickm02L (approved):
The totals add up and everything looks great! Thanks for fixing the alignment as well, looks clean! 🚢

chiaberry (approved):
thanks for making quick work of this, this is nitpicky and similar to what Frank pointed out, something that predates your work, but the fact that the panels dont line up bugs me
Screenshot 2024-01-11 at 8 52 02 AM

atd-vzv/src/views/summary/CrashesByMode.js
mddilley: awesome! 🚀 I re-read my comment and was being really unclear at the end of yesterday. 😴 When I was talking about the "final row hard coded", I really meant there is a magic number used to assign a value to `paddingBottom` earlier in the code. I think that will lead to the fix. 🤞

atd-vzv/src/views/summary/CrashesByMode.js
```diff
@@ -339,7 +358,7 @@ const CrashesByMode = () => {
                                   {chart.data.datasets.map(
                                     (mode, modeIterator) => {
                                       let paddingBottom =
-                                        modeIterator === 4 ? "pb-1" : "pb-0";
+                                        modeIterator === 4 ? "pb-0" : "pb-0";
```
mddilley, line 361: To avoid further confusion down the road with the meaning of this magic number 4, could you remove this `paddingBottom` var altogether? 🧹 Thanks again for getting these totals in here - it really ties this viz together 🙏

roseeichelmann: yeah i definitely did not mean to leave this in here, was gonna continue working on yalls requests today after doing a quick commit last night :facepalm:

chiaberry:
@johnclary the black bar mystery issue: https://github.com/cityofaustin/atd-data-tech/issues/11370

atd-vzv/src/views/summary/CrashesByMode.js
mddilley: oh gotcha - just me being eager here! 😅

roseeichelmann:
I'm still doing some pixel pushing trying to get the cells to line up :sweat_smile:  but in the mean time @johnclary i pushed up an update that memoizes the `getYearTotalsArray()` function and i agree that it would be nice to revisit this component in a future issue and address the `getModeData` function as well

atd-vzv/src/views/summary/CrashesByMode.js
```diff
@@ -352,6 +371,11 @@ const CrashesByMode = () => {
                                       );
                                     }
                                   )}
+                                  <hr className="my-0"></hr>
+                                  <p className={`h6 text-center my-1 pb-1`}>
+                                    {data.datasets &&
+                                      getYearTotalsArray()[yearIterator]}
```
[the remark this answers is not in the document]
roseeichelmann: definitely not outside the scope imo now that u pointed that out i could not live with it haha

atd-vzv/src/views/summary/CrashesByMode.js
```diff
@@ -174,6 +174,21 @@ const CrashesByMode = () => {
     datasets: !!chartData && createTypeDatasets(),
   };
 
+  // Get an array of annual totals for the selected crash type
+  const getYearTotalsArray = useMemo(() => {
+    const yearTotalsArray = yearsArray().map((year, index) => {
+      let currentYearTotal = 0;
+      data.datasets &&
```
roseeichelmann, line 181: on first render `data.dataset` is null so this will fail unless i do this check

johnclary: thanks @roseeichelmann for making this change! this looks great.

Worth noting that because `data` is not memoized, this hook is still going to run more than it needs to, but that problem feels out of scope.

johnclary (approved):

roseeichelmann:
@johnclary i tested it and i think its only running twice (console log inside of the function), once when `data.datasets` is null and once when we actually get `data.datasets`. is that not right?

johnclary:
@roseeichelmann it also runs when you interact with the mode toggles, because that causes data to recalculate 🙃

mddilley (approved):
Looks great! Thanks for making the spin off issue too. It would be great for someone to run back through this code and freshen it up. 🚢

atd-vzv/src/views/summary/CrashesByMode.js
mddilley: +1 - a small improvement could be to update the dependency array with `data` and throw a `if (!data?.datasets) return;` at the beginning of the function passed to useMemo to skip the rest of the code when there is no dataset. Not a huge improvement, and I think we can punt until we pay off the debt here.

atd-vzv/src/views/summary/CrashesByMode.js
roseeichelmann: ahh i never responded to this -- yeah code sandwich is accurate lol. theres often that choice between picking an elegant/less lines of code way vs something that is easier to read/follow but more wordy... sometimes i like the easy to read way but i need to check out `reduce` some time :p

PR #1344, opened 2024-01-13
Fix WSOD in VZE when interacting with data widgets on crash listing page

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/15130

## Testing
**URL to test:** 
https://deploy-preview-1344--atd-vze-staging.netlify.app/#/crashes

**Steps to test:**
1. Go to the crashes page, erase the start date like you are gonna type in a new one, yay no more WSOD. If you click away without typing in a new date it just auto selects the day before the day you already had clicked.
2. Repeat with the end date picker.

---
#### Ship list
- [x] Code reviewed

### DIFF
--- a/atd-vze/src/Components/GridDateRange.js
+++ b/atd-vze/src/Components/GridDateRange.js
@@ -69,10 +69,10 @@ const GridDateRange = ({
    * @returns {string}
    */
   const formatDate = (date, fallbackValue) => {
-    let formattedDate = format(date, "yyyy-MM-dd");
+    let formattedDate = date ? format(date, "yyyy-MM-dd") : "Invalid date";
 
     if (formattedDate === "Invalid date") {
-      formattedDate = format(fallbackValue, "yyyy-MM-dd");
+      formattedDate = format(new Date(fallbackValue), "yyyy-MM-dd");
     }
     return formattedDate;
   };
--- a/atd-vze/src/Components/GridDateRange.js
+++ b/atd-vze/src/Components/GridDateRange.js
@@ -64,16 +64,15 @@ const GridDateRange = ({
 
   /**
    * Returns a date in a valid SQL format.
-   * @param {string} date - The string to be transformed
-   * @param {string} fallbackValue - The value to use if the formattedDate is invalid
+   * @param {Date} date - The date object to be formatted
+   * @param {string} fallbackValue - The value to use if date is null
    * @returns {string}
    */
   const formatDate = (date, fallbackValue) => {
-    let formattedDate = date ? format(date, "yyyy-MM-dd") : "Invalid date";
+    const formattedDate = date
+      ? format(date, "yyyy-MM-dd")
+      : format(new Date(fallbackValue), "yyyy-MM-dd");
 
-    if (formattedDate === "Invalid date") {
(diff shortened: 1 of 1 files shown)


atd-vze/src/Components/GridDateRange.js
```diff
@@ -64,16 +64,15 @@ const GridDateRange = ({
...
-   * @param {string} fallbackValue - The value to use if the formattedDate is invalid
+   * @param {Date} date - The date object to be formatted
+   * @param {string} fallbackValue - The value to use if date is null
    * @returns {string}
    */
   const formatDate = (date, fallbackValue) => {
-    let formattedDate = format(date, "yyyy-MM-dd");
+    const formattedDate = date
+      ? format(date, "yyyy-MM-dd")
+      : format(new Date(fallbackValue), "yyyy-MM-dd");
```
roseeichelmann, line 74: previously if the date was null this errored out bc `format()` expects a valid date

atd-vze/src/Components/GridDateRange.js
```diff
@@ -114,6 +113,7 @@ const GridDateRange = ({
           // Prevent user from selecting start date after current date
           maxDate={maxDate}
           minDate={minDate}
+          strictParsing={true}
```
roseeichelmann, line 116: fixes some weird behavior where if you start erasing with your keyboard `onChange()` will fire not just once you've completely emptied the textbox but at various intervals, for example if the date was "12/20/2020" and you start erasing the characters once you get to "1" `onChange()` fires and sets the date to "01/01/2024".

johnclary: ah, interesting that this component supports this behavior!

i was _just_ dealing with this in Moped, and ended up writing [this handler](https://github.com/cityofaustin/atd-moped/blob/98892dc964d8fa5630c69c7e62df5536056e4f8c/moped-editor/src/components/forms/ControlledDateField.js#L5C1-L13C3) to test when a date was valid, because sometimes it fires off random partial date strings that throw an error.

roseeichelmann: @johnclary yeah this was super convenient and i only figured out about this prop by combing thru their github issues :sweat_smile:

chiaberry: 🕵🏼 nice find!

patrickm02L (approved):
I like how it falls back to the previous day to prevent it tripping the WSOD. 🔙

frankhereford (approved):
Rose, you fixed it! This looks great - thank you for this. 🚢 🚢 🚢 

@patrickm02L - I think we should spin off another issue from this work to enhance this text input to accept 1-digit dates and months and maybe 2 digit years. Right now, if you start typing in `1/1/2024` it won't parse; it expects `01/01/2024` which isn't very human-oriented. But for sure, this is outside the scope of this issue! I'm curious what you think.

mddilley (approved):
Works great to fix the WSOD! I tried to break it every way that I could and failed. 🚀 🚢 

+1 to Frank's suggestion for a spinoff issue. The pre-existing `useEffects` are hard to follow too, and, at a glance, it feels like that logic could be moved into `onChange` handlers. For another day!

PR #1348, opened 2024-01-18
Add new near trauma center fields to crashes table 

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/15279

I couldn't find anything about the new fields in the [CRIS release notes](https://austininnovation.slack.com/archives/CM9MK950S/p1705084971737399?thread_ts=1705084752.767399&cid=CM9MK950S), I wonder where else we would find info on added fields? I figured out what fields we needed to add by looking at the`record_json` in the `atd_txdot_changes` table and it was just these two having to do with nearby trauma centers.

Follow up question to this PR is -- are these fields that would be of interest to the VZ team and should we display them in the VZE on the Crash details page?

## Testing
**URL to test:** 
Local

**Steps to test:**
1. Check out this branch locally, replicate a fresh copy of prod data & start up the graphql engine.
2. Start VZE locally and observe how if you go to the Changes tab, click on a record, select some fields, and try to "Save Selected Changes" like it says in the issue, you will get a graphql error.
3. Now run the up migrations.
```
hasura migrate apply --up 1
hasura metadata apply
```
5. Try to "Save Selected Changes" on some records in the Changes tab now, you should get no errors!
6. Test running the down migration
`hasura migrate apply --down 1`
7. Test editor user, you should not see the Changes tab in the nav bar anymore.

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vzd/metadata/databases/default/tables/public_atd_txdot_crashes.yaml
+++ b/atd-vzd/metadata/databases/default/tables/public_atd_txdot_crashes.yaml
@@ -245,8 +245,6 @@ insert_permissions:
         - milepoint_2
         - mpo_id
         - nbr_of_lane
-        - near_trauma_center_distance
-        - near_trauma_center_id
         - non_injry_cnt
         - nonincap_injry_cnt
         - obj_struck_id
@@ -934,8 +932,6 @@ update_permissions:
         - milepoint_2
         - mpo_id
         - nbr_of_lane
-        - near_trauma_center_distance
-        - near_trauma_center_id
         - non_injry_cnt
         - nonincap_injry_cnt
         - obj_struck_id
--- a/atd-vzd/metadata/databases/default/tables/public_atd_txdot_crashes.yaml
+++ b/atd-vzd/metadata/databases/default/tables/public_atd_txdot_crashes.yaml
@@ -245,6 +245,8 @@ insert_permissions:
         - milepoint_2
         - mpo_id
         - nbr_of_lane
+        - near_trauma_center_distance
+        - near_trauma_center_id
         - non_injry_cnt
         - nonincap_injry_cnt
         - obj_struck_id
@@ -473,6 +475,8 @@ select_permissions:
         - milepoint_2
         - mpo_id
         - nbr_of_lane
+        - near_trauma_center_distance
+        - near_trauma_center_id
         - non_injry_cnt
         - nonincap_injry_cnt
         - obj_struck_id
@@ -701,6 +705,8 @@ select_permissions:
         - milepoint_2
         - mpo_id
         - nbr_of_lane
+        - near_trauma_center_distance
+        - near_trauma_center_id
         - non_injry_cnt
         - nonincap_injry_cnt
         - obj_struck_id
@@ -928,6 +934,8 @@ update_permissions:
         - milepoint_2
         - mpo_id
         - nbr_of_lane
+        - near_trauma_center_distance
+        - near_trauma_center_id
         - non_injry_cnt
         - nonincap_injry_cnt
         - obj_struck_id
--- a/atd-vzd/migrations/default/1705620819250_crashes_add_near_trauma_center_fields/up.sql
+++ b/atd-vzd/migrations/default/1705620819250_crashes_add_near_trauma_center_fields/up.sql
@@ -1,0 +1,3 @@
+ALTER TABLE public.atd_txdot_crashes
+ADD COLUMN near_trauma_center_id integer NULL,
+ADD COLUMN near_trauma_center_distance double precision NULL;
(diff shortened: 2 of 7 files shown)


atd-vzd/migrations/default/1705620819250_crashes_add_near_trauma_center_fields/up.sql
```diff
@@ -0,0 +1,4 @@
+-- Add new trauma center fields from CRIS to the crashes table
+ALTER TABLE public.atd_txdot_crashes
+ADD COLUMN near_trauma_center_id integer NULL,
+ADD COLUMN near_trauma_center_distance double precision NULL;
```
roseeichelmann, line 4: i looked at all the recent records and their `record_json` fields in the `atd_txdot_changes` table to make the assumption that `near_trauma_center_id` will always be an integer and `near_trauma_center_distance` seems to always be a number with 3 digits. I thought [double precision](https://www.postgresql.org/docs/current/datatype-numeric.html) was a safe type choice bc we use it for a lot of other floating point fields.

frankhereford:
Hi Rose - thanks for adding this! I wanted to mention that if you want to get a copy of data as we receive it from CRIS, you can either ask CRIS to send you an export and grab it off the SFTP endpoint, or you can find the CSVs in S3. I don't expect there will be anything there that you have not already found, but I wanted to let you know about that S3 resource. I had to grab a copy of CSVs just this week when I wanted to find out what had caused a CRIS import to take too long and it was really helpful. Thanks again for this patch!!

johnclary (commented):
I am not running the local VZ setup, but I did want to ask: do there need to be any hasura permission changes here? i guess so long as only admins are using this change detection thing, the answer is "no"?

roseeichelmann:
@johnclary you're totally right, i have the worst habit of forgetting to update hasura permissions for new columns

roseeichelmann:
When testing with the Edit only VZE test user i found another bug separate from this PR - when you try to "Save Selected Changes" you get the following error `field 'delete_atd_txdot_changes' not found in type: 'mutation_root'`. I realized this is because editors do not have any insert/update/delete permissions on the changes table. I'm guessing it's been like this for a long time and since Xavier is the only one that goes thru the changes and hes admin user its never came up?

Should I address that in this issue or another? And should "editor" users even have access to the changes tab?

chiaberry:
that's a good question Rose, @patrickm02L , should editors be able to access the changes tab?

atd-vzd/metadata/databases/default/tables/public_atd_txdot_crashes.yaml
```diff
@@ -245,6 +245,8 @@ insert_permissions:
         - milepoint_2
         - mpo_id
         - nbr_of_lane
+        - near_trauma_center_distance
+        - near_trauma_center_id
```
mddilley, line 249: Another question for @patrickm02L, will these new fields be shown in the UI and be editable? I'm thinking that, if the answer is no or if we don't know, then maybe we shouldn't add insert/update permissions yet? I also acknowledge that we aren't in a great state as far as existing permissions matching what the UI does...

mddilley (commented):
This fixes the bug! 🧹 🐛 Hiding the changes tab from the editors in the UI feels like the quickest/patchiest way to handle the editor role not having the needed permissions for `atd_txdot_changes` to me since it is broken anyway. I'll approve this once we get some input from Patrick.

patrickm02L:
> that's a good question Rose, @patrickm02L , should editors be able to access the changes tab?

Teams Xavier, he said it's not necessary for Editors to see the Changes Tab. He also said, that he's the only one that does anything with the Changes Tab - he has admin permissions

atd-vzd/metadata/databases/default/tables/public_atd_txdot_crashes.yaml
roseeichelmann: Yeah so based off Patricks comment it sounds like i could remove Editors from accessing the changes tab, and then in that case i could also not add insert/update/delete permissions for editors yet. I am also wondering about whether will these fields be shown in the UI and/or be editable @patrickm02L

patrickm02L: @Rose, yes please turn off permissions for Editors. No direction for editing in the UI right now, but I can follow up with Xavier - which will probably be a separate Issue.

roseeichelmann:
Okay this should be good to review now i fixed the permissions for Editor (now it only has select permissions) and removed Editors from seeing the Changes tab.

atd-vze/src/_nav.js
mddilley, line 1: Thanks for cleaning up this file! 🧹 It is interesting that we are seeing a lot of the editor role specific UI pieces thinning out. Glad you caught that the editor permissions issue in this one.

atd-vze/src/_nav.js
```diff
@@ -36,17 +36,13 @@ export const navigation = roles => {
     ],
   };
 
-  // Editor nav itmes
```
roseeichelmann, line 39: hehe typo begone

mddilley (approved):
Tested with all roles, and I see no conflict resolution bug and that only admin+ roles can access the changes tab. 🚀 🚢

chiaberry (approved):
cleaning up the routes is great! thanks for working with all the requests around this

frankhereford (approved):

PR #1353, opened 2024-01-26
Unhoused investigation / tooling to update the VZDB with data straight out of CRIS CSVs

frankhereford:
## Intent

This is a small, simple python script which can be used to rapidly go through large volumes of CRIS CSVs and search for certain strings in certain fields. It has potential uses in verify and debugging data consistency problems and can easily be adapted for any number of CRIS CSV parsing needs. In conjunction with `xargs`, I used it recently to go through 10 years of CRIS data in just a few seconds.

## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/14985

## Testing

Testing in this case I believe is best done via a review of the code. It was used successfully to set the PEH flag recently for `primarypersons`, https://github.com/cityofaustin/atd-data-tech/issues/14985. 

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-toolbox/cris_export_parser/parse_data.py
+++ b/atd-toolbox/cris_export_parser/parse_data.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python3
+
+import csv
+import argparse
+
+
+def parse_csv(file_path, field_names, substrings):
+    with open(file_path, "r") as file:
+        reader = csv.DictReader(file)
+        for row in reader:
+            for field in field_names:
+                if field in row:
+                    for substring in substrings:
+                        if substring.lower() in row[field].lower():
+                            print(
+                                {
+                                    "Crash_ID": row.get("Crash_ID", ""),
+                                    "Prsn_Nbr": row.get("Prsn_Nbr", ""),
+                                    "Prsn_Name": f"{row.get('Prsn_First_Name', '')} {row.get('Prsn_Last_Name', '')}",
+                                    "Drvr_City_Name": row.get("Drvr_City_Name", ""),
+                                    "Drvr_Street_Name": row.get("Drvr_Street_Name", ""),
+                                }
+                            )
+                            break
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Parse a CSV file.")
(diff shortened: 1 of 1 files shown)


frankhereford:
@patrickm02L Should we pursue getting this into the repo?

patrickm02L:
> @patrickm02L Should we pursue getting this into the repo?

Yes, let's move it to the toolbox.

atd-toolbox/cris_export_parser/parse_data.py
```diff
@@ -0,0 +1,98 @@
...
+                                cur.execute(update)
+                                conn.commit()
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Parse a CSV file.")
+    parser.add_argument("-i", "--input", help="Input file path", required=True)
+    args = parser.parse_args()
+
+    fields = [
```
mddilley, line 84: Not a change needed but this list seems like it could be moved to be a [constant at the module level](https://peps.python.org/pep-0008/#constants). I think that sounds right but I've been trying to get back in the swing of Python for the ETL i was working on. Certainly doesn't change how this program runs.

atd-toolbox/cris_export_parser/parse_data.py
```diff
@@ -0,0 +1,98 @@
...
+                                count = f"""
+select count(*) as count
+from atd_txdot_primaryperson
+where true
+and crash_id = {row.get("Crash_ID")} 
+and unit_nbr = {row.get("Unit_Nbr")}
+and prsn_nbr = {row.get("Prsn_Nbr")};"""
+
+                                cur.execute(count)
+                                count_check = cur.fetchone()
```
mddilley, line 49: cool, so this checks that a record match exists before trying to update its `peh_fl`and logs it if there isn't a match and is skipped 👍

mddilley (approved):
All looks good to me! 🚢 into the 🧰 🚀

roseeichelmann (approved):
Looks good, thanks for adding this script to the toolbox :))

patrickm02L (approved):
👍

PR #1355, opened 2024-01-29
Street/satellite layer toggle doesn't work in Crash location edit map

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/15193

This PR removes the street/satellite layer toggle from the Crash location edit map. It also combines some shared map layers in the maps into a new component.

## Testing
**URL to test:** 
https://deploy-preview-1355--atd-vze-staging.netlify.app/

**Steps to test:**
1. Go to a Crash view and look at the map in the top left corner
2. You should see a token warning on the map and, in your network dev tools, you should see 401 responses from a
`https://api.nearmap.com/tiles` endpoint. These requests for tiles should be happening and should be failing since we can't add a wildcard in URLs added to the allow list for our NearMap token. When this merges, we should see NearMap aerial tiles in all map views in these test steps in staging VZE. (if you are testing locally, we fallback to the Mapbox satellite layer)
3. You should also see street labels when you explore the map.
4. Now, click **Edit Coordinates**, and you should see that the toggle UI for the satellite and street layer is removed. NearMap tile requests should also be returning 401s here and street labels should be present.
5. Last, go to a Location view and look at the map in the top left. Same NearMap 401s and street labels should be present.

---
#### Ship list
- [x] Code reviewed


atd-vze/src/views/Locations/LocationMap.js
```diff
@@ -168,12 +171,7 @@ export default class LocationMap extends Component {
         mapboxApiAccessToken={TOKEN}
       >
         {/* add nearmap raster source and style */}
-        {!isDev && (
-          <>
-            <Source {...LOCATION_MAP_CONFIG.sources.aerials} />
-            <Layer {...LOCATION_MAP_CONFIG.layers.aerials} />
-          </>
-        )}
+        {!isDev && <LabeledAerialSourceAndLayer />}
```
mddilley, line 174: We didn't show the street labels on top of the NearMap tiles here previously, but I think that we should. What do y'all think?

mddilley:
@patrickm02L I noticed when testing as a read-only user that they can use the edit crash coordinates map in the UI. Want me to create a spinoff or fix in this PR? Doesn't seem like a large lift to me.

patrickm02L (approved):
Really great instructions on how to view the token, 401, street names and removal of the street view!

Crashes Example

Screen Shot 2024-01-29 at 7 39 24 PM

Location Example

Screen Shot 2024-01-29 at 7 39 51 PM


So long to the streets! 🛣️

atd-vze/src/views/Locations/LocationMap.js
roseeichelmann: absolutely! i think the street labels are always super helpful :sparkles:

roseeichelmann (approved):
yay thanks for cleaning this up Mike looks great!! :broom: :ship: also thanks for the nice test steps and letting us know to expect the 401s and token warnings, that has thrown me off in the past.

atd-vze/src/views/Locations/LocationMap.js
roseeichelmann: i actually am seeing them in staging though on locations?

frankhereford (approved):
Looks great - everything checks out, and I think this an improvement on all the maps. Thank you Mike!

atd-vze/src/views/Locations/LocationMap.js
mddilley: @roseeichelmann thank you for catching this! 🕵️‍♂️ The street layer was in here a few lines below. The new `LabeledAerialSourceAndLayer` component introduced a duplicate so I removed the old one that was on its own. 🙏

mddilley:
Going to merge this and check that the NearMap tiles are showing in staging. Thanks for the reviews, y'all! 🙏

PR #1370, opened 2024-02-01
Control access to certain information by user email domain

frankhereford:
## Associated issues

closes https://github.com/cityofaustin/atd-data-tech/issues/15536

## Changed behavior

Hide the CR3 download button and the fatality review board information from non `austintexas.gov` email address holders. Additionally, the notes component is hidden on both the crash and location details page.

## Tag-along change

I adjusted the label for the CR3 download button to read "CR3" instead of "CR-3."

## Local testing

Because we all have an `austintexas.gov` address, the easiest way to test this is to spin it up locally and comment out the line where we `setIsCOA(true)` in `atd-vze/src/auth/authContext.js` to simulate not having an `austintexas.gov` address. The default value of the state variable is false, so not setting it to true simulates a non-COA user.

With that default state value, as if you were a non-`austintexas.gov` user, check to see that the following are not visible to you:

* [ ] Notes component on locations page
* [ ] Notes component on crash page
* [ ] Fatality Review Board component on the crash page
* [ ] Download CR3 button on the crash page

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vze/src/views/Crashes/Crash.js
+++ b/atd-vze/src/views/Crashes/Crash.js
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useState, useEffect } from "react";
 import { Link } from "react-router-dom";
 import {
   Card,
@@ -55,6 +55,16 @@ const calculateYearsLifeLost = people => {
 };
 
 function Crash(props) {
+  const [isCOA, setIsCOA] = useState(false);
+  useEffect(() => {
+    const storedEmail = window.localStorage.getItem("hasura_user_email");
+    if (storedEmail) {
+      if (storedEmail.toLowerCase().endsWith("@austintexas.gov")) {
+        setIsCOA(true);
+      }
+    }
+  }, []);
+
   const crashId = props.match.params.id;
   const { loading, error, data, refetch } = useQuery(GET_CRASH, {
     variables: { crashId },
@@ -283,6 +293,7 @@ function Crash(props) {
         </Col>
         <Col xs="12" md="6" className="mb-4">
           <CrashDiagram
+            isCOA={isCOA}
             crashId={crashId}
             isCr3Stored={cr3StoredFlag === "Y"}
             isTempRecord={tempRecord}
@@ -304,7 +315,7 @@ function Crash(props) {
           <CrashCollapses data={data} props={props} />
         </Col>
       </Row>
-      {shouldShowFatalityRecommendations && (
+      {isCOA && shouldShowFatalityRecommendations && (
         <Row>
           <Col>
             <Recommendations crashId={props.match.params.id} />
--- a/atd-vze/src/views/Crashes/Crash.js
+++ b/atd-vze/src/views/Crashes/Crash.js
@@ -292,7 +292,7 @@ function Crash(props) {
           />
         </Col>
       </Row>
-      {!!investigatorNarrative ? (
+      {isCOA && !!investigatorNarrative ? (
         <Row>
           <Col>
             <CrashNarrative investigatorNarrative={investigatorNarrative} />
--- a/atd-vze/src/views/Crashes/Crash.js
+++ b/atd-vze/src/views/Crashes/Crash.js
@@ -62,7 +62,7 @@ function Crash(props) {
         setIsCOA(true);
       }
     }
-  }, []);
+  }, [user]);
 
   const crashId = props.match.params.id;
   const { loading, error, data, refetch } = useQuery(GET_CRASH, {
--- a/atd-vze/src/views/Crashes/Crash.js
+++ b/atd-vze/src/views/Crashes/Crash.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from "react";
+import React, { useState } from "react";
 import { Link } from "react-router-dom";
 import {
   Card,
@@ -56,14 +56,9 @@ const calculateYearsLifeLost = people => {
 
 function Crash(props) {
   const { getRoles, user } = useAuth0();
-  const [isCOA, setIsCOA] = useState(false);
-  useEffect(() => {
-    if (user && user.email) {
-      if (user.email.toLowerCase().endsWith("@austintexas.gov")) {
-        setIsCOA(true);
-      }
-    }
-  }, [user]);
+
+  const isCOA =
+    !user?.email && user.email.toLowerCase().endsWith("@austintexas.gov");
 
   const crashId = props.match.params.id;
   const { loading, error, data, refetch } = useQuery(GET_CRASH, {
--- a/atd-vze/src/views/Crashes/Crash.js
+++ b/atd-vze/src/views/Crashes/Crash.js
@@ -58,7 +58,7 @@ function Crash(props) {
   const { getRoles, user } = useAuth0();
 
   const isCOA =
-    user?.email && user.email.toLowerCase().endsWith("@austintexas.gov");
+    !!user?.email && user.email.toLowerCase().endsWith("@austintexas.gov");
 
   const crashId = props.match.params.id;
   const { loading, error, data, refetch } = useQuery(GET_CRASH, {
--- a/atd-vze/src/views/Crashes/Crash.js
+++ b/atd-vze/src/views/Crashes/Crash.js
@@ -55,6 +55,7 @@ const calculateYearsLifeLost = people => {
 };
 
 function Crash(props) {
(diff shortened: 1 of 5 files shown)


atd-vze/src/views/Crashes/Crash.js
```diff
@@ -55,6 +55,16 @@ const calculateYearsLifeLost = people => {
 };
 
 function Crash(props) {
+  const [isCOA, setIsCOA] = useState(false);
+  useEffect(() => {
+    const storedEmail = window.localStorage.getItem("hasura_user_email");
```
frankhereford, line 60: See https://github.com/cityofaustin/atd-vz-data/blob/61869d6f920b359a7d8aad3302c4699e973d1c2d/atd-vze/src/auth/authContext.js#L70 for where this is set.

atd-vze/src/views/Crashes/Crash.js
```diff
@@ -55,6 +55,16 @@ const calculateYearsLifeLost = people => {
...
+    if (storedEmail) {
+      if (storedEmail.toLowerCase().endsWith("@austintexas.gov")) {
+        setIsCOA(true);
```
frankhereford, line 63: To get this state set to `true`, you have to have an austintexas.gov address. Otherwise, this state variable `isCOA` will continue to have the default value of `false`. To test this PR and simulate not having an `austintexas.gov` address, just comment out this line.

atd-vze/src/views/Crashes/Crash.js
```diff
@@ -283,6 +293,7 @@ function Crash(props) {
         </Col>
         <Col xs="12" md="6" className="mb-4">
           <CrashDiagram
+            isCOA={isCOA}
```
frankhereford, line 296: is our state variable down as a prop to the `<CrashDiagram>` component so it can hide the CR3 download button based on it.

frankhereford:
> Going to be great to have a way to lock this down. I do think that we should look into checking the user email from the internal session state instead of local storage which could be edited on the client end.

Genius. I'm on it - thank you for this superb suggestion.

atd-vze/src/views/Crashes/Crash.js
```diff
@@ -71,7 +80,7 @@ function Crash(props) {
   const [formData, setFormData] = useState({});
   const [isEditingCoords, setIsEditingCoords] = useState(false);
 
-  const { getRoles } = useAuth0();
+  const { getRoles, user } = useAuth0();
```
frankhereford, line 83: Thanks to @mddilley for the suggestion to check the email as stored in the session state, not local storage, which can be easily modified by a user.

atd-vze/src/views/Crashes/Crash.js
mddilley: @frankhereford I'm taking a look at this, and I'm pretty sure that we can get away without the `useEffect` and the `isCOA` state since `isCOA` can be derived directly from `user` below.

atd-vze/src/views/Crashes/Crash.js
```diff
@@ -55,6 +55,15 @@ const calculateYearsLifeLost = people => {
...
 
 function Crash(props) {
+  const [isCOA, setIsCOA] = useState(false);
+  useEffect(() => {
+    if (user && user.email) {
+      if (user.email.toLowerCase().endsWith("@austintexas.gov")) {
+        setIsCOA(true);
+      }
+    }
+  }, [user]);
```
frankhereford, line 65: Another thank you to @mddilley for catching my omission of `user` from the dependency array. 🙏

atd-vze/src/views/Crashes/Crash.js
frankhereford: Mike and I looked at this, and I believe that the useEffect is needed to cause the component to re-render when the `user` authentication data is loaded.

frankhereford: Further @mddilley stuck with me and helped me realize that he had been absolutely correct from the get-go. Thank you!

mddilley (commented):
Tested this locally, and it works! 🚀 🙏 I just thought about Notes (Location and Crash views) as another part of the app that might need protection. I'll approve after some input from @patrickm02L.

frankhereford:
@mddilley wrote:

> Also, I think that your question in Slack about doing this elsewhere in the app is really on point. Makes me think that extracting some of this logic into a helper or hook would be worthwhile so that it could be dropped into components.

I agree with Mike 100% here. This fix is a temporary patch, and I entirely support making a spin-off issue to pull this hard-coded logic out of these two components and embed it in the auth context for the entire application. That will make it accessible for any content restriction at any place in the app.

This needs to get up there, but that is the long term solution.

atd-vze/src/views/Crashes/CrashDiagram.js
```diff
@@ -50,9 +50,9 @@ const CrashDiagram = props => {
         <Row className="d-flex align-items-center">
           <Col>Crash Diagram</Col>
           <Col className="d-flex justify-content-end">
-            {props.isCr3Stored ? (
+            {props.isCOA && props.isCr3Stored ? (
               <Button color="primary" onClick={requestCR3}>
-                Download CR-3 PDF
+                Download CR3 PDF
```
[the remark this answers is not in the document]
patrickm02L: O-kay. Let's do it!

patrickm02L:
> Tested this locally, and it works! 🚀 🙏 I just thought about Notes (Location and Crash views) as another part of the app that might need protection. I'll approve after some input from @patrickm02L.

I looked at the Notes (Location and Crash views) in DBeaver and I didn't see anything that immediately struck me as PII, but I do agree it would be good to protect. If it's an easy fix, let's do it, but if not I can create an Issue.

frankhereford:
@patrickm02L wrote:

> If it's an easy fix, let's do it, but if not I can create an Issue.

NP at all, I'll add it to this PR right now.

frankhereford:
@mddilley wrote:

> Makes me think that extracting some of this logic into a helper or hook would be worthwhile so that it could be dropped into components.

With the inclusion of content in the Locations details page, it made sense to go ahead and pull the logic that sets `isCOA` into the auth context, so this PR has been updated to do that, including the test instructions. Thanks!

mddilley (commented):
Tested all views and see the content not showing when I override my COA status. I'm sounding like a broken record but what about the Crash Narrative section and the crash diagram since these are extracted of the CR3? All this code looks great so super ready to approve!

frankhereford:
This has been updated to include hiding the crash narrative as well.

roseeichelmann:
So far looking good as I'm reviewing but one thing that jumped out at me since it's something I worked on in the past -- in the people section of the related records card, if the person had a fatal injury you can view their name. This should probably be restricted to only COA emails as well. Though if they are read only users they already won't be able to see this component but still, extra protection to this PII?

frankhereford:
> if the person had a fatal injury you can view their name.

@roseeichelmann, I think you've observed another piece of PII, and I am going to include that in what we omit for non COA users. Thank you!!

roseeichelmann (approved):
Awesome thanks for the quick updates Frank as we kept expanding the scope, this looks good to go to me!!

frankhereford:
> Also maybe after this is merged before giving out any access we can test it again but this time with a dummy user we make with a Non COA email?

I think that's a great idea, and we also have set Xavier up with a second account using a non-COA address, so he also will have that ability to review what it looks like to an outside user too.

johnclary:
thank you, Frank, for working on this, and thanks Mike and Rose for the thoughtful reviews.

what exactly are we trying to mitigate here, and how bad it would be if our mitigations failed? e.g., what would happen if an ACC user were to gain access to a CR3? what about all the CR3s?

we need to be really careful when we go down the road of security through obfuscation. whatever these risks might be, does the Vision Zero team understand them?

i'm opposed to adding client-side JS to obfuscate data. we already have major permission deficiencies with the admin user role usage, and now we're layering on new problems with client-side hacks.  can we just have the ACC students sign a piece of paper that says "I will not do x, y, or z under penalty of being in a lot of trouble"?

frankhereford:
This PR was intended as a stop-gap measure to buy us some time and mitigate exposure to data with something to do quick. We no longer have that need, and as John pointed out, this is a poor semblance of securing the data. I am going to close this, because I don't want to merge it, and if we need to bring it back, we can. Thanks y'all.

PR #1371, opened 2024-02-06
Add new lookup tables and values to get VZDB up to date with CRIS v27

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/15263

Got to reuse @frankhereford code from https://github.com/cityofaustin/atd-vz-data/blob/refactor-vze-for-ldm/atd-toolbox/materials_from_ldm_conversion/inspect_lookup_table_data.py and tweaked it ever so slightly to create up and down migration files to easily copy/paste into the hasura console to create migrations. This is now a script that lives in our toolbox that can be run whenever there is a new CRIS release or we feel we aren't up to date.

## Testing
**URL to test:** 
Local

**Steps to test:**
1. Check out this branch locally
2. Get a fresh copy of prod data to run locally using the replicate-db command
3. Run the hasura up migrations
```bash
hasura migrate apply --up 2
hasura metadata apply
``` 
5. Run the hasura down migrations
```bash
hasura migrate apply --down 2
``` 
6. If you want you can even test the script yourself but it takes a really long time like at least 20 min i think? 
7. You can also just look through the script and make sure it makes sense, is readable etc.

---
#### Ship list
- [ ] Code reviewed


roseeichelmann:
Some follow up questions I have:

We have a lot of tables that we aren't tracking in hasura. Should we be tracking more? What goes into deciding whether or not to track a table?

Is the VZ team interested in seeing some of these new additions in the VZE, like for example the nearest trauma center for a crash? What about autonomous level for a unit? I held off on creating relationships between the crashes table and the trauma_centers_lkp table for example bc thats how we currently seem to handle things unless a column is actually being queried in the VZE. Is this okay or should we make these relationships for lookup tables even if the data isn't being used/queried anywhere else?

mddilley:
@roseeichelmann awesome questions. I think not tracking tables is fine until we need to query them, create relationships, or auto-generate migrations for changes. We can always track as we go like if the VZ team wanted to use the new lookups in this PR. I agree that this feels closest to what we do now. I could also see tracking everything if we wanted to heavily lean into the Hasura console as a explorer tool. But, yeah, keeping them untracked until needed feels in line with our past decisions!

atd-vzd/migrations/default/1707751774791_update_lookup_tables_v27/up.sql
mddilley, line 1: Looks great! One suggestion that could help us read these in the future is to print out comments and add newlines between tables getting inserts or updates that would join with the rest of the sql generated. It seems possible but let me know if this is a bigger lift than I'm thinking!

atd-vzd/migrations/default/1707751774791_update_lookup_tables_v27/up.sql
```diff
@@ -0,0 +1,531 @@
...
+insert into public.atd_txdot__agency_lkp (agency_id, agency_desc) values (10221, 'CITY OF MESQUITE');
+insert into public.atd_txdot__agency_lkp (agency_id, agency_desc) values (10222, 'CITY OF ARLINGTON');
+insert into public.atd_txdot__agency_lkp (agency_id, agency_desc) values (10223, 'TOM GREEN COUNTY');
+insert into public.atd_txdot__agency_lkp (agency_id, agency_desc) values (10224, 'FORT BEND COUNTY DISTRICT ATTORNEY''S OFFICE');
+insert into public.atd_txdot__autonomous_level_engaged_lkp (autonomous_level_engaged_id, autonomous_level_engaged_desc) values (3, 'CONDITIONAL AUTOMATION');
+insert into public.atd_txdot__autonomous_level_engaged_lkp (autonomous_level_engaged_id, autonomous_level_engaged_desc) values (4, 'HIGH AUTOMATION');
+insert into public.atd_txdot__autonomous_level_engaged_lkp (autonomous_level_engaged_id, autonomous_level_engaged_desc) values (5, 'FULL AUTOMATION');
+update public.atd_txdot__autonomous_level_engaged_lkp set autonomous_level_engaged_desc = 'AUTOMATION LEVEL UNKNOWN' where autonomous_level_engaged_id = 6;
+update public.atd_txdot__drvr_lic_cls_lkp set drvr_lic_cls_desc = 'AUTONOMOUS' where drvr_lic_cls_id = 95;
```
mddilley, line 24: Looked at this one randomly, and it is interesting that this value changed from `NOT REPORTED` to `AUTONOMOUS`. Seems like something that could affect reporting if the VZ team ever looked at autonomous vs non-autonomous based on this field. Goes to show what a huge challenge this all is with CRIS changing existing lookups!

Edit: Actually, could you check on the rows that follow this? Just struck me as out of the ordinary that CRIS would update some of these lookup table's `NOT REPORTED` row to `AUTONOMOUS` like ethnicity and injury severity.

mddilley (commented):
So cool that existing code come into use here! New tool in the toolbox and awesome to see some migration generation in action. I ran the script succesfully and made a few suggestions that are **NOT*** dealbreakers.

I did find some updates to existing lookup tables in the migration that stood out to me. Could you check them out?

atd-vzd/migrations/default/1707751774791_update_lookup_tables_v27/up.sql
roseeichelmann: Yeah i mean im not sure... the script does seem to be working properly bc i checked the actual lkp export and CRIS really did change all these values to `AUTONOMOUS`... for drivers license there are already other lkp values for `OTHER` and `UNKNOWN` so seems like maybe unknown is functioning as their new `NOT REPORTED` for example. same for ethnicity and injury, there are `UNKNOWN` values so i could see how `UNKNOWN` and `NOT REPORTED` are very similar and maybe unnecessary to have both, hence swapping not reported for autonomous. does seem like something we should point out to the VZ team for the things they can see in the UI like injury and ethnicity, thanks for flagging this!!!

roseeichelmann: side note i find it funny that there are autonomous options for people records now.. i would imagine autonomous cars would just not have a person record for the driver lol?

mddilley: +1 - so strange! And, also why wouldn't they add to the table instead of replacing an existing value? I guess that is part of us being consumers of the data and not having a ton of insight into these decisions. Thanks for looking into this! 🙏

roseeichelmann:
Thanks for the great feedback @mddilley , just finished pushing up all your suggestions!

mddilley (approved):
Retested the script and migrations, looked at the new commits, and all of this looks ✨ to me! Really cool to be able to start documenting these lookup table changes in migrations and having a way to undo them as well. 🚢 🚀

johnclary (approved):
👀 🚢

PR #1373, opened 2024-02-16
[Bug] Vision Zero - some changes that are not able to be made due to multiple types of data conflicts

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/15424

## Testing
**URL to test:** 
You need to test this with prod data on your local db so that you can test the conflict resolution system with the records in the changes table, there are none for staging.

**Steps to test:**
1. First you can see the current bug by checking out master and seeing how you cant save any of the changes for a change record.
2. Now check out this branch, go to the changes tab. Filter the date to go back to 2013 so you have a lot of change records to test.
3. Test selecting some fields and saving them, test discarding records, test selecting all fields and saving. See that the change record is deleted after you save.

This feature is still quite broken in ways that i can go into deeper in these comments, im not sure how long this feature has been broken but its not functioning how we expect i dont think. This issue just tackles the bug with not being able to use the feature at all due to strings.

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vze/src/queries/crashes_changes.js
+++ b/atd-vze/src/queries/crashes_changes.js
@@ -1,6 +1,4 @@
 import { gql } from "apollo-boost";
-
-/// TODO: have we been keeping this up to date with new columns?
 /**
  * A template that is used to gather the original crash record
  * and all records of pending changes.
@@ -235,22 +233,18 @@ export const CRASH_MUTATION_DISCARD = `
  * or Crashes. The function name matches a Hasura insertion method.
  * @type {string}
  */
-
-export const RECORD_MUTATION_UPDATE = `
-        %FUNCTION_NAME%(         
-            objects: [
-                %UPDATE_FIELDS%
-            ],
-            on_conflict: {
-              constraint: %CONSTRAINT_NAME%,
-              update_columns: [
-                updated_by
-                %SELECTED_COLUMNS%
-              ]
-            }
-        ) {
-          affected_rows
-        }
+export const CRASH_MUTATION_UPDATE = gql`
+  mutation updatesCrashes(
+    $crashId: Int!
+    $changes: atd_txdot_crashes_set_input
+  ) {
+    update_atd_txdot_crashes_by_pk(
+      pk_columns: { crash_id: $crashId }
+      _set: $changes
+    ) {
+      crash_id
+    }
+  }
 `;
 
 /**
@@ -262,20 +256,10 @@ export const RECORD_DELETE_CHANGE_RECORDS = `
       delete_atd_txdot_changes (
           where: {
             record_id: { _eq: $crashId }
-            record_type: { _in: ["crash", "unit", "primaryperson", "person"] }
+            record_type: { _in: ["crash", "unit", "primaryperson", "person", "charges"] }
           }
       ) {
         affected_rows
       }
-  }
-`;
-
-/**
- * Dummy GraphQL query
- * @type {string}
- */
-export const UPSERT_MUTATION_DUMMY = `
-  mutation dummyQuery {
-    __typename
   }
 `;
--- a/atd-vze/src/queries/crashes_changes.js
+++ b/atd-vze/src/queries/crashes_changes.js
@@ -228,9 +228,7 @@ export const CRASH_MUTATION_DISCARD = `
 `;
 
 /**
- * This is a GraphQL query template that updates existing records
- * of a specific type. E.g., it expects one or many Units, or Persons
- * or Crashes. The function name matches a Hasura insertion method.
+ * This is a GraphQL query that updates existing crash records.
  * @type {string}
  */
 export const CRASH_MUTATION_UPDATE = gql`
--- a/atd-vze/src/queries/crashes_changes.js
+++ b/atd-vze/src/queries/crashes_changes.js
@@ -1,201 +1,203 @@
 import { gql } from "apollo-boost";
 
-
+/// TODO: have we been keeping this up to date with new columns?
 /**
  * A template that is used to gather the original crash record
  * and all records of pending changes.
  * @type {gql}
  */
 export const GET_CRASH_CHANGE = gql`
-    query FindCrash($crashId: Int) {
-        atd_txdot_crashes(where: { crash_id: { _eq: $crashId } }) {
-            crash_id
-            crash_fatal_fl
-            cmv_involv_fl
-            schl_bus_fl
-            rr_relat_fl
-            medical_advisory_fl
-            amend_supp_fl
-            active_school_zone_fl
-            crash_date
-            crash_time
-            case_id
-            local_use
-            rpt_cris_cnty_id
-            rpt_city_id
-            rpt_outside_city_limit_fl
-            thousand_damage_fl
-            rpt_latitude
-            rpt_longitude
-            rpt_rdwy_sys_id
-            rpt_hwy_num
-            rpt_hwy_sfx
-            rpt_road_part_id
-            rpt_block_num
-            rpt_street_pfx
-            rpt_street_name
-            rpt_street_sfx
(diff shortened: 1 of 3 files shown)


atd-vze/src/queries/crashes_changes.js
```diff
@@ -1,201 +1,201 @@
 import { gql } from "apollo-boost";
-
-
 /**
  * A template that is used to gather the original crash record
  * and all records of pending changes.
  * @type {gql}
  */
 export const GET_CRASH_CHANGE = gql`
-    query FindCrash($crashId: Int) {
```
roseeichelmann, line 10: this is prettier formatting

atd-vze/src/queries/crashes_changes.js
```diff
@@ -264,20 +256,10 @@ export const RECORD_DELETE_CHANGE_RECORDS = `
       delete_atd_txdot_changes (
           where: {
             record_id: { _eq: $crashId }
-            record_type: { _in: ["crash", "unit", "primaryperson", "person"] }
+            record_type: { _in: ["crash", "unit", "primaryperson", "person", "charges"] }
```
roseeichelmann, line 259: this is why the changes table is full of old stale `record_type = charges` records, bc they were never being deleted.

atd-vze/src/views/Crashes/CrashChange.js
```diff
@@ -54,47 +48,35 @@ function CrashChange(props) {
   const [differentFields, setDifferentFields] = useState([]);
   const [showFieldsDiffOnly, setShowFieldsDiffOnly] = useState(true);
   // Modals
-  const [approveAllChanges, setApproveAllChanges] = useState(false);
-  const [discardAllChanges, setDiscardAllChanges] = useState(false);
-  const [clearAllSelections, setClearAllSelections] = useState(false);
+  const [approveAllModalOpen, setApproveAllModalOpen] = useState(false);
+  const [discardAllModalOpen, setDiscardAllModalOpen] = useState(false);
+  const [unselectAllModalOpen, setUnselectAllModalOpen] = useState(false);
```
roseeichelmann, line 53: updating these state names so its more obvious they are for handling the modals

atd-vze/src/views/Crashes/CrashChange.js
```diff
@@ -523,19 +502,35 @@ function CrashChange(props) {
...
+    updateCrashRecord({
+      variables: {
+        crashId: crashId,
+        changes: changesObject,
+      },
+    })
+      .then(() => {
+        refetch();
+      })
+      .catch(error => console.error(error));
```
roseeichelmann, line 533: was able to simplify this logic a lot and cut down on some of the code and functions that were being used previously

roseeichelmann:
How the conflict resolution system is currently working:

From the cris import script, all records going into the changes table are being entered as `record_type = crash` despite many of the record json changes being for person table or units changes.

The view `atd_txdot_changes_view` that the crashes changes page uses queries all change records that are of type `record_type = crash`

But the records that have `null` for `crash_date` aka change records for units or persons don’t show up in the UI bc we have the date calendar select and `null` dates don’t show up

But you can manually find these change records for units/persons by searching in the URL, though the UI is all broken for them like so:

image

For changes from CRIS to persons/units/charges records, we simply havent been displaying them in the UI at all. I'm very confused about whether any of this was by design or just slipped through the cracks unnoticed.

johnclary (approved):
@roseeichelmann i have tested this and as far as i can tell you have managed to restore this feature to its previously working state.

the amount of code you slashed gives me pause, only because i don't understand why we wouldn't take an approach similar to yours in the first place, but as far as I can tell you've been judicious in your refactoring.

really nice work—thank you! 

as you point out, this feature has some really major flaws. let's try to stay focused on the data model overhaul which should render this feature obsolete. 

🚢 🚢 🚢 🚢

johnclary:
not relevant to any of this work, but the date search functionality (which is applied whether you like it or not) is searching on the `crash_date` column in the changes view, which is extremely confusing because the only timestamp shown in the table is the `created_timestamp` of the change record. this appears to be because we hardcode `crash_date` into the search settings of the grid table component 😑

atd-vze/src/views/Crashes/CrashChange.js
```diff
@@ -103,6 +85,8 @@ function CrashChange(props) {
    */
   useEffect(() => setRecordData(data), [data]);
 
+  console.log(differentFieldsList, "different fields list");
```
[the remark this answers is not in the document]
roseeichelmann: yup! heh

atd-vze/src/queries/crashes_changes.js
chiaberry: can we delete all record_type = charges records?

chiaberry: I mean from the existing table, to fix what was being overlooked

atd-vze/src/views/Crashes/CrashChange.js
```diff
@@ -523,19 +504,34 @@ function CrashChange(props) {
...
    */
   const executeUpdateCrashRecord = async () => {
-    const updateQueries = await generateUpdateQuery();
-    setSaveStatus("updating crash record");
-    setUpsertRecordQuery(
-      gql`
-        ${updateQueries}
-      `
-    );
-    await sleep(1000);
```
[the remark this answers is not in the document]
mddilley: ooo nice catch. I guess it was intended to give time to read the UI status update text, but it also adds like 3 seconds to each conflict resolution rep! ⏳

atd-vze/src/queries/crashes_changes.js
```diff
@@ -229,30 +225,26 @@ export const CRASH_MUTATION_DISCARD = `
...
+    $crashId: Int!
+    $changes: atd_txdot_crashes_set_input
+  ) {
+    update_atd_txdot_crashes_by_pk(
+      pk_columns: { crash_id: $crashId }
+      _set: $changes
+    ) {
+      crash_id
+    }
+  }
```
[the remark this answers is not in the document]
mddilley: I agree that this is perfect for replicating the only part of this that has been working for a long while now and isn't a regression. I think that the original idea to make all changes look like crash changes even if they were unit or person ones made it hard to see that this was breaking over time.

It is a really nice catch! 🎉

atd-vze/src/queries/crashes_changes.js
mddilley: imo they can go. if we backfill the charges before the new data model to get them back in the crash page UI, all of the changes will be irrelevant anyways.

atd-vze/src/views/Crashes/CrashChange.js
```diff
@@ -601,241 +593,6 @@ function CrashChange(props) {
       });
   };
 
-  /**
-   * Returns true if the current field (key) is PII
-   * @param {string} recordType - The record type
-   * @param {string} key - The name of the field
-   * @returns {boolean}
-   */
-  const isFieldPII = (recordType, key) => {
```
mddilley, line 610: This caught my eye, but then I saw that the crash records don't have any fields in the PII config that this uses. 👍

mddilley (approved):
Tested this locally with some of the example record IDs in the issue, and it works just like in production and without bug (narrative with double quotes updated ✅ and change row deleted ✅). 🎉 Removing large chunks of code can be scary but I think that you've more than proven that the existing component isn't working as originally intended. Our code now reflects the actually functionality of this tool. 🙏

atd-vze/src/views/Crashes/CrashChange.js
roseeichelmann: same, also curious why PII fields were removed? why wouldnt we want those changes updated?

PR #1383, opened 2024-02-21
Role typo fix going into master

frankhereford:
## Associated issues

- https://github.com/cityofaustin/atd-data-tech/issues/15935

Slack discussion here: https://austininnovation.slack.com/archives/CM9MK950S/p1708539841539509.

## H/T

Thanks to @mddilley for diagnosing this and getting it fixed in production ⚡ fast.

## Testing

https://deploy-preview-1383--atd-vze-staging.netlify.app/#/

* You would need to add a user to the VZE application, and opt to not specify a role.
  * In doing so, the default will be used, which is now going to be `readonly` and not `readOnly`. 
* If this works as intended, and the users page does not fail with a runtime error, then ✅.
* Be sure to clean up any changes you make in Auth0.

Due to the very small amount of code that changed, I also am comfortable with y'all 👀ing this one and not making changes in Auth0, but either is great with me.

---
#### Ship list
- [ ] Code reviewed


atd-vze/src/views/Users/Users.js
```diff
@@ -96,7 +96,7 @@ const UserRow = ({ user }) => {
           ? format(parseISO(user.last_login), "MM/dd/yyyy")
           : "Never"}
       </td>
-      <td>{rules[user.app_metadata.roles[0]].label}</td>
+      <td>{rules[user.app_metadata.roles[0]]?.label}</td>
       <td>
```
johnclary, line 100: @frankhereford one really minor change. how about adding a fallback text string here so it doesn't render `undefined`. i think rendering `"Error"` would suffice. Just to make it clear that something went horribly wrong 🤷

johnclary (approved):

johnclary (approved):

PR #1392, opened 2024-03-05
Updates fetch policy to use cache-and-network everywhere

johnclary:
## Associated issues

- https://github.com/cityofaustin/atd-data-tech/issues/15884

I noticed the Crashes and Locations never refetch, so I went ahead and updated the policy everywhere. Here are the [Apollo docs](https://www.apollographql.com/docs/react/data/queries/#setting-a-fetch-policy) on this topic.

## Testing
**URL to test:** Netlify

Bounce around between the various pages and views in the app, making sure hit the Crashes, Locations, and Fatalities lists. Those pages should remain hydrated as you navigate away and then back to them.

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vze/src/views/Fatalities/Fatalities.js
+++ b/atd-vze/src/views/Fatalities/Fatalities.js
@@ -12,11 +12,6 @@ import {
 
 // Our initial query configuration
 let queryConf = {
-  options: {
-    useQuery: {
-      fetchPolicy: "no-cache",
-    },
-  },
   table: "view_fatalities",
   single_item: "crashes",
   showDateRange: true,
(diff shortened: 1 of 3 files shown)


atd-vze/src/views/Fatalities/Fatalities.js
mddilley, line 1: This is reminding me that someone in a past sprint review mentioned not being able to tell when this table is loading when paginating it since it takes longer than the other too. Maybe we could add a spinner somewhere to be more transparent about when data is loading like we recently did in Moped.

mddilley (approved):
I tested all three views, and I see the cache-and-network behavior happening. ✨ Thanks for updating! 🙏 🚀

atd-vze/src/views/Fatalities/Fatalities.js
johnclary: oh thanks for mentioning that. yes—for sure. i have been meaning to open a table enhancements epic 😈

PR #1393, opened 2024-03-07
Align CRIS import to the editable fields of the VZE's crash page

frankhereford:
## Associated issues

This PR is related to this work: https://github.com/cityofaustin/atd-data-tech/issues/16157.

## Testing

Local:
* Create a `env` file for the CRIS import with the 3 usual suspect entries to enable 1PW API access
```
OP_API_TOKEN=
OP_CONNECT=
OP_VAULT_ID=
```
* Spin up the DB + graphql endpoint
* Replicate production data into the DB
* Get a handle on what's inside the changes table, with something like this:
```
select count(*)
from atd_txdot_changes;
```
* Execute:
```
docker compose run cris-import
```
* Inspect the changes table and the fact tables to the CSVs that were extracted and see if you're happy with the results.

---
#### Ship list
- [ ] Code reviewed


frankhereford:
All of the changes that I want to do are completed here; I'm just waiting for an execution of the script to run end to end before I put it up for review.

frankhereford:
My local simulation of the import ran today without issue. It took about 90 minutes locally.

johnclary (commented):
@frankhereford this looks great—thank you! one last request, would you mind double checking if the period (`.`) in [these entries](https://github.com/cityofaustin/atd-vz-data/blob/24ff334f95deb128b81cce2715a5523b1d9604dc/atd-etl/cris_import/lib/mappings.py#L55) is needed or if they can be removed?

frankhereford:
@johnclary wrote:
>  would you mind double checking if the period (`.`) in [these entries](https://github.com/cityofaustin/atd-vz-data/blob/24ff334f95deb128b81cce2715a5523b1d9604dc/atd-etl/cris_import/lib/mappings.py#L55) is needed

Sure thing John, you bet. I'm starting another import run with the explicit table removed. I grabbed the extract that we got this morning (Sunday's reprocessing) which is the smallest we've received over the last few days to do a quick import to check it. It looks good, so I've removed it. 

I was trying to figure a reason why I stuck it in there -- I may have been trying to be explicit about the table since this is a column name that is used in all of our fact tables (query below). Maybe I had a join in there at some point that required it, but I don't think we do any longer. It's outta there!

```
SELECT 
    table_name
FROM 
    information_schema.columns
WHERE 
    table_name IN (
        SELECT 
            table_name
        FROM 
            information_schema.tables
        WHERE 
            table_type = 'BASE TABLE'
    )
    AND column_name = 'sus_serious_injry_cnt';
```

atd-etl/cris_import/development_extracts/.gitignore
frankhereford, line 1: I stuck this `.gitignore` in here to hold open this directory when we check this out to remind us about this feature which is very handy for local testing, and to prevent committing any extracts too, as well.

johnclary (approved):
thanks for doing that! this looks great. let's gooo 🚢 

Reminder that this needs to make it's way into production in order for our ETL image to rebuild

atd-etl/cris_import/development_extracts/.gitignore
johnclary: nice 🙏

frankhereford:
@johnclary wrote:

> Reminder that this needs to make it's way into production in order for our ETL image to rebuild

Thank you!! I needed that reminder :pray: 

https://github.com/cityofaustin/atd-vz-data/pull/1394

mddilley (approved):
The updates look good to me! Thanks for investigating, documenting, and getting the important ETL fields aligned with the editable ones! 🙏 🚢 🚀

PR #1399, opened 2024-03-14
Update CrashEditCoordsMap

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/15867

**Please review these PRs in this order: https://github.com/cityofaustin/atd-vz-data/pull/1398, https://github.com/cityofaustin/atd-vz-data/pull/1399, https://github.com/cityofaustin/atd-vz-data/pull/1400**

This one reimplements the `CrashEditCoords` with `react-map-gl` v7 and prevents map zoom from happening on page scroll.

## Testing
**URL to test:** 
Local

Steps to test:

1. FYI you will need to run `npm i` in the editor folder when switching back and forth between this branch and others without the map library update
2. Go to a crash page and then click on the **Edit Coordinates** button to show the crash location edit map
3. See that there are zoom and fullscreen controls in the top left just like in staging and production
4. ~Try scrolling while hovering the mouse over the crash map. You should see a prompt to use a cooperative gesture to zoom the map in and out.~ This is outdated. It should zoom on scroll.
5. Drag the map, and you should see the marker animations happening just like they do in staging and production. Also notice that the values in the latitude/longitude inputs in the form below the map update when the marker is moved.
6. After moving the marker, click the **Reset** button and the marker and map should jump to the original location of the crash before you started editing.
7. Click the **Cancel** button and the edit map should close and not save a new position.
8. Open the edit map again, move the marker, and then click **Save**. You should see that all three buttons now disable while the update request is happening. You might have to throttle your network speed in browser to see this in action. The edit map closes after the update succeeds.
9. Verify that the location updated by refreshing the page and checking it in the crash map.

---
#### Ship list
- [x] Code reviewed


atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
mddilley, line 1: The biggest update in this file is the old `react-map-gl` version required us to ask a cursor state if the map was being dragged. The new version exposes the Mapbox events called `onDragStart`, `onDrag`, and `onDragEnd` to hook into and manage dragging state.

Also, the geocoder is removed. I talked to John about it, and he is okay with nixing it for now and modifying https://github.com/cityofaustin/atd-data-tech/issues/15175 to capture the steps needed to use the native Mapbox geocoder plugin that `react-map-gl` v7 enables us to do like in Moped.

atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
```diff
@@ -1,189 +1,130 @@
...
+        </Marker>
+        {/* add nearmap raster source and style */}
+        {!isDev && <LabeledAerialSourceAndLayer />}
+      </MapGL>
+      <CrashEditLatLonForm
+        {...markerCoordinates}
+        handleFormSubmit={handleFormSubmit}
+        handleFormReset={handleFormReset}
+        handleFormCancel={handleFormCancel}
+        isSubmitting={isSubmitting}
```
mddilley, line 124: I added this `isSubmitting` state so we can disable the form buttons while the new form values save. Open to any feedback here since it was out of scope.

frankhereford (approved):
# ✅ This PR  

Mike, it's another winner. I've gone through your testing instructions locally, and I have read over your code. Everything looks great and makes sense. Your map experience is really showing here. ✨ 

I'm looking forward to the exciting conclusion to the trilogy in the next PR!

### Things not about this PR that came to mind

* I think we should consider rounding the latitude and longitude values when we show them below the edit map to 6 decimal places. At our latitude, a `0.000001 x 0.000001` area of decimal degrees is about the size of a postcard. We can store whatever resolution postGIS settles on, but I think we don't need show the user anything beyond what is helpful.

mddilley:
@frankhereford Thanks for mentioning that - I agree that the # of decimal places is out of control (and looks pretty sloppy in the editor form too)! We punted on it in the past, but I think that your point about using what PostGIS goes to is 🎯 🙏 and I'll add that into this PR.

atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
```diff
@@ -1,189 +1,130 @@
...
+      <MapGL
+        ref={mapRef}
+        initialViewState={{
+          latitude: latitude_primary || defaultInitialState.latitude,
+          longitude: longitude_primary || defaultInitialState.longitude,
+          zoom: defaultInitialState.zoom,
+        }}
+        style={{ width: "100%", height: "50vh" }}
+        {...mapParameters}
+        cooperativeGestures={true}
```
johnclary, line 105: i think we might want to leave this this off (`false`) when editing the coordinates. going on the assumption that you probably want scroll capture in edit mode. does that sound right?

atd-vze/src/helpers/map.js
```diff
@@ -123,3 +123,11 @@ export const LabeledAerialSourceAndLayer = () => {
...
   );
 };
+
+/**
+ * Shorten a coordinate to 6 decimal places
+ * @param {string|number} coordinate - latitude or longitude
+ * @returns {string} - truncated coordinate
+ */
+export const truncateCoordinate = coordinate =>
+  parseFloat(coordinate).toFixed(6);
```
johnclary, line 133: nice—i like de-crufting the super "precise" lat/lons that we get from the input

johnclary (approved):
looks great; works great. really awesome work, Mike! i think we should remove coop gestures for this map. otherwise this needs to ship 🚢 🚢 🚢 🚢 🚢 🚢

atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
mddilley: that sounds right to me too, and thanks for catching. copypasta double strike! (test steps and code) 🍝

atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
```diff
@@ -1,189 +1,129 @@
...
-    });
+const CrashEditCoordsMap = ({
+  data,
+  mapGeocoderAddress,
+  crashId,
+  refetchCrashData,
+  setIsEditingCoords,
+}) => {
+  const { user } = useAuth0();
+  const { email = null } = user;
```
roseeichelmann, line 29: oo i dont think ive seen this syntax before, could you explain whats going on here?

atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
roseeichelmann: this is a really nice touch, great attention to detail!! :sparkles:

atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
mddilley: definitely! destructuring allows for default values to be assigned in case that a value isn't present. Pretty handy but I've stumbled when working with deeply nested objects before https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#default_value

roseeichelmann (approved):
this all worked perfectly, and i really like some of the details you added like disabling the buttons on save and truncating the coords! :ship: :ship:

PR #1403, opened 2024-03-18
[Enhancement] Increase change log modal width and wrap text

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/15894

This makes a few updates to the change log modal to prevent horizontal scrolling from happening on screens with plenty of real estate to show the change log data.

## Testing
**URL to test:** 
https://deploy-preview-1403--atd-vze-staging.netlify.app/

**Steps to test:**
1. Find a crash with change log entries like: https://deploy-preview-1403--atd-vze-staging.netlify.app/#/crashes/19448665
2. Go to the bottom of the page and click **Compare** on one of the change log rows
3. When the modal opens, notice that it takes up the majority of the width available in your browser window
4. Also, notice as you decrease the browser width, the text in the current and previous columns wrap and break to prevent side-scrolling early on. `cr3_file_metadata` and `investigator_narrative_ocr` are a good indicators of this behavior. The OCR narrative was specifically causing uneven table column widths between the current and previous values.

---
#### Ship list
- [x] Code reviewed

### DIFF
--- a/atd-vze/src/views/Crashes/CrashChangeLog.js
+++ b/atd-vze/src/views/Crashes/CrashChangeLog.js
@@ -165,8 +165,12 @@ class CrashChangeLog extends Component {
       content = <p>No changes found for this record.</p>;
     } else {
       modal = (
-        <Modal isOpen={this.state.modal} className="mw-100 mx-5">
-          <ModalHeader>Record Differences</ModalHeader>
+        <Modal
+          isOpen={this.state.modal}
+          toggle={this.closeModal}
+          className="mw-100 mx-5"
+        >
+          <ModalHeader toggle={this.closeModal}>Record Differences</ModalHeader>
           <ModalBody>{this.state.modalBody}</ModalBody>
           <ModalFooter>
             <Button color="secondary" onClick={this.closeModal}>
--- a/atd-vze/src/views/Crashes/CrashChangeLog.js
+++ b/atd-vze/src/views/Crashes/CrashChangeLog.js
@@ -148,10 +148,14 @@ class CrashChangeLog extends Component {
         <tr key={`recordHistory-${i}`}>
           <td>{item.original_record_key}</td>
           <td>
-            <Badge color="primary">{String(item.original_record_value)}</Badge>
+            <Badge color="primary" className="text-wrap">
+              {String(item.original_record_value)}
+            </Badge>
           </td>
           <td>
-            <Badge color="danger">{String(item.archived_record_value)}</Badge>
+            <Badge color="danger" className="text-wrap">
+              {String(item.archived_record_value)}
+            </Badge>
           </td>
         </tr>
       );
@@ -193,7 +197,7 @@ class CrashChangeLog extends Component {
       content = <p>No changes found for this record.</p>;
     } else {
       modal = (
-        <Modal isOpen={this.state.modal} className={this.props.className}>
+        <Modal isOpen={this.state.modal} className="mw-100">
           <ModalHeader>Record Differences</ModalHeader>
           <ModalBody>{this.state.modalBody}</ModalBody>
           <ModalFooter>
--- a/atd-vze/src/views/Crashes/CrashChangeLog.js
+++ b/atd-vze/src/views/Crashes/CrashChangeLog.js
@@ -148,12 +148,12 @@ class CrashChangeLog extends Component {
         <tr key={`recordHistory-${i}`}>
           <td>{item.original_record_key}</td>
           <td>
-            <Badge color="primary" className="text-wrap">
+            <Badge color="primary" className="text-wrap text-break">
               {String(item.original_record_value)}
             </Badge>
           </td>
           <td>
-            <Badge color="danger" className="text-wrap">
+            <Badge color="danger" className="text-wrap text-break">
               {String(item.archived_record_value)}
             </Badge>
           </td>
@@ -197,7 +197,7 @@ class CrashChangeLog extends Component {
       content = <p>No changes found for this record.</p>;
     } else {
       modal = (
-        <Modal isOpen={this.state.modal} className="mw-100">
+        <Modal isOpen={this.state.modal} className="mw-100 mx-5">
           <ModalHeader>Record Differences</ModalHeader>
           <ModalBody>{this.state.modalBody}</ModalBody>
           <ModalFooter>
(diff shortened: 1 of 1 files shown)


atd-vze/src/views/Crashes/CrashChangeLog.js
```diff
@@ -118,37 +118,47 @@ class CrashChangeLog extends Component {
...
+      if (key === "__typename") {
         return undefined;
       }
       return value;
-    }
+    };
 
     // For each entry created in the diff array, generate an HTML table row.
-    let modalItems = diff.map(item => {
+    let modalItems = diff.map((item, i) => {
```
mddilley, line 128: this fixes a key warning. Index is okay here since the order of these items won't change after initial render.

atd-vze/src/views/Crashes/CrashChangeLog.js
```diff
@@ -118,37 +118,47 @@ class CrashChangeLog extends Component {
...
-        </td>
-        <td>
-          <Badge color="danger">{String(item.archived_record_value)}</Badge>
-        </td>
-      </tr>
-      )
+        <tr key={`recordHistory-${i}`}>
+          <td>{item.original_record_key}</td>
+          <td>
+            <Badge color="primary" className="text-wrap text-break">
```
mddilley, line 151: new classes here to handle the word wrap

atd-vze/src/views/Crashes/CrashChangeLog.js
```diff
@@ -187,7 +197,7 @@ class CrashChangeLog extends Component {
       content = <p>No changes found for this record.</p>;
     } else {
       modal = (
-        <Modal isOpen={this.state.modal} className={this.props.className}>
+        <Modal isOpen={this.state.modal} className="mw-100 mx-5">
```
mddilley, line 200: There is a `modal-xl` class in Bootstrap 4 but it wasn't super wide so I made this take up the whole width and added some margin prevent it from rendering all the way to the edge.

atd-vze/src/views/Crashes/CrashChangeLog.js
```diff
@@ -211,7 +221,7 @@ class CrashChangeLog extends Component {
             </thead>
             <tbody>
               {this.props.data.atd_txdot_change_log.map(record => (
-                <tr key={`changelog-${record.id}`}>
+                <tr key={`changelog-${record.change_log_id}`}>
```
mddilley, line 224: One more key warning fix

frankhereford (approved):
Mike, thank you for this. This makes this tool instantly much more useful and easier to use. As it is, this is already a huge improvement and is shippable. 🚢 

I was wondering about what you thought this change, and totally no big if you think it's a bad idea.

You can see in this screenshot, that the horizontal scrollbar is still there, however, your changes make it largely moot, since it seems to be able to show >95% of the width no matter how I resize the viewport. 

has_scroll_bar

Do you think it would be possible to change the `overflow-x` value in some version of this deeply specified tag:
```
body > div:nth-child(12) > div > div.modal.fade.show > div > div > div.modal-body > section > div
```
Maybe something like this as the selector? I don't have a copy of the DB replicated with changelog data handy; I have not tested, so take me with a grain of salt!
```
div.modal.fade.show > div.modal-body
```
Here's maybe a better demonstration of what I'm trying to convey:


https://github.com/cityofaustin/atd-vz-data/assets/3653206/563e82fd-194d-41e5-90d8-d65abfcf12e8

Please let me know if I can help explain this better; I feel like I've done a pretty lackluster job here. I also know this is a style set deep in bootstrap on a tag deep in a component we didn't write, so I totally understand if this is more trouble than it's worth.

chiaberry (approved):

chiaberry:
This is such an improvement over what exists now

@frankhereford is there a place in vze where you can override styles? or are you suggesting to add a class to that nested div? I cannot remember how the classes work there, versus how they work in material ui. 

fwiw, I think the scrollbar is fine, even if it is vestigal

frankhereford:
@chiaberry wrote:

> @frankhereford is there a place in vze where you can override styles? or are you suggesting to add a class to that nested div? I cannot remember how the classes work there, versus how they work in material ui.

Great question Chia, I had to look into it myself to figure out a way. I am suggesting a new class to target the div that contains the table of changed fields, yea. Here's what I came up with when looking at how it might be doable. I'm not trying to say this is the best way, but it's a way that seems pretty OK -- what do you think?

https://github.com/cityofaustin/atd-vz-data/compare/md-15894-change-log-modal..apply-custom-css

mddilley:
@frankhereford @chiaberry just catching up here. thanks for the reviews, and thanks Frank for catching the sneaky scrollbar 🙏 I'm going to push an update that will have a similar effect to your example, Frank. 

I want to give John a chance to check if this is what he envisioned in the first place too. The static column widths are great for large screens, but I'm thinking that we might want a scrollbar on smaller screen? Not sure.

atd-vze/src/views/Crashes/CrashChangeLog.js
```diff
@@ -168,7 +168,7 @@ class CrashChangeLog extends Component {
         <h6>Edited Date: {this.timeConverter(record.update_timestamp)}</h6>
         <h6>Updated by: {record.updated_by || "Unavailable"}</h6>
         &nbsp;
-        <Table responsive>
+        <Table responsive className="overflow-hidden">
```
frankhereford, line 171: Such an elegant solution - thanks @mddilley ! 🙌

roseeichelmann (approved):
Huggeeeee improvement! Thanks mike it looks so much better now! :sparkles:

johnclary (changes requested):
@mddilley this is a really nice improvement! thanks for cleaning up our code along the way.

i have one bit of scope creep, which is not critical. would you mind adding a close button to the modal header, in addition to the existing button in the footer? now that this modal is scrollable i found that i got a tiny bit lost trying to close it 😅

mddilley:
@johnclary good catch on the modal! when i compared it to the confirmation modal, we were missing the `toggle` prop in the modal and modal header. This gives us the X icon, closing with the esc key, and clickaway to close. Once the deploy is updated, this is ready for another look. 🙏

johnclary (approved):
niiiiice. much improved! 🚢 🚢 🚢 🚢 🚢 🚢

mddilley:
thanks, everybody! merging

PR #1409, opened 2024-03-20
Fix bug where searching without a column chosen doesn't show warning on first search

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/15454

This fixes a bug where the GridTable search didn't show the expected messages or search behavior until a local storage configuration was initialized in the browser local storage.

## Testing
**URL to test:** 
https://deploy-preview-1409--atd-vze-staging.netlify.app/#/crashes

**Steps to test:**
1. Go to the test link and then delete the key called `jestersavedCrashesConfig`. I think that we should capture the debt around this saved config that I mentioned [here](https://github.com/cityofaustin/atd-data-tech/issues/15454#issuecomment-2010765513). Frank and John had some good ideas when we patched this back in the day, and we've addressed something like this in Moped now too.
2. Refresh the crashes page, and you should see that crash ID is selected in the search dropdown. You should also see that the search button is disabled since the search value input isn't complete. There will also be a warning message to enter a search value.
3. Change the chosen field in the dropdown to Case ID, and the click the **Reset** button. The dropdown should reset to show the default field, crash ID, and the warning message will be back.
4. Try these same steps in staging, and you will see that the UI lets you search before selecting a field and does not show a message about the missing field until the local storage search config is initialized.
5. Check the Locations list loads normally since this PR makes a small update to GridTable.

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vze/src/Components/GridTableSearch.js
+++ b/atd-vze/src/Components/GridTableSearch.js
@@ -89,12 +89,6 @@ const GridTableSearch = ({
   return (
     <Col md={12}>
       <Form className="form-horizontal" onSubmit={handleSearchSubmission}>
-        {!isFieldSelected && isSearchValueEntered && (
-          <Alert color="warning">Please provide a field to search.</Alert>
-        )}
-        {isFieldSelected && !isSearchValueEntered && (
-          <Alert color="warning">Please enter a value to search.</Alert>
-        )}
         <FormGroup>
           <InputGroup>
             <Input
--- a/atd-vze/src/Components/GridTableSearch.js
+++ b/atd-vze/src/Components/GridTableSearch.js
@@ -88,8 +88,11 @@ const GridTableSearch = ({
   return (
     <Col md={12}>
       <Form className="form-horizontal" onSubmit={handleSearchSubmission}>
-        {!isFieldSelected && searchFieldValue && (
+        {!isFieldSelected && isSearchValueEntered && (
           <Alert color="warning">Please provide a field to search.</Alert>
+        )}
+        {isFieldSelected && !isSearchValueEntered && (
+          <Alert color="warning">Please enter a value to search.</Alert>
         )}
         <FormGroup>
           <InputGroup>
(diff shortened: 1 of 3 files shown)


atd-vze/src/Components/GridTableSearch.js
```diff
@@ -91,9 +88,12 @@ const GridTableSearch = ({
   return (
     <Col md={12}>
       <Form className="form-horizontal" onSubmit={handleSearchSubmission}>
-        {!isFieldSelected && searchFieldValue && (
+        {!isFieldSelected && isSearchValueEntered && (
           <Alert color="warning">Please provide a field to search.</Alert>
         )}
+        {isFieldSelected && !isSearchValueEntered && (
+          <Alert color="warning">Please enter a value to search.</Alert>
+        )}
```
mddilley, line 96: This was the most low-effort way to indicate that a search value is missing which is the opposite situation that already had a warning message to address. The warning messages can still only be cleared by clicking the reset button.

I think this seems okay but would like some input. 🙏

roseeichelmann (approved):
Nice catch with this bug and i think the fix works great! :ship:

johnclary (commented):
@mddilley thanks for cleaning this up and making the feature work as it was intended.

in terms of level of effort, what would you think about defaulting the search field to `crash_id` when there is no filter in local storage? we would also reset the search field to  to `crash_id` when someone clicks the **Clear** button. 

i'd also like to rename the button from **Clear** to **Reset** since it does not only clear the search input—it resets all the advanced search toggles as well. 

i know we need to make more substantial improvements to the search experience, so if you think it makes sense to table this for later that's cool with me 👍

mddilley:
@johnclary That sounds good to me, and I don't think it is a lot of effort at all. 🙏 I'll make those updates now and re-request review.

johnclary (approved):
this looks great, Mike. thanks for removing the banner entirely. we will revisit this and the [other issues you flagged](https://github.com/cityofaustin/atd-data-tech/issues/15454#issuecomment-2010765513) a little further down the road.

PR #1412, opened 2024-03-22
Adds tooltip to travel mode component

tillyw:
## Associated issues
fixes https://github.com/cityofaustin/atd-data-tech/issues/15987

## Testing
**URL to test:** 
https://deploy-preview-1412--atd-vzv-staging.netlify.app/

**Steps to test:**
- click the new info icon to the right of the "By Travel Mode" header and confirm the modal pops up with the correct information!
- as requested, there should no longer be a fade-in animation on any of the info popover modals ✨ no starwipe i'm sorry ❌ ❇️ 😢 

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vzv/src/Components/Popover/InfoPopover.js
+++ b/atd-vzv/src/Components/Popover/InfoPopover.js
@@ -39,6 +39,7 @@ const InfoPopover = ({ config }) => {
         isOpen={modal}
         toggle={toggle}
         zIndex={1305} // Set z-index to supercede SideDrawer and SideMapControlDateRange components
+        fade={false}
         scrollable
         autoFocus
       >
--- a/atd-vzv/src/views/summary/CrashesByMode.js
+++ b/atd-vzv/src/views/summary/CrashesByMode.js
@@ -6,6 +6,8 @@ import styled from "styled-components";
 import { format } from "date-fns";
 
 import CrashTypeSelector from "./Components/CrashTypeSelector";
+import InfoPopover from "../../Components/Popover/InfoPopover";
+import { popoverConfig } from "../../Components/Popover/popoverConfig";
 import { colors } from "../../constants/colors";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import {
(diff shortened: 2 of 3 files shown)


atd-vzv/src/views/summary/CrashesByMode.js
```diff
@@ -222,7 +224,9 @@ const CrashesByMode = () => {
     <Container className="m-0 p-0">
       <Row>
         <Col>
-          <h2 className="text-left, font-weight-bold">By Travel Mode</h2>
+          <h2 className="text-left, font-weight-bold">By Travel Mode{" "}
```
mddilley, line 227: Could you removed the comma next to this `text-left` class since you are in here? That snuck its way in there somewhere in the past.

Also, no change needed, but I noticed we use a space character around the app to create space between elements. Probably something we could have reached to CSS styles to achieve!

mddilley (approved):
Looks great, and I love the new info! ⭐✨ 🚢

atd-vzv/src/Components/Popover/InfoPopover.js
```diff
@@ -39,6 +39,7 @@ const InfoPopover = ({ config }) => {
         isOpen={modal}
         toggle={toggle}
         zIndex={1305} // Set z-index to supercede SideDrawer and SideMapControlDateRange components
+        fade={false}
```
johnclary, line 42: thank you!

atd-vzv/src/views/summary/CrashesByMode.js
johnclary: thanks for pointing that out, Mike. i would also rather see spacing addressed here with a padding class rather than `" "`, but i guess we should tackle that separately so that we can fix it everywhere at once?

tillyw: thanks, y'all! @mddilley I dropped the comma, thanks for noticing that! and +1 to fixing the spacing using css throughout the app. going to go ahead and merge for the next release !

PR #1414, opened 2024-03-26
Show a banner at the top of the screen when using a non-production environment

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/15892

## Testing
**URL to test:** 
Local and netlify. https://deploy-preview-1414--atd-vze-staging.netlify.app/#/crashes

**Steps to test:**
1. Check out the netlify link, youll see a banner at the top saying you are in a staging environment.
2. Check out the branch locally and start up your local DB and VZE, you will see a banner saying you are in a local environment.
3. Lmk thoughts on color preferences for the banner.

---
#### Ship list
- [ ] Code reviewed


atd-vze/src/containers/DefaultLayout/DefaultHeader.js
```diff
@@ -21,64 +22,84 @@ const propTypes = {
 
 const defaultProps = {};
 
+const EnvAlertBanner = () => {
+  const env = process.env.REACT_APP_HASURA_ENDPOINT.includes("staging")
+    ? "staging"
+    : process.env.NODE_ENV === "production"
+    ? "production"
+    : "local";
```
roseeichelmann, line 30: is this nested ternary too gnarly? :sweat_smile: just using NODE_ENV shows production for netlify and isnt helpful for specifying local vs staging. using it in combo with REACT_APP_HASURA_ENDPOINT i could specify all the diff environments

johnclary: ya know, i kinda think we should go ahead and use a dedicated env variable for this like we do in Moped. i think we can just add `REACT_APP_ENV` to all of the `.env` files. it may come in handy again down the road anyway...

mddilley: I don't think the ternary is too gnarly at all. Not likely that we'll have more than these envs. I was originally going to say the staging check was alright, but John's review pushed my opinion over to updating the env files.

Also, making the alert yellow (`warning`) in the local env would be a cool touch to match up what we see in Moped Editor. That is not blocking this feature though - love the banner!

mddilley (approved):
Thanks for adding this. It will be nice to have the same visual cue as we do in Moped! 🙌 🚀 🚢

atd-vze/src/containers/DefaultLayout/DefaultHeader.js
mddilley: @johnclary after your requested change is implemented, we may want to create a spin-off to tidy up the existing `isDev` var sprinkled around the app 🏗️

frankhereford (approved):

atd-vze/src/containers/DefaultLayout/DefaultHeader.js
johnclary: oh, thanks for mentioning that. will do 👍 🙏

atd-vze/.env
```diff
@@ -5,3 +5,4 @@ REACT_APP_AUTH0_CLIENT_ID=2qbqz2sf8L9idBOwn0d5YA9efNgQbL7c
 REACT_APP_MAPBOX_TOKEN="pk.eyJ1Ijoiam9obmNsYXJ5IiwiYSI6ImNqemU4OGQzaTBlZjczYm1rZWg1c2lidTcifQ.2eooq4Q7dv_1Le5la9D96Q"
 REACT_APP_NEARMAP_KEY="YzI5YTM1ZWMtMWRkNy00NGU4LTkxMWQtNDI3ZDcxNjVjMzg3"
 REACT_APP_CR3_API_DOMAIN="https://atd-vz-api-staging.austinmobility.io"
+REACT_APP_ENV="local"
```
roseeichelmann, line 8: new env variables! john and i decided to just put `local` here even though when you spin up the local stack you could point to staging data depending on what `REACT_APP_HASURA_ENDPOINT` is commented out. maybe we want to think about doing away with the staging option here, but i do find it helpful for now when i just have a front end code change that i want to check without pulling a new copy of prod data and starting up the local db which can take some time.

mddilley (approved):
Tested the new code, and it works great! Sooo nice to have these env env vars and the banner colors too. 🎨 🙏 🚀 🚢 🚀

johnclary (approved):
thanks @roseeichelmann! 🚢 🚢 🚢

PR #1418, opened 2024-03-28
[Bug] The the location polygon is not visible on the location map

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/16527

This PR adds a `beforeId` prop to the `LabeledAerialSourceAndLayer` component so that we can control where it falls in the stack of map layers if needed. I also added a mock layer so we can see if it is working correctly or not when running the local stack. Not sure if that should be permanent, but it gave me more confidence that this will work once merged to staging. I could see it being confusing when there is a big blue layer for no apparent reason.

## Testing
**URL to test:** 
Local

**Steps to test:**
1. Start the local stack and go to a location page
2. You should see that there is a blue see-through layer that falls behind the yellow outline of the location polygon.
3. To test this further, you can add/remove the `beforeId` prop [here](https://github.com/cityofaustin/atd-vz-data/blob/f7dc2cfb100827164ea0f44722c3507bd9bb145d/atd-vze/src/views/Locations/LocationMap.js#L63) and adjust the `fill-opacity` to `1` [here](https://github.com/cityofaustin/atd-vz-data/blob/f7dc2cfb100827164ea0f44722c3507bd9bb145d/atd-vze/src/helpers/map.js#L130) to make the order really clear.
4. In staging and production, the NearMap aerial tiles will show and use the same `beforeId` prop.
5. Last, check the crash view and crash location edit maps to make sure they still load as expected.
6. In the deploy preview, we can add the same behavior once we have our environment config more ironed out in https://github.com/cityofaustin/atd-vz-data/pull/1414.
7. Long term, we can [check with CTM to see if we can get wildcards into our NearMap key allow list](https://austininnovation.slack.com/archives/CM9MK950S/p1711643282116059?thread_ts=1711638110.820439&cid=CM9MK950S)

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vze/src/Components/GridTableSearch.js
+++ b/atd-vze/src/Components/GridTableSearch.js
@@ -12,7 +12,6 @@ import {
   DropdownToggle,
   DropdownItem,
   DropdownMenu,
-  Alert,
 } from "reactstrap";
 
 // TODO add query operators to each field to better fit data types (_eq, etc.)?
--- a/atd-vze/src/helpers/map.js
+++ b/atd-vze/src/helpers/map.js
@@ -1,5 +1,7 @@
 import React from "react";
 import { Source, Layer } from "react-map-gl";
+import { isDev } from "./environment";
+import { colors } from "../styles/colors";
 
 const TOKEN = process.env.REACT_APP_MAPBOX_TOKEN;
 // This API key is managed by CTM. Contact help desk for maintenance and troubleshooting.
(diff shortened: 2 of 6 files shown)


atd-vze/src/Components/GridTableSearch.js
```diff
@@ -12,7 +12,6 @@ import {
   DropdownToggle,
   DropdownItem,
   DropdownMenu,
-  Alert,
```
mddilley, line 15: compile warning leftover from one of my PRs 👋

atd-vze/src/helpers/map.js
```diff
@@ -111,13 +113,43 @@ export const LOCATION_MAP_CONFIG = {
...
+};
+
+/** Source and layer to display NearMap aerials with street labels on top.
+ * Localhost and deploy preview URLs are not on the NearMap tile API key allow list.
+ * To test locally, this returns a mock layer to test the layer ordering.
+ * Adjust the opacity in the mockPolygonDataLayer paint object for further testing.
+ * @param {string} beforeId - layer id to place these layers before
+ */
+export const LabeledAerialSourceAndLayer = ({ beforeId }) => {
+  return isDev ? (
```
mddilley, line 141: we were conditionally rendering this component in each use, but I think it makes more sense for it to consider environment itself now. idk

atd-vze/src/helpers/map.js
```diff
@@ -1,5 +1,7 @@
 import React from "react";
 import { Source, Layer } from "react-map-gl";
+import { isDev } from "./environment";
```
johnclary, line 3: funny to see this come up immediately after you mentioned it yesterday 😅

atd-vze/src/helpers/map.js
johnclary: yes—i agree!

johnclary (approved):
nice! thanks for stitching this up, and i really appreciate that you added a testing path. 

i would lean toward leaving the mock layer in the app but having it disabled by default. but, i'm of two minds. leaving it on by default is a good reminder that we have a missing layer in the env we're testing 🤷 

in either case—i will follow-up with CTM about our allow list so this can hopefully become a non-issue.

thanks for your quick work on, Mike!! 🚢 🚢 🚢 🚢 🚢 🚢

atd-vze/src/helpers/map.js
mddilley: 😅 heh yep, it was top of mind since we mostly use it in this map code! now we are down to 2 uses in the app. 👍

johnclary:
@mddilley i'm merging this so that I can wrap up testing 🙏

PR #1419, opened 2024-03-29
Thread cr3 downloader & clean-up `stdout` with a verbose flag

frankhereford:
## Associated issues

This PR intends to close: https://github.com/cityofaustin/atd-data-tech/issues/16556

## Testing 🤔 

🎗️ First, please rebuild with `docker compose build`, both when you start to test and when you switch back to `master`.

I guess, maybe we should just test it for a few weeks while we download CR3s? 
Maybe trade days? 

We can decide on this in a dev sync. I'm not positive what is best and easiest. 

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-etl/cr3_download/cr3_download.py
+++ b/atd-etl/cr3_download/cr3_download.py
@@ -154,14 +154,6 @@ with ThreadPoolExecutor(max_workers=max_workers) as executor:
     for future in futures:
         future.result()
 
-
-# for crash_record in crashes_list_without_skips:
-#     process_crash_cr3(
-#         crash_record,
-#         CRIS_BROWSER_COOKIES,
-#         skipped_uploads_and_updates,
-#     )
-
 print("\nProcess done.")
 
 if skipped_uploads_and_updates:
--- a/atd-etl/cr3_download/cr3_download.py
+++ b/atd-etl/cr3_download/cr3_download.py
@@ -11,156 +11,180 @@ any CR3 files associated.
 
 import os
 import time
-import json
+import argparse
 from concurrent.futures import ThreadPoolExecutor
 
-from process.helpers_cr3 import *
-
-from onepasswordconnectsdk.client import Client, new_client
+from process.helpers_cr3 import process_crash_cr3, get_crash_id_list
+from onepasswordconnectsdk.client import new_client
 import onepasswordconnectsdk
 
 
-# Start timer
-start = time.time()
+def load_secrets(one_password_client, vault_id):
+    """Load required secrets from 1Password."""
+    required_secrets = {
+        "HASURA_ENDPOINT": {
+            "opitem": "Vision Zero graphql-engine Endpoints",
+            "opfield": "production.GraphQL Endpoint",
+            "opvault": vault_id,
+        },
+        "HASURA_ADMIN_KEY": {
+            "opitem": "Vision Zero graphql-engine Endpoints",
+            "opfield": "production.Admin Key",
+            "opvault": vault_id,
+        },
+        "AWS_ACCESS_KEY_ID": {
+            "opitem": "CR3 Download IAM Access Key and Secret",
+            "opfield": "production.accessKeyId",
+            "opvault": vault_id,
+        },
+        "AWS_SECRET_ACCESS_KEY": {
+            "opitem": "CR3 Download IAM Access Key and Secret",
+            "opfield": "production.accessSecret",
+            "opvault": vault_id,
+        },
+        "AWS_DEFAULT_REGION": {
+            "opitem": "CR3 Download IAM Access Key and Secret",
+            "opfield": "production.awsDefaultRegion",
+            "opvault": vault_id,
+        },
+        "ATD_CRIS_CR3_URL": {
+            "opitem": "Vision Zero CRIS CR3 Download",
+            "opfield": "production.ATD_CRIS_CR3_URL",
+            "opvault": vault_id,
+        },
+        "AWS_CRIS_CR3_BUCKET_NAME": {
+            "opitem": "Vision Zero CRIS CR3 Download",
+            "opfield": "production.AWS_CRIS_CR3_BUCKET_NAME",
+            "opvault": vault_id,
+        },
+        "AWS_CRIS_CR3_BUCKET_PATH": {
+            "opitem": "Vision Zero CRIS CR3 Download",
+            "opfield": "production.AWS_CRIS_CR3_BUCKET_PATH",
+            "opvault": vault_id,
+        },
+    }
+
+    return onepasswordconnectsdk.load_dict(one_password_client, required_secrets)
 
 
-# Get 1Password secrets from environment
-ONEPASSWORD_CONNECT_HOST = os.getenv("OP_CONNECT")
-ONEPASSWORD_CONNECT_TOKEN = os.getenv("OP_API_TOKEN")
-VAULT_ID = os.getenv("OP_VAULT_ID")
-
-# Setup 1Password server connection
-one_password_client = new_client(ONEPASSWORD_CONNECT_HOST, ONEPASSWORD_CONNECT_TOKEN)
-
-# Get secrets from 1Password
-REQUIRED_SECRETS = {
-    "HASURA_ENDPOINT": {
-        "opitem": "Vision Zero graphql-engine Endpoints",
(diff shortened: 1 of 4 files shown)


atd-etl/cr3_download/cr3_download.py
```diff
@@ -5,145 +5,186 @@
...
+                process_crash_cr3_threaded,
+                crash_record,
+                CRIS_BROWSER_COOKIES,
+                skipped_uploads_and_updates,
+                args.verbose,
+            )
+            futures.append(future)
+
+        for future in futures:
+            future.result()
```
frankhereford, line 171: This is the meat of the changes. Almost all of this is linting / or changing the indention / block level.

atd-etl/cr3_download/docker-compose.yml
```diff
@@ -1,9 +1,9 @@
-version: '3'
-
 services:
   cr3_download:
     build: .
     volumes:
       - .:/app
     env_file:
       - ./.env
+    # entrypoint: /bin/bash
```
frankhereford, line 8: I want to leave in this breadcrumb to aid in recalling how to get a shell easily on this image.

atd-etl/cr3_download/process/helpers_cr3.py
frankhereford, line 1: I was tempted to bring these functions into the main file; I'm not really sold on the benefit of having them split out, but I didn't. All the changes here are me passing down a `verbose` boolean to control the amount of output the script makes.

chiaberry:
I'm looking forward to testing this out when I'm on cr3 duty next week, and I'll think about what updates need to be made to the readme / gitbook instructions while I use it

atd-etl/cr3_download/cr3_download.py
```diff
@@ -5,145 +5,186 @@
...
-is obtained from Hasura, and it is contingent to records that do not have
+is obtained from Hasura, and it is made up of the records that do not have
 any CR3 files associated.
 """
 
 import os
 import time
-import json
+import argparse
+from concurrent.futures import ThreadPoolExecutor
```
johnclary, line 15: i think this is a great choice for this use case. [asyncio](https://docs.python.org/3/library/asyncio.html) would probably add even more zippiness, but it would introduce substantial refactoring (and 3rd party libs) to get all the `async` operations working 👍 🚢

johnclary (changes requested):
i have been running this script all week, and @frankhereford this is such a massive QOL improvement—thank you!!

it would be nice to take this opportunity to clean up the logging, but this is otherwise ready to ship.

thanks again, @frankhereford! 🚢 🚢 🚢 🚢 🚢

atd-etl/cr3_download/process/helpers_cr3.py
```diff
@@ -152,17 +161,20 @@ def process_crash_cr3(crash_record, cookies, skipped_uploads_and_updates):
...
             print(f"\nFile {download_path} is not a pdf - skipping upload and update")
+            with open(download_path, "r") as file:
+                print(file.read())
+            time.sleep(10)
             skipped_uploads_and_updates.append(crash_id)
         else:
-            upload_cr3(crash_id)
-            update_crash_id(crash_id)
+            upload_cr3(crash_id, verbose)
+            update_crash_id(crash_id, verbose)
```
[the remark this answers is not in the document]
frankhereford: Yes! I love this sort of thing. I'm more than happy to bring this into scope. I normally am a little bit lax about leaving stuff hanging around on disk when it's in an ephemeral docker container, but still, I totally like the idea of doing it as right as possible. Thanks for thinking about this!

atd-etl/cr3_download/cr3_download.py
frankhereford: Total agreement from me. This started as a very "quick'n'dirty" hack, and since it was totally a non-requested change, I was going for whatever was quickest. I boxed my time to only what I could reasonably expect to gain back by making the changes for the 30K downloads we had queued up. Since this seems to be working, I'm inclined to agree with your sentiment to leave it be and 🚢

frankhereford:
@chiaberry wrote:

> I'm looking forward to testing this out when I'm on cr3 duty next week, and I'll think about what updates need to be made to the readme / gitbook instructions while I use it

Oh, such a good call Chia! I embedded a little micro-documentation behind a `-h` flag, but it absolutely deserves an update to the gitbook docs. 🙏 Thank you for suggesting this, and I'm 100% going to include it prior to shipping.

mddilley (approved):
This script has more than proven itself! I like John's suggested improvements and enjoyed reading about different directions that we could take this in the future. Thanks for leveling this script up! 🍄 I feel ready to run it next week now that I've read through the code. 🙏

atd-etl/cr3_download/cr3_download.py
```diff
@@ -5,145 +5,186 @@
...
+
+def main():
+    # Parse command-line arguments
+    parser = argparse.ArgumentParser(description="CRIS - CR3 Downloader")
+    parser.add_argument(
+        "-t",
+        "--threads",
+        type=int,
+        default=1,
+        help="Number of concurrent downloaders (default: 1)",
```
chiaberry, line 93: so the default concurrent downloader is 1, but the docker-compose entrypoint overrides this to be 5

`entrypoint: /app/cr3_download.py -t 5`

wondering if the default should just be 5? or take the `-t 5` part off of the entrypoint?

frankhereford:
@johnclary wrote:

> 👍 would you make it a bread loaf and add a comment explaining that it enables a shell?

Yep, great idea. I extended the documentation that @chiaberry added with this commit. https://github.com/cityofaustin/atd-vz-data/pull/1419/commits/31d4fb78b00b501ca7ea9b45ca355210c46c2d63

atd-etl/cr3_download/cr3_download.py
```diff
@@ -5,145 +5,186 @@
...
+    )
+    parser.add_argument(
+        "-v", "--verbose", action="store_true", help="Enable verbose logging"
```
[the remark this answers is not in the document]
frankhereford: Sounds good. I am not going to try to persist the logs back out of the container, but I do think it's a helpful pattern to explicitly define what level a log entry belongs to. https://github.com/cityofaustin/atd-vz-data/pull/1419/commits/e27c3a9738a0aadf8fcab1d653a62d9382c76448

atd-etl/cr3_download/cr3_download.py
```diff
@@ -5,145 +5,186 @@
...
+        crashes_list_without_skips = [
+            x for x in crashes_list if x["crash_id"] not in known_skips
+        ]
+
+    except Exception as e:
+        crashes_list_without_skips = []
+        print(f"Error, could not run CR3 processing: {str(e)}")
+
+    if args.verbose:
+        print(f"Length of queue: {len(crashes_list_without_skips)}")
```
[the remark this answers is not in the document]
frankhereford: You bet, I left that `info` level logging directive.

atd-etl/cr3_download/cr3_download.py
frankhereford: This is a great question Chia, thanks for asking about it and for catching that mismatch. I think there's no absolute right or wrong answer here, and fwiw, here's the reasoning that went into these choices. 

I think of the default in the program itself as being the default for when there is no contextual information about how the program is going to be used for what and by whom and that this default should represent the most conservative and safe choice, which is to just have one thread. 

Then, in the docker compose stack, I view that configuration as being in our own, specific context, and that it should reflect how we use the program. Under this thinking, that is where we can loosen up the configuration to a setting that is good for our particular use.

I realize that we don't really expect that there are other users of this software out there, and that it's open source for transparency and not for reuse, at least at this time, so I totally get that my reasoning is entirely conceptual and not practical, but at least that's what I was thinking when I picked those numbers.

frankhereford:
Based on @chiaberry's thoughtful request, I have added some more documentation to the gitbook page that describes the options for the program. Thanks!

With this, I believe this should be ready to be re-reviewed by John and any other devs who would like to approve it. 🙏

johnclary (approved):
@frankhereford thanks for updating the gitbook docs 🙏 

i agree we can revisit the logging cleanup later. and in this case we may be able to retire the script in favor of CRIS pdf deliver to S3 🤞

PR #1421, opened 2024-04-01
[BUG] VZV: Missing "Other" modes crashes on the map

mddilley:
## Testing
**URL to test:** 
TBD

**Steps to test:**
TBD

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vzv/src/views/map/Map.js
+++ b/atd-vzv/src/views/map/Map.js
@@ -6,7 +6,10 @@ import MapPolygonFilter from "./MapPolygonFilter";
 import MapCompassSpinner from "./MapCompassSpinner";
 import { createMapDataUrl, useMapEventHandler } from "./helpers";
 import { mapInit, travisCountyBboxGeoJSON, mapNavBbox } from "./mapData";
-import { crashGeoJSONEndpointUrl } from "../../views/summary/queries/socrataQueries";
+import {
+  crashGeoJSONEndpointUrl,
+  mapRequestFields,
+} from "../../views/summary/queries/socrataQueries";
 import {
   baseSourceAndLayer,
   fatalitiesDataLayer,
@@ -101,6 +104,7 @@ const Map = () => {
       filters,
       dateRange,
       mapPolygon,
+      mapRequestFields,
       mapTimeWindow
     );
 
--- a/atd-vzv/src/views/nav/SideMapTimeOfDayChart.js
+++ b/atd-vzv/src/views/nav/SideMapTimeOfDayChart.js
@@ -16,7 +16,7 @@ const fieldsToRequest = [
   "crash_date",
 ];
 
-export const SideMapTimeOfDayChart = ({ filters }) => {
+export const SideMapTimeOfDayChart = ({ timeWindowConfig }) => {
   const chartRef = useRef();
 
   const defaultBarColor = colors.dark;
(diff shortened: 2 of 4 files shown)


atd-vzv/src/views/map/Map.js
```diff
@@ -101,6 +104,7 @@ const Map = () => {
       filters,
       dateRange,
       mapPolygon,
+      mapRequestFields,
```
mddilley, line 107: I added a parameter to the helper that builds up the [Socrata $select parameter](https://dev.socrata.com/docs/queries/select.html) so we can control what fields are returned when this function is reused. We were requesting unneeded geo fields here previously.

atd-vzv/src/views/nav/SideMapTimeOfDayChart.js
```diff
@@ -33,36 +40,47 @@ export const SideMapTimeOfDayChart = ({ filters }) => {
...
-            accumulator[i]++;
+
+          const crashFatalities = parseInt(crash.death_cnt);
+          const seriousInjuries = parseInt(crash.sus_serious_injry_cnt);
+          const isCrashInTimeWindow =
+            crashHour >= timeWindow[0] && crashHour <= timeWindow[1];
+
+          if (isCrashInTimeWindow) {
+            accumulator[i] = accumulator[i] + crashFatalities + seriousInjuries;
+          }
```
mddilley, line 76: We used to add up the total number of crash records per time of day window that returned using the map filters for severity, mode, and date range. 

Now, we add up the number of fatalities and serious injuries for each crash record.

roseeichelmann (approved):
The code looks great to me and this works well, but i do find this change a little confusing. I think of the crash map with all the dots as showing data on crashes, not people, and when you filter to injuries, you are showing all the crashes with injuries. And the total number in the histogram should be how many crashes happened within that time frame, at least thats how i expect it to work. It seems weird to be going from crashes being shown on the map to number of people being shown in the histogram. If we do go with this change i think we need a tooltip to specify what is the true meaning of the numbers in the histogram ie. its the number of crashes or the number of people. 

Also this whole thing is making me realize that the meaning of "All" both in the map and summary pages is kind of ambiguous. I could see users thinking All means show all crashes or all people despite level of injury, but the meaning is really "show all fatality or serious injury crashes". Just some food for thought!

For example, maybe having just two buttons on the map for Fatalities and Serious Injuries and you can toggle both or just one and do away with the All button would make it more apparent that we are never displaying all the crashes that happen in Austin. (obvi a completely diff issue and out of scope for this)

mddilley:
@roseeichelmann thanks!, and I agree that it is worth thinking about ways to indicate exactly what is displayed in this viz. Getting oriented on what data was coming into the component and what is displayed wasn't super straightforward when i was working on this.

For the **All** button, I see exactly what you are saying. It does kind of imply "All Crashes with no filters" (neither) instead of "All crashes that have a fatality or serious injury" (both). I remembered that we changed this once before and found [this PR](https://github.com/cityofaustin/atd-vz-data/pull/735/files) that shows that we once had just the two buttons but [we changed it](https://github.com/cityofaustin/atd-data-tech/issues/2096) because it was confusing.  Not a lot of crumbs as to why it was considered confusing so I think it could use another look. That was a long time ago now!

Curious what @johnclary thinks too. Happy to keep iterating on this one!

chiaberry (approved):
Cosign on what Rose said -- the code looks great, I appreciate all the work you put in to cleaning up the code and refining the request. 

I do think its a little confusing to see three dots on the map, but then 5 in the histogram. But also think it should match what is on the charts on the VZV, so it makes sense to see 5 in the histogram. 

Text saying the histogram shows people and the map shows incidents would be helpful. I was also thinking if the dots were bigger if there were more than 1 person injured or killed, if the size of the dot corresponded to people.

frankhereford (approved):
I agree with the sentiment that the data being displayed in the histogram isn't clear. Additionally, the calendar icon in it is also misplaced or unclear in my opinion.

image

I'd suggest we can help reduce this confusion by striking that calendar icon and replacing it with a header for the histogram that says "Injuries."

However - for what this PR set out to do, it's 💯! Thank you Mike!

PR #1424, opened 2024-04-03
[Enhancement] Persist sidebar visibility between sessions

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/16006

This PR adds a local storage a setting for whether the app side bar is open or closed and uses it to update the app to reflect that value. CoreUI controls the side bar with a class added to or removed from the `body` tag of `index.html`. This falls outside of the DOM node that the React app is appended to so we can't use [React refs](https://react.dev/reference/react/useRef#manipulating-the-dom-with-a-ref) here. But, we can use the [MutationObserver API](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) and query selectors to reach outside the app and watch the class changes.

## Testing
**URL to test:** 
https://deploy-preview-1424--atd-vze-staging.netlify.app/

**Steps to test:**
1. Go to any view of the app
2. Check your local storage for a key called `isAppSideBarOpen`. On the first time this key is initialized, it should have a value of `true` and the sidebar should be open.
3. Refresh the page, the local storage value should remain `true`.
4. Now click on the hamburger icon on the side bar header to close the side bar. The local storage value should now be `false`.
5. Refresh the page or close the app and come back. When the app loads, it should load with the side bar closed.
6. Try updating the value of `isAppSideBarOpen` in your browser settings manually to either `true` or `false` and refresh. The app will use this value to either hide or show the side bar on load.
7. Try out different combinations, and your local storage value should always match the status of the side bar.

---
#### Ship list
- [x] Code reviewed

### DIFF
--- a/atd-vze/src/containers/DefaultLayout/DefaultLayout.js
+++ b/atd-vze/src/containers/DefaultLayout/DefaultLayout.js
@@ -26,22 +26,30 @@ const DefaultFooter = React.lazy(() => import("./DefaultFooter"));
 const DefaultHeader = React.lazy(() => import("./DefaultHeader"));
 
 const useIsAppSideBarOpen = () => {
-  const [isAppSideBarOpen, setIsAppSideBarOpen] = React.useState(null);
+  const localStorageIsAppSideBarOpen = localStorage.getItem("isAppSideBarOpen");
+  const doesLocalStorageExist = localStorageIsAppSideBarOpen !== null;
 
-  console.log(isAppSideBarOpen);
+  const [isAppSideBarOpen, setIsAppSideBarOpen] = React.useState(
+    doesLocalStorageExist ? JSON.parse(localStorageIsAppSideBarOpen) : true
+  );
 
-  // Sync state with classes on body-outside-app
+  // On mount, observe the body element for changes in the class attribute and update the state
+  // and local storage when CoreUI toggles the sidebar
   React.useEffect(() => {
     const elemToObserve = document.getElementById("body-outside-app");
 
     const observer = new MutationObserver(function(mutations) {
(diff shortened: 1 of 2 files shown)


johnclary (approved):
this worked for me at first, and then when i opened my debug console it stopped working, and so i debugged it some more, and then i finally realized that when my debug console is open, the `sidebar-lg-show` class is not active because the sidebar is autohiding. d'oh!


this is fantastic. and TIL about `MutationObserer`s. this is a really nice fix, @mddilley! thank you! 🚢 🚢 🚢 🚢 🚢

atd-vze/src/containers/DefaultLayout/DefaultLayout.js
```diff
@@ -25,10 +25,69 @@ const DefaultAside = React.lazy(() => import("./DefaultAside"));
...
+  const [isAppSideBarOpen, setIsAppSideBarOpen] = React.useState(
+    doesLocalStorageExist ? JSON.parse(initialIsAppSideBarOpen) : true
+  );
+
+  // On mount, set an observer that watches the body element for changes in the class attribute
+  // and updates the state and local storage to match the classes used by CoreUI to show/hide the sidebar
+  React.useEffect(() => {
+    const elemToObserve = document.getElementById(appBodyId);
+
+    const observer = new MutationObserver(function(mutations) {
```
roseeichelmann, line 50: this is super cool! big TIL for me :)

roseeichelmann (approved):
Works like a charm, this was such a cool solution mike and nicely executed :raised_hands:

mddilley:
thanks y'all! Mergin' 🚢

PR #1426, opened 2024-04-04
Write SQL to create lookup tables in the new data model

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/16635

## Testing
**URL to test:** 
You could test on ur local db.

**Steps to test:**
1. Start up your local db, could be johns data model db even. Run the up and down migrations.
2. If you are using another db that already has lookup tables, you could probs first run the down sql to remove the tables to then test the up migrations sql.

---
#### Ship list
- [ ] Code reviewed


atd-toolbox/get_lookup_table_changes/README.md
roseeichelmann, line 1: oops i branched off this branch from my PR here... https://github.com/cityofaustin/atd-vz-data/pull/1395

didnt actually do any changes to the get_lookup_table_changes folder in this PR

atd-toolbox/insert_needed_lkp_tables/insert_lkp_tables.py
```diff
@@ -0,0 +1,127 @@
...
+
+def escape_single_quotes(input_string):
+    return input_string.replace("'", "''")
+
+
+def new_table(name):
+    return f"""
+    create table db.{name} (
+        id integer primary key,
+        label text not null
```
roseeichelmann, line 43: do we like `label`? `desc` cant be used bc its a pg special term

atd-toolbox/insert_needed_lkp_tables/insert_lkp_tables.py
```diff
@@ -0,0 +1,127 @@
...
+    return grouped_data
+
+
+def escape_single_quotes(input_string):
+    return input_string.replace("'", "''")
+
+
+def new_table(name):
+    return f"""
+    create table db.{name} (
```
roseeichelmann, line 41: will have to be updated when we decide on schema names :)

atd-toolbox/insert_needed_lkp_tables/insert_lkp_tables.py
roseeichelmann, line 1: This program is similar to get_lookup_table_changes.py but a lil diff bc dont actually need to connect to pg and only need to create insert statements for a select list of tables.. dont know if we will actually wanna carry this script into our repo but put it in this PR incase we do/so yall can see what i did

atd-toolbox/insert_needed_lkp_tables/insert_lkp_tables.py
```diff
@@ -0,0 +1,127 @@
...
+        if table in needed_lkp_tables:
+            extract_table_dict = extract_data[table]
+            changes.append(f"\n-- Adding table {table}")
+            changes.append(new_table(table))
+            down_changes.append(f"\n-- Dropping table {table}")
+            new_table_down = f"drop table if exists db.{table};"
+            down_changes.append(new_table_down)
+            for key in extract_table_dict:
+                insert = f"insert into db.{table} (id, label) values ({str(key)}, '{escape_single_quotes(extract_table_dict[key])}');"
+                changes.append(insert)
```
johnclary, line 110: @roseeichelmann this all looks great. one thing that would be good to add is a check to see if the script was able to find a table for each of the tables listed in `needed_lkp_tables`. in case I typo'd a lookup table name or referenced one that is supposed to be in the CSV but for whatever reason is not.

roseeichelmann: Ooo really good call with this comment.. i just added this functionality in and found that some of the tables are indeed not in the extract I was using:
['drvr_ethncty_lkp', 'movt_lkp', 'substnc_cat_lkp', 'substnc_tst_result_lkp', 'trvl_dir_lkp', 'veh_direction_of_force_lkp']
So i think we need an updated list

johnclary (changes requested):
@roseeichelmann this looks fantastic—thank you! i'd like to merge this, and then i'm going to make a few tweaks to get aligned with the migrations i've been writing for the new data model.

for now, would you go ahead and remove the actual migration files from this PR? we will re-generate them later.

roseeichelmann:
@johnclary sounds good, just removed them!

johnclary (approved):

PR #1427, opened 2024-04-08
Preserve old CR3 metadata when CRIS updates a crash

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/16684

## Testing

I'm not sure what the best way is to test the CRIS import ETL. Looking for guidance here :) 

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-etl/cris_import/lib/sql.py
+++ b/atd-etl/cris_import/lib/sql.py
@@ -25,7 +25,7 @@ def get_pgfutter_path():
 
 def invalidate_cr3(pg, crash_id):
     invalidate_cr3_sql = f"""UPDATE public.atd_txdot_crashes 
-    SET cr3_stored_flag = 'N', cr3_file_metadata = null, cr3_ocr_extraction_date = null
+    SET cr3_stored_flag = 'N'
     WHERE crash_id = {crash_id}"""
     cursor = pg.cursor(cursor_factory=psycopg2.extras.RealDictCursor)
     cursor.execute(invalidate_cr3_sql)


chiaberry (approved):
didn't test but 🚢

atd-etl/cris_import/lib/sql.py
```diff
@@ -25,7 +25,7 @@ def get_pgfutter_path():
 
 def invalidate_cr3(pg, crash_id):
     invalidate_cr3_sql = f"""UPDATE public.atd_txdot_crashes 
-    SET cr3_stored_flag = 'N', cr3_file_metadata = null, cr3_ocr_extraction_date = null
+    SET cr3_stored_flag = 'N'
```
mddilley, line 28: i'm going to pick back up on this later so I wanted to drop some notes because I'm finding it hard to trace the dependencies between the cris import, cr3 download, narrative extract, and populate CR3 metadata scripts.

I'm looking at the [query that the CR3 metadata script uses](https://github.com/cityofaustin/atd-vz-data/blob/34066e5d3c79a479c5c2503106d38dd6439f4d09/atd-etl/populate_cr3_file_metadata/populate_cr3_file_metadata.py#L226-L233) and [the one that the OCR extract one uses](https://github.com/cityofaustin/atd-vz-data/blob/34066e5d3c79a479c5c2503106d38dd6439f4d09/atd-etl/cr3_extract_diagram/cr3_extract_diagram_ocr_narrative.py#L161-L177). I'll take a closer look at this later but wanted to share in case anyone else is looking at this.

frankhereford:
How does this know when a crash has been meaningfully updated?

atd-etl/cris_import/lib/sql.py
frankhereford: @mddilley - thanks for these great questions. 

Here's a little bit of info that might be helpful, at least in terms of the OCR script. Back when the OCR script was first put into place, I was behind a little bit of questionable database design. The `cr3_ocr_extraction_date` field is a nullable timestamp, and it's intended to be both the most recent time that OCR was completed on the CR3 and, if null, to mean that OCR is required / requested for the CR3.

It's the implied meaning of the `null` value that doesn't jump off the page for me, and I wish I had not opted for that choice now, but it's not entirely wrong either ... it's not great.

I certainly don't think that I've answered Mike's questions here, but it was a little bit of history that might be helpful.

johnclary: ugh, thanks @mddilley for mentioned those other scripts. i had only looked at the CR3 download query, but overlooked the obvious way that these other ETLs depend on this field. so this PR as written will have the effect of causing CR3s to never be re-OCR'd after they're downloaded anew.

i am going to cancel my review request and probably move this into the backlog. it feels like something that we can pick up again in the data model work.

mddilley: nah, i don't think that it is obvious at all! thanks for thinking through this, y'all! (i was way too distracted w/ the eclipse yesterday). I do think that we could get the switches right if we put all our heads together, but, yep, agreed that it will be nice to take another swing at this in the near future!

PR #1428, opened 2024-04-10
Add loading spinners to cards

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/16100

## Testing
**URL to test:** 
https://deploy-preview-1428--atd-vzv-staging.netlify.app/

**Steps to test:**
1. The spinners happen super fast sometimes so you may need to throttle your network to actually see anything. But just go to the summary page of the VZV and look for the new spinners while the summary cards are loading

Looking for design feedback so lmk what yall think. Heres a screenshot I took of when all the spinners are going so yall can see it easier:

image

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vzv/src/views/summary/CrashesByMode.js
+++ b/atd-vzv/src/views/summary/CrashesByMode.js
@@ -1,7 +1,7 @@
 import React, { useEffect, useState, useRef, useMemo } from "react";
 import axios from "axios";
 import { Bar } from "react-chartjs-2";
-import { Container, Row, Col } from "reactstrap";
+import { Container, Row, Col, Spinner } from "reactstrap";
 import styled from "styled-components";
 import { format } from "date-fns";
 
@@ -224,8 +224,8 @@ const CrashesByMode = () => {
     <Container className="m-0 p-0">
       <Row>
         <Col>
-          <h2 className="text-left font-weight-bold">By Travel Mode{" "}
-            <InfoPopover config={popoverConfig.summary.byMode} />
+          <h2 className="text-left font-weight-bold">
+            By Travel Mode <InfoPopover config={popoverConfig.summary.byMode} />
           </h2>
         </Col>
       </Row>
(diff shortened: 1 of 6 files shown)


atd-vzv/src/views/summary/CrashesByMode.js
```diff
@@ -239,160 +240,170 @@ const CrashesByMode = () => {
...
-                                legendColors[i] !== "dimgray"
-                                  ? legendColorsClone.splice(i, 1, "dimgray")
-                                  : legendColorsClone.splice(
-                                      i,
-                                      1,
-                                      chartColors[i]
-                                    );
-                                setLegendColors(legendColorsClone);
-                              };
+      {!!data.datasets ? (
```
roseeichelmann, line 246: really only adding this but the shift is causing a whole bunch of diff changes

atd-vzv/src/views/summary/CrashesByMode.js
```diff
@@ -239,160 +240,170 @@ const CrashesByMode = () => {
...
+                    },
+                  }}
+                />
+              </Container>
+            }
+          </Col>
+        </Row>
+      ) : (
+        <h1>
+          <ColorSpinner />
```
roseeichelmann, line 404: new spinner here

atd-vzv/src/views/summary/CrashesByMode.js
mddilley, line 246: Way out of scope, but it is interesting to think about how we could use these spinners as a fallback for when the ETL fails every once in a blue moon and the API returns an empty array of data records. We'll probably have a chance to avoid that problem altogether with the new data model, though! 🤞

mddilley (approved):
Tested this out while throttling my connection, and this is a huge improvement! Especially with the newly-evened-out card heights. It is fun to watch everything snap into place. ✨ 🚢

chiaberry (approved):
I throttled down to slow 3G in chrome, cleared my cache and saw the spinners. Much better experience than seeing empty graphs and charts. 

🚢

johnclary (approved):
Such a nice improvement. thank you!! 🚢

PR #1429, opened 2024-04-11
Check for CR3 metadata when showing the download CR3 button

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/16759

## Testing
**URL to test:** Local

I believe you'll need to test locally because I was not able to locate a record in staging that is missing CR3 file metadata.

**Steps to test:**
1. Find a crash that has a stored CR3 but no file metadata: 
```sql
select crash_id from atd_txdot_crashes
  where cr3_stored_flag = 'Y'
  and cr3_file_metadata is null limit 1;```

2. Visit the crash details page for this crash and verify that the **DOWNLOAD CR3 PDF** button is not visible

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vze/src/views/Crashes/Crash.js
+++ b/atd-vze/src/views/Crashes/Crash.js
@@ -278,7 +278,7 @@ function Crash(props) {
         <Col xs="12" md="6" className="mb-4">
           <CrashDiagram
             crashId={crashId}
-            isCr3Stored={cr3StoredFlag === "Y"}
+            isCr3Stored={cr3StoredFlag === "Y" && !!cr3FileMetadata}
             isTempRecord={tempRecord}
             cr3FileMetadata={cr3FileMetadata}
           />

### CODE REVIEW

File: atd-vze/src/views/Crashes/Crash.js [hunk 1]
```diff
@@ -278,7 +278,7 @@ function Crash(props) {
         <Col xs="12" md="6" className="mb-4">
           <CrashDiagram
             crashId={crashId}
-            isCr3Stored={cr3StoredFlag === "Y"}
+            isCr3Stored={cr3StoredFlag === "Y" && !!cr3FileMetadata}
```
mddilley, line 281: @johnclary I was reviewing this, and I wanted to double check that this is working like how we think. I checked out https://visionzero.austin.gov/editor/#/crashes/15208414 in prod. There is a pdf in S3 for it, but it matches `cr3_stored_flag = Y` and `cr3_file_metadata IS NULL`. I might be misunderstanding when we want to hide the button.

roseeichelmann: thats an interesting point mike... thats also the crash that i checked out, i didnt realize we actually do have a cr3 for it. in this case it would be nice to be able to download that cr3 even though the metadata column is null?

roseeichelmann: also, so far xavier letting us know when hes getting the `The specified key does not exist` error has helped flag underlying issues in our import script or cr3 downloader.. so maybe just hiding the button could also bite us and let things going wrong under the hood slip thru the cracks

roseeichelmann (approved):
The test steps work but i also have the same questions as mike about whether this is hiding the button when we wouldnt want it to be hidden as well..

atd-vze/src/views/Crashes/Crash.js
johnclary: ugh, thank you. i got myself confused about the underlying issue w/ downloading CR3s from prod. the real problem is/was that the `cr3_stored_flag` was just plain wrong. the existing behavior is what we want. i'm closing this 😞

PR #1435, opened 2024-04-12
Multiple NO DATA options in Body Style edit drop down

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/13602

Realized that the real weirdness here is that we had added a custom lookup value to `atd_txdot__veh_body_styl_lkp` at some point with `id = -1` and `desc = 'NO DATA'`. (I'm positive its a custom value, i checked and double checked the extracts and it was the only one without eff beg and end dates) This is confusing bc for the rest of the Related Records table `'NO DATA'` on the front end means `null` on the back end. I removed this custom lookup value and created a permanent `NO DATA` option on all the editable drop down fields which just maps to null so users can null out any values. I think this makes the most sense as a fix, it seems like we were trying to create a custom lookup value that means null instead of just using null.

## Testing
**URL to test:** 
Local

**Steps to test:**
1. Check out this branch locally, apply the up migration.
2. Start up VZE, go to a crash and observe that now all of the editable fields in the Related Records table have a NO DATA option which really just means null in the database. Also observe that if you select NO DATA for body style and click on the edit tool again, you wont see two NO DATA options in the drop down anymore.
3. Test selecting the NO DATA option for all the diff values in the related records table and double check in the database that the value is `null`
4. Test down migration

---
#### Ship list
- [x] Code reviewed


atd-vze/src/views/Crashes/RelatedRecordsTable.js
```diff
@@ -208,10 +208,7 @@ const RelatedRecordsTable = ({
                                     }
                                     type="select"
                                   >
-                                    {/* Show a NO DATA option only when formatValue is displayed. */}
-                                    {formatValue(row, field) === "NO DATA" && (
-                                      <option value={null}>NO DATA</option>
-                                    )}
+                                    <option value={""}>NO DATA</option>
```
roseeichelmann, line 211: always have a null option :sparkles:

atd-vze/src/views/Crashes/RelatedRecordsTable.js
```diff
@@ -271,7 +268,8 @@ const RelatedRecordsTable = ({
                             )}
 
                             {!isEditing &&
-                              (fieldConfig.fields[field].badge ? (
+                              (fieldConfig.fields[field].badge &&
+                              formatValue(row, field) != "NO DATA" ? (
```
roseeichelmann, line 272: for injury severity we dont want a badge if its null, we want the styled NO DATA value to match the rest of the table

atd-vzd/migrations/default/1712941514048_delete_no_data_option/up.sql
```diff
@@ -0,0 +1,4 @@
+-- Remove the custom "NO DATA" lookup option, instead we should just use null to mean null
+DELETE FROM public.atd_txdot__veh_body_styl_lkp WHERE veh_body_styl_id = -1;
+
+UPDATE public.atd_txdot_units SET veh_body_styl_id = null WHERE veh_body_styl_id = -1;
```
roseeichelmann, line 4: lolz that awkward moment when we dont have foreign key constraints so i was able to do the first deletion and then there were still units that had -1 for veh_body_styl_id even though we dont have a lookup value for that anymore :woozy_face:

roseeichelmann: also btw this second UPDATE command only effects 3 record units. there are only 3 units where the body style has an id of -1

atd-vze/src/views/Crashes/RelatedRecordsTable.js
mddilley: also, i was initially confused about seeing `""` here as the value prop until i traced back to `handleInputChange` and see that we handle converting to `null` there. Thanks for working inside the bounds of this code. 🙏

mddilley (approved):
Works great to fix the bug. 👋 🐛 I replicated in staging, and I cannot on this branch. 🙌 It will be interesting to see how we deal with the -1 lookup values in the new model. fwiw I saw them in other tables like `atd_txdot__prsn_type_lkp` but not in the lookup csvs in the extracts that I have locally from the last couple days. 🚢 🚀 🚢

atd-vze/src/views/Crashes/RelatedRecordsTable.js
roseeichelmann: oo good catch thanks mike! :monocle_face:

johnclary (approved):
@roseeichelmann i'm really happy to see these changes. i was just scratching my head at the three `-1`'s i came across while working on the new units table in the data model. this is great.

i'm struggling a bit with the implications of this VZE change for the new data model. in the new model, it's not possible to override a CRIS value with a user-provided `null` value, because we always fallback to the CRIS value when the user-provided value is `null`.

on the other hand—we've not discussed this—but under the new data model i think we will need some way to enable a VZE user to nullify/reset the user-provided value so that the CRIS value can take precedent again. so this "NO DATA" option could actually provide that functionality.

all that to say, i'm good with this change, but we may need to revisit it in the near future.

🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

johnclary:
> if any other parts of the app are ill-equipped to handle a null in some of these related record columns. I don't think so because I nulled out everything possible, but I wanted to say it out loud.

thanks @mddilley — i also tested this and double-checked the column definitions and i also think we're good.

roseeichelmann:
@johnclary yeah definitely will need to revisit this with the new data model, and i agree something like this could provide the functionality to nullify the vze edit and "revert" to a cris value.

PR #1436, opened 2024-04-15
Thread cr3 file metadata ETL

frankhereford:
## Associated issues

This PR aims to close https://github.com/cityofaustin/atd-data-tech/issues/16746.

## Testing

Local testing for this one.

* Spin up local DB and replicate production data down onto it. 
* Create a `env` file or get one from me
* `docker compose build`
* `docker compose run metadata`
* `./populate_cr3_file_metadata.py -s 100` for example. Try `-h` to explore the options.

---
#### Ship list
- [ ] Code reviewed


docker-compose-docker-volume.yml
frankhereford, line 1: This is a tag-along change. We need to, non-urgently, remove the deprecated, `version` designation from our compose files. The same goes for the other two compose files.

atd-etl/populate_cr3_file_metadata/docker-compose.yaml
frankhereford, line 1: Basic compose stack to improve DX; now you can define an `env` file and then `docker compose run metadata` to get dropped into a terminal where you can run the program in the same environment it might expect to have been provided by airflow.

atd-etl/populate_cr3_file_metadata/env_template
frankhereford, line 1: Here's where you would get these values: https://github.com/cityofaustin/atd-airflow/blob/production/dags/vz_populate_cr3_metadata.py#L20-L45. Also, you can point it at your local stack with something like this for the endpoint: 
```
HASURA_ENDPOINT=http://host.docker.internal:8084/v1/graphql
```

NB: It takes a little extra fiddling to work around that magic DNS name on linux. I have no idea why Docker desktop provides it but good ol' normal docker does not. 🤯

atd-etl/populate_cr3_file_metadata/populate_cr3_file_metadata.py
```diff
@@ -3,14 +3,50 @@
...
 
+def main():
+    """
+    Main Loop
+    """
+    parser = argparse.ArgumentParser(description='Process some CR3s for metadata')
+    parser.add_argument('-t', '--threads', type=int, default=25, help='Number of threads')
+    parser.add_argument('-v', '--verbose', action='store_true', help='Verbose output')
+    parser.add_argument('-s', '--batch-size', type=int, default=os.getenv('PDF_MAX_RECORDS', 100), help='Batch size')
+    parser.add_argument('-a', '--all', action='store_true', help='Process all records')
```
frankhereford, line 24: Naming arguments and picking the right ones is always a challenge, but I like these OK. In airflow, we'll run this like:

```
populate_cr3_file_metadata.py -v 
```

And while we're catching up manually on our local machines, something like:

```
./populate_cr3_file_metadata.py -a
```

atd-etl/populate_cr3_file_metadata/populate_cr3_file_metadata.py
```diff
@@ -3,14 +3,50 @@
...
+    todo_count = get_todo_count() if args.all else args.batch_size
+    print(f"\nItems to process: {todo_count}\n")
+
+    with ThreadPoolExecutor(max_workers=args.threads) as executor:
+        futures = []
+        for crash_id in get_records(limit=PDF_MAX_RECORDS):
+            futures.append(executor.submit(process_record, crash_id=crash_id, args=args))
+
+        if not args.verbose:
+            for future in tqdm(futures, total=todo_count):
```
frankhereford, line 45: ✨ fancy progress bar. this library is a gem.

atd-etl/populate_cr3_file_metadata/.gitignore
```diff
@@ -1 +1,3 @@
 *pdf
+env
```
mddilley, line 2: Thanks for adding this! Makes me think that we should add a `.dockerignore` file in this project folder too just in case someone ever pushes from their own machine.

atd-etl/populate_cr3_file_metadata/populate_cr3_file_metadata.py
mddilley: TIL!

atd-etl/populate_cr3_file_metadata/docker-compose.yaml
mddilley: Yay! Thanks for making this easier to run locally. Could you add the tiniest of READMEs with this command for the next person who needs to run this locally? Actually, you could probably almost completely copypasta [this one](https://github.com/cityofaustin/atd-vz-data/blob/master/atd-etl/cris_import/README.md#local-testing).

mddilley (commented):
Awesome to see the threading spread to another script! 🧵 I took a quick look and added so small comments about docs. I know this process is screaming along right now so I'll take a closer look at the code this afternoon.

frankhereford:
@mddilley wrote:

> Could you add the tiniest of READMEs with this command for the next person who needs to run this locally?

Great idea, you bet! How's this look? https://github.com/cityofaustin/atd-vz-data/tree/thread-cr3-file-metadata/atd-etl/populate_cr3_file_metadata

https://github.com/cityofaustin/atd-vz-data/pull/1436/commits/0510864b7abc5ec7cedb1465fe0d0bfe496801ad

atd-etl/populate_cr3_file_metadata/populate_cr3_file_metadata.py
```diff
@@ -146,15 +182,16 @@ def get_file_metadata(crash_id: int) -> dict:
...
     """
     Controls the process for a single file
     """
     if not is_crash_id(crash_id):
         print(f"Invalid crash_id: {crash_id}")
         return False
 
-    print(f"Processing crash_id: {crash_id}")
+    if args.verbose:
+        print(f"Processing crash_id: {crash_id}")
```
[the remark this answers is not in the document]
frankhereford: Yea i totally like the idea of returning to these scripts and giving them a once-over in terms of logging. It'd be good, I think, to chat as a group about what we want to see in our logging and then, with that shared goal, come back through all the VZ etls and make them similar where we can. Adopt `argparse`, `logging`, threading, and `docker compose`. @johnclary - I'm curious how you feel about that - maybe something we can put on the slow cooker in the backlog for when we're caught up?

johnclary (approved):
@frankhereford looks great to me. would you PR the needed Airflow changes against this same issue: https://github.com/cityofaustin/atd-data-tech/issues/16746?

🚢 🚢 🚢 🚢 🚢

PR #1440, opened 2024-04-30
New data model

johnclary:
This PR holds the main collector branch for the new data model.


frankhereford (changes requested):
This is to prevent merging of this branch which will be the collector branch for VZ model work.

johnclary:
👋 Thanks for reviewing this PR. Heads-up that I will be working on down migrations right away, since this wasn't necessary up until just now, when I moved everything into the `public` schema. 

For time being, please focus your review attention on the conventions, assumptions, and functionality being put forward by this branch that may at odds with our vision for the data model, or will be difficult to undo down the road. That said, no detail is too small :)

johnclary:
also—sorry for more noise. I do not think there is any need to review the files in `./atd-toolbox` at this time. the `atd-vzd` and `atd-vze` changes are what matters :)

atd-vzd/migrations/default/1708558767621_views/up.sql
johnclary, line 1: I've reworked this view by spinning out a separate `crash_injury_counts` view. That work is captured in #1447

atd-toolbox/data_model/python/sql_templates/crashes_list_view.sql
```diff
@@ -0,0 +1,148 @@
...
+    select
+        crashes.crash_id,
+        sum(injury_severities.unkn_injry) as unkn_injry_cnt,
+        sum(
+            injury_severities.nonincap_injry
+        ) as nonincap_injry_cnt,
+        sum(injury_severities.poss_injry) as poss_injry_cnt,
+        sum(injury_severities.non_injry) as non_injry_cnt,
+        sum(injury_severities.sus_serious_injry) as sus_serious_injry_cnt,
+        sum(injury_severities.fatal_injury) as atd_fatality_cnt,
```
roseeichelmann, line 55: its a bit confusing to me that `atd_fatality_cnt` is the only one prepended with `atd`, but i know thats because there is also the `cris_fatality_cnt` and the `apd_fatality_cnt`

atd-toolbox/data_model/python/sql_templates/crashes_list_view.sql
```diff
@@ -0,0 +1,148 @@
...
+    group by
+        crashes.crash_id
+),
+
+geocode_sources as (
+    select
+        cris.crash_id,
+        coalesce(
+            (cris.latitude is null or cris.longitude is null),
+            false
```
roseeichelmann, line 70: is this coalesce necessary cuz i dont see how either of these arguments could ever return null?

atd-toolbox/data_model/python/sql_templates/crashes_list_view.sql
```diff
@@ -0,0 +1,148 @@
...
+    geocode_sources.has_no_cris_coordinates,
+    case
+        when (cris_fatal_injury_counts.cris_fatality_cnt is null) then 0
+        else cris_fatal_injury_counts.cris_fatality_cnt
+    end as cris_fatality_cnt,
+    case
+        when (law_enforcement_fatality_num is not null)
+            then cris_fatality_cnt
+        else 0
+    end as apd_fatality_cnt,
```
roseeichelmann, line 131: could you walk me through this logic here? why does `law_enforcement_fatality_num` come into play, and also what is the meaning of `apd_fatality_cnt` according to the vz team and how do they want to use it? also i feel like for clarity `ytd_law_enforcement_fatal_crash` would maybe be a better name? [heres](https://app.gitbook.com/o/-LzDQOVGhTudbKRDGpUA/s/-M4Ve3sp7qA5cPXha0B4/data-dictionary/~/comments?context=post&node=ab6556b2e2ed454684bc6ca1bc53b666) the description of the column

roseeichelmann:
I still have some more to dig into but here are some notes I have so far:

- I like your handling of `peh` in the `cris_import` script and dropping the names if not a fatality. What about for updates, where inj_sev is updated to a fatality, then do we want to carry over the name? Or is that getting too complicated?
- So great that you cleaned up the injury counts and now they are handled automatically in the `crashes_list` view by summing up the counts of injury severities from the person tables instead of the vz team having to keep everything up to date in multiple places. Now they only have to QA in one place, under the injury severity column of the persons table.
- What about prefixing column names with `vz_` if they are custom so we can quickly know if something is not from cris? That would help sooo much imo and i know we talked about doing something like that

atd-toolbox/data_model/python/sql_templates/crashes_list_view.sql
johnclary: ooh, thanks for pointing me to those docs. so for this PR, after i got this basic working view together, i decided to split off this work [into a separate PR](https://github.com/cityofaustin/atd-vz-data/pull/1447). I also wrote up new definitions of all the metrics in the [corresponding issue](https://github.com/cityofaustin/atd-data-tech/issues/17219).

thanks for your suggestion on `ytd_law_enforcement_fatal_crash`. i've been struggling with what to name that column and i really like including `ytd`. so maybe i'll go with `ytd_law_enforcement_fatal_crash_num`?

in any case, i agree the method for `apd_fatality_cnt` is confusing and it's the one i have the most doubts about. let's take that convo to the other PR 👍

johnclary:
thanks @roseeichelmann for your review!

> * I like your handling of `peh` in the `cris_import` script and dropping the names if not a fatality. What about for updates, where inj_sev is updated to a fatality, then do we want to carry over the name? Or is that getting too complicated?

My thinking was that if a VZ user edits a person injury severity to a fatality, they will need to input the person's name by manually copying it from the CR3 pdf. If the injury severity change comes via CRIS import, then the name will be picked up in the ETL and correctly added to the person record ✅  

> * So great that you cleaned up the injury counts and now they are handled automatically in the `crashes_list` view by summing up the counts of injury severities from the person tables instead of the vz team having to keep everything up to date in multiple places. Now they only have to QA in one place, under the injury severity column of the persons table.

Yeah! There are still some open questions about the best way to approach some of the metrics. And for that we have [this WIP PR](https://github.com/cityofaustin/atd-vz-data/pull/1447).

> * What about prefixing column names with `vz_` if they are custom so we can quickly know if something is not from cris? That would help sooo much imo and i know we talked about doing something like that

Yes, I haven't forgotten about it this idea. I think I'm a little on the fence about because I'd like a more foolproof way for us to maintain our list of CRIS vs VZ columns. Maybe i'll open a little side branch just so we can see what this looks like.

The alternative i think still prefer is having a `column_metadata` table in the db where we track this info, plus other CRIS import and VZE-related metadata (e.g. we could track which columns are in the public vs private CRIS extract). This table would basically replace the function of the current google sheet 🤔

roseeichelmann:
@johnclary I definitely think a column_metadata table is something we should have, but when im just scrolling through the database or a table i think i would find it helpful to see a column and know that its custom based off the name, but im defs a fan of descriptive column names. curious what everyone else thinks! yeah maybe a spin off branch to test out how it feels could be good

atd-toolbox/data_model/python/tests/tests.sql
```diff
@@ -0,0 +1,26 @@
...
+    crash_id, latitude, longitude, created_by, updated_by, cris_schema_version)
+    values (1, 30.2800238, -97.743370, 'cris', 'cris', '2023');
+insert into public.units_cris (unit_nbr, crash_id, cris_schema_version) values (1, 1, '2023');
+insert into public.people_cris (prsn_nbr, unit_nbr, crash_id, cris_schema_version) values (1, 1, 1, '2023');
+insert into public.charges_cris (prsn_nbr, unit_nbr, crash_id, charge, cris_schema_version) values (1, 1, 1, 'whatever', '2023');
+
+-- location null, in_austin_full_purpose false, council_district null
+update public.crashes_cris set active_school_zone_fl = TRUE, latitude = null where crash_id = 1;
+update public.units_cris set vin = 'abcdefg' where id = 1;
+update public.people_cris set drvr_city_name = 'austin' where id = 1;
```
roseeichelmann, line 12: at first i thought the triggers on units/people werent working bc i was searching for where `crash_id = 1`. finally realized this was updating where `id = 1` :sweat_smile: all the diff id's still bother me but i think its the lesser of two evils. it does make sense to be keeping our own custom ids

johnclary: yeah 😵 we're going to be doing a lot of joining on people > units > crashes. i think we might just make ourselves a handy view that adds crash ID (and whatever else) to people records. in fact we'll probably need one for the socrata ETLs anyway. i also think we should find a way to subtly incorporate the person and unit IDs into the VZE so that it can be super easy to debug VZE issues 🤔

mddilley (commented):
looks like we are ready to move on this to me! it was fun to read through the python and see how it generated a lot of these files. so satisfying to look back on everything that led up to this! 😎

atd-toolbox/data_model/python/cris_import.py
```diff
@@ -0,0 +1,427 @@
...
+    affected_rows
+  }
+}
+"""
+
+mutations = {
+    "crashes": CRASH_UPSERT_MUTATION,
+    "units": UNIT_UPSERT_MUTATION,
+    "persons": PERSON_UPSERT_MUTATION,
+    "charges": CHARGES_INSERT_MUTATION,
```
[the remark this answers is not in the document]
johnclary: that's right. the issue we're dealing with is that we have no way to uniquely identify a charge record in the CRIS extract, so the only way we can guarantee we have the right data is to purge it. 

so we just have a `charges_cris` table for charges (no `_edits` or unified table), and a crash’s charges are deleted and re-inserted whenever a crash is updated. There is no change log for charges. 

This still feels a little clumsy to me. It might be slightly better to soft-delete these records? Maybe it's not worth overthinking. In any case, I've added this to the "Documentation" todos. thanks for flagging it.

atd-toolbox/data_model/python/cris_import.py
```diff
@@ -0,0 +1,427 @@
...
+                        ),
+                        None,
+                    )
+
+                    set_default_values(records, {"is_primary_person": False})
+
+                    pp_records = lower_case_keys(load_csv(p_person_file["path"]))
+
+                    for record in pp_records:
+                        set_peh_field(record)
```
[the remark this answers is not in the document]
johnclary: 👍 cool—yes. and this python is still quick 'n dirty. i do look forward to cleaning up this branching logic and whatnot.

atd-toolbox/data_model/python/tests/tests.sql
```diff
@@ -0,0 +1,26 @@
+-- location A187DE10F4, in_austin_full_purpose true, council_district 9, engineering_area 2
+insert into public.crashes_cris (
+    crash_id, latitude, longitude, created_by, updated_by, cris_schema_version)
+    values (1, 30.2800238, -97.743370, 'cris', 'cris', '2023');
+insert into public.units_cris (unit_nbr, crash_id, cris_schema_version) values (1, 1, '2023');
+insert into public.people_cris (prsn_nbr, unit_nbr, crash_id, cris_schema_version) values (1, 1, 1, '2023');
+insert into public.charges_cris (prsn_nbr, unit_nbr, crash_id, charge, cris_schema_version) values (1, 1, 1, 'whatever', '2023');
+
+-- location null, in_austin_full_purpose false, council_district null
```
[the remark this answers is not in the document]
johnclary: sounds like a bug—i'll take take a look 🙏

atd-toolbox/data_model/python/tests/tests.sql
johnclary: ah, yes—these tests were designed to be run before the CRIS import increments all the sequences 😑 
i will try to re-write these in a way that isn't so fragile 🤔

atd-toolbox/data_model/python/cris_import.py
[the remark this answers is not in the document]
mddilley: yep, also, the fact that charges have been broken for a long while without disruption to the VZ team's work also supports not worrying too much. the only hesitation that i have too is if this is the only exception to a soft delete that we have in the model and why that is.

mddilley (approved):

atd-toolbox/data_model/python/tests/tests.sql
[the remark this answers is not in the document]
johnclary: ✅ fixed https://github.com/cityofaustin/atd-vz-data/pull/1440/commits/8176bf4f4089bc9c9996d5b3e67982700f94b854

roseeichelmann:
Something else ive run into that i wanna mention - i still feel like the `crash_id` should be carried into the edits tables. i was looking for a unit by the id in the `units_edits` table and i found it unhelpful that the `crash_id` column was null because i was trying to validate that i was editing the right unit in this specific crash.. idk maybe im still thinking about all this wrong and in a crash id-centric way?

atd-toolbox/data_model/python/cris_import.py
```diff
@@ -0,0 +1,427 @@
...
+def main():
+    overall_start_tme = time.time()
+
+    print("downloading column metadata...")
+    column_metadata = load_data(COLUMN_ENDPOINT)
+
+    files_todo = get_files_todo()
+
+    extract_ids = list(set([f["extract_id"] for f in files_todo]))
+    # assumes extract ids are sortable oldest > newest by filename. todo: is that right?
```
frankhereford, line 291: I'm not sure this is true. Here are some extracts which I recently pulled:

image

Because CRIS puts a schema number in the second element of the name, I think it becomes undetermined, as an extract commonly can have multiple sets of CSVs in them for when the data spans CRIS schemas. The following element is the date of the request, so you 'mostly' get them in the order you requested when you sort, but not reliably. 

While I think the assertion that you can't count on the order to be useful is true, the nitty-gritty details are always hard to pin down on CRIS, so this comment is YMMV for better or worse.

johnclary: >  because the third element of the CSVs' filename is the date stamp of the extract request, so sorting them will just get back the order they were requested / received from CRIS

thanks—this is the assumption i had made so it's good to hear you confirm. it's not clear from this comment, but my concern was indeed to make sure that we always process extracts in the order they were processed by CRIS. if we process extracts in the wrong order, we could theoretically update a crash record with old data.

atd-vzd/migrations/default/1715960005005_people/up.sql
```diff
@@ -0,0 +1,102 @@
...
+);
+
+create table public.people (
+    id integer primary key,
+    drvr_city_name text,
+    drvr_drg_cat_1_id integer references lookups.substnc_cat_lkp (id) on update cascade on delete cascade,
+    drvr_zip text,
+    ems_id integer references public.ems__incidents (id) on update cascade on delete set null,
+    est_comp_cost_crash_based integer generated always as (case when (prsn_injry_sev_id = 1) then 250000 when (prsn_injry_sev_id = 2) then 3000000 when (prsn_injry_sev_id = 3) then 200000 when (prsn_injry_sev_id = 4) then 3500000 else 20000 end) stored,
+    is_primary_person boolean,
```
roseeichelmann, line 76: do we wanna have some more in-depth convos about columns that should be not be nullable? for example, i feel like this is a column that would be problematic to be null. wondering if we want to consider having more strict constraints, theres probably many other column that could be argued should have a not null constraint

mddilley (approved):
Yayyyyyyyyy ship it! 🚢 🙌 😎 💪 🚀 🍾

chiaberry (approved):
✨ its happening!

frankhereford (approved):
My testing and learning are on-going, but I wanted to ✅✅✅ here so there's not even a temptation to wait for me. I will comment when I'm done testing, but this is more than ready to get merged into main. This is a huge milestone, and I can't wait to see this launch! Thank you @johnclary and @roseeichelmann -- this is really just tremendous.

johnclary:
Thanks y'all, here we go....

PR #1441, opened 2024-04-30
Clean up atd_txdot_locations table

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/16779

The only columns we need to keep:
location_id
description
last_update
latitude
longitude
location_group
scale_factor
geometry
street_level
is_intersection (per Xavier's request)
council_district - used in `locations_with_crash_injury_counts` view that powers the Locations page in the VZE

## Testing
**URL to test:** 
Local

**Steps to test:**
1. Checkout this branch, start up your local db and run the migrations.
2. Make sure the `atd_txdot_locations` table only has the columns in the list above
3. Run the VZE locally, make sure nothing broke with the removal of these columns. Test the locations page, click on a location. The aerial map doesnt work locally (for me.. does it work for yall?) so dont expect it to be there.
4. Test updating the crash coordinates of a crash in the VZE.

---
#### Ship list
- [ ] Code reviewed


atd-vzd/migrations/default/1714576842105_delete_crufty_locations_columns/up.sql
```diff
@@ -0,0 +1,51 @@
+-- This is an old crufty view that needs to be dropped bc it has dependencies with some columns we are dropping
+DROP VIEW atd_txdot_locations_with_centroids;
```
roseeichelmann, line 2: this view can be dropped it is cruft and not used anywhere! i inventoried it [here](https://docs.google.com/spreadsheets/d/1oHzKt8Cvd1BGpZ7KJl4dfBDHGuSl7gs4yN0OM7AJMBA/edit#gid=1171053547)

atd-vzd/migrations/default/1714576842105_delete_crufty_locations_columns/up.sql
```diff
@@ -0,0 +1,51 @@
...
+DROP COLUMN IF EXISTS transit_score,
+DROP COLUMN IF EXISTS level_1,
+DROP COLUMN IF EXISTS level_2,
+DROP COLUMN IF EXISTS level_3,
+DROP COLUMN IF EXISTS level_4,
+DROP COLUMN IF EXISTS level_5,
+DROP COLUMN IF EXISTS metadata,
+DROP COLUMN IF EXISTS unique_id,
+DROP COLUMN IF EXISTS geometry,
+DROP COLUMN IF EXISTS unknown_injury_count;
```
roseeichelmann, line 51: :wave: :broom: :sparkles:

johnclary (changes requested):
@roseeichelmann i'm getting a graphQL query error when i try to edit a crash's coordinates. 

Here's the Hasura response:
```json
{
    "errors": [
        {
            "message": "database query error",
            "extensions": {
                "path": "$",
                "code": "unexpected"
            }
        }
    ]
}
```

I think it may be the `update_cr3_location()` function that's failing. looks like it references the `geometry` column. i wonder if there is a good reason why VZE uses `shape` and VZD uses `geometry` 😑

johnclary:
after discussion with the team, our preference is to use the `geometry` column every where and ditch `shape` instead

atd-vzd/migrations/default/1715033725867_drop_crufty_function/up.sql
roseeichelmann, line 1: Surfaced these bc they all used the old `shape` attribute of the locations table but I can't find these functions being used anywhere, so time for some clean up :broom:

atd-vzd/migrations/default/1714576842105_delete_crufty_locations_columns/up.sql
frankhereford: So many columns that were working, temporary columns from when I wrote the polygon stuff back when the earth was cooling. This is so great, thank you!!!

johnclary (approved):

PR #1444, opened 2024-05-05
Use simplified engineering areas geometries

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/17135

## Testing
**URL to test:** Local

It's worth noting that in the new data model we are going to add a trigger to persist the engineering area ID on the crash record, the same as we do location, council district, and jurisdiction. So this change will add a noticeable performance  boost for the current fatalities view, and in the data model this change will provide a performance boost to the location-based trigger.  

**Steps to test:**
1. Spin up a prod snapshot.

2. Benchmark the `view_fatalities` query speed. It runs in ~350ms on my machine:

```sql
select * from view_fatalities;
```

3. Take note of the current engineering area fatality counts. These should be the same before and after this change:

```sql
select engineering_area, count(engineering_area) from view_fatalities group by engineering_area;
```

4.  Delete the existing engineering areas from the db:

```sql
delete from engineering_areas where true;
```

5. Save a copy of `secrets_template.js` as `secrets.js` and run the script to insert the new engineering areas:

```shell
# ./atd-toolbox/engineering_areas_load
$ node insert_areas.js local
``` 

6. Test the improved query speed and check the engineer area fatality counts. They should match the results from earlier.

```sql
select engineering_area, count(engineering_area) from view_fatalities group by engineering_area;
```

---
#### Ship list
- [ ] Code reviewed


roseeichelmann (approved):
Using `EXPLAIN ANALYZE` the view_fatalities query speed on the hasura console was 947ms. After running the script and using the simplified layer it was 266ms :partying_face: Also loaded up the VZE and the fatalities page loads up so fast now :') im sure the vz team will be rly happy with this update

Also for anyone else that runs into this issue, make sure you `nvm use` in this directory first to switch to the right node version. The VZE uses v16 and this directory uses v18 i was getting an error before i figured that out.

mddilley (approved):
Nice!!! I also see that the query is significantly faster. 🏎️💨 The fatality counts also matched up before and after. Thanks for updating while we work towards the new model. 🚢 🚀 🙌

johnclary:
thanks for the quick reviews! i'm going to go ahead and run this on staging and prod. it will be nice to have this lil' boost deployed ahead of the tomorrow's sprint review.

PR #1445, opened 2024-05-06
Refactor CRIS import script to use S3 and log activities

frankhereford:
## Associated issues

This PR aims to advance the following issue: https://github.com/cityofaustin/atd-data-tech/issues/16953.

It completes the second bullet point, and on merger, allows the final two points to become possible.

## Mechanics

There is a S3 bucket named `vision-zero-cris-exports`. It has versioning turned on, so we can always recover a lost or changed file, if we were to need to -- at least for 180 days. After 180 days, things start to get removed due to lifecycle rules in place.

In the bucket, you'll find 3 folders, one for development, staging and production. Within each of those, you'll see an inbox, where CRIS will upload extracts, and a processed folder, where we will keep our, well, processed extracts. 

Additionally, you'll see a new table in the DB called `cris_import_log`, which will keep some logs of the import events, one record per processing attempt per file.

## Testing

This command has been real useful in development, and I imagine it may be useful in testing. It assumes you have the `aws` command set up with your credentials.

```bash
aws s3 mv s3://vision-zero-cris-exports/development/processed/extract_2023_20240417155400400_99726_20240401-20240405_HAYSTRAVISWILLIAMSON.zip s3://vision-zero-cris-exports/development/inbox/                                                                                
```

Local testing for this one. To get going, spin up your local stack to where you have at least a DB and a `graphql-engine` endpoint running. ⚠️ Be sure to apply the new migration that comes along with this PR.

```bash
cd atd-etl/cris_import;
docker compose build --no-cache;
```

There's a file in the processed folder for development, and you can use the above to move it into the inbox or via the S3 console works just fine too. 

```bash
docker compose run cris-import;
```

* Did it run? 

* ⚠️ Don't forget to rebuild your cris import docker image after you switch back to `master` if you're intending to use it for something other than testing this work. 🐳 

---
#### Ship list
- [ ] Code reviewed


atd-etl/cris_import/.gitignore
```diff
@@ -1 +1,2 @@
 env
+*sqlite
```
frankhereford, line 2: I find that I end up with these sqlite files floating around from inspecting them off S3. This prevents them from ever getting into `git`, but if they do, it's not a huge deal. No secrets in there.

atd-etl/cris_import/Dockerfile
```diff
@@ -5,7 +5,7 @@ RUN ln -snf /usr/share/zoneinfo/America/Chicago /etc/localtime && echo America/C
 RUN apt-get update && apt-get install -y tzdata
 
 RUN apt-get -y upgrade
-RUN apt-get install -y aptitude magic-wormhole vim black python3-pip
+RUN apt-get install -y aptitude magic-wormhole vim black python3-pip sqlite3 file
```
frankhereford, line 8: Add some tooling to the image

atd-etl/cris_import/cris_import.py
```diff
@@ -83,21 +87,25 @@ def main():
...
+
     local_mode = False
     if bool(glob.glob("/app/development_extracts/*.zip")):
         local_mode = True
 
     zip_location = None
+    database_location = None
     if not local_mode:  # Production
-        zip_location = download_archives()
+        zip_location, database_location = download_s3_archive()
```
frankhereford, line 99: Drop in replacement function that handles getting the archives. There's an extra bit of information that comes out of this step now which is where the SQLite DB file is saved.

atd-etl/cris_import/cris_import.py
```diff
@@ -83,21 +87,25 @@ def main():
...
         zip_location = specify_extract_location()
 
     if not zip_location:
         return
 
-    extracted_archives = unzip_archives(zip_location)
-    for archive in extracted_archives:
+    extracted_archives = unzip_archives(zip_location, database_location)
+    for archive_data in extracted_archives:
+        archive = archive_data[1]
```
frankhereford, line 108: The `unzip_archives` function returns an array of tuples now, instead of just an array of strings representing the locations of the extracted zip files' temp directories. The tuples contain that string of the location on the file system, but also now the ID of the record for that extract in the SQLite DB. We'll use that later on to mark successful imports.

atd-etl/cris_import/cris_import.py
```diff
@@ -107,9 +115,45 @@ def main():
             typed_token = align_db_typing(trimmed_token)
             align_records_token = align_records(typed_token)
             clean_up_import_schema(align_records_token)
+        mark_extract_as_imported(archive_data[0], database_location)
```
frankhereford, line 118: Here we are using the ID of the record to form up the `update` statement that marks a successful import of the extract.

atd-etl/cris_import/cris_import.py
```diff
@@ -727,13 +777,12 @@ def group_csvs_into_logical_groups(extracted_archives, dry_run):
     files = os.listdir(str(extracted_archives))
     logical_groups = []
     for file in files:
-        if file.endswith(".xml"):
+        if file.endswith(".xml") or file == "crashReports":
```
frankhereford, line 780: Extracts can have a folder in them of PDFs now, and we don't want to treat the folder with the code for CSVs in this block, so this skips over that folder.

atd-etl/cris_import/docker-compose.yml
```diff
@@ -1,5 +1,3 @@
-version: '3'
-
```
frankhereford, line 2: Tag along change; `docker compose` not `docker-compose`, no longer need version designation. Same as many other PRs. LMK if you have any questions on this.

atd-etl/cris_import/uploads.sqlite.schema
frankhereford, line 1: This memorializes the form of the default, empty database used to track import state for a environment.

atd-etl/cris_import/cris_import.py
```diff
@@ -107,9 +115,45 @@ def main():
             typed_token = align_db_typing(trimmed_token)
             align_records_token = align_records(typed_token)
             clean_up_import_schema(align_records_token)
+        mark_extract_as_imported(archive_data[0], database_location)
     if not local_mode:  # We're using a locally provided zip file, so skip these steps
-        remove_archives_from_sftp_endpoint(zip_location)
-        upload_csv_files_to_s3(archive)
+        upload_sqlite_to_s3(database_location)
```
frankhereford, line 120: Here is where we put our modified SQLite DB back on S3. This system includes no provisions for locking this file, and as a result, there can never be two CRIS import scripts running at once. This is a given in our nominal operation, but it is worth mentioning.

atd-etl/cris_import/cris_import.py
```diff
@@ -60,6 +62,8 @@ def main():
     global GRAPHQL_ENDPOINT_KEY
     global SFTP_ENDPOINT_SSH_PRIVATE_KEY
```
johnclary, line 63: it looks like perhaps all three of these variables may no longer be in use

atd-etl/cris_import/cris_import.py
```diff
@@ -83,21 +87,25 @@ def main():
     GRAPHQL_ENDPOINT_KEY = secrets["graphql_endpoint_key"]
     SFTP_ENDPOINT_SSH_PRIVATE_KEY = secrets["sftp_endpoint_private_key"]
 
+    S3_EXTRACT_BUCKET = secrets["s3_extract_bucket"]
```
johnclary, line 90: how do you feel about using the existing vision zero database bucket instead of create a new bucket for this task? the old bucket looks a bit crufty, but i think we could safely purge past extracts from it.

atd-etl/cris_import/cris_import.py
```diff
@@ -225,48 +274,94 @@ def specify_extract_location():
...
+    except botocore.exceptions.ClientError as e:
+        # If the DB does not exist, copy it from the default directory on S3
+        if e.response["Error"]["Code"] == "404":
+            default_db_key = "default_database/uploads.sqlite"
+            copy_source = {"Bucket": bucket, "Key": default_db_key}
+            s3.copy(copy_source, bucket, db_key)
+
+    uploads_prefix = f"{DEPLOYMENT_ENVIRONMENT}/uploads/"
+    # Get list of all objects in the bucket with the specified prefix
+    objects = s3.list_objects(Bucket=bucket, Prefix=uploads_prefix)
```
johnclary, line 300: this is going to run into the limitation that you must paginate through the object list beyond the first thousand results, but i would not be concerned about this if we were moving processed extracts to a separate subdirectory.

johnclary (commented):
@frankhereford—thanks for making such quick work of this critical upgrade!

it's cool to see this in action, and i love sqlite, but i have doubts whether or not going down the path of using a databsae is an improvement over moving processed extracts into an "archive" subdirectory. can you expand a bit on the issue you're trying to workaround by introducing the db? 

if we do need to  use a DB to track the extract state, i wonder if we might want to use our postgres database so we don't have to worry about it de-syncing 🤔

atd-etl/cris_import/cris_import.py
```diff
@@ -225,45 +292,94 @@ def specify_extract_location():
...
+            # Extract the schema from the filename
+            schema = int(object_name.split("_")[1])
+
+            # Check if the object already exists in the database
+            cursor.execute(
+                "SELECT * FROM cris_import_log WHERE object_path = ? AND object_name = ?",
+                (object_path, object_name),
             )
-        except:
-            print("No files to copy..")
```
johnclary, line 255: @frankhereford what do you think about throwing an exception if there are no files to copy? i think this should be a pretty rare occurrence for our prod deployment, right?

frankhereford: Given the recent events, I think this is an exceptional idea and I'll roll that into this PR. Great call.

atd-etl/cris_import/cris_import.py
frankhereford: The graphql endpoint ones need to persist, but the SFTP endpoint key is outta here!

atd-etl/cris_import/cris_import.py
frankhereford: We spoke about this, and I believe we settled on using this new bucket on account of having to dole out a credential to access it to another agency and keeping a maximally limited scope on that permission.

atd-etl/cris_import/cris_import.py
frankhereford: We spoke about this, and with the lifecycle rules in place, we were in a good place re: 1000 entries at a time, but we're in a doubly good place now because the script has been updated to move the files into a processed folder too.

atd-etl/cris_import/cris_import.py
frankhereford: This will throw an exception now if it has no work to do -- thank you for this super good suggestion! https://github.com/cityofaustin/atd-vz-data/pull/1445/commits/df608325cbc7fb10108159713ad662397835bf3b

atd-etl/cris_import/uploads.sqlite.schema
johnclary, line 1: I think this file can be removed

atd-etl/cris_import/cris_import.py
```diff
@@ -225,45 +270,107 @@ def specify_extract_location():
...
+            source = {"Bucket": bucket, "Key": upload_path}
+            destination = os.path.join(
+                f"{DEPLOYMENT_ENVIRONMENT}/processed/", upload[2]
+            )
+
+            # Copy the file from the source to the destination
+            s3.copy(source, bucket, destination)
+
+            # Delete the file from the source
+            s3.delete_object(Bucket=bucket, Key=upload_path)
```
johnclary, line 367: would it be best to wait until the import has been completely processed before moving it into the `processed` dir? so move these bucket operations to the `mark_extract_as_imported()` function.

atd-etl/cris_import/cris_import.py
```diff
@@ -225,45 +270,107 @@ def specify_extract_location():
...
+                    """
+                    INSERT INTO cris_import_log (object_path, object_name, first_seen)
+                    VALUES (%s, %s, %s)
+                    """,
+                    (object_path, object_name, datetime.datetime.utcnow()),
+                )
+
+        # Query all uploads where import_attempted = 0
+        cursor.execute("SELECT * FROM cris_import_log WHERE import_attempted is false")
+        uploads_to_import = cursor.fetchall()
```
johnclary, line 340: instead of `uploads_to_import` — is `extracts_to_process` (or similar) a more accurate var name?

atd-etl/cris_import/cris_import.py
```diff
@@ -225,45 +270,107 @@ def specify_extract_location():
...
+        # Create a new temporary directory
+        import_dir = tempfile.mkdtemp()
+
+        # Download each upload into the new temporary directory
+        for upload in uploads_to_import:
+            upload_path = os.path.join(
+                upload[1], upload[2]
+            )  # Construct the full S3 path
+            upload_file_path = os.path.join(import_dir, upload[2])
+            s3.download_file(bucket, upload_path, upload_file_path)
```
johnclary, line 355: what would happen if we got into a situation where the DB thinks there should be a certain extract in the bucket, but for whatever reason that extract is not present? i assume `s3.download_file` would throw an error, right?

atd-vzd/migrations/default/1715094801442_add_import_log_table/up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE cris_import_log (
+    id SERIAL PRIMARY KEY,
+    object_path TEXT NOT NULL,
+    object_name TEXT NOT NULL,
+    first_seen TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    import_attempted BOOLEAN NOT NULL DEFAULT FALSE,
+    import_successful BOOLEAN NOT NULL DEFAULT FALSE,
+    import_time TIMESTAMPTZ DEFAULT NULL
```
johnclary, line 8: how do you feel about `created_at` and `updated_at` (or `completed_at`) for these timestamp column names? 

do you think it would it be worth going ahead and creating a compound index on (`object_path`, `object_name`)? perhaps that's overkill since we're talking about 365 rows/year on average. 

similarly, and i guess this would accomplish the indexing automatically, do we want to add a unique constraint against those two columns?

johnclary (changes requested):
@frankhereford this is looking great! just a few minor things, most of which are questions :)

atd-etl/cris_import/cris_import.py
johnclary: all credit to @chiaberry

atd-etl/cris_import/uploads.sqlite.schema
frankhereford: it's outta there! https://github.com/cityofaustin/atd-vz-data/pull/1445/commits/abf7f0676e1b300e1fc1cdb1ac877f61f4630981

atd-etl/cris_import/cris_import.py
frankhereford: @johnclary I agree with this, but only if we uphold the current functionality that based on the `cris_import_log` table, we never try to automatically import an extract twice. When one fails, I want that extract to go into a "needs human assessment and possible intervention" state and not keep trying to do the import over and over. Can you give me a 👍 if you agree on that, and if so, I'll make the change. Otherwise, let's discuss.

atd-etl/cris_import/cris_import.py
frankhereford: 💯 Yes -- this naming made a lot more sense with my old S3 structure, and it's nonsensical with the enhancements we've done. This symbol is renamed. https://github.com/cityofaustin/atd-vz-data/pull/1445/commits/2baf51c249f8c2b751f60f99a32583f2aa4eb1a3

atd-etl/cris_import/cris_import.py
frankhereford: Great idea! I'll have it check for successfully getting the zip down onto the local disk and to use `magic` to make sure it's a valid zip. https://github.com/cityofaustin/atd-vz-data/pull/1445/commits/01ec3ae7493da82518633e6238f3129a49299a36

atd-etl/cris_import/cris_import.py
frankhereford: John and I got to sync on about this, and we came to agree on his assertion that the log table is valuable, but that we should entirely rely on the state of the system being tracked in the S3 by the locations of files. The log is just a log, and I'll update the program to reflect this.

atd-etl/cris_import/cris_import.py
frankhereford: Additionally, the log table is not going to be used to track state -- it's just a log now -- and the existence of files in s3 is the only truth about what files are there or not, and that makes sense to me now. Thanks John 🙏

atd-etl/cris_import/cris_import.py
frankhereford: This rework has been complete. The log table is now genuinely a log table and holds no state for the ETL. It's written to, and it is never read from -- a log.

Unlike some of the first invocations of this functionality, every record in this table now represents a processing attempt. 

@johnclary -- I had mentioned when we spoke today that I was against putting columns in this table that don't have functionality built out to populate them. I'm walking that back, at least for a little bit, because I want to memorialize the ideas that I have for them. I'll do so in the `up.sql` migration. If you want to drop those columns and spin off issues to make them again later when we code out functionality to populate them, I'm 1,000% behind that.

atd-etl/cris_import/cris_import.py
frankhereford: The log is a real log, write and never read now. State is entirely held in the S3 filesystem.

atd-vzd/migrations/default/1715094801442_add_import_log_table/up.sql
frankhereford: Yep. The audit field names are a billion times better. They are adopted. ✨ 

Regarding additional indexes & constraints, given the changes since this comment was written, I think we're where we want to be. The log records a record for each processing attempt, so there is no longer an expectation for unicity of the object path + name.

atd-etl/cris_import/Dockerfile
```diff
@@ -5,7 +5,7 @@ RUN ln -snf /usr/share/zoneinfo/America/Chicago /etc/localtime && echo America/C
 RUN apt-get update && apt-get install -y tzdata
 
 RUN apt-get -y upgrade
-RUN apt-get install -y aptitude magic-wormhole vim black python3-pip
+RUN apt-get install -y aptitude magic-wormhole vim black python3-pip libmagic1 file
```
frankhereford, line 8: We're using `libmagic` to check that the zips are zips through inspection of the binary file.

atd-etl/cris_import/cris_import.py
```diff
@@ -225,45 +295,91 @@ def specify_extract_location():
...
 
-    Returns path of temporary directory as a string
-    """
+    s3 = session.client("s3")
+    bucket = S3_EXTRACT_BUCKET
 
     with SshKeyTempDir() as key_directory:
         write_key_to_file(
-            key_directory + "/id_ed25519", SFTP_ENDPOINT_SSH_PRIVATE_KEY + "\n"
+            key_directory + "/id_ed25519", DB_BASTION_HOST_SSH_PRIVATE_KEY + "\n"
```
frankhereford, line 309: This is a real confusing part of this diff. It's managed to intermingle two parts of the program that did wildly different things:
* Connecting to the sftp host to rsync files down
* Connect to the bastion to tunnel into the DB server

It's bonkers how it diff's like this.

atd-vzd/migrations/default/1715094801442_add_import_log_table/up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE cris_import_log (
+    id SERIAL PRIMARY KEY,
+    object_path TEXT NOT NULL,
+    object_name TEXT NOT NULL,
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    completed_at TIMESTAMPTZ DEFAULT NULL,
+    records_processed JSONB DEFAULT '{}',
```
frankhereford, line 7: Yes, this is really Frank writing this. I think this is a good place for a JSON blob in lieu of a one-to-many table FK relationship because I want the log to fit neatly in one table. I'm seeing this field getting populated with a dict that has the keys of the CRIS schema year and the value being the number of crashes which we processed. I think if we develop that metadata to be any more complex, we should bail out of the JSONB field and switch to proper DB tables with normal, normal relationships.

atd-vzd/migrations/default/1715094801442_add_import_log_table/up.sql
```diff
@@ -0,0 +1,9 @@
...
+    outcome_status TEXT NULL
```
frankhereford, line 8: This column is intended to hold a short, textual message that indicates how the program finished. It will default to something like "nominal completion," and for all error states, each one can have a text string that goes into here that's a breadcrumb for the dev who is trying to figure out what went wrong and why. The breadcrumb strings can come from the string argument of the `raise()` function if we stick in some sort of error handling. I'm not 100% i know exactly how this goes down in python, but I'm pretty hopeful that it's possible.

frankhereford, line 8: Again, repeating something from way above -- these two columns are /not/ getting populated in the code at the time of writing, and I'd be very happy to remove them as cruft. We can create spin off issues to enhance this code to populate them, but really, we should focus that effort on the CRIS importer in the new data model, perhaps?

frankhereford:
@johnclary, thank you /so much/ for all of your thoughtful feedback; I appreciate your ability to step back and see the big picture. 

No rush on any sort of review here. I am going to baby any CRIS extracts we get (or have to get via `download`ing) though for the time being until we're ready to roll this out. It's really no trouble at all, esp after working on this -- for a hot second, I have all of the moving parts under control.

atd-etl/cris_import/cris_import.py
```diff
@@ -276,15 +392,48 @@ def unzip_archives(archives_directory):
...
+            cursor = pg.cursor()
+            cursor.execute(
+                "select id from cris_import_log where object_name = %s order by created_at desc limit 1",
+                (filename,),
+            )
+            id = cursor.fetchone()[0]
+            extracted_csv_directories.append((id, extract_tmpdir, filename))
+
+            pg.commit()
+            pg.close()
```
johnclary, line 434: while testing with multiple extract in the directory, this block threw an error because `pg.close()` closes the connection after the first file is processed. the fix is to de-indent `pg.close()` 👍

johnclary (changes requested):
@frankhereford i was able to run the script following your instructions, with just a few minor things:
- don't forget to apply migrations after you start the db ;)
- one line of code needs to be de-indented

besides that—it would be nice to have a few comments in the db migrations. awesome work—thank you!

atd-etl/cris_import/cris_import.py
frankhereford: Great catch, thank you. I've outdented those back into their correct block. https://github.com/cityofaustin/atd-vz-data/pull/1445/commits/b9ce5eb4f5350148340d035d2bbf510654526e58

atd-vzd/migrations/default/1715094801442_add_import_log_table/up.sql
frankhereford: You bet, great idea. Here are some comments for each column in this table. Thanks! https://github.com/cityofaustin/atd-vz-data/pull/1445/commits/e2ca36ef9fd1b427e422502bfb29c3c720d81bc9

atd-etl/cris_import/lib/mappings.py
```diff
@@ -77,7 +77,7 @@ def no_override_columns():
             "prsn_nbr",  # key column
             "prsn_type_id",  # key column (but why?)
             "prsn_occpnt_pos_id",  # key column (but why?)
-            "injury_severity",
+            "prsn_injry_sev_id",
```
johnclary, line 80: this was an issue we identified which is unrelated to this PR. i'm going to start spin out a separate task of investing whether or not we've overwritten the `prsn_injry_sev_id` in cases where we had not intended to 😵

johnclary (approved):
great stuff—ship it! 🙏 🚢

mddilley (approved):
I tested this locally and see the log, S3 objects, and import all working just like the code says it does. The local dev experience is so slick now!!! I really like the bucket path structure and how straightforward it makes the flow. It is going to be so nice when we sunset the SFTP server 👋 - thanks for getting us there! 🙌 🚀

PR #1448, opened 2024-05-10
[BUG] Crash coordinates render as NaN when they should be blank

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/17104

This makes a few small bug fixes to the crash coordinates edit map. It:
- makes the form show **None** when there are no crash coordinates
- gives the marker default coordinates so it always show on the map when it opens
- fixes an issue where the alert message about a crash not having coordinates pushes the crash map out of the card it should stay inside

Side note - i found an interesting gist about `!!` vs `Boolean()` [here](https://gist.github.com/arthurvi/66cb1e2bcfc92f99f465e0db04264367). 

## Testing
**URL to test:** 
https://deploy-preview-1448--atd-vze-staging.netlify.app/

**Steps to test:**
1. Go to https://deploy-preview-1448--atd-vze-staging.netlify.app/#/crashes/11157589 to a crash without coordinates and no crash diagram
2. See that the map card shows an alert and that the map is contained in the card body
3. Also see that the map has height even though there is no crash diagram
4. Click the **Edit Coordinates** button and see that the form shows **None** in both inputs
5. Go to https://deploy-preview-1448--atd-vze-staging.netlify.app/#/crashes/19540057 to see a crash with coordinates
6. Click the **Edit Coordinates** button and see that the form shows the coordinates of the crash
7. Test updating and saving any crash besides the one above without coordinates to make sure it still works. Trying to keep the example above intact. 🙏

---
#### Ship list
- [x] Code reviewed

### DIFF
--- a/atd-vze/src/views/Crashes/Crash.js
+++ b/atd-vze/src/views/Crashes/Crash.js
@@ -168,7 +168,8 @@ function Crash(props) {
     data &&
     data?.atd_txdot_crashes.length > 0 &&
     data?.atd_txdot_crashes[0]["location_id"];
-  const notEditingCoords = !isEditingCoords && latitude && longitude;
+  const hasCoordinates = Boolean(latitude) && Boolean(longitude);
+  const notEditingCoords = !isEditingCoords && hasCoordinates;
 
   return !data?.atd_txdot_crashes?.length ? (
     <Page404 />
@@ -237,7 +238,7 @@ function Crash(props) {
                   )
                   <br />
                   Geocode Provider:{" "}
-                  {latitude && longitude
+                  {hasCoordinates
                     ? geocodeMethod.name
                     : "No Primary Coordinates"}
                 </Col>
@@ -255,7 +256,7 @@ function Crash(props) {
               </Row>
             </CardHeader>
             <CardBody style={{ minHeight: "350px" }}>
-              {(!latitude || !longitude) && (
+              {!hasCoordinates && (
                 <Alert color="danger">
                   Crash record is missing latitude and longitude values required
                   for map display.
--- a/atd-vze/src/views/Crashes/Crash.js
+++ b/atd-vze/src/views/Crashes/Crash.js
@@ -168,7 +168,7 @@ function Crash(props) {
     data &&
     data?.atd_txdot_crashes.length > 0 &&
     data?.atd_txdot_crashes[0]["location_id"];
-  const hasCoordinates = Boolean(latitude) && Boolean(longitude);
+  const hasCoordinates = !!latitude && !!longitude;
 
   return !data?.atd_txdot_crashes?.length ? (
     <Page404 />
--- a/atd-vze/src/views/Crashes/Crash.js
+++ b/atd-vze/src/views/Crashes/Crash.js
@@ -254,7 +254,10 @@ function Crash(props) {
                 </Col>
               </Row>
             </CardHeader>
-            <CardBody className="d-flex flex-column">
+            <CardBody
+              className="d-flex flex-column"
+              style={{ minHeight: "400px" }}
+            >
               {!hasCoordinates && (
                 <Alert color="danger">
                   Crash record is missing latitude and longitude values required
--- a/atd-vze/src/views/Crashes/Crash.js
+++ b/atd-vze/src/views/Crashes/Crash.js
@@ -169,7 +169,6 @@ function Crash(props) {
     data?.atd_txdot_crashes.length > 0 &&
     data?.atd_txdot_crashes[0]["location_id"];
   const hasCoordinates = Boolean(latitude) && Boolean(longitude);
-  const notEditingCoords = !isEditingCoords && hasCoordinates;
 
   return !data?.atd_txdot_crashes?.length ? (
     <Page404 />
@@ -255,14 +254,14 @@ function Crash(props) {
                 </Col>
               </Row>
             </CardHeader>
-            <CardBody style={{ minHeight: "350px" }}>
+            <CardBody className="d-flex flex-column">
               {!hasCoordinates && (
                 <Alert color="danger">
                   Crash record is missing latitude and longitude values required
                   for map display.
                 </Alert>
               )}
-              {notEditingCoords ? (
+              {!isEditingCoords ? (
                 <CrashMap data={data.atd_txdot_crashes[0]} />
               ) : (
                 <CrashEditCoordsMap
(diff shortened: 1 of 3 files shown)


atd-vze/src/views/Crashes/Crash.js
```diff
@@ -168,7 +168,7 @@ function Crash(props) {
     data &&
     data?.atd_txdot_crashes.length > 0 &&
     data?.atd_txdot_crashes[0]["location_id"];
-  const notEditingCoords = !isEditingCoords && latitude && longitude;
```
mddilley, line 171: This was causing an edge case where not having lat/lon for a crash made the edit map show on load. We want to always start with the non-edit map. You can see the bug in prod here -> https://visionzero.austin.gov/editor/#/crashes/19389730

atd-vze/src/views/Crashes/Crash.js
```diff
@@ -254,14 +254,14 @@ function Crash(props) {
                 </Col>
               </Row>
             </CardHeader>
-            <CardBody style={{ minHeight: "350px" }}>
-              {(!latitude || !longitude) && (
+            <CardBody className="d-flex flex-column">
```
mddilley, line 257: This fixes a bug where the map was escaping its parent card when the "no coordinates" alert showed. There is a screenshot in the issue. Now, the map will stretch to fit the remaining height in the card.

atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
```diff
@@ -31,8 +31,8 @@ const CrashEditCoordsMap = ({
   const mapRef = React.useRef();
   const [isDragging, setIsDragging] = React.useState(false);
   const [markerCoordinates, setMarkerCoordinates] = React.useState({
-    latitude: latitude_primary,
-    longitude: longitude_primary,
+    latitude: latitude_primary || defaultInitialState.latitude,
+    longitude: longitude_primary || defaultInitialState.longitude,
```
mddilley, line 35: Another small fix so that the marker has coordinates when the crash does not. Previously, the marker would only appear after moving the map since it is captured on a map drag event. Now, the initial location of the marker is the same lat/lon used to initialize the map when adding coordinates to a crash that has none.

atd-vze/src/views/Crashes/Crash.js
mddilley: disregard - starting a new comment since we need the `minHeight` afterall 🙃

atd-vze/src/views/Crashes/Crash.js
```diff
@@ -254,14 +254,17 @@ function Crash(props) {
                 </Col>
               </Row>
             </CardHeader>
-            <CardBody style={{ minHeight: "350px" }}>
-              {(!latitude || !longitude) && (
+            <CardBody
+              className="d-flex flex-column"
+              style={{ minHeight: "400px" }}
```
mddilley, line 259: This fixes a bug where the map was escaping its parent card when the "no coordinates" alert showed. There is a screenshot in the issue of this issue. Now, the map will stretch to fit the remaining height in the card. The min height is still needed so the map doesn't shrink down when there is no crash diagram available to stretch the parent row to have height.

Makes me wonder what making the height of this row fixed might look like so it doesn't look different on every crash? There might be a good reason to keep it this way, though.

Makes me wonder what a fixed height for both of these could look like!

atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
mddilley: ignore this too - caused another bug 🐛

atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
```diff
@@ -108,7 +108,12 @@ const CrashEditCoordsMap = ({
       >
         <FullscreenControl position="top-left" />
         <NavigationControl position="top-left" showCompass={false} />
-        <Marker {...markerCoordinates}>
+        <Marker
+          latitude={markerCoordinates.latitude || defaultInitialState.latitude}
+          longitude={
+            markerCoordinates.longitude || defaultInitialState.longitude
+          }
```
mddilley, line 115: Another small fix so that the marker has coordinates when the crash does not. Previously, the marker would only appear after moving the map since it is captured on a map drag event. Now, the initial location of the marker is the same lat/lon used to initialize the map when adding coordinates to a crash that has none.

atd-vze/src/views/Crashes/Crash.js
johnclary: I think that's a great idea. There is some weird re-sizing that happens on save as well. I think we can revisit this in the near-ish future as we take a closer look at this whole page's layout 🙏

johnclary (approved):
Much better! Thanks for including these other upgrades as well 🙏 🚢🚢🚢🚢

atd-vze/src/views/Crashes/Crash.js
mddilley: i was questioning how it was supposed to work as well! now that you mention it - the original intention probably was to open the edit map when coordinates were missing. 🤔

atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
```diff
@@ -72,7 +72,7 @@ const CrashEditCoordsMap = ({
     // Reset marker to original coordinates or default fallback
     const originalMarkerCoordinates = {
       latitude: latitude_primary || defaultInitialState.latitude,
-      longitude: longitude_primary || defaultInitialState.latitude,
```
chiaberry, line 75: 😬

chiaberry (approved):
Nice fixes!

atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
roseeichelmann: erm :grimacing: is right lol nice catch mike

atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
roseeichelmann: nice touch!

roseeichelmann (approved):
Really nice fix and code looks great! Also oops i definitely edited the no coordinates example.. my bad :sweat_smile: this is why you read all the instructions first

atd-vze/src/views/Crashes/Maps/CrashEditCoordsMap.js
mddilley: +1 to 😬 that was my oooopsie. initializing the map way outta Austin! 🔭

mddilley:
@roseeichelmann all good and thanks for testing! i'm merging so no more reviews needed. ✅ 🙏

PR #1450, opened 2024-05-14
Patch production with S3 enabled CRIS importer (to get the migration run in production)

frankhereford:
This is a patch on production to bring the S3 CRIS import enhancements online. This ETL work comes with a migration, hence rushing this out the door so we can update the ETL.

This PR is related to [PR 1445](https://github.com/cityofaustin/atd-vz-data/pull/1445).

I have observed that there are no other migrations in `master` that are not in production, so there should be no out-of-order conflicts, but please 👀 that assertion to make sure you agree.

---
#### Ship list
- [ ] Code reviewed


frankhereford:
@johnclary Because this feature branch came off `master` and not `production`, it includes some work that made it into master after our last release party but before I started working on this. It looks like a toolbox script is getting to make it into `production` early. That OK with you?

johnclary (approved):
Yep—all of these changes look good to me. Definitely no harm in bringing that toolbox script along. And the `atd-etl` updates are already deployed via GH action. ship it! 🚢 🙏

PR #1451, opened 2024-05-14
Crash details page: header

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/17181

## Testing
**URL to test:** 
Locally

**Steps to test:**

1. Check out this branch
2. Follow [the steps](https://github.com/cityofaustin/atd-vz-data/pull/1440) to start the data model collector branch and run the CRIS import ETL
3. Review the code, see that the header data is now pulling from the new crashes table
4. In the VZE click on diff crashes and make sure the headers are rendering correctly

---
#### Ship list
- [ ] Code reviewed


atd-vze/src/views/Crashes/Crash.js
```diff
@@ -149,19 +157,22 @@ function Crash(props) {
...
     cr3_file_metadata: cr3FileMetadata,
     investigator_narrative_ocr: investigatorNarrative,
   } = !!data?.atd_txdot_crashes[0] ? data?.atd_txdot_crashes[0] : {};
 
+  const {
+    crash_injury_metrics: { cris_fatality_count: deathCount },
+    crash_injury_metrics: { sus_serious_injry_count: seriousInjuryCount },
+    address_primary: primaryAddress,
+    address_secondary: secondaryAddress,
+  } = crashData?.crashes_by_pk ? crashData?.crashes_by_pk : {};
```
roseeichelmann, line 174: as we work on the crash details page more and more, this will grow and eventually we will be able to ditch the other query.

atd-vze/src/queries/crashes.js
```diff
@@ -118,6 +118,21 @@ export const GET_CRASH = gql`
...
+  query CrashDetails($crashId: Int!) {
+    crashes_by_pk(crash_id: $crashId) {
+      crash_id
+      address_primary
+      address_secondary
+      crash_injury_metrics {
+        cris_fatality_count
+        sus_serious_injry_count
+        years_of_life_lost
+      }
```
roseeichelmann, line 131: the only way i could think of incrementally working on the crash details page is to create another query and be using both for rendering the app, and with each PR this query will grow and handle more of the querying

atd-vze/src/queries/crashes.js
```diff
@@ -118,6 +118,21 @@ export const GET_CRASH = gql`
...
 `;
 
+export const GET_CRASH = gql`
+  query CrashDetails($crashId: Int!) {
+    crashes_by_pk(crash_id: $crashId) {
+      crash_id
+      address_primary
+      address_secondary
+      crash_injury_metrics {
+        cris_fatality_count
```
johnclary, line 128: instead of `cris_fatality_count`, we should use `vz_fatality_count`. my bad—i typo'd this in the issue description

atd-vze/src/views/Crashes/Crash.js
```diff
@@ -44,7 +44,15 @@ import {
...
+  });
+  const {
+    loading: crashLoading,
+    error: crashError,
+    data: crashData,
+    refetch: crashRefetch,
+  } = useQuery(GET_CRASH, {
     variables: { crashId },
   });
   const {
```
johnclary, line 58: you can use `years_of_life_lost` from the `crashData` and ditch the code that calculates this metric from people. looks like there will be a nice chunk of code to slash, actually ✂️

atd-vze/src/views/Crashes/Crash.js
johnclary: yep—this is great!

johnclary (changes requested):
just a little change needed. i love the way you've set us up to transition to the new data sources! 🙌

atd-vze/src/views/Crashes/Crash.js
roseeichelmann: ahh thanks for catching this! i meant to circle back to it and then completely forgot :face_with_spiral_eyes:

atd-vze/src/views/Crashes/Crash.js
```diff
@@ -220,9 +204,7 @@ function Crash(props) {
         </Col>
         <Col xs="12" sm="6" md="4">
           <Widget02
-            header={`${
-              totalYearsOfLifeLost === null ? "--" : totalYearsOfLifeLost
-            }`}
+            header={`${yearsOfLifeLost === null ? "--" : yearsOfLifeLost}`}
```
johnclary, line 207: interesting to see `--`. i think you can simplify this to `yearsOfLifeLost || 0`. i don't know why we would prefer `--` over `0`

atd-vze/src/views/Crashes/PeopleDetailsCard.js
```diff
@@ -24,10 +24,13 @@ const PeopleDetailsCard = ({ isExpanded, toggleAccordion, ...props }) => {
   const crashId = props.match.params.id;
 
   const { data: lookupSelectOptions } = useQuery(GET_PERSON_LOOKUPS);
-  const { data, refetch } = useQuery(GET_PEOPLE, {
+  const { loading, error, data, refetch } = useQuery(GET_PEOPLE, {
     variables: { crashId },
   });
 
+  if (loading) return "Loading...";
+  if (error) return `Error! ${error.message}`;
```
johnclary, line 32: thanks for adding this 👍

johnclary (approved):
requesting a tiny lil change but adding my approval as well. good stuff! 🚢

atd-vze/src/views/Crashes/Crash.js
roseeichelmann: done! i dont even think its possible for yearsOfLifeLost to be null with the way its calculated but its good to have this catch just in case

atd-vze/src/queries/crashes.js
tillyw: this makes sense!

tillyw (approved):
so excited to see the new data model in action!! applied the migrations and metadata and ran the cris import script totally smoothly and was able to test out a number of crashes to confirm the headers are displaying correctly. code looks great too, i love the iterative approach for transitioning to the new query - looking forward to the catharsis of deleting the old query. thanks y'all!

atd-vze/src/views/Crashes/Crash.js
mddilley: +1 - love the destructuring! I think that this could use some more default values for the nested keys like `vz_fatality_count` because it will break if it ever hits that empty object fallback. It probably never will because of the loading state check further up in the component, but it is something that stuck out because it has bitten me in the past.

mddilley (approved):
I tested this locally, and every crash that I looked at matches up with the figures that I'm seeing in the list view. Thanks for improving the code too. It is awesome to see the metrics view make the code clearer overall! 🚀 🙌

PR #1455, opened 2024-05-16
bump paths-filter version and use  setting

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/17354

The [paths-filter docs](https://github.com/dorny/paths-filter/tree/master?tab=readme-ov-file#examples) has an example called `Long lived branches: Detect changes against the most recent commit on the same branch before the push` that covers the behavior that i think we want, by adding  `base: ${{ github.ref }}` to our settings.

## Testing

I think we need to merge this to master and then prod. Got any better ideas?

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/.github/workflows/build_docker_images.yml
+++ b/.github/workflows/build_docker_images.yml
@@ -40,12 +40,13 @@ jobs:
           password: ${{ secrets.DOCKER_TOKEN }}
 
       - name: "Checkout"
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: "Paths filter"
-        uses: dorny/paths-filter@v2
+        uses: dorny/paths-filter@v3
         id: changes
         with:
+          base: ${{ github.ref }}
           filters: |
             afd_ems:
               - 'atd-etl/afd_ems_import/**'
--- a/atd-etl/socrata_export/README.md
+++ b/atd-etl/socrata_export/README.md
@@ -14,3 +14,7 @@ $ docker compose run --build -it socrata_export python socrata_export.py
 
 In production, they will be run from a DAG which handles starting the containers with
 the needed environment and other parameters.
+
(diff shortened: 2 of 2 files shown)


atd-etl/socrata_export/README.md
mddilley, line 1: thanks for adding this 🙏

.github/workflows/build_docker_images.yml
```diff
@@ -40,12 +40,13 @@ jobs:
...
       - name: "Checkout"
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: "Paths filter"
-        uses: dorny/paths-filter@v2
+        uses: dorny/paths-filter@v3
         id: changes
         with:
+          base: ${{ github.ref }}
```
frankhereford, line 49: Ah, the magic. Thank you and @chiaberry for y'all's insight in this. I was struggling to understand all the moving parts of this git operation and what was going on, to say the least. 🙏

frankhereford (approved):
I don't have any better ideas John, I think your deployment makes sense to me, and at worst, once it's in both places, if we're not satisfied that it's tested sufficiently, we'll be in a place where we can test more easily.

🚢🚢🚢

mddilley (approved):
I say YOLO 🚀 🚢  The config makes sense to me, and I agree that merging from master to prod sounds like a good way to test this! 🧪

mddilley:
@johnclary actually, i assumed that you are going to PR this into prod like a patch. Does that match up with what you are thinking?

johnclary:
@mddilley not exactly what i was thinking but that's a great idea 🙏 

thanks @chiaberry for untangling what was going on with this issue!!

chiaberry (approved):
long live the branch!

PR #1463, opened 2024-05-24
Crash details page: charges related records table

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/17319

I also updated the queries this page uses so that all the components on the crash details page can pull data from one top level crash query. The query refetches whenever any edits are made so the header widgets will update if for example an injury severity changes from non fatal to fatal.

## Testing
**URL to test:** 
Local

**Steps to test:**
1. Follow [the steps](https://github.com/cityofaustin/atd-vz-data/pull/1440) to start the data model collector branch and run the CRIS import ETL, or if you already have it set up use that!
2. Apply the hasura migrations and metadata from this branch
3. Go to a crash with charges, see that the charges table is now pulling data from the new data model.
4. Test updating a persons injury severity and see that the header widgets are updating automatically as well.

---
#### Ship list
- [ ] Code reviewed


atd-vze/src/queries/crashes.js
```diff
@@ -129,6 +129,79 @@ export const GET_CRASH = gql`
...
+          id
+          label
+        }
+      }
+      charges_cris {
+        unit_nbr
+        prsn_nbr
+        citation_nbr
+        charge_cat_id
+        charge
```
roseeichelmann, line 203: decided to make one query for the crash details page that trickles down to all the components related to that crash instead of having separate queries for people, units, charges, etc. thoughts?

atd-vze/src/views/Crashes/ChargesDetailsCard.js
```diff
@@ -47,6 +49,8 @@ const ChargesDetailsCard = ({ chargesData, isExpanded, toggleAccordion }) => {
                 (charge, i) =>
                   charge.charge !== "NO CHARGES" && (
                     <tr key={`charges-${i}`}>
+                      <td>{charge.unit_nbr}</td>
+                      <td>{charge.prsn_nbr}</td>
                       <td>{charge.charge}</td>
                       <td>{charge.charge_cat_id}</td>
```
roseeichelmann, line 55: noticed that the `charge_cat_id` is null for every entry in the csvs you provided @johnclary ?

atd-vze/src/views/Crashes/Crash.js
```diff
@@ -292,7 +292,13 @@ function Crash(props) {
       )}
       <Row>
         <Col>
-          <CrashCollapses data={data} props={props} />
+          <CrashCollapses
+            data={crashData.crashes_by_pk}
+            refetch={refetch}
+            loading={loading}
+            error={error}
+            props={props}
```
roseeichelmann, line 300: passing all of this down thru the components now so we can use just the one top level crash query

atd-vze/src/views/Crashes/ChargesDetailsCard.js
johnclary: ah yes—that is just CRIS being CRIS. it's always blank. it can be removed from the VZE ❌

roseeichelmann: i see that we did get values for that in our old table, so is that just a field theyve since dropped?

johnclary: according to the CRIS extract specification it was deprecated in 2010:
Screenshot 2024-05-29 at 12 48 20 PM

so let's go ahead and remove it from the VZE 👍

roseeichelmann: Cool added Citation Number as a field to the table

johnclary (approved):
at long last—charges are back!! yay 🎉 

🚢 🚢 🚢 🚢  thank youuu

atd-vze/src/queries/people.js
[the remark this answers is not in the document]
roseeichelmann: went ahead and deleted it in the next PR for data tables :p

PR #1468, opened 2024-06-04
Allow multiple related record tables to be expanded at the same time

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/17560

## Testing
**URL to test:** 
Local

**Steps to test:**
1. Follow instructions in https://github.com/cityofaustin/atd-vz-data/pull/1440 to start the data model collector branch and run the CRIS import ETL, or if you already have it set up use that!
2. Navigate to a crash, see that you can now open all the related records tables at the same time. They should all default to open. Test editing fields.

---
#### Ship list
- [ ] Code reviewed


atd-vze/src/views/Crashes/UnitDetailsCard.js
```diff
@@ -35,29 +29,22 @@ const UnitDetailsCard = ({
 
   const { data: lookupSelectOptions } = useQuery(GET_UNIT_LOOKUPS);
 
-  if (loading) return "Loading...";
-  if (error) return `Error! ${error.message}`;
-
-  return loading ? (
-    <span>Loading...</span>
```
roseeichelmann, line 42: i dont thinkkk we actually need this here

atd-vze/src/views/Crashes/PeopleDetailsCard.js
```diff
@@ -16,19 +16,10 @@ import { primaryPersonDataMap, personDataMap } from "./peopleDataMap";
...
-  loading,
-  ...props
-}) => {
-  const { data: lookupSelectOptions } = useQuery(GET_PERSON_LOOKUPS);
+const PeopleDetailsCard = ({ data, refetch, ...props }) => {
+  const [isOpen, setIsOpen] = useState(true);
 
-  if (loading) return "Loading...";
-  if (error) return `Error! ${error.message}`;
+  const { data: lookupSelectOptions } = useQuery(GET_PERSON_LOOKUPS);
```
johnclary, line 22: i think there is a hidden race condition here where if you tried to edit a person or unit record before the lookups query resolves, you would hit an error.  i'm not sure if we want to block rendering of the related records table until we have the lookup values, or handle this further down the tree where they are used by the related records table component 🤷 

either way, i think it would be good to add some kind of handler to prevent this edge case where the lookups query fails or is low. i guess you could also move the lookups queries into the single mega query we use for the crash data 😈

johnclary (changes requested):
wow—this is such an improvement. and another class component bites the dust! it would be nice to handle the lookup query loading/error states, otherwise i think this is good to go! 🙏

atd-vze/src/views/Crashes/PeopleDetailsCard.js
roseeichelmann: i think it makes sense to handle it by blocking rendering of related records until the query is finished?

atd-vze/src/views/Crashes/UnitDetailsCard.js
chiaberry: also a few lines above it, it would return anyway "Loading...."

should that be removed too?

chiaberry: Oh duh, now I see that you added those lines in this PR -- disregard my removal comment!!

chiaberry (approved):

atd-vze/src/views/Crashes/CrashCollapses.js
mddilley, line 1: one more class component down! 🗑️

atd-vze/src/views/Crashes/UnitDetailsCard.js
mddilley: also took me a sec to see it is the lookup data loading state 🙃 Love how all of this works now!

atd-vze/src/views/Crashes/Crash.js
```diff
@@ -271,13 +273,19 @@ function Crash(props) {
...
       <Row>
         <Col>
-          <CrashCollapses
-            data={crashData.crashes_by_pk}
-            refetch={crashRefetch}
-            loading={loading}
-            error={error}
-            props={props}
-          />
+          <Card>
```
mddilley, line 276: Super small detail but this card wrapper is giving this section an extra wide border and some white background that is revealed by the border radius where contained cards meet.

mddilley (approved):
This is such a cool enhancement!! I can collapse/expand each one separately. 🪗 🚀

atd-vze/src/views/Crashes/Crash.js
johnclary: 🙏 thank you — i just cleaned this up in my the custom ID branch: https://github.com/cityofaustin/atd-vz-data/pull/1465/commits/a4de86929be7e794d46d5fc7313bf43978dfba23

johnclary:
in the interest of branch sanity i'm going to merge this one. thank you @roseeichelmann and reviewers!

PR #1470, opened 2024-06-06
Crash details page: crash diagram + narrative

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/17183

## Testing
**URL to test:** 
Local

**Steps to test:**
1. Follow instructions in https://github.com/cityofaustin/atd-vz-data/pull/1440 to start the data model collector branch and run the CRIS import ETL, or if you already have it set up use that!
2. Because all of the cr3 columns the crash diagram component uses are currently `null` until we do a back fill in the new `crashes` table, to test you will need to manually copy over some fields to a record.
3. In your db client, find a crash record in the new `crashes` table. Now find that same crash in the old `atd_txdot_crashes` table. Copy over the following columns from `atd_txdot_crashes` to the record in `crashes`: `cr3_stored_flag` (now a boolean instead of a string) and `cr3_file_metadata`
4. Now if you go to the crash details page for that crash id you should see the crash diagram there. You should also see a button to download CR3 pdf though you wont actually be able to locally.
5. Observe that the crash narrative is also pulling from the new table

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vze/src/views/Crashes/Crash.js
+++ b/atd-vze/src/views/Crashes/Crash.js
@@ -124,11 +124,8 @@ function Crash(props) {
   const {
     latitude_primary: latitude,
     longitude_primary: longitude,
-    cr3_stored_flag: cr3StoredFlag,
     temp_record: tempRecord,
     geocode_method: geocodeMethod,
-    cr3_file_metadata: cr3FileMetadata,
-    investigator_narrative_ocr: investigatorNarrative,
   } = !!data?.atd_txdot_crashes[0] ? data?.atd_txdot_crashes[0] : {};
 
   const {
@@ -137,6 +134,9 @@ function Crash(props) {
     address_primary: primaryAddress,
     address_secondary: secondaryAddress,
     crash_injury_metrics_view: { years_of_life_lost: yearsOfLifeLost },
+    investigator_narrative: investigatorNarrative,
+    cr3_stored_flag: cr3StoredFlag,
+    cr3_file_metadata: cr3FileMetadata,
   } = crashData?.crashes_by_pk ? crashData?.crashes_by_pk : {};
 
   const mapGeocoderAddress = createGeocoderAddressString(data);
@@ -256,7 +256,7 @@ function Crash(props) {
         <Col xs="12" md="6" className="mb-4">
           <CrashDiagram
             crashId={crashId}
-            isCr3Stored={cr3StoredFlag === "Y"}
+            isCr3Stored={cr3StoredFlag}
             isTempRecord={tempRecord}
             cr3FileMetadata={cr3FileMetadata}
           />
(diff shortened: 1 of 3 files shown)


atd-vze/src/views/Crashes/CrashDiagram.js
```diff
@@ -119,27 +124,27 @@ const CrashDiagram = props => {
...
-              href={"https://cris.dot.state.tx.us/"}
-              target={"_blank"}
-              rel="noopener noreferrer"
-            >
-              CRIS website
-            </a>{" "}
-            for the latest status of this crash.
-          </div>
         ) : (
+          // ) : props.isTempRecord ? (
```
roseeichelmann, line 128: just commenting this out for now until we figure out how we are going to handle temp records in the new db

atd-vze/src/views/Crashes/CrashDiagram.js
```diff
@@ -14,15 +14,20 @@ import axios from "axios";
 import { TransformWrapper, TransformComponent } from "react-zoom-pan-pinch";
 import { isDev } from "../../helpers/environment";
 
-const CrashDiagram = props => {
+const CrashDiagram = ({
+  crashId,
+  isCr3Stored,
+  isTempRecord,
+  cr3FileMetadata,
+}) => {
```
johnclary, line 22: 🙏

johnclary (approved):
yep—this is perfecto 💯 

side note: i noticed that the `cr3_stored_flag` and `cr3_file_metadata` columns are missing from the `crashes_edits` table, which was not intentional. and like you mentioned, Rose, we can drop the `investigator_narrative_ocr` column. i will include those changes in my next batch of column migrations 👍 

🚢 🚢 🚢 🚢 🚢

atd-vze/src/views/Crashes/Crash.js
```diff
@@ -137,6 +134,9 @@ function Crash(props) {
     address_primary: primaryAddress,
     address_secondary: secondaryAddress,
     crash_injury_metrics_view: { years_of_life_lost: yearsOfLifeLost },
+    investigator_narrative: investigatorNarrative,
+    cr3_stored_flag: cr3StoredFlag,
+    cr3_file_metadata: cr3FileMetadata,
   } = crashData?.crashes_by_pk ? crashData?.crashes_by_pk : {};
```
chiaberry, line 140: this is what i was mentioning in dev sync, that the nested entries will error out when its the fallback `{}` .

chiaberry (approved):
great test steps, I see the diagram and a button!

atd-vze/src/views/Crashes/Crash.js
johnclary: thanks for flagging this Chia! i have a fix staged in my ID column branch.

johnclary:
I am merging this!

PR #1477, opened 2024-06-21
Use `crash_timestamp` and `prsn_death_timestamp` columns

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/17954

This ended up being a bit more of a lift than i was expecting, but the majority of this diff is just codegen and linting 😅 

## Testing
1. Start your data model stack as usual and run the CRIS import ETL.
2. Check that all of the VZE pages load and interact normally
3. Pick a few crashes at random and compare the crash date you see locally versus in prod. They should match if you're testing from US/Central. Otherwise you will notice that the crash date is always rendered in local time. (That change is not specific to this PR, but I have fixed up the formatting of the timestamp strings).
4. Query the people table and compare  `prsn_death_date` and `prsn_death_time` to  `prsn_death_timestamp`
---
#### Ship list
- [ ] Code reviewed


atd-vzd/migrations/default/1715960019005_crashes_list_and_fatality_metrics_views/up.sql
```diff
@@ -222,9 +222,15 @@ select
     lookups.collsn_lkp.label as collsn_desc,
     geocode_status.is_manual_geocode,
     geocode_status.has_no_cris_coordinates,
+    to_char(
+        public.crashes.crash_timestamp at time zone 'US/Central', 'YYY:MM:DD'
+    ) as crash_date_ct,
+    to_char(
+        public.crashes.crash_timestamp at time zone 'US/Central', 'HH24:MI:SS'
+    ) as crash_time_ct,
```
johnclary, line 230: xavier mentioned wanting these, and we're going to want to expose these fields in crash exports to replicate old functionality, so i went ahead and created them here and will add them to the crashes export in a follow up when I take on https://github.com/cityofaustin/atd-data-tech/issues/17944

atd-vzd/migrations/default/1715960022005_column_metadata_table/up.sql
johnclary, line 1: i changed up the logic in the helper script that flags if a column comes in via the CRIS ETL. this made it easier to clean up the cris import script, which will be cleaned up in a much more substantial way in the not too distant future.

atd-toolbox/data_model/python/column_metadata.json
johnclary, line 1: ugh—linting! there are a few small changes in here which define the new `crash_timestamp` and `prsn_death_timestamp` columns.

atd-vzd/migrations/default/1715960003005_crashes/up.sql
johnclary, line 1: except for the views, all of the changes to the migration files were generated via Python as a result of the changes to the column metadata 👍

atd-vze/src/Components/GridTable.js
johnclary, line 1: While working on this PR i realized that we still need to do some more ** light ** refactoring of the grid table/gqlabstract to properly handle timezones when searching by date. at the moment the crashes list is querying the `crash_timestamp` column with a local date that is being treated as a UTC date.

atd-vzd/migrations/default/1715960019005_crashes_list_and_fatality_metrics_views/up.sql
roseeichelmann: sounds/looks perfect!

roseeichelmann (approved):
This works great! Nice that the Last Updated and Crash Date are now rendering in the same format on the crash details page :ship:

PR #1478, opened 2024-06-26
CRIS import and CR3 ETLs

johnclary:
## Associated issues

- https://github.com/cityofaustin/atd-data-tech/issues/18037
- https://github.com/cityofaustin/atd-data-tech/issues/18038

This is the new CRIS import, complete with CR3 pdf processing. This is ready for review, but please keep in mind these follow-up todos which I intend to address in follow-up issues (pending your feedback + approval):

- Archive the old ETLs and rename this ETL subdirectory
- Expand readme
- Consolidate / clean up our various S3 buckets
- Revisit / extend the logging in the `cris_import_log` table, and probably rename that table to `_cris_import_log`
- Github CI, 1pass integration, airflow deployment, etc.

### Setup

1. Start your local Vision Zero stack (database + Hasura + editor) using a recent copy of production

2. from the `./atd-vzd` directory, apply migrations and metadata:

```shell
$ hasura migrate apply
$ hasura metadata apply
```

3. Grab a copy of the environment file from our 1pass dev vault. The item is named `Env file for the Vision Zero new data model CRIS import ETL`. Save it as `.env` in the `./atd-etl/data_model` directory

4. Build the docker image—you only need to do this once

```shell
# from ./atd-etl/data_model
$ docker compose build
```

5. Start the Docker container and drop into it's shell

```shell
# from ./atd-etl/data_model
$ docker compose run cris_import
```

6. It's time to run the CRIS import script! you're going to test a few different CLI options, described below.

### End-to-end CRIS import

This will download each extract available in the S3 `./inbox`, unzip it, load the CSV crash records into the database, crop crash diagrams out of the CR3 PDFs, and upload the CR3 pdfs and crash diagrams to the s3 bucket.

```shell
# from the cris_import container's shell
$ ./cris_import.py --csv --pdf --s3-download --s3-upload 
```

- Open your VZE and verify that the crashes list loads normally
- Open a crash details page and verify that the crash diagram renders properly
- The CR3 download button should be enabled, but clicking it will result in an error. This is a known issue, because the CR3 API is pointed at a different bucket
- Scroll down to the change log and observe that the change log includes updates to `cr3_processed_at` and `cr3_stored_fl` fields
- Head to the AWS console and locate the PDFs and diagrams in the S3 bucket: `vision-zero-new-data-model-dev/dev/cr3s`. Observe that the crash diagrams and PDFs have a **Last modified** timestamps that track with when you ran the import script.
- Use your SQL client to query the `cris_import_log` table. Verify that there are new entries for each extract you processed

```sql
select * from cris_import_log order by id desc;
```

### Local import

This will process the extract zips that were downloaded to your `./extracts` directory during the previous step. CSVs will be loaded ino the db, and crash diagrams will be extracted but not uploaded to S3.

```shell
$ ./cris_import.py --csv --pdf
```

### Archive and un-archive zips

The script can archive the extract zips by moving them from `./inbox` to `./archive` once they have been processed. This is intended for the production deployment, where the `./inobx` functions as a work queue.

```shell
# download zips from ./inbox, process csvs, and archive the zips when done
$ ./cris_import.py --csv --s3-download --s3-archive
```

Restore the zips to the `./inbox` using the helper script

```shell
$ python _restore_zips_from_archive.py
```

### Other flags to test

```shell
# process any unzipped extract CSVs you have in your `./extracts` directory
$ ./cris_import.py --skip-unzip --csv

# process pdfs with more workers and in debug mode
$ ./cris_import.py --pdf --skip-unzip --s3-upload --workers 8 --verbose
```


atd-etl/data_model/utils/files.py
```diff
@@ -0,0 +1,163 @@
...
+    return f"{round(bytes/1000000)}mb"
+
+
+def get_extract_name_from_file_name(file_key):
+    """Parse the full extract name from the file name"""
+    return os.path.basename(file_key).replace(".zip", "")
+
+
+def get_unzipped_extracts_local():
+    """Gets a list of sudirectories in the <LOCAL_EXTRACTS_DIR>, making the assumption that any subdirectory
```
johnclary, line 30: i initially wrote this by making use of temporary directories, but i found that to be pretty complicated to implement in combination with the option of supplying a zip file locally, and i also found it quite annoying to debug the PDF processing when files were being saved to an ephemeral temp dir.

for the production Airflow + docker deployment, local directories will essentially be treated the same as temporary directories, since containers will be removed after each run. so i think this is ok?

atd-etl/data_model/utils/process_pdfs.py
```diff
@@ -0,0 +1,165 @@
...
+def process_pdf(extract_dir, filename, s3_upload, index):
+    """Handles processing of one CR3 PDF document.
+
+    Args:
+        extract_dir (str): the local path to the current extract
+        filename (str): the filename of the CR3 PDF process, e.g. 123456.pdf
+        s3_upload (bool): if the diagram and PDF should be uploaded to the S3 bucket
+        index (int): the index ID of this pdf among all PDFs being processed
+    """
+    logger.info(f"Processing {filename} ({index})")
```
johnclary, line 81: if you'd like to test the error handling, you can raise an exception here:

```python
raise ValueError(f"something went wrong")
```

atd-etl/data_model/utils/process_pdfs.py
johnclary, line 1: there are a few important changes in this new version of the PFD processor:
- we are no longer OCR-ing crash narratives. these are included in the CRIS extract now
- we're using just two fields to track the state of the CR3 PDF and diagram: `cr3_processed_at` and `cr3_stored_fl`. We can query to see if we have the most current CR3 PDF by comparing the `cr3_processed_at` versus the most recent entry in the`crashes_cris` change log
- we are no longer storing crash diagrams using a `uuid`. they are being saved into the S3 bucket as `<crash_id>.jpeg`.
- we are storing crash diagrams as JPEGs instead of PNGs, since they are smaller files and we have no need for transparency.

atd-toolbox/data_model/python/10_make_core_tables.py
```diff
@@ -90,6 +90,8 @@ def make_column_sql(columns, full_table_name):
             elif full_table_name == "charges":
                 fk_table_name = "people"
             fk_cascade = "on update cascade on delete cascade"
+        elif constraint and "not null default false" in constraint and "_edits" in full_table_name:
+            constraint = ""
```
johnclary, line 94: fixes a bug in our DB schema in which we had default values in our `_edits` schema

atd-etl/data_model/utils/process_pdfs.py
```diff
@@ -0,0 +1,174 @@
...
+            filename = pdfs[index]
+            errors.append([filename, e])
+
+    if errors:
+        logger.error(
+            f"Encountered {len(errors)} error(s) processing PDFs. Logging up to 10 errors and raising the first..."
+        )
+        for filename, e in errors[0:10]:
+            logger.info(f"Error processing {filename}: {e}")
+        raise errors[0][1]
```
johnclary, line 170: because of concurrency, we will not know if the PDF processing is bombing until _all_ PDFs have been processed.

this isn't a major concern for processing a normal-sized extract of ~200 crashes, but really not ideal if we were processing thousands of PDFs. like would take us a few hours to find out we had a misconfigured env var.

i didn't want to over-engineer this off the bat, but one improvement we could make would be to process the PDFs in chunks of 100, inspecting each chunk for errors before running another batch.

atd-etl/data_model/README.md
roseeichelmann, line 1: Would it be helpful to put somewhere in the README what to expect if you just run the cris_import.py script without any CLI args? Like it will just process local zip files, and will skip unzipped folders.

atd-etl/data_model/utils/process_csvs.py
```diff
@@ -120,35 +57,56 @@ def get_cris_columns(column_metadata, table_name):
 
 
 def make_upsert_mutation(table_name, cris_columns):
-    """Sets the column names in the upsert on conflict array"""
-    upsert_mutation = mutations[table_name]
+    """Retrieves the graphql query string for a record mutation
+
+    Args:
+        table_name (str): `crashes`, `units`, `charges`, or `people`
+        cris_columns (list): a list of column names that should updated if a given record
```
roseeichelmann, line 64: most minorrr of typos "should updated"

atd-etl/data_model/utils/graphql.py
```diff
@@ -0,0 +1,180 @@
...
+def make_hasura_request(*, query, variables=None):
+    """Make a POST request to the graphql API.
+
+    Requires HASURA_GRAPHQL_ENDPOINT env var.
+
+    Args:
+        query (str): the graphql query
+        variables (dict, optional): the query variables. Defaults to None.
+    Raises:
+        HasuraAPIError: If the API the response JSON not contain a top-level `data` property
```
roseeichelmann, line 118: typo. btw loving all this documentation though

atd-etl/data_model/utils/graphql.py
```diff
@@ -0,0 +1,180 @@
...
+
+import requests
+
+from utils.exceptions import HasuraAPIError
+
+ENDPOINT = os.getenv("HASURA_GRAPHQL_ENDPOINT")
+ADMIN_SECRET = os.getenv("HASURA_GRAPHQL_ADMIN_SECRET")
+
+COLUMN_METADATA_QUERY = """
+query ColumnMetadata {
```
roseeichelmann, line 12: this `_column_metadata` table is sooo clutch im so glad we are doing this! reminding me of our recent convos on documentation, would be good to have a protocol written up somewhere on adding/deleting columns that lists the steps, ie. having to add new columns to this table. also is there a constraint to make sure all columns in the tables are in the column metadata table?

atd-vzd/migrations/default/1715960022005_column_metadata_table/up.sql
roseeichelmann, line 1: would it be possible to add PG comments to this table for the columns, ie. what is the diff between `column_name` and `column_label`

atd-etl/data_model/utils/process_pdfs.py
```diff
@@ -0,0 +1,174 @@
...
+        is_new_cr3_form (bool): if the CR3 is in the 'new' format
+        extract_dir (str): the local directory in which to save the file
+
+    Returns:
+        str: diagram_full_path - the full path to the saved diagram, including it's name
+        str: diagram_filename - the name of diagram file, .e.g 12345678.jpeg
+    """
+    bbox = DIAGRAM_BBOX_PIXELS["new"] if is_new_cr3_form else DIAGRAM_BBOX_PIXELS["old"]
+    diagram_image = page.crop(bbox)
+    # todo: is it ok to swith to JPEG (as is done here) and save 75% disk space?
```
roseeichelmann, line 59: methinks yes!!

atd-toolbox/data_model/python/10_make_core_tables.py
roseeichelmann: im a bit confused about what `constraint` means here

johnclary: sorry, the code for these migration helper scripts is pretty spaghettified. 

This script is parsing something like this:

```json
 {
    "column_name": "prsn_exp_homelessness",
    "data_type": "boolean",
    "data_type_constraint": "not null default false",
    "foreign_column_cascade": "",
    "foreign_column_name": "",
    "foreign_table_name": "",
    "foreign_table_schema": "",
    "is_cris_column": true,
    "is_edits_column": true,
    "is_existing_foreign_table": false,
    "is_primary_key": false,
    "is_unified_column": true,
    "note": "",
    "record_type": "people",
    "source": "cris"
  }
```

and turning it into column sql like 

```
prsn_exp_homelessness boolean not null default false
```

so `constraint` in this case is whatever text is in `data_type_constraint` in the json blob

roseeichelmann (approved):
this looks sooo great! all the test steps worked perfectly, im loving all the documentation and the diff argument options for the script. this was so much work and you did a great job of making this complex code as organized and simple to follow as possible. so awesome to have this huge chunk of work towards the data model done!! :rocket:

atd-toolbox/data_model/python/10_make_core_tables.py
johnclary: but in the `people_edits` table we don't want the column to have a default value, so i added that extract logic to remove the default value for boolean columns in the `_edits` tables.

atd-toolbox/data_model/python/column_metadata.json
mddilley, line 1: @johnclary there are duplicates of fields in this config json and in the `_column_metadata` table. Is this expected?

atd-vzd/migrations/default/1715960022005_column_metadata_table/up.sql
mddilley: when i exploring the `_columns_metadata` table, i also wondered about the difference between these two columns and ran `SELECT * from _column_metadata WHERE column_name != column_label` with no rows returned. Will these two columns always have the same value?

atd-vze/src/views/Crashes/CrashDiagram.js
```diff
@@ -56,122 +49,125 @@ const CrashDiagram = ({
...
-              <Button color="primary" onClick={requestCR3}>
-                Download CR-3 PDF
-              </Button>
-            ) : (
-              <div></div>
-            )}
+            <Button
+              color="primary"
+              onClick={requestCR3}
+              disabled={!isCr3Stored}
```
mddilley, line 55: love this change 🙌

mddilley (approved):
I ran through all the steps, and I see everything happening just as described!! There is some really nice Python in here 🐍 💯, and the consistent docs throughout is just really nice. I made a few small comments, and there might be one issue (or i could be misunderstanding) that I mentioned you in. Either way, the process is successful, and I think 🚢  it.

atd-etl/data_model/utils/files.py
```diff
@@ -0,0 +1,146 @@
...
+                    "local_zip_file_path": os.path.join(
+                        LOCAL_EXTRACTS_DIR, os.path.basename(key)
+                    ),
+                    "file_size": item.get("Size"),
+                    "extract_name": get_extract_name_from_file_name(key),
+                }
+            )
+
+    if not (len(extracts)) and ENV != "dev":
+        raise IOError("No extract zips found in S3 bucket")
```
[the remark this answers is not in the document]
johnclary: yes! and to simplify things i'm going to remove the env check (`ENV != "dev"`). i will also add a similar exception for PDF processing

atd-etl/data_model/utils/files.py
johnclary: nice—thanks for doing that research 🙌

atd-vzd/migrations/default/1715960022005_column_metadata_table/up.sql
johnclary: thanks for flagging the various issues with this table. it still needs to be fully fleshed out and and tightened up. to address the feedback from all your comments here:
- dupe column entries: this is an artifact of when we were using the excel sheet and had duplicate person-level column names, one for primary and one for the non-primary person tables. i will clean this up.
- column constraints: i'll double check we have them in place
- column comments: yes, for sure. i've been saving comments for later in our development because it will be tedious to manage them in the migrations. i will won't let this fall through the cracks.
- `column_name` vs `column_label`: the vision for this was that `column_label` could hold the label we render in the VZE, like  the `"crash_id"` column has the label `"Crash ID"`. i will drop this column and we are already tracking a follow-up todo to expand how we use this table if we want to.
- keeping this column up to date/current: for one thing, we will have to add columns to this table if we want to bring them in via CRIS import. for other, non-cris import columns, we can either (1) decide not to track them in this table or (2) set up an automated test that validates that we are tracking all of our columns here by comparing this data to the `information_schema.columns` built-in view. i'm filing this concern under the "change management" todos in our doc.

atd-etl/data_model/utils/process_pdfs.py
johnclary: 👍 will open a follow-up issue

atd-etl/data_model/README.md
johnclary: yes—done! 🙏

johnclary:
thanks again for the feedback. i am re-requesting review after having made the following changes:
- modify CLI to require at least one of `--csv` or `--pdf`. running `$ cris_import.py` without any other args will throw an error 👍 
- expand the readme
- raise an error if no pdfs are processed
- fix a bug where the `created_by` and `updated_by` columns were being removed and therefore not being set to `cris`
- fix some typos

i'm going to leave the `_column_metadata` improvements for a separate PR.

atd-etl/data_model/README.md
```diff
@@ -35,23 +35,60 @@ Create your environment by saving a copy of the `env_template` file as `.env`. T
...
-## Usage
+## Usage examples
 
 The only script that should be run directly is `cris_import.py`. It supports the following CLI args:
 
 ```shell
-  --csv              Only process CSV files
-  --pdf              Only process CR3 pdfs
+  --csv              Process CSV files. At least one of --csv or --pdf is required.
+  --pdf              Process CR3 pdfs. At least one of --csv or --pdf is required.
```
roseeichelmann, line 50: im liking this update you made to require --csv or --pdf and you can use both arguments at once. its more intuitive :))

atd-etl/data_model/README.md
```diff
@@ -35,23 +35,60 @@ Create your environment by saving a copy of the `env_template` file as `.env`. T
...
   --s3-upload        Upload cr3 pdfs and digrams to S3 bucket
   --s3-archive       If using --s3-download, move the processed extracts from ./inbox to ./archive when done
   --skip-unzip       Only process files that are already unzipped in the local directory
   --verbose, -v      Sets logging level to DEBUG mode
   --workers <int>    The number of concurrent workers to use when processing PDFs
 ```
+
+### Producton run
+
+This is the expected invocation during a production deployment. It will download any extracts available in S3, load the CSV crash records into the database, crop crash diagrams out of the CR3 PDFs, upload the CR3 pdfs and crash diagrams to the s3 bucket.
```
roseeichelmann: love that you added this production run example as well

atd-etl/data_model/README.md
```diff
@@ -35,23 +35,60 @@ Create your environment by saving a copy of the `env_template` file as `.env`. T
...
+
+Process any extract zips in your local `./extracts` directory. CSVs will be loaded ino the db, and crash diagrams will be extracted but not uploaded to S3.
+
+```shell
+$ ./cris_import.py --csv --pdf
+```
+
+### Un-archive
+
+During local devleopment, you may want to restore testore the zips to the `./inbox` after archiving them. Use the helper script for that. It will prompt you for confirmation before executing this step, since the production bucket archive may contain hundreds of extacts.
```
roseeichelmann: some typos here

roseeichelmann:
i notice when you restore from archive the extracts get copied over to the inbox but stay in the archived folder too. i guess that makes sense as expected behavior but just wanted to point it out incase yall think it makes more sense to move them?

roseeichelmann (approved):
this looks great john, i like the updates you made to the readme and the CLI modifications. i think i already mentioned this in a data model meeting but wanna flag it again so we/i dont forget, but the columns in the `cris_import_log` table `records_processed` and `outcome_status` are currently all null/empty and nothing is being done with them, i kinda forgot where our convo landed with that but i think you maybe said those columns are artifacts from the old table and we arent sure if we want to keep them around?

atd-etl/data_model/README.md
```diff
@@ -0,0 +1,94 @@
+# CRIS Import ETL
+
+This ETL manages the processing and importing of TxDOT CRIS data into the Vision Zero database.
+
+All data processing is managed by a single script, `cris_import.py` which processes both CSV files and CR3 PDF crash reports. The script supports a number of CLI arguments to handle a number of data processing scenarios.
+
+## Quick start
+
+Follow these steps to run the ETL locally. All interactions with AWS S3 occur with against a single bucket which has subdirectores for the `dev` and `prod` environments. If you set your `ENV` to `dev` you can safely run this ETL's S3 operations.
```
chiaberry, line 9: "All interactions with AWS S3 occur with against a single" 
either remove with or against

chiaberry: actually what about removing this part, since its repeated below under environment and then for here just having it say "Set your ENV to dev to safely run this ETL locally"

atd-etl/data_model/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+Process any extract zips in your local `./extracts` directory. CSVs will be loaded ino the db, and crash diagrams will be extracted but not uploaded to S3.
+
+```shell
+$ ./cris_import.py --csv --pdf
+```
+
+### Un-archive
+
+During local devleopment, you may want to restore testore the zips to the `./inbox` after archiving them. Use the helper script for that. It will prompt you for confirmation before executing this step, since the production bucket archive may contain hundreds of extacts.
```
chiaberry, line 80: a few typos here, "testore" and extacts

atd-vze/src/views/Crashes/CrashDiagram.js
```diff
@@ -56,122 +49,125 @@ const CrashDiagram = ({
...
-            }}
-          >
-            {({ zoomIn, zoomOut, resetTransform, ...rest }) => (
-              <>
-                <Row className="my-2">
-                  <Col>
-                    <ButtonGroup>
-                      <Button color="primary" onClick={zoomIn}>
-                        <i className="fa fa-search-plus"></i>
+      {!diagramError && (
```
chiaberry, line 62: ✨

chiaberry (approved):
I tested the different flags and this is gonna be such a good upgrade to our workflow 🥇

atd-etl/data_model/README.md
mddilley: +1 🚀 🙌

mddilley (approved):
I reviewed the latest commits, and it all looks great. Thanks for the new details in the docstrings and readme too. Super helpful to know the combinations of CLI args and how they are used in the production runs. 🚢

johnclary:
> i notice when you restore from archive the extracts get copied over to the inbox but stay in the archived folder too. i guess that makes sense as expected behavior but just wanted to point it out incase yall think it makes more sense to move them?

i did this partially out of laziness, but was also thinking it's probably nice to have that breadcrumb of seeing the extract in the `./archive` and knowing it was processed. happy to change this though 🤷

johnclary:
@roseeichelmann i am tracking a few follow-up todos for the `cris_import_log`. will not forget them! 🙏

PR #1480, opened 2024-06-27
Bump Hasura version and VZE and VZV package versions

johnclary:
I forgot to include this in the v1.48.0 release :/


mddilley (approved):
And i forgot to check for that in the review... 🚢 🙏 🙏

PR #1483, opened 2024-07-03
Remove unused components from VZE

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/17392

Ended up finding a lot more unused components than just the Charts

## Testing
**URL to test:** 
Locally

**Steps to test:**
1. Check out this branch and start your data model DB
2. Check around the app to make sure nothing is broken
3. Admire all the red :heart_eyes: 

---
#### Ship list
- [ ] Code reviewed


johnclary:
@roseeichelmann would you mind pointing this PR against `data-model-v2` instead of `master`?

atd-vze/src/routes.js
```diff
@@ -71,70 +42,6 @@ const routes = roles => [
   { path: "/dev/theme", exact: true, name: "Theme", component: Colors },
```
johnclary, line 42: any reason not to get rid of these other pieces? Notifications, Theme, Dev...?

johnclary (changes requested):
I tested this locally after merging it with the data model collector branch—all looks great!

I guess, fwiw, it might actually be helpful to preserve all of this for a future product manager who wanted to [browse all the UI components](https://visionzero.austin.gov/editor/#/dev), but I would rather see a tool like this hosted in a separate fork or something. As it is, this feature bogs down the code base and I'm very happy to see it removed. 

It looks like there are a few more files/routes that can be deleted, then this will be good to go! 🙌

atd-vze/src/routes.js
roseeichelmann: doneee

johnclary (approved):
😍

atd-vze/src/views/Widgets/Widget02.js
mddilley, line 1: @roseeichelmann I'm getting this error when starting VZE. I merged the latest from `data-model-v2`, but maybe I should be pulling some newer code in from somewhere. Let me know, and I'll turn this into an approval.

Screenshot 2024-07-30 at 10 50 54 AM

mddilley (commented):
Thanks for the the cleanup!! 🧹 🙏 I found one error when starting this locally, but I'm not sure if I have the latest code in here.

atd-vze/src/views/Widgets/Widget02.js
johnclary: oops—it does appear some of those widgets were load bearing 😮

roseeichelmann: thanks for catching that mike! i just brought back the two widgets we actually use

mddilley (approved):
I retested, and every view loaded for me! Thanks again for carving out this code, Rose! 🙏 🚢  🎉

mddilley:
Almost certainly a product of the merge party 🥳, but I saw this one compile warning.

Screenshot 2024-08-01 at 10 34 54 AM

johnclary (approved):
it's all working for me! thanks @mddilley for catching those busted widgets.

🚢 
🚢 
🚢 
🚢 
🚢

johnclary:
> Almost certainly a product of the merge party 🥳, but I saw this one compile warning.

ah yes—i saw that this is already fixed in the soft deletes branch 👍 🙏

PR #1484, opened 2024-07-05
 Ability to create temporary crashes in the new data model

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/18039

## Testing
**URL to test:** 
Local

**Steps to test:**
1. Check out this branch and follow instructions in https://github.com/cityofaustin/atd-vz-data/pull/1440 to start the data model, apply migrations and run the CRIS import script
2. Go to the create temp crash page under Tools in the sidebar
3. Note that we now have a crash timestamp field and the address fields auto capitalize your input
4. Test adding temp crashes with various combos of units and fatalities/injuries
5. When your temp crash is successfully created, a banner will pop up at the top with a link to it
6. Temp crashes will have a `record_locator` value of `T<id>` normal crashes will have a `record_locator` value of just their `<id>`
7. This new `record_locator` field is what is now used for routing
8. In the URL if ur lazy like me u dont need to uppercase the "T", the id will be converted to upper case before querying the database so "http://localhost:3000/#/crashes/t32449" for example will work
9. Test going to the temp records crash details page, editing the units, people, etc. Make sure everything is working

---
#### Ship list
- [ ] Code reviewed


atd-vzd/migrations/default/1715960022005_column_metadata_table/up.sql
roseeichelmann, line 1: would it be possible to add postgres comments to this table for the columns, ie. what is the diff between `column_name` and `column_label`

atd-vze/src/views/Tools/CreateCrashRecord.js
```diff
@@ -253,16 +205,12 @@ const CreateCrashRecord = ({ client }) => {
           );
```
johnclary, line 205: `fatalityCountSum` and `susSeriousInjuryCountSum` can be removed bc they're not used 🚮

atd-vze/src/views/Tools/CreateCrashRecord.js
```diff
@@ -131,14 +97,14 @@ const CreateCrashRecord = ({ client }) => {
       .query({
         query: gql`
           {
```
johnclary, line 99: interesting that we query the DB to check if the case ID is unique, and yet we don't enforce that as a DB constraint.

i was _just_ talking to Xavier about this topic and asked if we should enforce that `case_id` + `investigat_agency_id` pair be unique, and he said no. he actually prefers to check this field for inconsistencies. 

so let's just remove this db query altogether 👍

atd-vze/src/views/Tools/CreateCrashRecord.js
```diff
@@ -253,16 +205,12 @@ const CreateCrashRecord = ({ client }) => {
           );
 
           const crashVariables = {
-            atd_fatality_count: fatalityCountSum,
-            address_confirmed_primary: primaryAddress,
-            address_confirmed_secondary: secondayAddress,
+            rpt_street_name: primaryStreetName.toUpperCase(),
+            rpt_sec_street_name: secondaryStreetName.toUpperCase(),
```
johnclary, line 209: the form bombs here if you try to submit it before filling in these fields 💣

atd-vze/src/views/Tools/CreateCrashRecord.js
```diff
@@ -344,46 +290,23 @@ const CreateCrashRecord = ({ client }) => {
...
+                name="timestamp-input"
+                placeholder="timestamp"
+                value={crashTimestamp}
+                invalid={isFieldInvalid(crashTimestamp)}
+                onChange={e => setCrashTimestamp(e.target.value)}
               />
-              {isFieldInvalid(crashDate) ? (
-                <FormFeedback>Crash Date can't be blank.</FormFeedback>
+              {isFieldInvalid(crashTimestamp) ? (
+                <FormFeedback>Crash Timestamp can't be blank.</FormFeedback>
```
johnclary, line 306: this feedback never appears because we are using a default value for `crashTimestamp`.

weirdly, i don't see that default value in the data input when i load the form. possibly because it's not formatted correctly as a string (e.g. `2024-07-18T13:05`) ?

johnclary: ah yeah now i'm seeing the warning message:
```
react-dom.development.js:1734 The specified value "2024-07-15" does not conform to the required format.  The format is "yyyy-MM-ddThh:mm" followed by optional ":ss" or ":ss.SSS".
```

atd-vze/src/views/Tools/CreateCrashRecord.js
```diff
@@ -27,33 +27,31 @@ import { useAuth0 } from "../../auth/authContext";
...
   const { user } = useAuth0();
 
+  const userEmail = user.email;
+
   const unitsInitialState = [
-    { unit_desc_id: 1, atd_fatality_count: 0, sus_serious_injry_cnt: 0 },
+    { unit_desc_id: 1, fatality_count: 0, sus_serious_injry_cnt: 0 },
   ];
 
   const formInitialState = {
```
johnclary, line 36: this can be moved out of the component since it's constant

atd-vze/src/views/Tools/CreateCrashRecord.js
```diff
@@ -398,8 +321,8 @@ const CreateCrashRecord = ({ client }) => {
                 id="primary-address-input"
                 name="primary-address-input"
                 placeholder="ex: S 900 AUSTIN AVE"
-                value={primaryAddress}
-                onChange={e => setPrimaryAddress(e.target.value)}
+                value={primaryStreetName?.toUpperCase()}
```
johnclary, line 324: i am getting this classic error when i type into this field: `A component is changing an uncontrolled input of type text to be controlled`.

i think the problem is that `{primaryStreetName?.toUpperCase()` returns `undfined`. if you change this to `{primaryStreetName?.toUpperCase() || ""`} it should fix the problem

atd-vze/src/views/Tools/CreateCrashRecord.js
```diff
@@ -253,16 +205,12 @@ const CreateCrashRecord = ({ client }) => {
...
+            rpt_street_name: primaryStreetName.toUpperCase(),
+            rpt_sec_street_name: secondaryStreetName.toUpperCase(),
             case_id: caseId,
-            crash_date: crashDate,
-            crash_fatal_fl: fatalityCountSum > 0 ? "Y" : "N",
-            crash_id: tempId,
-            crash_time: crashTime,
-            sus_serious_injry_cnt: susSeriousInjuryCountSum,
-            updated_by: user.email,
+            crash_timestamp: crashTimestamp,
```
johnclary, line 211: there's a little timezone bug here, because the user-provided timestamp is being stored in UTC. the way that i've handled this in the past, and i think it's sound, is to create a `Date` instance from the user-input string, and then re-stringify it as an ISO stamp:

```javascript
new Date(crashTimestamp).toISOString()
```

atd-vze/src/views/Crashes/Crashes.js
```diff
@@ -55,7 +55,7 @@ const Crashes = () => {
       minDate={minDate}
       roleSpecificColumns={adminColumns}
       hasSpecificRole={isAdminOrItSupervisor}
-      defaultSearchField="crash_id"
+      defaultSearchField="record_locator"
```
johnclary, line 58: i'm glad you changed this, although i'm not quite sure what effect this is having. i can open a follow-up issue to make whatever changes are needed to get temp records into the crashes list

atd-vze/src/views/Tools/CreateCrashRecord.js
```diff
@@ -151,31 +117,21 @@ const CreateCrashRecord = ({ client }) => {
 
           // Build an array of persons objects formated as a string
           // so the String can be interpolated into the gql tag syntax.
-          let personObjects = "[";
-          let unitObjects = "[";
+          let unitObjects = "";
```
johnclary, line 120: if it can be done easily enough, it would be nice to move this payload-building code into a separate function just to keep things readable

atd-vze/src/views/Tools/CreateCrashRecord.js
```diff
@@ -185,66 +141,62 @@ const CreateCrashRecord = ({ client }) => {
...
+            unit_desc_id: ${Number(unit.unit_desc_id)},
+            cris_schema_version: "2023",
+            updated_by: "${userEmail}",
+            created_by: "${userEmail}",
+            people_cris: {
+              data: [
+                ${personObjects}
+              ]
+            }
+          }`);
```
johnclary, line 164: this works fine, but i'd prefer to to see the entire crash record object be included as a single `crashes_cris_insert_input` graphQL variable. it's just that the string-formatting tends to be error-prone and a bit difficult to maintain.

atd-vze/src/views/Tools/CreateCrashRecord.js
roseeichelmann: should we still check that there is a valid/not blank case id?

johnclary (changes requested):
@roseeichelmann this looks really good! it's so awesome to see this all working together!

in addition to the other feedback, i have encountered this very annoying problem when trying to create a unit of type `MICROMOBILITY DEVICE` (id=`177`).

```
Error: Error: GraphQL error: Check constraint violation. new row for relation "units_cris" violates check constraint "units_cris_unit_desc_id_check"
```

the problem is that we have forbidden this value from being used in the CRIS schema because it's a custom VZ lookup value! 🙃 

if we want to solve this at the db level, i think we could modify that check constraint so that `unit_desc_id`  = `177` is allowed if the user name is not `cris` or `system` or whatever. 🤷

atd-vze/src/views/Tools/CreateCrashRecord.js
johnclary: i think we should enforce that case ID, crash timestamp, and primary address are not null in the form. separate issue coming to make these constraints at the db level, too 👍

atd-vze/src/views/Tools/CreateCrashRecord.js
```diff
@@ -271,10 +219,8 @@ const CreateCrashRecord = ({ client }) => {
               variables: crashVariables,
             })
             .then(res => {
-              // Increment the temp ID so the user can resuse the form.
-              setTempId(tempId + 1);
               setSuccessfulNewRecordId(
-                res.data.insert_atd_txdot_crashes.returning[0].crash_id
+                res.data.insert_crashes_cris.returning[0].id
               );
               resetForm();
```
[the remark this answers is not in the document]
roseeichelmann: yeah i was thinking a similar thing, another option would be to just redirect to the top of the page so u can see the green banner that says crash record creation successful and links u to the details page?

atd-vze/src/views/Crashes/Crashes.js
roseeichelmann: i had to update this to `record_locator` otherwise the page breaks because we arent using `crash_id` in the `crashGridTableParameters` anymore. what this does is set the field `record_locator` from the crash grid table parameters to be the default field being searched in the search bar when the page loads. though the label in the search bar still says "Search by crash id"

atd-vze/src/views/Tools/CreateCrashRecord.js
roseeichelmann: fixed dis

atd-vze/src/views/Tools/CreateCrashRecord.js
roseeichelmann: moving them out of the component caused a bug when adding new units to the form and basically messed up the units state array so i think these can stay here?

roseeichelmann:
@johnclary thanks for your super helpful feedback, i think i addressed all of it except for the re-routing thing which im still happy to do if you think thats better than the green banner at the top that we currently have

atd-toolbox/data_model/python/40_make_lookup_seeds.py
```diff
@@ -27,7 +27,7 @@ def escape_single_quotes(input_string):
 
 unit_desc_lkp_custom_values = """alter table lookups.unit_desc_lkp add constraint unit_desc_owner_check check ((id < 177 and source = 'cris') or (id >= 177 and source = 'vz'));
 insert into lookups.unit_desc_lkp (id, label, source) values (177, 'MICROMOBILITY DEVICE', 'vz');
-alter table public.units_cris add constraint units_cris_unit_desc_id_check check (unit_desc_id < 177);
+alter table public.units_cris add constraint units_cris_unit_desc_id_check check ((unit_desc_id < 177) or (created_by <> 'cris' and created_by <> 'system'));
```
roseeichelmann, line 30: allows users to add micromobility as a unit type for temp record creations

atd-vze/src/views/Crashes/PeopleDetailsCard.js
```diff
@@ -61,6 +61,7 @@ const PeopleDetailsCard = ({ data, refetch, ...props }) => {
             fieldConfig={primaryPersonDataMap[0]}
             data={primaryPersonData}
             sortField={"unit_nbr"}
+            secondSortField={"id"}
```
roseeichelmann, line 64: fixes a bug where if the peoples card had multiple people in it, every time you edited a person the order would change which was really jarring bc we were only sorting on `unit_nbr`. now we sort first by `unit_nbr` and then person id

atd-vze/src/views/Crashes/RelatedRecordsTable.js
```diff
@@ -120,7 +121,11 @@ const RelatedRecordsTable = ({
         </thead>
         <tbody>
           {data
-            .sort((a, b) => (a[sortField] > b[sortField] ? 1 : -1))
+            .sort(
+              (a, b) =>
+                a[sortField] - b[sortField] ||
+                a[secondSortField] - b[secondSortField]
```
roseeichelmann, line 127: sorting magic

atd-vze/src/views/Tools/CreateCrashRecord.js
```diff
@@ -93,295 +139,237 @@ const CreateCrashRecord = ({ client }) => {
...
+      .mutate({
+        mutation: INSERT_BULK,
+        variables: {
+          crash_data: {
+            rpt_street_name: primaryStreetName?.toUpperCase(),
+            rpt_sec_street_name: secondaryStreetName?.toUpperCase(),
             case_id: caseId,
-            crash_timestamp: crashTimestamp,
+            rpt_city_id: 22,
+            crash_timestamp: new Date(crashTimestamp).toISOString(),
```
johnclary, line 185: i tested this in Chrom and Safari and it works as expected 👍 🙏

johnclary (changes requested):
@roseeichelmann your changes look excellent—thank you! would you mind moving that new function out of the component, and yes to forcing a page direct on submit, please!

atd-vze/src/views/Tools/CreateCrashRecord.js
roseeichelmann: firefox works too :partying_face:

johnclary:
after discussion with Rose, i moved the alert banners to the bottom of the form 👍

johnclary (approved):
🚢 🚢 🚢 🚢

atd-toolbox/data_model/python/40_make_lookup_seeds.py
roseeichelmann: yeaahhhh :sweat_smile:

atd-vze/src/views/Tools/CreateCrashRecordTable.js
```diff
@@ -190,7 +194,10 @@ const CreateCrashRecordTable = () => {
         <ModalHeader toggle={toggleModalDelete}>
           Delete this record?
         </ModalHeader>
-        <ModalBody>Are you sure you want to delete crash id <strong>{deleteId}</strong>? This cannot be undone.</ModalBody>
+        <ModalBody>
+          Are you sure you want to delete crash id <strong>{deleteId}</strong>?
+          This cannot be undone.
+        </ModalBody>
```
chiaberry, line 200: should this be updated since its a soft delete now?

chiaberry:
I'm running into issues trying to delete a temp record

```
{
    "errors": [
        {
            "message": "invalid input syntax for type integer: \"T3\"",
            "extensions": {
                "path": "$",
                "code": "data-exception"
            }
        }
    ]
}
```

Also the "updated at" in this table doesn't seem to reflect when the record was updated, but rather when it was created. 
Screenshot 2024-07-23 at 10 21 06 AM

I was trying different ways to break the form and I can input negative fatalities and serious injuries (you can in vze prod now too, well at least in staging) but the good news is in this new data model version, the fatalities cards at the top wont show negative numbers while in staging....
Screenshot 2024-07-23 at 10 23 19 AM

atd-vze/src/views/Tools/CreateCrashRecordTable.js
```diff
@@ -138,38 +144,36 @@ const CreateCrashRecordTable = () => {
...
                         <td>{item.updated_by}</td>
-                        <td>{item.last_update}</td>
+                        <td>{item.updated_at}</td>
                         <td>
                           <Button
                             color="danger"
                             className="btn-pill"
                             size={"sm"}
-                            onClick={() => openModalDelete(item.crash_id)}
+                            onClick={() => openModalDelete(item.record_locator)}
```
chiaberry, line 176: I tracked down the error im getting when I try to delete here, its sending the record locator "T32449" to the delete mutation on line 36 that is expecting an Int!

chiaberry (approved):
Rose let me know that deletes will be handled in a later issue/PR, 🚢

johnclary:
i am merging this into the collector so that i can start resolving conflicts with this and the cris import family tree

PR #1488, opened 2024-07-16
Fixes miscellaneous loose ends

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/18274
- https://github.com/cityofaustin/atd-data-tech/issues/17686

### Summary of changes

- Clean up column metadata table
  - remove duplicate columns, which has no impact on our actual DB schema and was an artifact of the old google sheet that served as the source data for this table
  - add postgre comments about this table
- Rename `law_enforcement_fatality_num` to `law_enforcement_ytd_fatality_num` (for clarity)
- Charges
  - Drop the `charge_cat_id` column and `charge_cat_lkp` table. this column was deprecated after schema year 2010.
  - Exclude `NO CHARGES` records from the db. Per Xavier this is ok and it's cutting back on hundreds of thousands of records
- Update our crash location trigger to assign L5 polygons as needed. This replicates behavior in the old `update_cr3_location` function
- Update the VZE crash list to search on the timezone-adjusted `crash_date_ct` column, thereby addressing a timezone bug in our crash search
- Add a calculated field called `est_total_person_comp_cost` which is the sum of all person comp costs

### Testing

1. Start your local stack VZ stack and load some records with the CRIS import ETL.  [This PR](https://github.com/cityofaustin/atd-vz-data/pull/1478) has detailed instructions to run the CRIS import, and here's a quick docker command:

```shell
# from the ./atd-etl/data_model directory
$ docker run -it --rm --env-file .env --network host -v $PWD:/app data_model_cris_import ./cris_import.py --csv --s3-download
```

2. In the VZE, make sure the crash list loads normally. Test out the date filter and sorting. 

3. Go to the crash details page—it should normally. The charges table should look normal (hopefully you can find a crash that does have charges to verify this)

4. Go to the locations list and details and make sure that loads normally.

5. You can query the DB to verify that we have crashes associated with locations of both group 1 and 2:

```sql
SELECT
    crashes.id,
    loc.location_group
FROM
    crashes
    LEFT JOIN atd_txdot_locations loc ON crashes.location_id = loc.location_id
LIMIT 1000;
```

I think that's it 👍


atd-vzd/migrations/default/1715960023005_socrata_export_views/up.sql
```diff
@@ -92,7 +92,7 @@ where
 create or replace view socrata_export_people_view as (
     select
         people.id as id,
-        people.id as person_id, --legacy support
+        people.id as person_id, --this column exists for backwards compatibility
```
roseeichelmann, line 95: wait, why do we have two person id columns?

atd-vze/src/views/Crashes/crashGridTableParameters.js
```diff
@@ -58,13 +58,13 @@ export const crashGridTableColumns = {
...
-//     label_table: "Unit Description",
-//     type: "String",
-//     hidden: true,
-//   },
+  "units { unit_desc_lkp { label } }": {
+    searchable: false,
+    sortable: false,
+    label_table: "Unit Description",
+    type: "String",
+    hidden: true,
```
roseeichelmann, line 66: why is this hidden?

atd-toolbox/data_model/python/sql_templates/crashes_list_view.sql
```diff
@@ -45,7 +45,7 @@ create or replace view person_injury_metrics_view as (
                         people.prsn_injry_sev_id = 4
                         or people.prsn_injry_sev_id = 99
                     )
-                    and crashes.law_enforcement_fatality_num is not null
+                    and crashes.law_enforcement_ytd_fatality_num is not null
```
roseeichelmann, line 48: glad we are updating the name of this column for clarity :pray:

roseeichelmann (approved):
these changes look great, thanks for keeping track of all the little details we've been wanting to come back around to over time. test steps worked but i noticed the location details page notes arent working, i dont think that was introduced with this PR though so :ship:

atd-vzd/migrations/default/1715960023005_socrata_export_views/up.sql
johnclary: only because i feel bad "breaking" the public dataset if folks have been using the `person_id` column. maybe that's not worth it though 🤷

atd-vze/src/views/Crashes/crashGridTableParameters.js
johnclary: it is hidden because we don't render it in directly in the UI, but it it is used for the charts on the location details page. i had i had forgotten that i commented it out completely and it in fact did break the location details page, we just didn't notice 🙃

johnclary:
@roseeichelmann thanks for flagging the issue with the location notes. i apparently noticed that too and fixed it in https://github.com/cityofaustin/atd-vz-data/pull/1489 👍

atd-vzd/migrations/default/1715960023005_socrata_export_views/up.sql
roseeichelmann: ahhh i see. i dont have a heavy opinion on this :shrug:

PR #1490, opened 2024-07-18
 DB changes to support the fatalities UI

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/17711

## Testing
**URL to test:** 
local

**Steps to test:**
1. Follow steps in https://github.com/cityofaustin/atd-vz-data/pull/1478 to start up the data model and run the cris import.
2. Start your VZE, go to the Fatalities page.
3. Go to the crash details page of the crashes on the fatalities page, edit the people in the crash to add more fatal injuries then go back to the fatalities page to see that reflected. Test adding FRB status and recommendations to the crash, see that in the Fatalities page table. FYI weird bug with recommendations where you have to add a status or partner first before you can add a rec or update, looks like its a preexisting bug but just noticed it with this PR.
4. On the Fatalities page test sorting columns and search by victim name, crash id, etc. _( Also annoying bug with the search where if you type in non ints to an int search field like crash id it bombs, and then because of the way we store search fields in local storage you cant reset the field without clearing your cookies :melting_face: )_
6. Test exporting the rows to .csv using the download button

---
#### Ship list
- [ ] Code reviewed


atd-toolbox/data_model/python/sql_templates/fatalities_view.sql
```diff
@@ -0,0 +1,48 @@
...
+    AS ytd_fatality,
+-- get ytd fatal crash, partition by year and sort by date timestamp.
+-- records with the same crash_id will get "tie" rankings thus making
+-- this column count each crash rather than each fatality
+DENSE_RANK() OVER (
+    PARTITION BY EXTRACT(year FROM crashes.crash_timestamp) 
+    ORDER BY crashes.crash_timestamp ASC, crashes.crash_id) 
+    AS ytd_fatal_crash,
+crashes.case_id,
+crashes.law_enforcement_fatality_num,
```
roseeichelmann, line 41: still think `law_enforcement_fatality_num` is a kinda confusing name for the column bc its a ytd fatal crash number not fatality i.e its counting crashes not persons. but the vz team knows what it means so :shrug: 

`law_enforcement_fatal_crash_num`?

atd-toolbox/data_model/python/sql_templates/fatalities_view.sql
```diff
@@ -0,0 +1,48 @@
...
+  crashes.rpt_street_pfx,
+  crashes.rpt_street_name,
+  '(',
+  crashes.rpt_sec_block_num,
+  crashes.rpt_sec_street_pfx,
+  crashes.rpt_sec_street_name,
+  ')') AS location,
+to_char(
+  public.crashes.crash_timestamp at time zone 'US/Central', 'YYYY-MM-DD'
+  ) as crash_date_ct,
```
roseeichelmann, line 24: maybe instead we should just use the crash_timestamp and have the whole crash timestamp be in the Crash Date column like we do on the Crashes page 
image

atd-toolbox/data_model/python/sql_templates/fatalities_view.sql
johnclary: i fixed this in https://github.com/cityofaustin/atd-vz-data/pull/1488!

johnclary: don't sweat it here—i'll resolve conflicts. there are a decent number at this point

atd-toolbox/data_model/python/90_fatalities_view.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def load_sql_template(name):
+    with open(name, "r") as fin:
+        return fin.read()
+
+
+def main():
+    indexes_sql = load_sql_template("sql_templates/fatalities_view.sql")
+    migration_path = make_migration_dir("indexes")
```
johnclary, line 12: looks great! would you rename this from `indexes` to `fatalities_view` ?

roseeichelmann: oh yeah,  fixed this!

atd-toolbox/data_model/python/sql_templates/fatalities_view.sql
```diff
@@ -0,0 +1,47 @@
+drop table fatalities cascade;
```
chiaberry, line 1: I was in a tangled mess of migrations and so I downloaded a fresh copy of prod and did the up migrations, that worked fine, but when I went back down and tried to apply migrations again, I got an error here

`"message": "table \"fatalities\" does not exist",`

so I changed this to `drop table if exists fatalities cascade;`

i probably messed something up in my migrations snafu, but thought id mention it anyway.

chiaberry (approved):
Thanks for pointing out the bug where adding recommendation text without coordination partners silently fails ({"message": "unexpected variables in variableValues: rec_text"}) should that/could that be fixed in VZ now or should it wait til  data model delivery?

besides that, test steps worked out!

atd-toolbox/data_model/python/sql_templates/fatalities_view.sql
johnclary: @roseeichelmann would you mind making that change to the migration like Chia did? would be nice to handle this edge case

atd-vzd/migrations/default/1715960026005_fatalities_view/up.sql
```diff
@@ -0,0 +1,48 @@
...
+    ORDER BY crashes.crash_timestamp ASC, crashes.crash_id) 
+    AS ytd_fatal_crash,
+crashes.case_id,
+crashes.law_enforcement_fatality_num,
+crashes.engineering_area
+  from 
+    people 
+  left join units on people.unit_id = units.id
+  left join crashes on units.crash_id = crashes.id
+ where crashes.in_austin_full_purpose = true AND people.prsn_injry_sev_id = 4;
```
johnclary, line 47: @roseeichelmann like you mentioned [here](https://github.com/cityofaustin/atd-vz-data/pull/1485#discussion_r1691877263), yes definitely you should exclude crashes where `private_drive_fl = true`

atd-vze/src/views/Fatalities/fatalityGridTableParameters.js
```diff
@@ -38,7 +38,7 @@ export const fatalityGridTableColumns = {
     label_table: "YTD Fatalities",
     type: "Int",
   },
-  crash_date: {
+  crash_date_ct: {
```
johnclary, line 41: 👍

atd-toolbox/data_model/python/create_migrations.sh
```diff
@@ -17,3 +17,4 @@ set -e
 ./60_make_views.py
 ./70_make_column_metadata_table.py
 ./80_update_notes_and_recs.py
+./90_fatalities_view.py
```
johnclary, line 20: ✨ 🙏

johnclary (changes requested):
@roseeichelmann this all looks fantastic! 🙌 

One minor thing, should we remove the old triggers and functions that handled the inserts into the fatalities table? looks like we have `fatality_insert` and `update_fatality_soft_delete` functions, and if you delete those with `cascade` it should delete the triggers as well.

johnclary (approved):
looks great!!!! 🙌  🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

PR #1493, opened 2024-07-25
Make the Dashboard page compatible with the new data model

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/18328

The issue suggests using the crashes list view but the way this currently works is we are supposed to reflect the same values in this VZE that are in the socrata dataset, so i thought it made the most sense to use the `socrata_export_crashes_view`

## Testing
**URL to test:** 
Local

**Steps to test:**
1. Check out this branch, start up your data model db and apply hasura migrations/metadata and run the cris import script if necessary
2. Open up the hasura console and navigate to the `socrata_export_crashes_view`, sort by death count, years of life lost, and suspected serious injuries to see what we would expect the widgets on the dashboard to say. Keep in mind this is only for crashes that happened in 2024.
3. Start your VZE, go to the dashboard, and confirm you are seeing the numbers you would expect to see in the widgets.
4. Now go create some temp crash records with fatalities/serious injuries to further test the dashboard widget. Remember that the socrata dataset is only counting crashes that happened at least 14 days ago. So in your crash creation the date needs to be todays date - 14 days or farther back.
5. Also, you will need to update the private drive flag to `NO` in the crash details page of the temp record in order for it to factor into the socrata ytd totals. **_Sidenote - lets not let this requirement slip through the cracks so the VZ team isnt wondering why their temp crashes arent making it into the socrata totals_**
6. Update your computer to a different time zone. You should still see the same numbers in your widgets bc everything should be using UTC time and not expecting the user to be in CT.

---
#### Ship list
- [ ] Code reviewed


atd-vze/src/queries/dashboard.js
```diff
@@ -1,92 +1,22 @@
...
-          { crash_time: { _is_null: false } }
-        ]
-        _or: [
-          { in_austin_full_purpose: { _eq: true } }
-          { _and: [{ city_id: { _eq: 22 } }, { position: { _is_null: true } }] }
-        ]
-      }
+  query GetCrashesYTD($yearStart: String, $yearEnd: String) {
+    socrata_export_crashes_view_aggregate(
+      where: { crash_timestamp: { _lt: $yearEnd, _gte: $yearStart } }
```
roseeichelmann, line 8: i think i should update this to less than or equal to here? so if a crash was on `2024-01-01T00:00:00` it would get counted for this year?

atd-vze/src/views/Tools/CreateCrashRecord.js
```diff
@@ -347,13 +347,13 @@ const CreateCrashRecord = ({ client }) => {
                     fade={false}
                   >
                     {/*eslint-disable-next-line*/}
-                    <i className="fa fa-check-circle" /> Successfsully
-                    created crash{" "}
+                    <i className="fa fa-check-circle" /> Successfully created
```
roseeichelmann, line 350: lil typo :smile:

atd-vze/src/views/Tools/CreateCrashRecordTable.js
```diff
@@ -23,7 +23,10 @@ import {
 
 const GET_TEMP_RECORDS = gql`
   query getTempRecords {
-    crashes(where: { is_temp_record: { _eq: true } }) {
+    crashes(
+      where: { is_temp_record: { _eq: true } }
+      order_by: { record_locator: asc }
```
roseeichelmann, line 28: was annoying me that the temp records table wasnt being sorted so it was hard to find ur last record creation

atd-vze/src/views/VZDashboard/VZDashboard.js
```diff
@@ -5,41 +5,31 @@ import { useQuery } from "@apollo/react-hooks";
...
 
 function VZDashboard() {
   const year = new Date().getFullYear();
-  const yearStart = `${year}-01-01`;
+  const yearStart = `${year}-01-01T00:00:00`;
   // We use the same end date as VZV so VZE widget totals match VZV widgets
-  const yearEnd = format(subDays(new Date(), 14), "yyyy-MM-dd");
+  const yearEnd = subDays(new Date(), 14)
+    .toISOString()
+    .substring(0, 19); // to match the ISO string format in the socrata view crash_timestamp column
```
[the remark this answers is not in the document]
johnclary: i think you can do away `yearEnd` completely, since we have the 14-day cutoff baked into the socrata view 🤔

atd-vze/src/views/VZDashboard/VZDashboard.js
```diff
@@ -5,41 +5,31 @@ import { useQuery } from "@apollo/react-hooks";
...
+import { subDays } from "date-fns";
 
 import { GET_CRASHES_YTD } from "../../queries/dashboard";
 
 import bi_logo from "../../assets/img/brand/power_bi_icon_white_on_transparent.png";
 
 function VZDashboard() {
   const year = new Date().getFullYear();
-  const yearStart = `${year}-01-01`;
+  const yearStart = `${year}-01-01T00:00:00`;
```
johnclary, line 16: little TZ bug here—to get the `year` value you want to use `new Date().getUTCFullYear()` since you're constructing a UTC timestamp

atd-vze/src/views/Tools/CreateCrashRecordTable.js
johnclary: do you want to sort this by `desc` instead, so the most recent records are at the top?

also, i could have sworn you had configured the query to refetch after form submit so that the temp records table gets refreshed. that's not happening anymore. maybe i broke that when i patched your branch?? 🙈

johnclary (changes requested):
i ran through all the tests—it works great. love that you went with the socrata export view for this!

chiaberry:
fwiw I didnt see the temp records table refresh when i tested last week

atd-vze/src/views/VZDashboard/VZDashboard.js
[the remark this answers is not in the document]
roseeichelmann: hah you're so right wow what a silly oversight on my part :sweat_smile:

roseeichelmann:
@johnclary ready for re-review, i dont think there was ever refetch functionality but i just added that in!

atd-vze/src/views/Tools/CreateCrashRecordTable.js
```diff
@@ -50,20 +38,15 @@ const DELETE_TEMP_RECORD = gql`
...
   const [crashSearch, setCrashSearch] = useState("");
   const [modalOpen, setModalOpen] = useState(false);
   const [deleteId, setDeleteId] = useState(null);
   const [feedback, setFeedback] = useState(null);
-  const { loading, error, data } = useQuery(GET_TEMP_RECORDS, {
-    fetchPolicy: "no-cache",
-  });
   const [deleteRecord] = useMutation(DELETE_TEMP_RECORD);
 
-  useEffect(() => {
```
roseeichelmann, line 64: dont need this useEffect

atd-vze/src/views/Tools/CreateCrashRecordTable.js
```diff
@@ -73,26 +56,26 @@ const CreateCrashRecordTable = () => {
...
-        setCrashesData({ atd_txdot_crashes: newData });
-        toggleModalDelete();
-      })
-      .catch(err => {
-        setFeedback(String(err));
-        setDeleteId(null);
-      });
-  };
+  // /**
+  //  * Deletes all the temporary records in the database
```
roseeichelmann, line 60: just commenting this out till we handle temp deletes in a following PR

johnclary (approved):
niceeee—thanks for adding in the refetch!

🚢 🚢 🚢 🚢 🚢 🚢 🚢

PR #1496, opened 2024-08-01
Rename crash ID columns

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/18461

This PR makes two changes:
- Renames `crash_id`  to `cris_crash_id` on the `crashes` tables
- Renames `crash_id` to `crash_pk` on any table that references the `crashes.id` column

## Testing

1. The CRIS import ETL: reminder that this currently lives in `./atd-etl/data_model`
2. All of the VZE viewing and editing interactions: crashes, crash details, locations, location details, temp records
3. I'd prefer you don't test the Socrata export ETL because I'm using the staging datasets for debugging crash data.

---
#### Ship list
- [ ] Code reviewed


atd-toolbox/data_model/python/sql_templates/crash_notes.sql
johnclary, line 1: this was cruft—we were not using it

atd-vzd/metadata/databases/default/tables/public_charges_cris.yaml
```diff
@@ -2,17 +2,17 @@ table:
   name: charges_cris
   schema: public
 object_relationships:
-  - name: crashes_cri
```
mddilley, line 5: goodbye cri!!!! 🥲

mddilley (commented):
I read through the diff and the word "crash" officially looks and sounds weird to me now! 😄 Thanks for updating all of these - no small feat!!! The local stack came up for me, and I'll finish testing through the UI tomorrow.

mddilley (commented):
Looks great, John! I ran into a few issues but I think that they are probably spin-offs, and I can quickly turn this into a ✅

- Location view
    - ~I’m seeing a mismatch between the data in the Details card of the Location page and the CR3 crashes table at the bottom. In my import, the location 0C096FD90D shows 149 crashes but the table at the bottom lists 0 crashes with any combination of filters. I checked the DB views, and I see the same results.~ Ahhhh this is me forgetting again that the total in that card is CR3 + non-CR3 crashes, and I was only considering the CR3 count 🤦‍♂️
- Units and People accordion
    - Units: Type, Body style, Direction - NO DATA option doesn’t save. The mutation is passing `null` but is not saving as expected.
    - People: Sex and Race/Ethnicity - Same issue
- Small edge case export bug:
    - In the locations list, click the export icon and press **Cancel** while the spinner is still spinning. Click the export icon again quickly and 💥

johnclary:
thanks @mddilley for that thorough testing! i'll take a look these momentarily

johnclary:
thanks @mddilley—i am tracking a new todo to look further into that location export bug. i could not reproduce it in prod, so that's not a great sign :/

for the units and people accordions, we discussed that what you're seeing is the effect of the CRIS-provided values superseding the `null` edits values. we have decided to rename the `NO DATA` option to `RESET` via https://github.com/cityofaustin/atd-data-tech/issues/18575

mddilley (approved):
🚢 🚀 😎

roseeichelmann (approved):
Approving based on my review of https://github.com/cityofaustin/atd-vz-data/pull/1499 :))

PR #1502, opened 2024-08-07
Write SQL to backfill the crash_pk column for crash recommendations and notes

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/18504

## Testing
**URL to test:** 
Local data model

**Steps to test:**
1. Start up your local data model. To truly test this you need to import a decent chunk of crashes using the cris_import script. I used the 10 year extracts John linked in this [slack thread](https://austininnovation.slack.com/archives/CM9MK950S/p1723046093254839?thread_ts=1722874844.778289&cid=CM9MK950S) 
2. After importing all the crashes, run the SQL commands in this PR to backfill the `crash_pk` column in the crash notes and recommendations tables. If you get an error saying there are still null values in the `crash_pk` column of recommendations/crash_notes then thats because your import didnt contain all of the crashes with recommendations or notes so, you can kinda basically ignore that maybe remove those commands with the faith that John and I both tested with full crash imports and we will make sure this is right for launch :crossed_fingers: 
3. In your VZE test going to different crash records using their cris crash ids and seeing the corresponding notes and recommendations there on the page.

---
#### Ship list
- [ ] Code reviewed


roseeichelmann:
After we run these commands we may need to check back in and see if there are any remaining records where crash_pk is still null and see what we want to do about it? and we will need to add some constraints to make `crash_pk` a non-nullable column. lmk if i should address that in this PR @johnclary

atd-toolbox/data_model/python/post_migration_scripts/notes_recs_crash_pk_backfill.sql
```diff
@@ -0,0 +1,34 @@
...
+
+-- Drop old columns
+ALTER TABLE recommendations DROP COLUMN atd_txdot_crashes_crash_id;
+
+ALTER TABLE crash_notes DROP COLUMN atd_txdot_crashes_crash_id;
+
+-- Each record must always have an associated crash_pk
+ALTER TABLE recommendations ALTER COLUMN crash_pk SET NOT NULL;
+
+ALTER TABLE crash_notes ALTER COLUMN crash_pk SET NOT NULL;
```
johnclary, line 34: i think it would be best to execute these two `SET NOT NULLS` before we drop the `atd_txdot_crashes_crash_id` column, that way we know 100% of records have been assigned a new ID before we drop the old data.

johnclary (changes requested):
looks good! just one minor change in the execution order

johnclary (approved):
🚢 🚢 🚢 🚢 🚢

PR #1505, opened 2024-08-09
Misc loosed ends part `III` - The final chapter (?)

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/18581

This PR:
- Updates Socrata export ETL and datasets to rename `crash_id` to `cris_crash_id`, exclude deleted columns, and add `is_deleted` and `is_temp_record` columns. Just some loose ends, eh?
- Enforce `unit_nbr`, `crash_timestamp` not-null values in a few places. I scanned all of our tables, and I don't feel strongly about enforcing any other not nulls 🤷 
- Move audit fields into `_column_metadata` table. these columns were in a separate migration just as an artifact of how the data model code evolved over time 🧹 

## Testing

You guessed it—run some data into your local stack and get out your mag 🔍 

---
#### Ship list
- [ ] Code reviewed


atd-vzv/src/views/summary/CrashesByYear.js
```diff
@@ -113,7 +113,7 @@ const CrashesByYear = () => {
           <hr />
         </Col>
       </Row>
-      {avgData.length > 0 && currentYearData.length > 0 ? (
+      {avgData && currentYearData ? (
```
johnclary, line 116: allows the loading spinner to stop if we just don't have any data for the current year

atd-toolbox/data_model/python/column_metadata.json
```diff
@@ -1359,7 +1600,23 @@
...
+    "foreign_table_name": "",
+    "foreign_table_schema": "",
+    "is_cris_column": false,
+    "is_edits_column": false,
+    "is_existing_foreign_table": false,
+    "is_primary_key": false,
+    "is_unified_column": true,
+    "note": "",
+    "record_type": "crashes",
+    "source": "cris"
```
roseeichelmann, line 1618: what exactly does `source` mean here? i would think "db generated field" would be more fitting which i see as the source for some other generated columns

atd-toolbox/data_model/python/sql_templates/audit_fields.sql
```diff
@@ -1,42 +1,3 @@
...
-alter table public.crashes add column created_by text not null default 'system';
-alter table public.crashes add column updated_by text not null default 'system';
-alter table public.units add column created_at timestamptz not null default now();
-alter table public.units add column updated_at timestamptz not null default now();
-alter table public.units add column created_by text not null default 'system';
-alter table public.units add column updated_by text not null default 'system';
-alter table public.people add column created_at timestamptz not null default now();
-alter table public.people add column updated_at timestamptz not null default now();
-alter table public.people add column created_by text not null default 'system';
-alter table public.people add column updated_by text not null default 'system';
```
roseeichelmann, line 38: you are just deleting all this in favor of having them in the up migrations for these tables instead right?

atd-toolbox/data_model/python/sql_templates/socrata_export_views.sql
```diff
@@ -86,16 +88,20 @@ left join lateral (
     limit 1
 ) as unit_aggregates on true
 where
-    crashes.in_austin_full_purpose = true and crashes.private_dr_fl = false
+    crashes.is_deleted = false
```
roseeichelmann, line 91: :rocket:

atd-toolbox/data_model/python/10_make_core_tables.py
```diff
@@ -92,7 +92,8 @@ def make_column_sql(columns, full_table_name):
             fk_cascade = "on update cascade on delete cascade"
         elif constraint and "not null default false" in constraint and "_edits" in full_table_name:
             constraint = ""
-
+        elif constraint == "not null" and column_name in ['unit_nbr', 'crash_timestamp'] and "_edits" in full_table_name:
+            constraint = ""
```
roseeichelmann, line 96: can you refresh me on the purpose of this script/what this is doing? and is this line in particular making sure that `crash_timestamp` and `unit_nbr` can be null in the edits table?

roseeichelmann (approved):
These changes all look great to me. I tried to think through any other fields that should maybe have a not null constraint on them -- what about `is_primary_person` in the people table? Other than that i couldnt find any others. Yay for cleaning up loose ends :broom: :sparkles:

atd-toolbox/data_model/python/10_make_core_tables.py
johnclary: yes! this is the script that generates all of the `create table` migrations for crashes/units/people/charges. it does this by parsing out the `column_metadata.json` file. 

and just like you said, this `elif` block makes sure that those fields can be null in the edits table.

atd-toolbox/data_model/python/column_metadata.json
johnclary: thanks for catching this—i will fix this. this property gets translated to the `is_imported_from_cris` column in the `_column_metadata_table`, which in turn is used to generate the upsert mutations in the CRIS import ETL.

so we can set this to any value we want, except `cris`, and it will be properly ignored by the ETL.

it will be nice when we don't to detail with this JSON and the spaghetti migration-gen scripts anymore 😅

atd-toolbox/data_model/python/sql_templates/audit_fields.sql
johnclary: correct!

PR #1506, opened 2024-08-09
Use Reset instead of NO DATA option in dropdowns

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/18575

## Testing
**URL to test:** 
Local data model

**Steps to test:**
1. Go to a crash, edit some values in the Units/Peoples tables. Then edit them again and click the RESET option. This will reset these edit values to be NULL in the edits table, so CRIS values will propagate to the top again. 
2. We should create a spin off issue to really address how we want to handle this long term; functionally and stylistically. For example, maybe be able to hover over a tooltip or something and see what the CRIS value is if there is a VZ edit covering it. And then have a reset option only if there is a VZ edit. Also currently can only reset with drop down fields, cant reset an age for example rn.

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vze/src/views/Crashes/RelatedRecordsTable.js
+++ b/atd-vze/src/views/Crashes/RelatedRecordsTable.js
@@ -218,7 +218,7 @@ const RelatedRecordsTable = ({
                                     }
                                     type="select"
                                   >
-                                    <option value={""}>NO DATA</option>
+                                    <option value={""}>RESET</option>
                                     {lookupOptions[
                                       fieldConfig.fields[field].lookupOptions
                                     ].map(option => {


mddilley (approved):
I read the diff but did not run locally. Looks like it will work to me! ↩️ 🙏 🚢 🚀 

I fully agree with your step 2, and I like the ideas that you proposed! I hadn't thought about how there isn't a "reset" option for other editable fields that don't have dropdowns right now.

johnclary (approved):
thanks @roseeichelmann! i opened https://github.com/cityofaustin/atd-data-tech/issues/18607 to track the follow-up work 🙏

PR #1507, opened 2024-08-15
Remove page and record counts from grid table pagination controls

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/18622

## Testing
**URL to test:** 
Local data model

**Steps to test:**
1. Go to the Crashes page, see that the total records and total pages counts arent there anymore. Do you notice a faster loading time if you have a big cris import loaded in your database?
2. The next page button should be disabled if the current page total record count is less than the limit, ie. rows per page. Test this. 
3. This also means that if the last page happens to have exactly the number of records that is the limit, the user will be able to click the next page button and the next page will just have 0 records. There is really no way around this without querying aggregate total record counts, so this is a bug we will have to live with for the time being. Good news is its not super likely that the last page should have exactly the number of rows per page the user picks.
4. When the above happens, test that the next page button is disabled when the current page has 0 records.
5. Test the Locations list page as well.
6. Test that nothing broke on the Location details page, this also contains a gridTable.

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vze/src/Components/GridTable.js
+++ b/atd-vze/src/Components/GridTable.js
@@ -229,6 +229,7 @@ const GridTable = ({
   const handleRowClick = e => {
     const rowNumber = parseInt(e.target.value);
     setLimit(rowNumber);
+    changePage(1);
   };
 
   /**
--- a/atd-vze/src/Components/GridTable.js
+++ b/atd-vze/src/Components/GridTable.js
@@ -182,7 +182,9 @@ const GridTable = ({
     } else if (sortColumn === col) {
       // Else if the current sortColumn is the same as the new
       // then invert values and repeat sort on column
-      sortOrder === "desc_nulls_last" ? setSortOrder("asc") : setSortOrder("desc_nulls_last");
+      sortOrder === "desc_nulls_last"
+        ? setSortOrder("asc")
+        : setSortOrder("desc_nulls_last");
     } else if (sortColumn !== col) {
       // Sort different column after initial sort, then reset
       setSortOrder("desc_nulls_last");
@@ -377,7 +379,7 @@ const GridTable = ({
     query.setOrder(sortColumn, sortOrder);
   }
 
-  // Mange LIMIT & OFFSET
+  // Manage LIMIT & OFFSET
   query.limit = limit;
   query.offset = offset;
 
@@ -395,14 +397,9 @@ const GridTable = ({
   if (error) return `Error! ${error.message}`;
 
   let dataEntries = [];
-  let totalRecords = 0;
-  let totalPages = 1;
 
   // If we have data
   if (data[query.table]) {
-    totalRecords = data[query.table + "_aggregate"]["aggregate"]["count"];
-    totalPages = Math.ceil(totalRecords / limit);
-
     // DataEntries: For each item in the data array, generate a row with each column
     data[query.table].map((row, index) =>
       dataEntries.push(
@@ -526,15 +523,13 @@ const GridTable = ({
                     moveBack={moveBackPage}
                     pageNumber={page}
                     limit={limit}
-                    totalRecords={totalRecords}
-                    totalPages={totalPages}
                     handleRowClick={handleRowClick}
+                    recordsCount={dataEntries.length}
                   />
                   {columnsToExport && (
                     <GridExportData
                       query={query}
                       columnsToExport={columnsToExport}
-                      totalRecords={totalRecords}
                       roleSpecificColumns={roleSpecificColumns}
                       hasSpecificRole={hasSpecificRole}
                     />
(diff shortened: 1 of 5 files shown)


atd-vze/src/Components/GridTable.js
```diff
@@ -526,15 +523,13 @@ const GridTable = ({
                     moveBack={moveBackPage}
                     pageNumber={page}
                     limit={limit}
-                    totalRecords={totalRecords}
-                    totalPages={totalPages}
                     handleRowClick={handleRowClick}
+                    recordsCount={dataEntries.length}
```
roseeichelmann, line 527: this is the recordcount of the current page

atd-vze/src/queries/gqlAbstract.js
```diff
@@ -27,13 +27,6 @@ gqlAbstractTableName (
     gqlAbstractFilters
 ) {
     gqlAbstractColumns
-},
-gqlAbstractTableAggregateName (
-    gqlAbstractAggregateFilters
-) {
-    aggregate {
-      count
```
roseeichelmann, line 35: this is what was slowing down the query :wave:

atd-vze/src/views/Crashes/crashGridTableParameters.js
```diff
@@ -60,75 +60,74 @@ export const crashGridTableColumns = {
   },
 };
 
-
 export const locationCrashGridTableColumns = {
```
roseeichelmann, line 63: this is all linting with prettier

atd-vze/src/Components/GridTable.js
```diff
@@ -526,15 +523,13 @@ const GridTable = ({
                     moveBack={moveBackPage}
                     pageNumber={page}
                     limit={limit}
-                    totalRecords={totalRecords}
-                    totalPages={totalPages}
                     handleRowClick={handleRowClick}
```
mddilley, line 526: One small bug that i noticed which I think already exists is that you can page ahead, update the the row count, and then end up on a page that doesn't exist like if you go from 10 rows per page to 100 rows per page for example. 

Resetting the pagination could be a way to handle this, but I think a spinoff is good since it is reproducible in the current prod editor.

mddilley, line 526: One small bug that i noticed which I think already exists is that you can page ahead, update the the row count, and then end up on a page that doesn't exist like if you go from 10 rows per page to 100 rows per page for example. 

Resetting the pagination in this handler could be a way to handle this, but I think a spinoff is good since it is reproducible in the current prod editor.

atd-vze/src/Components/GridTable.js
```diff
@@ -377,7 +379,7 @@ const GridTable = ({
     query.setOrder(sortColumn, sortOrder);
   }
 
-  // Mange LIMIT & OFFSET
```
mddilley, line 380: MANGE 😅 🙏

atd-vze/src/queries/gqlAbstract.js
mddilley, line 1: for a moment, i wondered if we should keep the aggregate totals around in the abstract and make them optional, but I think that this is great. 🧹  Especially if we find ourselves moving away from gqlabstract in the future like in Moped. It would also be straightforward enough to paste it back in if needed.

mddilley (approved):
I tested the pagination, and it works so well! Thanks for sharing about it yesterday, and it will awesome to reap the benefits of a leaner query. I left one comment about an existing bug that can be reproduced in prod, but I think this is 🚢 🚀

atd-vze/src/Components/GridTable.js
mddilley: this is a duplicate of my above comment - not sure what happened 🙃

mddilley:
@johnclary Since I was checking the pagination queries, I noticed that I'm not seeing the Apollo caching behavior that i was expecting. The `cache-and-network` setting made me think that I should see the requests stop when paging through the same data without updating any filters. I can't remember if there is caching config that we might need to update. It also may have always worked this way! Either way, probably not something to worry about before the launch.

atd-vze/src/Components/GridTable.js
johnclary: great catch! i went ahead and patched that because it seems to have turned out to be a one-liner fix: https://github.com/cityofaustin/atd-vz-data/pull/1507/commits/6d31590c3ef3714b16318cecbc675ff1ae2955e6

atd-vze/src/queries/gqlAbstract.js
johnclary: yes, i agree. we can bring them back if we need to 🙏

johnclary (approved):
awesome—this works perfectly!

i am sad to see these details disappear from the UI; it's definitely helpful contextual information. but it really is a 2x reduction in query time as far as i can tell.

we will bring it back!

mddilley (approved):
Retested locally with the page reset, and it works great! 🚢 🚀

PR #1510, opened 2024-08-16
Remove Inconsistent KSI and changes views from the DB

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/18423

I wasn't sure if i should add these drop commands to the post migration scripts folder or if i should do all the other fancy stuff and add them to sql templates and create down/up migrations? @johnclary lmk what you think makes the most sense

## Testing
**URL to test:** 
Local data model

**Steps to test:**
1. Run the commands to drop the views, thats it!

---
#### Ship list
- [ ] Code reviewed


mddilley (commented):
I ran these commands, and they work but the Hasura metadata also needs cleanup or else it ends up being inconsistent with the DB schema. That might be part of a different task, but I wanted to check!

atd-toolbox/data_model/python/sql_templates/recommendations_and_notes_up.sql
johnclary, line 1: we now have to sets of migrations to split out from this file where i have crammed them. i will do that in a follow up very soon

johnclary (approved):
@mddilley thanks for flagging the metadata—i fixed that up ✅

mddilley (approved):
Spun this up locally, and I no longer see the two views or the metadata associated with them. Thanks! 🚢 🚀

PR #1511, opened 2024-08-19
Clean up ETL directory and Docker image CI

johnclary:
## Associated issues

- https://github.com/cityofaustin/atd-data-tech/issues/18438

## Testing

1. Before you pull, you probably want to make a copy of your env file at `atd-etl/data_model/.env`
2. After you pull, delete the old `env` filed fro `./atd-etl/cris_import` and save your `.env` file there
3. Start your local stack
4. Run the cris import from `./atd-etl/cris_import`:

```shell
 docker run -it  --rm --env-file .env --network host -v $PWD:/app data_model_cris_import ./cris_import.py --csv  --pdf --s3-download --verbose
```

4. Look at the github action changes and see if those look ok 🔍


atd-etl/cris_import/.gitignore
```diff
@@ -1 +1,3 @@
-env
```
mddilley, line 1: We'll have to remember to delete our existing `env` files since only `.env` is ignored at the top level of the project. I just did mine because I'm paranoid that I'll forget and accidentally leak my 1PW token.

mddilley (approved):
I ran the export locally, and it succeeded. I read through the file updates and deletions. Nothing sticks out to me, and I think the GH action diff looks good too. 🚢 💨

atd-etl/cris_import/.gitignore
johnclary: ah—thank you. i had the luxury of deleting the old directory and renaming the new directory, so didn't encounter this. i just added `env` to the root `.gitignore` and i will add note to the test steps to remove that. will also drop a note in slack post-go-live

PR #1515, opened 2024-08-21
Fixes WSOD on location details page for non-AFP location polygons

johnclary:
- https://github.com/cityofaustin/atd-data-tech/issues/18715

This change fixes a bug that bombs the location details page when loading locations which are outside of the Austin Full Purpose jurisdiction or are so-called "Level 5" polygons.

The reason this bug came about is because  we consolidated the various views that provide location-level data into one which excluded non-afp and L5 polygons.  To fix this without creating yet another view, I went ahead and added an advanced filter config that is similar to the one we use on the crashes page, which is a kind of hack/workaround for being able to set default filters within the gridtable / gqlabstract magic.

I think this gets us close enough to the current/previous production state 👍 

By the way, I did test list view performance in staging with this modified view and it has no impact on query time.

## Testing
 Local

**Steps to test:**
1. Start your stack, apply migrations and run some CRIS import. 
2. Use your SQL client to query an `atd_txdot_locations` record that has no `council_district`. Take note of the location ID can try searching for it, and opening it, in the VZE.
3. I'm testing with location ID `24A1E99089` which is broken in staging.

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vzd/migrations/default/1724206665257_locations_list_afp/up.sql
+++ b/atd-vzd/migrations/default/1724206665257_locations_list_afp/up.sql
@@ -1,0 +1,60 @@
+drop view if exists locations_list_view;
+create view locations_list_view as
+with cr3_comp_costs as (
+    select
+        location_id,
+        sum(est_comp_cost_crash_based) as cr3_comp_costs_total
+    from crashes_list_view group by location_id
+),
+
+cr3_crash_counts as (
+    select
+        location_id,
+        count(location_id) as crash_count
+    from
+        crashes
+    where
+        crashes.private_dr_fl = false
+        and crashes.location_id is not null
+        and crashes.crash_timestamp > (now() - '5 years'::interval)
+    group by
+        location_id
+),
+
+non_cr3_crash_counts as (
+    select
+        location_id,
+        count(location_id) as crash_count,
+        count(location_id) * 10000 as noncr3_comp_costs_total
+    from
+        atd_apd_blueform
+    where
+        atd_apd_blueform.location_id is not null
+        and atd_apd_blueform.date > (now() - '5 years'::interval)
+    group by
+        location_id
+)
+
+select
+    locations.location_id,
+    locations.description,
+    locations.council_district,
+    locations.location_group,
+    coalesce(
+        cr3_comp_costs.cr3_comp_costs_total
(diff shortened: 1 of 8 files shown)


atd-vze/src/Components/DataTable.js
```diff
@@ -64,7 +64,7 @@ const DataTable = ({
                       let fieldValue = fieldConfigObject.relationshipName
                         ? data[fieldDataTable][
                             fieldConfigObject.relationshipName
-                          ][field]
+                          ]?.[field]
```
johnclary, line 67: this prevents the WSOD but hides the underling issue

atd-vze/src/views/Locations/locationGridTableParameters.js
```diff
@@ -0,0 +1,32 @@
+export const locationsGridTableAdvancedFilters = {
```
johnclary, line 1: copied from the crashes grid table filters. TBH i don't expect anyone to use these, but the "invert_toggle_state" setting has the affect of causing these filters to be applied by default, which is what we want.

atd-vzd/migrations/default/1724206665257_locations_list_afp/up.sql
johnclary, line 1: i will miss those python scripts that generated these automatically 😓

atd-vzd/migrations/default/1724259574465_crash_position_index_gist/up.sql
johnclary, line 1: squeezing this little change in here. i accidentally had a `btree` configured on this geometry column, which does absolutely nothing

atd-vze/src/views/Locations/locationGridTableParameters.js
mddilley: this makes sense to me! retain the ability to nav to any location in the database but also feature those of most interest in the list view 👍

mddilley (approved):
I tested the migrations and was able to replicate with location id `24A1E99089` in staging and locally. I can now navigate to that location page and also see it in the location list view if I include outside AFP from the advanced filters. 🚢 🚀

chiaberry (approved):
bye bye wsod

johnclary:
thans, y'all!!!!

PR #1518, opened 2024-08-21
Better error handling when there are no crashes to process

johnclary:
## Associated issues

- https://github.com/cityofaustin/atd-data-tech/issues/18719

This issue cropped up when processing ~30 daily CRIS extracts. It seems that occasionally, but not always, the `crashReports` directory is completely missing from the extracts with no crashes. The CSV files are always present.

This script makes the following changes:
- Skips PDF processing entirely when (1) the script is invoked with both `--csv` and `--pdf` and (2) there were no crashes found in the CSV
- Adds a final test at the end of processing to make sure that the number of PDFs processed matches the number of crashes processed via CSV (again, only when the script is invoked with both `--csv` and `--pdf`)
- Raises an error if no extracts are retrieved from the S3 bucket. This was an earlier oversight.

## Testing

I've left ~30 extracts in the dev inbox which you can use for testing.

1. Start your local stack
2. Run the cris import with the `--s3-download` command:

```shell
$ ./cris_import.py --csv --pdf --s3-download
```

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-etl/cris_import/cris_import.py
+++ b/atd-etl/cris_import/cris_import.py
@@ -25,8 +25,9 @@ def main(cli_args):
 
     logger.info(f"{len(extracts_todo)} extract(s) to process")
 
-    if not extracts_todo:
-        return
+    if cli_args.s3_download and not extracts_todo:
+        # always short circuit if we find nothing in S3
+        raise Exception("No extracts found in S3 bucket")
 
     for extract in extracts_todo:
         records_processed = {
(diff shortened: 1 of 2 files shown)


atd-etl/cris_import/cris_import.py
```diff
@@ -25,8 +25,9 @@ def main(cli_args):
 
     logger.info(f"{len(extracts_todo)} extract(s) to process")
 
-    if not extracts_todo:
-        return
+    if cli_args.s3_download and not extracts_todo:
+        # always short circuit if we find nothing in S3
+        raise Exception("No extracts found in S3 bucket")
```
johnclary, line 30: Eeek—I thought we already had a check like this in place but it must have dropped off during one of the later refactors

atd-etl/cris_import/utils/process_pdfs.py
```diff
@@ -139,10 +139,8 @@ def process_pdfs(extract_dir, s3_upload, max_workers):
         for filename in os.listdir(os.path.join(extract_dir, "crashReports"))
         if filename.endswith(".pdf")
     ]
-    pdf_count = len(pdfs)
 
-    if not pdf_count:
-        raise IOError("No PDFs found in extract")
```
johnclary, line 145: we don't need this check anymore since we're handling this earlier in the script

atd-etl/cris_import/cris_import.py
chiaberry: I also could have sworn we were raising an exception already

chiaberry (approved):
I tested, but didnt test it when there were no crashes to process.

frankhereford (approved):
Pull up them anchors and 🚢🚢🚢🚢

johnclary:
Thank you for your reviews!

PR #1519, opened 2024-08-21
Add directory existence check to CRIS import script

frankhereford:
# Issue

While standing up the VZ data model locally, I found that there are a heterogeneous set of CRIS extracts in the dev inbox; some have PDFs in them, some don't, it appears. 

This small diff adds a check to `utils.process_pdfs()` to return early if there is not a directory of PDFs to process for a given zip.

# Associated issues

_nada_

# Testing

Import the CRIS extracts currently in the dev inbox.

---
#### Ship list
- [ ] Code reviewed


frankhereford:
@johnclary already had a PR cooking with this fix and others: https://github.com/cityofaustin/atd-vz-data/pull/1518. Thanks!

PR #1522, opened 2024-08-22
Drop all legacy tables, views, functions and metadata

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/18752
- https://github.com/cityofaustin/atd-data-tech/issues/13950

## Testing

Start your local stack and run the migrations. Then start the VZE and make sure it works.

In practice, I think we may want to run these migrations by hand—I am worried about the Hasura CLI or the github action runner timing out.

---
#### Ship list
- [ ] Code reviewed


chiaberry (approved):
ran migrations, tested vze by checking different pages, added a temp record, searching etc.

atd-vzd/migrations/default/1724353502869_remove_legacy_objects/up.sql
```diff
@@ -0,0 +1,189 @@
...
+drop table if exists polygons;
+drop view if exists view_location_crashes_by_manner_collision;
+drop view if exists view_location_injry_count_cost_summary;
+drop view if exists view_vzv_by_mode;
+drop view if exists view_vzv_by_month_year;
+drop view if exists view_vzv_by_time_of_day;
+drop view if exists view_vzv_demographics_age_sex_eth;
+drop view if exists view_vzv_header_totals;
+drop table if exists cr3_mainlanes;
+drop table if exists non_cr3_mainlanes;
```
mddilley, line 151: @johnclary i was skimming through these and this one caught my eye. If the way we calculate non-CR3 locations hasn't change, then I think that we need to keep this table since it is referenced in `update_noncr3_location()` which is used in the trigger on insert or update of records in the `atd_apd_blueform` table.

atd-vzd/migrations/default/1724353502869_remove_legacy_objects/up.sql
mddilley, line 1: I ran into a migration error since my snapshot had the `1724521377803_fix_loc_comp_cost_date_filter`. I was able to fix by pulling the latest from `master` branch with no issues but wanted to note in case someone gets confused.

mddilley (commented):
I ran into one table that I think we need to keep around. I also ran into an issue with the migrations (might be the new I stopped for now but can pick back up on this one later today. The public schema looks much tidier! ✨

atd-vzd/migrations/default/1724353502869_remove_legacy_objects/up.sql
johnclary: such a fantastic catch—thank you. i guess i really need to test the afd + ems in addition to the blueform import. i will do that and strip out this table deletion 🙏

johnclary: thanks again @mddilley — we will not delete `non_cr3_mainlanes` ✅  i went through all of the trigger functions in the DB again to double check i didn't miss anything else.

johnclary:
I'm running this against staging now.

atd-vzd/migrations/default/1724353502869_remove_legacy_objects/up.sql
johnclary: thanks for flagging this—i bumped the migration timestamps

PR #1532, opened 2024-08-27
Move victim name to the right of people table

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/18816

## Testing
**URL to test:** 
[Netlify](https://deploy-preview-1532--atd-vze-staging.netlify.app/)

**Steps to test:**
1. For people with fatal injuries, see that the victim name column is now to the far right of the table. Check this for primary and non primary people.

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vze/src/views/Crashes/VictimNameField.js
+++ b/atd-vze/src/views/Crashes/VictimNameField.js
@@ -1,5 +1,5 @@
 import React from "react";
-import { Button, Input, Form, Row, Col, Label } from "reactstrap";
+import { Button, Input, Row, Col, Label } from "reactstrap";
 
 // This component should be role-restricted to users with Admin or IT Supervisor permissions
 // Role check currently happens in the shouldRenderVictimName function of personDataMap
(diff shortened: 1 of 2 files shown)


johnclary (changes requested):
this looks good, but there is a little horizontal scrollbar that seems to be visible at every page width. can you squash whatever is tripping that?

Screenshot 2024-08-28 at 12 04 33 PM

atd-vze/src/views/Crashes/VictimNameField.js
```diff
@@ -40,7 +40,7 @@ const VictimNameField = ({
   return (
     <td>
       {isEditing && (
-        <Form>
```
johnclary, line 43: can we keep the form element here, too? just as a best practice.

johnclary (approved):
horizontal scrollbar is gone! would you mind adding the `<form>` back in? then ship it! 🚢 🚢 🚢 🚢

chiaberry (approved):
✅

PR #1540, opened 2024-09-03
Implement soft deletes for crash notes

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/18788

Users cant delete recommendations only edit, so we dont need to implement soft-deletes for them.

## Testing
**URL to test:** 
Local

**Steps to test:**
1. Check out this branch, start your local db, apply the migrations and metadata.
2. Go to a crash, add a note and delete your note. Make sure you can still find the note in the `crash_notes` table but the is_deleted column will say true.
3. Run the down migrations

---
#### Ship list
- [ ] Code reviewed


chiaberry (approved):
deleted my note, found it in the db table and tested the down migration

✅

atd-vze/src/Components/Notes/Notes.js
```diff
@@ -30,7 +30,7 @@ const Notes = ({
   // declare mutation functions
   const [addNote] = useMutation(INSERT_NOTE);
   const [editNote] = useMutation(UPDATE_NOTE);
-  const [deleteNote] = useMutation(DELETE_NOTE);
+  const [deleteNote] = useMutation(SOFT_DELETE_NOTE);
```
johnclary, line 33: Screenshot 2024-09-05 at 11 09 23 AM

It's not quite right that we show the "Are you sure?" text and note text with the same styling. What do you think about not showing the note text at all in the modal?

johnclary (approved):
It works perfectly! So nice to have this in place—thank you!

I would like your feedback on a suggestion I made about the styling of the delete confirmation modal.

Also—I noticed this modal closes immediately, regardless of if the delete mutation resolved. I will open a backlog issue to hook the modal close into the mutation request , just like @chiaberry did in https://github.com/cityofaustin/atd-moped/pull/1399

atd-vze/src/Components/Notes/Notes.js
roseeichelmann: yeah i am very down to remove the note text completely!

johnclary: thanks, Rose! 🚢

atd-vze/src/Components/Notes/Notes.js
```diff
@@ -12,7 +12,7 @@ const Notes = ({
   notes,
   INSERT_NOTE,
   UPDATE_NOTE,
-  DELETE_NOTE,
+  SOFT_DELETE_NOTE,
```
mddilley, line 15: 🚀

mddilley (approved):
I tested this locally, and I could add, edit, and soft delete crash notes. I see the `is_deleted` column update as expected in the database too. Such an improvement and some nice code details along the way!! 🚀 🚢

PR #1542, opened 2024-09-04
Fix various things in the crash recommendations form

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/18468

## Testing
**URL to test:** 
[Netlify](https://deploy-preview-1542--atd-vze-staging.netlify.app/#/crashes)

**Steps to test:**
1. Go to our [staging website](https://visionzero-staging.austinmobility.io/editor/#/crashes) to see the buggy behavior of this component. Click on a crash without a fatality, update an injury to fatality to show the recommendation form. Try to add recommendation text or update, see that the mutation fails. Add a status, see that an empty partner record is created. Go to another crash, make it a fatality, and this time start by adding partners and then recommendation text, this actually works as expected.
2. Now test the netlify preview. See that the updates section isnt there until you create a recommendation. If you first create a recommendation text or status, a partner wont be created yet. Now add partners, status, etc.
3. Go to a new crash and this time add partners first, a recommendation record will be created when you add a partner. The update section will then display. Add text to all fields, status, etc. 

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/atd-vze/src/queries/recommendations.js
+++ b/atd-vze/src/queries/recommendations.js
@@ -15,9 +15,9 @@ export const GET_RECOMMENDATION_LOOKUPS = gql`
 
 export const INSERT_RECOMMENDATION = gql`
   mutation InsertRecommendation(
-    $recommendation_data: recommendations_insert_input!
+    $recommendationRecord: recommendations_insert_input!
   ) {
-    insert_recommendations(objects: [$recommendation_data]) {
+    insert_recommendations(objects: [$recommendationRecord]) {
       returning {
         crash_pk
         rec_update
--- a/atd-vze/src/queries/recommendations.js
+++ b/atd-vze/src/queries/recommendations.js
@@ -15,23 +15,9 @@ export const GET_RECOMMENDATION_LOOKUPS = gql`
 
 export const INSERT_RECOMMENDATION = gql`
   mutation InsertRecommendation(
-    $text: String
-    $update: String
-    $crashPk: Int
-    $userEmail: String
-    $recommendation_status_id: Int
-    $partner_id: Int
+    $recommendation_data: recommendations_insert_input!
   ) {
-    insert_recommendations(
-      objects: {
-        rec_text: $text
-        rec_update: $update
-        crash_pk: $crashPk
-        created_by: $userEmail
-        recommendation_status_id: $recommendation_status_id
-        recommendations_partners: { data: { partner_id: $partner_id } }
-      }
-    ) {
+    insert_recommendations(objects: [$recommendation_data]) {
       returning {
         crash_pk
         rec_update
--- a/atd-vze/src/views/Crashes/Recommendations/RecommendationMultipleSelectDropdown.js
+++ b/atd-vze/src/views/Crashes/Recommendations/RecommendationMultipleSelectDropdown.js
@@ -9,11 +9,18 @@ const RecommendationMultipleSelectDropdown = ({
   partners,
   fieldConfig,
   field,
+  doesRecommendationRecordExist,
 }) => {
   // Trigger mutation call to add partner upon click
   const handleOptionClick = selectedItem => {
     // Mutation expect lookup IDs as integers
-    const valuesObject = { [field]: parseInt(selectedItem.id) };
+    const valuesObject = !doesRecommendationRecordExist
+      ? {
+          recommendations_partners: {
+            data: { [field]: parseInt(selectedItem.id) },
+          },
+        }
+      : { [field]: parseInt(selectedItem.id) };
     onOptionClick(valuesObject);
   };
 
--- a/atd-vze/src/views/Crashes/Recommendations/RecommendationMultipleSelectDropdown.js
+++ b/atd-vze/src/views/Crashes/Recommendations/RecommendationMultipleSelectDropdown.js
@@ -9,18 +9,17 @@ const RecommendationMultipleSelectDropdown = ({
   partners,
   fieldConfig,
   field,
-  doesRecommendationRecordExist,
 }) => {
   // Trigger mutation call to add partner upon click
   const handleOptionClick = selectedItem => {
-    // Mutation expect lookup IDs as integers
-    const valuesObject = !doesRecommendationRecordExist
-      ? {
-          recommendations_partners: {
-            data: { [field]: parseInt(selectedItem.id) },
-          },
-        }
-      : { [field]: parseInt(selectedItem.id) };
+    // If the recommendation record doesnt exist yet we need to format the object for a nested insert
+    // to create a new recommendation and a new partner in the same request
+    const valuesObject = {
+      recommendations_partners: {
+        // Mutation expects lookup IDs as integers
+        data: { [field]: parseInt(selectedItem.id) },
+      },
+    };
     onOptionClick(valuesObject);
   };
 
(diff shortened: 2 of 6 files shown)


atd-vze/src/queries/recommendations.js
```diff
@@ -15,23 +15,9 @@ export const GET_RECOMMENDATION_LOOKUPS = gql`
...
-      objects: {
-        rec_text: $text
-        rec_update: $update
-        crash_pk: $crashPk
-        created_by: $userEmail
-        recommendation_status_id: $recommendation_status_id
-        recommendations_partners: { data: { partner_id: $partner_id } }
-      }
-    ) {
+    insert_recommendations(objects: [$recommendationRecord]) {
```
roseeichelmann, line 20: previously we had a bug where if you created a recommendation by just filling out the status, a recommendation record would be created along with a partner record that just had null as the partner id, so then you would end up with something that looks like this 
image
we dont want to create a partner record until the user actually adds a partner. so im changing this mutation to use an object variable that may or may not contain partner data so we arent creating empty partners

atd-vze/src/views/Crashes/Recommendations/Recommendations.js
```diff
@@ -164,20 +165,24 @@ const Recommendations = ({ crashPk, recommendation, refetch }) => {
...
-                label={"Updates"}
-                data={recommendation?.rec_update}
-                placeholder={"Enter updates here..."}
-                existingValue={getFieldValue(fieldConfig.fields.rec_update.key)}
-                field={fieldConfig.fields.rec_update.key}
-                doesRecommendationRecordExist={doesRecommendationRecordExist}
-                onAdd={onAdd}
-                onEdit={onEdit}
-              />
+          {doesRecommendationRecordExist && (
```
roseeichelmann, line 168: dont show the update section until a recommendation has been created

atd-vze/src/queries/recommendations.js
```diff
@@ -15,23 +15,9 @@ export const GET_RECOMMENDATION_LOOKUPS = gql`
 
 export const INSERT_RECOMMENDATION = gql`
   mutation InsertRecommendation(
-    $text: String
```
roseeichelmann, line 18: mutation was failing silently if you try to create a recommendation by first adding recommendation text because this mutation was expecting the variable `text` but we were feeding it `rec_text`

atd-vze/src/views/Crashes/Recommendations/RecommendationMultipleSelectDropdown.js
```diff
@@ -12,8 +12,14 @@ const RecommendationMultipleSelectDropdown = ({
...
   // Trigger mutation call to add partner upon click
   const handleOptionClick = selectedItem => {
-    // Mutation expect lookup IDs as integers
-    const valuesObject = { [field]: parseInt(selectedItem.id) };
+    // If the recommendation record doesnt exist yet we need to format the object for a nested insert
+    // to create a new recommendation and a new partner in the same request
+    const valuesObject = {
+      recommendations_partners: {
+        // Mutation expects lookup IDs as integers
+        data: { [field]: parseInt(selectedItem.id) },
```
roseeichelmann, line 20: alternatively i could have this `valuesObject` conditionally formatted based on whether `doesRecommendationRecordExist`, so it only gets this nested insert format if the rec record doesnt exist yet, and we wouldnt have to specify `valuesObject.recommendations_partners.data` in the `onAddPartner` function

atd-vze/src/views/Crashes/Recommendations/Recommendations.js
```diff
@@ -74,7 +75,7 @@ const Recommendations = ({ crashPk, recommendation, refetch }) => {
   const onAddPartner = valuesObject => {
     const recommendationPartnerRecord = {
       recommendationRecordId,
-      ...valuesObject,
+      ...valuesObject.recommendations_partners.data,
```
roseeichelmann, line 78: dont want that nested insert format if the recommendation record already exists and we are only inserting a partner record

roseeichelmann:
@johnclary the issue says to have visual feedback when the form is saving (does this mean a spinner?) so i guess you mean whenever the user clicks the add button or edits and saves, but none of our other components like notes or related records tables have any kind of feedback when you edit fields?

frankhereford (approved):
🚢🚢🚢 This is absolutely an improvement. Thank you @roseeichelmann!

I have been thinking about the possibilities you raised about using logic in the component to control the use of the nested record creation in the graphql. I am not sure, so don't give this too much weight, but I think I like it the way you have it. Simpler code, less deeply buried logic, and the application seems to be working as I would expect it. I like it!

chiaberry (approved):
i think I am following the alternative way of doing this, but the way you have it here makes sense to me. Thank you for fixing the silent failure, I tried adding things in different order and it all checks out

johnclary (changes requested):
Thanks @roseeichelmann!

I think that back when i created this issue i did not realize that editing the coordination partners or status input triggers an insert/update immediately—which was adding to my confusion about potential bugginess.

I request feedback in order to tighten up the save behavior of the text and text updates.

I agree we can leaving the loading feedback out of scope. If we were going to do a bigger refactor of this form, it would be great to rework this so that it behaves like one big form—you edit any field then click "Save" to save changes. I'd also like to make the parters and status inputs look like normal bootstrap inputs (with outlines, focus effects, etc), but the i can see multiselect library doesn't make that easy. It'd be nice to replace it with [react-select](https://react-select.com/) for a superior UX. I'll drop that in a backlog issue.

atd-vze/src/views/Crashes/Recommendations/RecommendationTextInput.js
```diff
@@ -42,7 +44,7 @@ const RecommendationTextInput = ({
   };
 
   // Recommendation record does not exist yet, adding a value will create record
-  const isAddingRecommendation =
+  const isCreatingRecord =
```
roseeichelmann, line 47: make it more apparent that we are talking about this is going to create a recommendation record, not talking about just adding recommendation text

atd-vze/src/views/Crashes/Recommendations/recommendationsDataMap.js
```diff
@@ -20,10 +20,5 @@ export const recommendationsDataMap = {
       label: "Updates",
       key: "rec_update",
     },
-    partner_id: {
-      label: "Coordination Partner:",
-      lookupOptions: "atd__coordination_partners_lkp",
-      key: "coord_partner_desc",
```
roseeichelmann, line 26: this was a duplication pretty much of the first object and its not actually doing anything?

atd-vze/src/views/Crashes/Recommendations/RecommendationTextInput.js
```diff
@@ -26,9 +26,11 @@ const RecommendationTextInput = ({
   };
 
   const handleSaveClick = () => {
-    const valuesObject = { [field]: inputValue };
-
+    const isEmptyString = !inputValue.trim();
+    // Save null to database if user entered empty string or only spaces/newlines
+    const valuesObject = { [field]: isEmptyString ? null : inputValue };
```
roseeichelmann, line 31: previously we were allowing empty strings to be saved to the database.

roseeichelmann:
@johnclary this is ready for re-review! i agree about making this one big form in the future. i disabled the save button until there is input and also made it so that users cant save empty strings to the database, including just spaces or newlines.

chiaberry:
I was re-testing this PR after the changes, and your comment made me want to try "deleting" the rec text, by updating it to be an empty string. It feels weird having the text flash again before the area turning into an input again. 

But I dont know how often the recommendation text would be deleted vs being updated. So maybe its not worth trying to fix the UX right now

PR #1545, opened 2024-09-09
Add a non-coa roadways layer to the database

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/15302

## Testing
**URL to test:** 
Locally

**Steps to test:**
1. Check out this branch, start your local db, run the migrations. 
2. You can just use the geojson i already created from running the first script and just test running the second script to insert the data into the db. Follow steps in the README and run
`node insert_roadways.js local`
4. Inspect the data in the table
5. Run the down migrations

---
#### Ship list
- [ ] Code reviewed


johnclary (approved):
Rose, this looks fantastic! It's ready to ship! 

Once I had this layer in the DB, I did a bunch of performance testing, with a focus on:

1. Do we need to index this layer's geometry column

I did not detect any benefit from this, probably because the entire geometry exists in a single row. I might be wrong on this and we can test in staging, but all of my tests indicate there is no benefit from an index here.

2. Should we break this layer into smaller features/rows, so that it can be indexed?

I tested this by clipping the layer into squares, using a few different cell sizes, plus indexing the geometry. To my surprise, this resulted in slower performance when testing the simple intersection of the crash position and any polygon row.

```sql
SELECT
    crash.id,
    crash.cris_crash_id,
    crash.address_primary,
    EXISTS (
        SELECT
            1
        FROM
            non_coa_roadways poly
        WHERE
            ST_Intersects(crash.position, poly.geometry)) AS is_non_coa_roadway
    FROM
        crashes crash
    LIMIT 100000;
```

I'd be happy to share the rest of my SQL if other folks want to test out the clipped/gridded version of this layer.


TLDR: ship it! 🚢

atd-vzd/migrations/default/1725907130269_create_non_coa_roadways_table/up.sql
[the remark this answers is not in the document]
roseeichelmann: done!

roseeichelmann:
@johnclary thanks for your thoughts on making this more performant. thats interesting that clipping and making more rows with indexing made the query slower :thinking: makes sense though that an index on a single row does add any benefit

johnclary (approved):
🚢 🚢 🚢 🚢

atd-toolbox/load_non_coa_roadways/process_roadways.js
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Params of note
+ * geometryPrecision=8: cap coordinates at 8 decimal places
+ * outSR=4326: use WGS84 coordinates
+ * f=json: return features as Esri json - although we could request geojson from the feature service, esri botches
+ * the geojson by creating features where holes should be. we need to use a third-party tool to convert esri to
+ * geojson
+ */
+ROADWAYS_URL =
```
mddilley, line 15: i noticed that an expired AGOL token is hardcoded in this URL. We probably don't intend to run this often (if ever again?), but it might be worth a noting that this request requires auth in case we find ourselves here again.

atd-toolbox/load_non_coa_roadways/process_roadways.js
```diff
@@ -0,0 +1,64 @@
...
+    }
+  });
+  return features;
+};
+
+const main = async () => {
+  const esriJson = await getRoadwaysJson();
+  const areaGeojson = arcgisToGeoJSON(esriJson);
+  const features = makeMultiPoly(areaGeojson.features);
+  filterAndFormatProperties(features);
```
mddilley, line 56: The script broke here since this function isn't defined. Same note as above that we may not intend to run this again, but I wanted to see how this worked and test it.

atd-toolbox/load_non_coa_roadways/process_roadways.js
roseeichelmann: yeahh unfortunately this layer is not public so i have to request a token that eventually expires to run this script. not sure how to best handle this for future reference, just a comment noting that i guess? and maybe i delete the part of the url that has the expired token?

mddilley: a comment makes sense to me! i think the token is a nice breadcrumb to keep around.

johnclary:
thanks @mddilley for testing the script 🙈

@roseeichelmann would you mind removing the busted function call?

roseeichelmann:
Okay @mddilley and @johnclary , i just removed that function call, added some comments, and ran both scripts again. All is working as expected :pray:

mddilley (approved):
Yay! Thanks for the extra context on what the future of this work looks like too. I enjoyed the discussion today. Looks great! 🚢 🙌

johnclary:
Thanks, Rose!

PR #1551, opened 2024-09-13
Ability to swap primary and secondary address

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/19016

## Testing
**URL to test:** 
[Netlify](https://deploy-preview-1551--atd-vze-staging.netlify.app/#/)

**Steps to test:**
1. Go to the crash details page, check out the new swap address button.
2. Click on it, confirm, watch the addresses swap. Also test cancelling.
3. There is a bug where if there is a null in one of the fields and not null value in the other field, on swap both of the primary and secondary will retain the not null value instead of one of them being null. This is because of how the layer data model is set up to ignore nulls in the edits table. We are addressing this in later work
4. Test a read only user, the button should not render

---
#### Ship list
- [ ] Code reviewed


atd-vze/src/Components/DataTable.js
```diff
@@ -46,7 +49,20 @@ const DataTable = ({
         return (
           <Col key={i} md="6">
             <Card key={section.title}>
-              <CardHeader>{section.title}</CardHeader>
+              <CardHeader>
+                <Row>
+                  <Col>{section.title}</Col>
+                  {section.button && (
```
roseeichelmann, line 55: alternatively this condition could just be based on the `section.title` if we dont plan on any other data tables ever having buttons... and then could remove the additional `button` attribute i added to `crashDataMap.js`

roseeichelmann: i updated this code to do the alternative option in this comment but not sure which i like more...

editor/src/views/Crashes/SwapAddressButton.js
```diff
@@ -0,0 +1,65 @@
...
+      .then(res => crashRefetch());
+  };
+
+  return (
+    <div className="float-right">
+      <Button color="primary" onClick={toggleModal} size="sm">
+        Swap Addresses
+        <i className="ml-2 fa fa-refresh" />
+      </Button>
+      <ConfirmModal
```
johnclary, line 52: is there a prop to disable the animation effect on this modal? i know this is re-used elsewhere but that's fine.

editor/src/views/Crashes/SwapAddressButton.js
```diff
@@ -0,0 +1,65 @@
+import React, { useState } from "react";
+import { Button } from "reactstrap";
+import ConfirmModal from "../../Components/ConfirmModal";
+import { UPDATE_CRASH } from "../../queries/crashes";
+
+const SwapAddressButton = ({ data, crashRefetch, ...props }) => {
```
johnclary, line 6: it'd be nice to tighten this up a tiny bit, and pass `crash` directly into this component and add `client` so that we don't need `...props`.

editor/src/views/Crashes/SwapAddressButton.js
```diff
@@ -0,0 +1,65 @@
...
+        variables: mutationVariables,
+      })
+      .then(res => crashRefetch());
+  };
+
+  return (
+    <div className="float-right">
+      <Button color="primary" onClick={toggleModal} size="sm">
+        Swap Addresses
+        <i className="ml-2 fa fa-refresh" />
```
johnclary, line 50: good icon 🙏

johnclary (changes requested):
My changes are really minor. This works great—thank you!

editor/src/views/Crashes/SwapAddressButton.js
chiaberry: good icon, but it could be cuter....

roseeichelmann: .... should i bring back `meow_blep`

roseeichelmann:
@johnclary i addressed your feedback and also disabled the animation, do you like it better that way?

atd-vze/src/Components/DataTable.js
johnclary: oops i missed this comment before. i think what you've got is just fine. we can revisit this if we need to make other carve-outs for special features in the future.

johnclary (approved):
👍 looks great! 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

atd-vze/src/Components/DataTable.js
mddilley: +1 - i had similar thoughts about an "action button" type prop or something, but I agree it is still early to know where this would go. Very cool to see it in the header!

editor/src/views/Crashes/SwapAddressButton.js
```diff
@@ -0,0 +1,65 @@
...
+const SwapAddressButton = ({ crash, crashRefetch }) => {
+  const [isConfirmModalOpen, setIsConfirmModalOpen] = useState(false);
+
+  const [updateCrash] = useMutation(UPDATE_CRASH);
+
+  const toggleModal = () => {
+    setIsConfirmModalOpen(!isConfirmModalOpen);
+  };
+
+  const mutationVariables = {
```
mddilley, line 16: i tested first and was curious how you implemented the data swap - very cool!

mddilley (approved):
I was able to swap a bunch of addresses 🔄, and I replicated the bug that you described with crash id `13620746`. Looks great, and I know this feature will be so useful! 🚢 🙌

PR #1560, opened 2024-09-19
Fix column order in units_cris_insert_rows()

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/19135

I'm not sure how this got through testing—I guess the extracts I tested with didn't have any data in these columns, and I didn't test the Create Crash Record form 😕 

## Testing

1. Start your local stack
2. Use the VZE to create a new crash record
4. Run the cris import:
 ```
docker run -it  --rm --env-file .env --network host -v $PWD:/app cris_import_cris_import ./cris_import.py --csv --s3-download
```

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/database/migrations/default/1726766736889_patch-units_cris_insert_rows/down.sql
+++ b/database/migrations/default/1726766736889_patch-units_cris_insert_rows/down.sql
@@ -69,16 +69,10 @@ BEGIN
         new.veh_damage_direction_of_force2_id,
         new.veh_damage_severity1_id,
         new.veh_damage_severity2_id,
-        new.veh_hnr_fl,
         new.veh_make_id,
+        new.veh_hnr_fl,
         new.veh_mod_id,
         new.veh_mod_year,
         new.veh_trvl_dir_id,
         new.vin
     );
-    -- insert new (editable) vz record (only record ID)
-    INSERT INTO public.units_edits (id) values (new.id);
-
-    RETURN NULL;
-END;
-$function$;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: database/migrations/default/1726766736889_patch-units_cris_insert_rows/down.sql [hunk 1]
```diff
@@ -0,0 +1,78 @@
...
+        new.updated_by,
+        new.veh_body_styl_id,
+        new.veh_damage_description1_id,
+        new.veh_damage_description2_id,
+        new.veh_damage_direction_of_force1_id,
+        new.veh_damage_direction_of_force2_id,
+        new.veh_damage_severity1_id,
+        new.veh_damage_severity2_id,
+        new.veh_make_id,
+        new.veh_hnr_fl,
```
johnclary, line 73: you can see the bug here in the down migration `new.veh_make_id` and `new.veh_hnr_fl` were swapped. this is a terrifying issue if, say, both of these were boolean types. it would have been very hard to catch that we were routing data to the wrong columns 😓

johnclary: we could absolutely write a helper script that generates this function based on the `_column_metadata` table. that is probably the right move 🤔

mddilley: oh yeah! this is so easy to overlook. i scanned the order of the insert and value fields in the new migration, and I see everything matching up! 🙌

mddilley (approved):
Works great - thanks for fixing! 🙌

If you add more changes, re-request my review, and I'll test again right away. 👍

frankhereford (approved):
Code changes are 👍 and the testing instructions ran without error here, so 🚢🚢🚢.

Please take me with a grain of salt though, because I have not entirely come to understand why the testing instructions actually test this, but I completely believe that they do.

database/migrations/default/1726766736889_patch-units_cris_insert_rows/down.sql
frankhereford: +1 to your idea of automating the creation of this (or any) function. code that writes code is always the hardest to read and write, but good gracious is it handy to have.

johnclary:
thank you!!!

PR #1564, opened 2024-09-20
Enable address editing

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/19014

## Testing

**URL to test:** 
[Netlify](https://deploy-preview-1564--atd-vze-staging.netlify.app/#/crashes)

**Steps to test:**
1. Go to a crash
2. Test editing the primary address/secondary address fields
3. When you type into an input your input should be auto capitalized
4. If you are editing and click on the swap address button before saving/canceling, you will be taken out of edit mode on that field
5. Test swapping the addresses

---

#### Ship list

- [ ] Code reviewed


editor/src/Components/DataTable.js
```diff
@@ -160,8 +173,10 @@ const DataTable = ({
                                         autoFocus
                                         name={field}
                                         id={field}
-                                        onChange={e => handleInputChange(e)}
-                                        defaultValue={fieldValue}
```
roseeichelmann, line 164: im updating all of these inputs to be [controlled components](https://blog.logrocket.com/controlled-vs-uncontrolled-components-in-react/) instead of uncontrolled

editor/src/Components/DataTable.js
```diff
@@ -177,8 +192,12 @@ const DataTable = ({
                                         name={field}
                                         id={field}
                                         type="text"
-                                        defaultValue={fieldValue}
-                                        onChange={e => handleInputChange(e)}
+                                        value={editValue
+                                          .toString()
+                                          .toUpperCase()}
```
roseeichelmann, line 197: updating these to be controlled components allows me to auto capitalize user input as they type

editor/src/views/Crashes/SwapAddressButton.js
```diff
@@ -3,12 +3,14 @@ import { Button } from "reactstrap";
...
 
-const SwapAddressButton = ({ data, crashRefetch, ...props }) => {
+const SwapAddressButton = ({ data, crashRefetch, setEditField, ...props }) => {
   const [isConfirmModalOpen, setIsConfirmModalOpen] = useState(false);
 
   const crash = data?.crash;
 
   const toggleModal = () => {
+    // make sure we are not in edit mode on a field
+    setEditField("");
```
roseeichelmann, line 13: so that if we are in edit mode on a field and click the swap address button, we get out of edit mode

editor/src/Components/DataTable.js
```diff
@@ -177,8 +192,12 @@ const DataTable = ({
                                         name={field}
                                         id={field}
                                         type="text"
-                                        defaultValue={fieldValue}
-                                        onChange={e => handleInputChange(e)}
+                                        value={editValue}
+                                        onChange={e =>
+                                          setEditValue(
+                                            e.target.value.toUpperCase()
```
roseeichelmann, line 198: auto capitalize user input

editor/package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "atd-vz-data",
-  "version": "1.47.0",
+  "version": "2.1.0",
```
roseeichelmann, line 3: from running `npm install` for the first time in awhile

chiaberry (approved):
I tried editing fields, quitting editing halfway through, opening them again, swapping addresses mid edit, and it all works as you described it should.

editor/src/Components/DataTable.js
```diff
@@ -158,8 +171,10 @@ const DataTable = ({
...
                                         id={field}
-                                        onChange={e => handleInputChange(e)}
-                                        defaultValue={fieldValue}
+                                        onChange={e =>
+                                          setEditValue(e.target.value)
+                                        }
+                                        value={editValue}
                                         type="select"
                                       >
                                         {selectOptions.map(option => (
```
johnclary, line 180: i just noticed we have an eslint error here because each <option> needs a unique key. would you mind setting the `key` prop on these 👇 ?

editor/src/Components/DataTable.js
```diff
@@ -33,13 +34,22 @@ const DataTable = ({
   const roles = getRoles();
   const isReadOnlyUser = isReadOnly(roles);
 
+  // Sets the state of the value for the current field being edited
+  const [editValue, setEditValue] = useState();
```
johnclary, line 38: eslint is barking because `null`/`undefined` is not a valid value to pass to `input` elements. we need an empty string as the default value and whenever you reset the value.

editor/src/Components/DataTable.js
```diff
@@ -33,13 +34,22 @@ const DataTable = ({
...
+
   // Import Lookup tables and aggregate an object of uiType= "select" options
   const { data: lookupSelectOptions } = useQuery(GET_LOOKUPS);
 
+  const handleEditClick = (field, fieldValue) => {
+    setEditField(field);
+    setEditValue(fieldValue);
```
johnclary, line 45: here to you need to set the edit value to `""` if it is null

editor/src/Components/DataTable.js
johnclary: i'm also getting this warning when i start editing a blank field:
> Warning: A component is changing an uncontrolled input of type text to be controlled.

i _think_ this might resolve itself if you fix make default input type a string.

johnclary (changes requested):
yay, address editing! 🙌 

and thanks for switching these over to controlled components. i'm seeing a few warnings that should be a quick fix.

also, one result of this change is that the previous value flashes on the screen while the save + refetch happens. can you see if there's a fix for this maybe in the `handleFieldUpdate` callback?

🙏

roseeichelmann:
@johnclary this is ready for re-review!

johnclary (changes requested):
Rose, I just noticed a little bug when editing a yes/no field. The dropdown defaults to "Yes", even when the current value is "No". 

E.g., here's the non-edit state:
Screenshot 2024-10-02 at 1 46 32 PM

And here's what is rendered after toggling the edit:
Screenshot 2024-10-02 at 1 46 35 PM

Can you check this?

mddilley (commented):
Looks great! I replicated the bug that John caught too, and, after that, this is good to go! I was able to edit every address field and all other select (non-boolean) and text type fields too.

roseeichelmann:
@johnclary @mddilley thanks for catching that guys! it had to do with the way i was checking for falsey values in general instead of explicitly checking if the edit value was null to set it to an empty string in the `handleEditClick`

johnclary (approved):
🚢 🚢 🚢 🚢

mddilley (approved):
Awesome! I tested again with the updated handling and it is ✨

PR #1543, opened 2024-09-06
Cleanup toolbox and update readme

johnclary:
## Associated issues

- https://github.com/cityofaustin/atd-data-tech/issues/18549

## Testing

Review these changes and let me know what you think 🙏 

---
#### Ship list
- [ ] Check migrations for any conflicts with latest migrations in master branch
- [ ] Confirm Hasura role permissions for necessary access
- [ ] Code reviewed
- [ ] Product manager approved

PR #1567, opened 2024-09-25
Restores helper script's ability to download db data

johnclary:
## Associated issues

As mentioned [in Slack](https://austininnovation.slack.com/archives/CM9MK950S/p1727286078220699), i broke the helper script when i cleaned up the toolbox in https://github.com/cityofaustin/vision-zero/pull/1543.

It looks like we can just use the existing `postgis` service to run the `psql` commands that the toolbox container was managing. I definitely want Frank to take a look before we merge this.

## Testing

**URL to test:** Local

**Steps to test:**
Run through your normal replicate DB steps and make sure the VZE behaves normally.

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/database/Dockerfile
+++ b/database/Dockerfile
@@ -1,3 +1,3 @@
-FROM frankinaustin/postgis-multiarch:14-3.3
+dFROM frankinaustin/postgis-multiarch:14-3.3
 RUN apt-get update
-RUN apt-get install -y aptitude magic-wormhole vim
+RUN apt-get install -y aptitude vim
--- a/database/Dockerfile
+++ b/database/Dockerfile
@@ -1,3 +1,3 @@
-dFROM frankinaustin/postgis-multiarch:14-3.3
+FROM frankinaustin/postgis-multiarch:14-3.3
 RUN apt-get update
 RUN apt-get install -y aptitude vim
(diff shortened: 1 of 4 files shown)


database/Dockerfile
```diff
@@ -1,3 +1,3 @@
-FROM frankinaustin/postgis-multiarch:14-3.3
+dFROM frankinaustin/postgis-multiarch:14-3.3
 RUN apt-get update
-RUN apt-get install -y aptitude magic-wormhole vim
```
johnclary, line 3: i am not a wormhole user but i realize others might find it important to keep this in the container. i was not up for troubleshooting this build issue. let me know what y'all prefer I do 🙏

docker-compose.yml
```diff
@@ -10,17 +10,6 @@ services:
       HASURA_GRAPHQL_DATABASE_URL: postgres://visionzero:visionzero@postgis:5432/atd_vz_data
       HASURA_GRAPHQL_ENABLE_CONSOLE: "true"
       HASURA_GRAPHQL_CONSOLE_ASSETS_DIR: /srv/console-assets
-  db-tools:
```
mddilley, line 13: ok cool, so this PR re-maps the dependencies of this container to `postgis` which is now being used for the snapshotting part of the local stack functionality. 👍

makes sense to me! i can't think of another dependency on this service.

mddilley (approved):
The local stack came up for me with all the expected tables excluded. I started up VZE and browsed the main views. All looks great to me, and works! ✅ 🚢

tillyw (approved):
tested and working as expected!

PR #1568, opened 2024-09-26
Return of the OCR'd crash narrative

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/19204 ♾️
- https://github.com/cityofaustin/atd-data-tech/issues/19229

## Testing

1. You will need a copy of the database that includes the `change_log_crashes_cris` table data. To include just this one change log table in your replica, comment out `change_log_crashes_cris` right [here](https://github.com/cityofaustin/vision-zero/blob/8dfc6b7bc51ccf14209bb2cc589ee2c79bec3bdf/vision-zero#L22) in the helper script.  

```python
# ./vision-zero helper script
TABLES_TO_IGNORE = [
    "public.change_log_crashes",
     #  "public.change_log_crashes_cris",
    "public.change_log_units_cris",
    "public.change_log_people_cris",
    "public.change_log_people",
    "public.change_log_units",
    "public.change_log_people_edits",
    "public.change_log_units_edits",
    "public.change_log_crashes_edits",
]
```

2. And now replicate like normal. This will download ~4gb of data:

```shell
./vision-zero replicate-db
```

3. Apply migrations and metadata

```
hasura migrate apply
hasura metadata apply
```

4. Now you will use the change log to backfill narratives that have been erased in recent weeks. Locate the commented out `update` statement at the bottom of the `1727451510064_preserve_crash_narrative` up migration ([here](https://github.com/cityofaustin/vision-zero/blob/004b670671f66e3650fa0a9e16161f5e565bbe3b/database/migrations/default/1727451510064_preserve_crash_narrative/up.sql#L30-L50)). Manually execute this update command in your sql client.

5. At this point there should be ~800 remaining crashes that are missing a narrative and need to be OCR'd. You can check this by querying the new SQL view:

```sql
select count(*) from view_crash_narratives_ocr_todo;
```

6. It's time to test the OCR script. head to the `./etl/cris_import` directory and rebuild the ETL docker image:

```shell
docker compose build
```

7. We need to set our environment to use the prod S3 bucket, because it contains all of the CR3 PDFs that we will process. The best way to do this is to save a copy of your existing `.env` file as `prod.local.env`, and set the `BUCKET_ENV` value to `prod` in this new file (this PR renames the `ENV` var to `BUCKET_ENV`). Make sure your Hasura endpoint is set to you local host.

8. Now it really is time to run the OCR script. Use this run command to run the OCR process with your new env file:

```shell
docker run -it  --rm --env-file prod.local.env --network host -v $PWD:/app cris_import_cris_import ./cr3_ocr_narrative.py --verbose
```

9. Nice! You can inspect the new narratives by querying the `crashes_edits` table:

```sql
select
    id,
    updated_at,
    updated_by,
    investigator_narrative
from
    crashes_edits
where
    investigator_narrative is not null;
```

10. Now we will test the CRIS import and make sure that our narrative-preserving trigger is working as expected. First, let's inspect a batch of crashes the should have investigator narratives, observing that the narrative is populated for all records:

```sql
select
    investigator_narrative
from
    crashes
where
    cris_crash_id in(18510430, 18548351, 18553517, 18553518, 18553519, 18553520, 18553522, 18553525, 18553527, 18553532, 18566225, 18569950, 18578152, 18600315, 18613803, 18615918, 18634346, 18640454, 18640455, 18945007, 19208347);
```

11.  I have placed a CRIS extract in the dev S3 inbox for testing (`extract_2023_20240823135638635_99726_20240920_HAYSTRAVISWILLIAMSON`), and this extract is missing a large number of crash narratives, including for those crashes listed above. Let's make sure those narratives are not going to be overwritten, thanks to the new DB trigger.  
 
Make sure your `BUCKET_ENV` is set to `dev` in your `.env` file, and run the CRIS import like so:

```
docker run -it  --rm --env-file .env --network host -v $PWD:/app cris_import_cris_import ./cris_import.py --csv --s3-download
```

15. Excellent—now query the DB again and make sure these crash narratives are still intact:

```sql
select
    investigator_narrative
from
    crashes
where
    cris_crash_id in(18510430, 18548351, 18553517, 18553518, 18553519, 18553520, 18553522, 18553525, 18553527, 18553532, 18566225, 18569950, 18578152, 18600315, 18613803, 18615918, 18634346, 18640454, 18640455, 18945007, 19208347);
```

That's it—thank you for testing this PR 👍 

---

#### Ship list

- [ ] Code reviewed


etl/cris_import/README.md
```diff
@@ -2,21 +2,23 @@
...
+- **`cr3_ocr_narrative.py`**: a utility script which extracts crash narrative data from CR3 PDFs using Optical Character Recognition (OCR)
+- **`_restore_zips_from_archive.py`**: a development helper script that moves CRIS extract zips from `./archive` to `./inbox` in the S3 bucket.
 
-Set your `ENV` to `dev` in order to safely run the S3 operations locally.
+## Quick start - CRIS import
 
 1. Start your local Vision Zero cluster (database + Hasura + editor).
 
-2. Save a copy of the `env_template` file as `.env`, and fill in the details.
+2. Save a copy of the `env_template` file as `.env`, and fill in the details. Set the `BUCKET_ENV` variable to `dev` in order to safely run the S3 operations locally.
```
johnclary, line 15: i decided to change the env var `ENV` to `BUCKET_ENV`, just to be super clear that this is the only S3 operations are affected by this variable. For example, you could set `BUCKET_ENV` to `prod` and still point your hasura endpoint/secret to local, and vice-versa.

etl/cris_import/cr3_ocr_narrative.py
```diff
@@ -0,0 +1,115 @@
...
+    # uncomment to save a copy of the pdf page image
+    # page.save(f"temp/{cris_crash_id}_page_{cr3_version}.jpeg")
+
+    bbox = NARRATIVE_BBOX_PIXELS[cr3_version]
+
+    logger.debug("Cropping narrative from PDF...")
+    narrative_image = page.crop(bbox)
+
+    # uncomment to save a copy of the cropped narrative image
+    # narrative_image.save(f"temp/{cris_crash_id}_narrative_{cr3_version}.jpeg")
```
johnclary, line 54: this is essential for debugging this, so i figured i'd leave it in here 🤷

etl/cris_import/cr3_ocr_narrative.py
```diff
@@ -0,0 +1,115 @@
...
+
+    logger.debug("Extracting narrative text...")
+    narrative = None
+    narrative = image_to_string(narrative_image, lang="eng")
```
johnclary, line 58: the old version of the script used a `try`/`except`/ block for this call, but it never errored for me, even when parsing some gnarly old garbage PDFs. my thinking was that for now it would be better to know if this breaks somehow, but if it becomes a nuisance we can easily revisit.

etl/cris_import/cr3_ocr_narrative.py
```diff
@@ -0,0 +1,115 @@
...
+    errors = []
+    for index, future in enumerate(futures):
+        try:
+            future.result()
+        except Exception as e:
+            # grab the filename from the pdf list
+            crash = todos[index]
+            errors.append([crash["cris_crash_id"], e])
+
+    if errors:
```
johnclary, line 104: this is the same behavior as the crash diagram extractor

etl/cris_import/utils/process_pdfs.py
```diff
@@ -35,24 +47,13 @@ def get_cr3_version(page, page_width):
...
-
-    for pixel in test_pixels:
-        rgb_pixel = page.getpixel(pixel)
-        if rgb_pixel[0] > 5 or rgb_pixel[1] > 5 or rgb_pixel[2] > 5:
-            # the PDF fails our pixel checks, so assume it's the
-            # earliest version
-            return f"v1_{page_size}"
-
-    return f"v2_{page_size}"
+    width, height = page.size
```
johnclary, line 50: i discovered this much easier method for getting the image size 👍

etl/cris_import/utils/settings.py
```diff
@@ -8,8 +8,16 @@
     "v2_large": (2496, 3036, 4836, 5464),
 }
 
+"""This is the bbox used to crop the crash narrative"""
+NARRATIVE_BBOX_PIXELS = {
+    "v1_small": (30, 928, 681, 1590),
+    "v1_large": (296, 3683, 2580, 5749),
+    "v2_small": (30, 791, 650, 1430),
+    "v2_large": (90, 3026, 2496, 5466),
+}
```
johnclary, line 17: it occurred to me that we might want to start storing this version in the database. another option would be to use the python PDF package to actually add metadata to the PDF itself indicating which version.

i spent quite a lot of time testing different ways to identify which PDF version is being used, including by parsing the CR3 from revision date using OCR. i learned two things. 1) there are many more form revisions than what we test for (at least 5 as far as i can tell) and 2) it is hard to reliable to distinguish between them.

I could expand on this and have more thoughts on it if y'all want to chat CR3 versions. My short summary is that what we're doing now captures 99% of what we need.

database/metadata/databases/default/tables/public_view_crash_narratives_ocr_todo.yaml
mddilley, line 1: nice! i watched the row count decrease as i ran the OCR script 🚀 Very cool to have the conditions for OCRing stored in a DB view this way.

database/migrations/default/1727451510064_preserve_crash_narrative/up.sql
```diff
@@ -0,0 +1,50 @@
...
+
+create or replace trigger crashes_cris_preserve_investigator_narrative_on_update
+before update on public.crashes_cris
+for each row
+when (
+    new.investigator_narrative is null and old.investigator_narrative is not null
+)
+execute procedure public.crashes_cris_set_old_investigator_narrative();
+
+comment on function public.crashes_cris_set_old_investigator_narrative is 'Sets the
```
mddilley, line 19: awesome - this comment answered my question before i could ask it! 🙏

mddilley (approved):
All the steps worked as expected, and I was able to see narratives OCR'd and verify that the narratives would not be overwritten. Thanks for sprucing this up and providing the SQL. Made testing a breeze. 🚢 🙌

etl/cris_import/env_template
chiaberry, line 1: here is where I need to update ENV to be BUCKET_ENV, right?

johnclary: yep!

etl/cris_import/README.md
chiaberry: I like this change, what do you think about updating the env_tamplate with the new var name?

chiaberry (approved):
Thanks for the test steps and the additions to the README. 

I welcome the return the of the OCR'd crash narrative

etl/cris_import/env_template
johnclary: ah, sorry—the diff on this is really confusing. i see now that i added the `BUCKET_NAME` var but failed to remove the old `ENV` var from this template.

chiaberry: oh duh, yes sorry, i was getting confused

johnclary: no no— i am the one who is confused. i will fix this now

database/migrations/default/1727375054583_narrative_ocr/up.sql
```diff
@@ -0,0 +1,39 @@
...
+        id,
+        cris_crash_id
+    from
+        crashes
+    where
+        cr3_stored_fl = TRUE
+        and investigator_narrative is NULL
+        and (
+            investigator_narrative_ocr_processed_at is NULL
+            or cr3_processed_at >= investigator_narrative_ocr_processed_at
```
roseeichelmann, line 28: so if we got a new cr3 that was processed and we have to OCR the investigator narrative again?

database/migrations/default/1727451510064_preserve_crash_narrative/up.sql
roseeichelmann: +1 loving the function comments

etl/cris_import/cr3_ocr_narrative.py
```diff
@@ -0,0 +1,115 @@
...
+    logger.debug(f"Extracted narrative: {narrative}")
+
+    variables = {
+        "id": crash_pk,
+        "updates": {
+            "investigator_narrative": narrative,
+            "investigator_narrative_ocr_processed_at": datetime.now(
+                timezone.utc
+            ).isoformat(),
+            "updated_by": "dts_automation",
```
roseeichelmann, line 69: should this be our standard naming convention going forward for automated tasks/etls making updates?

etl/cris_import/cr3_ocr_narrative.py
```diff
@@ -0,0 +1,115 @@
...
+    todos = make_hasura_request(query=NARRATIVES_TODO_QUERY)[
+        "view_crash_narratives_ocr_todo"
+    ]
+
+    logger.info(f"{len(todos)} to process")
+    if not todos:
+        return
+
+    futures = []
+    with ProcessPoolExecutor(max_workers=cli_args.workers) as executor:
```
roseeichelmann, line 89: so this whole block is for creating multiple parallel processes to run the `extract_narrative_pdf` function concurrently for the crashes in the todos list?

johnclary: yep!

etl/cris_import/cr3_ocr_narrative.py
johnclary: it would be nice to agree on a convention—i did feel 100% great about this and am very open to ideas.

chiaberry: you did feel 100% great? or are you looking for other ideas? 
would you want to label anything updated by our scripts with `dts_automation`, or make it more specific with which script updated it? like `cris_import_script`

johnclary: did not* feel 100% great. yeah, i was wondering if we see benefit in being specific about what task made the updates or just using the generic `dts_automation`.

PR #1575, opened 2024-10-10
Move layers into `geo` schema

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/19415

## Testing

**URL to test:** Local

We just want to make sure that VZ is business as usual.

1. Start your local stack, and apply migrations and metadata
2. Run the CRIS import, something like this should do the trick:

```
 docker run -it  --rm --env-file local.env --network host -v $PWD:/app cris_import_cris_import ./cris_import.py --csv --skip-unzip
```

3. Follow the [load_agol_layer tool instructions ](https://github.com/cityofaustin/vision-zero/tree/more-layer-updates/toolbox/load_agol_layer) to load a layer or two into your local instance

4. Start up the VZE. Edit a crash's location. Make sure the locations pages load.
5. In the VZE, test the upload non-cr3 tool

---

#### Ship list

- [ ] Code reviewed


johnclary:
👋 folks, this is a change that we discussed pretty briefly, so i'm curious for your feedback. i think the name `geo` is concise and self explanatory, but open to other ideas (including that we don't like using a schema at all of these tables).

one other thought is that i had doubts about whether to move `atd_txdot_locations` into this new schema. i decided to hold off on that because the table has so many dependencies. i think we can tackle that whenever we get to https://github.com/cityofaustin/atd-data-tech/issues/18629

roseeichelmann (approved):
i like these layers being in a separate schema, and `geo` sounds good to me. also i think `atd_txdot_locations` should eventually be moved into this schema but its fine to address that later :rocket:

database/migrations/default/1728581588450_create_geo_schema/up.sql
mddilley, line 1: So nice and clean thanks to Frank's recent squash 🧹 ✨ 

Screenshot 2024-10-25 at 2 05 37 PM

toolbox/load_agol_layer/src/utils.js
```diff
@@ -164,7 +164,7 @@ const getEsriJson = async (url) => {
  */
 const getTruncateMutation = (tableName) => `
   mutation Delete${tableName} {
-    delete_${tableName}(where: {}) {
+    delete_geo_${tableName}(where: {}) {
```
mddilley, line 167: Pretty cool that Hasura includes the schema in these generated mutations. I hadn't caught that detail before now.

johnclary:
thank you!!

PR #1579, opened 2024-10-15
Ability to global search by case ID

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/19478

I'm curious how folks feel about this. It can be a little clunky, but was the easiest path forward I could come up with. 

If you look in the commit history you will see that I played around with using [react-select](https://react-select.com/) to show search results in a dropdown menu, but i scrapped this because it was getting pretty complicated.

Screenshot 2024-10-15 at 3 48 29 PM

## Testing

**URL to test:** 
[Deploy preview](https://deploy-preview-1579--atd-vze-staging.netlify.app/#/crashes)

**Steps to test:**
1. Search with a valid crash ID
2. Search with an invalid crash ID, then edit the crash ID to be valid
3. Repeat with case ID
4. Try searching by **Crash ID** with a valid **Case ID**, then use the button to switch the search field to **Case ID**
5. Set the search field to **Case ID** and refresh the page to verify that your field selection is persisted

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/package-lock.json
+++ b/editor/package-lock.json
@@ -48,7 +48,6 @@
         "react-map-gl": "^7.1.7",
         "react-router-config": "^5.0.1",
         "react-router-dom": "^5.1.2",
-        "react-select": "^5.8.1",
         "react-test-renderer": "^16.8.6",
         "react-zoom-pan-pinch": "^1.6.1",
         "reactstrap": "^8.0.0",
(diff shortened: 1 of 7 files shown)


editor/package-lock.json
johnclary, line 1: this is the result of installing and uninstalling `react-select`

editor/src/helpers/environment.js
```diff
@@ -1 +1,8 @@
 export const isDev = window.location.hostname === "localhost";
+
+/**
+ * Use this code to apply breaking changes to local storage settings.
+ * when the code name is changed, user's will have their local storage
+ * reset. the code name is the VZ release name
+ */
+export const appCodeName = "asclepias";
```
chiaberry, line 8: I thought you were speaking in a secret code until I realized this was the VZ release name for 1.48

chiaberry:
I can do a screen recording or a screen share but heres something I came across while testing
1. toggle it to search by case id
2. put in something totally wrong, hit enter
3. see the case not found message. clear the field by highlighting and deleting (should you be able to click the red x to clear the input? thats not working for me)
4. paste in a valid case id. the error message comes up. (it still has the previous state maybe?)

Id like there to be a way to clear the search by clicking, because if i put an invalid search value in there, then navigate to another page, the error message stays up no matter how much i navigate, reminding me of my mistakes. 

I also swear that I've been able to put in a correct case id, go back to the crashes page and then put in an incorrect one and still be taken to the previous correct case id, but I cant recreate it every time. 

I like being able to toggle the crash id -> case id by clicking and it doesnt reset my search input. It wasnt immediately obvious to me that the toggle was clickable, but I think thats because i am not used to the core UI UX

chiaberry:
I was able to repeat the weird navigation thing that happens with the wrong input, i have a screen recording

johnclary:
thanks @chiaberry! i added a piece of a state that is making the error validation less wonky. would you give it another test?

> Id like there to be a way to clear the search by clicking,

yes, me too. and it doesn't help that the error validation shows an `X` in the search input that looks like it should be clickable. i think my most recent change makes this all less confusing, though. lmk what you think 🙏

chiaberry (approved):
As you know I tried my darnedest to break this, and now I can't. 🚢 

Its still a bummer that the red x isnt a button to clear the search input, but cest la vie. I thought about including the caret in the toggle button to make it more clear that its something that can be changed, but without an actual dropdown the caret seems misplaced. 

The local storage toggle setting is a nice touch.

editor/src/Components/CrashNavigationSearchForm.js
```diff
@@ -1,41 +1,144 @@
...
+    setHasSearchedTerm(false);
+  }, []);
+
+  const retrievedRecordLocator = data?.[globalSearchField]?.[0]?.record_locator;
+
+  /**
+   * Hook that redirects to crash page once we have found a
+   * crash record
+   */
+  useEffect(() => {
```
mddilley, line 76: I was able to replicate the bug that Mateo found, and I also think that it is an edge case that might lead to this `useEffect` taking on more work to resolve. I thought about how to tie the state updates more closely to the request trigger, and I remembered that we ran into issues with the [`onCompleted`](https://www.apollographql.com/docs/react/data/queries#queryhookoptions-interface-oncompleted) callback not working as expected in Moped.

[Rose wrote this snazzy code](https://github.com/cityofaustin/atd-moped/blob/46e3edc09bb901286ba9fc75551f494838ff324c/moped-editor/src/views/projects/projectsListView/useProjectListViewQuery/useCsvExport.js#L193-L204) that is a nice workaround for getting a solid chain of events after the async call. Might be helpful for the next iteration.

mddilley (approved):
Looks and works great! I was able to replicate the edge case that Mateo found, and I left some thoughts on a similar issue we ran into in Moped. I like the button toggling between the crash/case id since there are only two options. 🚢

editor/src/Components/CrashNavigationSearchForm.js
johnclary: thanks, Mike! (and Mateo for testing, too). i was just coming here to say that i think using the imperative `client` to make queries is the only way this feature is going to work. awesome to have this example from Rose 🙌 

looks like a lot of other folks have run into this issue, too, e.g. [this SO post](https://stackoverflow.com/questions/66020910/apollo-is-it-possible-to-clear-the-data-obj-after-calling-uselazyquery).

johnclary:
👋 i've requested re-review becausei rewrote this to make calls directly from the apollo client. thanks for bearing with me on this tricky feature 😅

chiaberry (approved):
Tried breaking again, this time also doing Mateo's test and couldnt.  
🚢

editor/src/Components/CrashNavigationSearchForm.js
```diff
@@ -1,41 +1,145 @@
...
             size="sm"
-            onClick={() => {
-              history.push(`/crashes/${crashSearchId}`);
-              setCrashSearchId("");
-            }}
+            onClick={onSearch}
           >
-            <i className="fa fa-arrow-right" />
+            {!isLoading && <i className="fa fa-search" />}
+            {isLoading && <Spinner style={{ height: "1rem", width: "1rem" }} />}
```
mddilley, line 142: ooo i like the spinner ✨

mddilley (approved):
I tested searching by case ID and crash ID, and I see everything working as expected whenever there is a match or not. 🚢 🙌 🚀

johnclary:
thanks for all the help on this 😎

PR #1582, opened 2024-10-21
Make VZE compatible with the two-layer model

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/19366

## Testing

**URL to test:** 
Local

**Steps to test:**
1. Check out this branch. Spin up recent copy of the prod database. Apply the hasura migrations and metadata.
2. Open the VZE. Check that all pages work, especially the crashes and location list pages, crash details page.
3. Test sorting and filtering on the crashes list page.
4. On the crashes details page test editing the coordinates, the Geocode Provider above the map should update to Manual Geocode whenever you edit coordinates that were previously provided by CRIS
5. Also test updating values in the related records table, data tables, etc.
6. Test swapping the primary and secondary addresses - the [bug](https://github.com/cityofaustin/vision-zero/pull/1551#issue-2525714232) that lead us to implementing the two layer data model is fixed now!! :partying_face: 

---

#### Ship list

- [ ] Code reviewed


database/migrations/default/1728066705364_drop_crashes_edits_update_crashes_cris_functions/down.sql
```diff
@@ -381,3 +381,141 @@ ALTER TABLE ONLY public.crashes_edits
 
 ALTER TABLE ONLY public.crashes_edits
     ADD CONSTRAINT crashes_edits_wthr_cond_id_fkey FOREIGN KEY (wthr_cond_id) REFERENCES lookups.wthr_cond(id) ON UPDATE CASCADE ON DELETE RESTRICT;
+
+
+-- Recreating these views
+
+
+CREATE OR REPLACE VIEW public.crashes_list_view
```
roseeichelmann, line 389: forgot to add these views back in the previous down migrations

roseeichelmann:
Something to discuss is how we want to handle "resetting" values now. Now if you edit a value in the related records table and click RESET, its just setting the value in the  unified table to null. Im happy to address this problem in this PR if yall have ideas!

I think we could provide a null option and a reset option, and the reset option would check whats in the cris table and input that into the unified table

database/migrations/default/1729527603827_recreate_crashes_list_view/up.sql
```diff
@@ -0,0 +1,100 @@
+-- Recreating this view and removing references to crashes_edits
+
+CREATE OR REPLACE VIEW public.crashes_list_view
+AS WITH geocode_status AS (
+         SELECT cris.id,
+            unified.latitude IS NOT NULL OR unified.longitude IS NOT NULL AS is_manual_geocode
+           FROM crashes_cris cris
+             LEFT JOIN crashes unified ON cris.id = unified.id
+        )
```
johnclary, line 9: for `is_manual_geocode`, this needs to be changed to check if the `unified` value is different from the `cris` value—in which case it should be `true`.

database/metadata/databases/default/tables/public_crashes_list_view.yaml
```diff
@@ -197,5 +197,4 @@ select_permissions:
         - rpt_street_sfx
         - crash_timestamp
       filter: {}
-      allow_aggregations: true
```
johnclary, line 200: i think this change is causing the location details page to bomb. i am seeing a failing network request with this error:
```
{
    "errors": [
        {
            "message": "field 'crashes_list_view_aggregate' not found in type: 'query_root'",
            "extensions": {
                "path": "$.selectionSet.crashes_list_view_aggregate",
                "code": "validation-failed"
            }
        }
    ]
}
```

database/metadata/databases/default/tables/public_crashes_list_view.yaml
```diff
@@ -11,7 +11,6 @@ object_relationships:
         remote_table:
           name: recommendations
           schema: public
-array_relationships:
   - name: units
```
johnclary, line 14: this deletion is throwing me off. looks like this is being changed to an `object` relationship, and yet nothing appears to be broken. can you double-check this?

johnclary (changes requested):
Rose, this is looking great! so fun to see the address swap bug disappear 😊 

i ran into a few hopefully-minor issues 🙏

database/metadata/databases/default/tables/public_crashes_list_view.yaml
roseeichelmann: weird, the locations page isnt bombing out for me. and its odd that we only had aggregations allowed for vz-admin and not for any other user types?

database/metadata/databases/default/tables/public_crashes_list_view.yaml
roseeichelmann: thanks for catching this, definitely should be an array relationship. just fixed that and i also updated the relationship to use `crashes_list_view.id` and `units.crash_pk` instead of `cris_crash_id`.

database/migrations/default/1729527603827_recreate_crashes_list_view/up.sql
roseeichelmann: :heavy_check_mark:  thank you!

database/metadata/databases/default/tables/public_crashes_list_view.yaml
johnclary: ah, ok. i tested this with an editor user, and the query is still failing but my editor is not bombing. this is causing the little widgets to not render on non-admin pages. so this may be a preexisting bug:

Screenshot 2024-10-23 at 3 55 41 PM

Screenshot 2024-10-23 at 3 57 36 PM

enabling aggregate query permissions for all users times on the crashes_list_view should fix this 👍

database/metadata/databases/default/tables/public_crashes_list_view.yaml
johnclary: nice—i'm still confused by how this was not breaking things, but it all looks great. i re-tested the crash list export to double-check there and everything looks good. thank you!

johnclary (approved):
i keep clicking around expecting to find something broken, because this has all gone so smoothly. amazing work—ship it! 🙌 🚢 🚢 🚢 🚢 🚢

database/metadata/databases/default/tables/public_crashes_list_view.yaml
```diff
@@ -68,57 +78,57 @@ select_permissions:
...
-        - toll_road_fl
-        - tot_injry_count
-        - traffic_cntl_id
-        - unkn_injry_count
-        - vz_fatality_count
-        - wthr_cond_id
-        - years_of_life_lost
+        - crash_timestamp
       filter: {}
-      allow_aggregations: true
```
johnclary, line 82: @roseeichelmann would you restore these aggregation permissions? then this is good to go

PR #1583, opened 2024-10-22
Add more spatial attributes to crashes

johnclary:
## Associated issues

- https://github.com/cityofaustin/atd-data-tech/issues/19434
- https://github.com/cityofaustin/atd-data-tech/issues/19554

This PR adds new spatial columns to the `crashes` table and updates our spatial attributes trigger to set them. Note that the task of backfilling these values will be handled via https://github.com/cityofaustin/atd-data-tech/issues/19555.

## Testing - local

1. Follow the [`load_agol_layer` tool instructions ](https://github.com/cityofaustin/vision-zero/tree/19434-jc-mor-spatial-attrs/toolbox/load_agol_layer) to load `apd_sectors`, `non_coa_roadways`, `signal_engineer_areas`, and `zip_codes` into your local instance
2. Apply Hasura migrations and metadata 
3. First we'll debug one record by manually editing its position. Run the following commands to see the trigger at work:

```sql
-- enable debugging
set client_min_messages to debug;
-- clear latitude from crashes_cris 
update crashes_cris set latitude = null where id = 160441;
-- reset latitude from crashes_cris
update crashes_cris set latitude = 30.19565639 where id = 160441; 
```

Observe the debug statements, which should look like this:
```sql
DEBUG:  Updating crashes record from CRIS update
DEBUG:  found location: 7DB65B49D7 compared to previous location: <NULL>
DEBUG:  in austin full purpose: t compared to previous: t
DEBUG:  council_district: 2 compared to previous: <NULL>
DEBUG:  engineering_area_id: 3 compared to previous: <NULL>
DEBUG:  signal_engineer_area_id: 2 compared to previous: <NULL>
DEBUG:  zipcode: 78744 compared to previous: <NULL>
DEBUG:  apd_sector_id: 10 compared to previous: <NULL>
DEBUG:  is_non_coa_roadway: f compared to previous: t
```

5. Now clear the `latitude` column and verify that the spatial columns are reset:

```sql
update crashes_cris set latitude = null where id = 160441; 
```

6. ✨ Bonus points ✨ Test this CRIS import. This is a bit tricky, because our spatial attribute trigger only fires when lat/lon values have change. So we're actually going to run the CRIS import script twice. I keep an unzipped extract in my `cris_import/extracts`, so it's always as easy as running this docker command:

```shell
 docker run -it  --rm --env-file local.env --network host -v $PWD:/app cris_import_cris_import ./cris_import.py --csv --skip-unzip
```

7. Great, now we're going to nullify the latitude column of all the crashes you just ran through the import. run these commands: 

```sql
-- remember to disable debug logging!
set client_min_messages to error;

-- nullify latitude for every crash that we just updated
update crashes_cris set latitude = null where updated_at >= CURRENT_DATE
```

8. Nice—now run this CRIS import again. once it completes we can check the results of the trigger.

```sql
--- this should return results!
SELECT
    id,
    is_non_coa_roadway,
    apd_sector_id,
    signal_engineer_area_id,
    zipcode
FROM
    crashes
WHERE
    is_non_coa_roadway = FALSE
    AND apd_sector_id IS NOT NULL
    AND signal_engineer_area_id IS NOT NULL
    AND zipcode IS NOT NULL
    AND council_district IS NOT NULL
    AND in_austin_full_purpose IS TRUE;
```
---

9. Lastly, run the down migrations to make sure they apply as expected. 

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/database/migrations/default/1729616317798_set_more_spatial_attributes/down.sql
+++ b/database/migrations/default/1729616317798_set_more_spatial_attributes/down.sql
@@ -1,4 +1,4 @@
--- revwer to 1728581588450_create_geo_schema
+-- revert to 1728581588450_create_geo_schema
 create or replace function public.crashes_set_spatial_attributes()
 returns trigger
 language plpgsql
--- a/database/migrations/default/1729616317798_set_more_spatial_attributes/down.sql
+++ b/database/migrations/default/1729616317798_set_more_spatial_attributes/down.sql
@@ -1,0 +1,143 @@
+-- revwer to 1728581588450_create_geo_schema
+create or replace function public.crashes_set_spatial_attributes()
+returns trigger
+language plpgsql
+as $function$
+begin
+    if (new.latitude is not null and new.longitude is not null) then
+        -- save lat/lon into geometry col
+        new.position = st_setsrid(st_makepoint(new.longitude, new.latitude), 4326);
+        --
+        -- get location polygon id
+        --
+        if (new.rpt_road_part_id != 2 and upper(ltrim(new.rpt_hwy_num)) in ('35', '183','183A','1','290','71','360','620','45','130')) then
+            -- use level 5 polygon
+            new.location_id = (
+                select
+                    location_id
+                from
+                    public.atd_txdot_locations
+                where
+                    location_group = 2 -- level 5
+                    and st_contains(geometry, new.position)
+                limit 1);
+        else
+            -- use the other polygons
+            new.location_id = (
+                select
+                    location_id
+                from
+                    public.atd_txdot_locations
+                where
+                    location_group = 1 -- not level 5
+                    and st_contains(geometry, new.position)
+                limit 1);
+        end if;
+
+        raise debug 'found location: % compared to previous location: %', new.location_id, old.location_id;
+        --
+        -- check if in austin full purpose jurisdiction
+        --
+        new.in_austin_full_purpose =  st_contains((select geometry from geo.atd_jurisdictions where id = 5), new.position);
+        raise debug 'in austin full purpose: % compared to previous: %', new.in_austin_full_purpose, old.in_austin_full_purpose;
+        --
+        -- get council district
+        --
+        new.council_district = (
+            select
+                council_district
+            from
+                geo.council_districts
+            where
+                st_contains(geometry, new.position)
+            limit 1);
+        raise debug 'council_district: % compared to previous: %', new.council_district, old.council_district;
+        --
+        -- get engineering area
+        --
+        new.engineering_area_id = (
+            select
+                engineering_area_id
+            from
+                geo.engineering_areas
+            where
+                st_contains(geometry, new.position)
+            limit 1);
+        raise debug 'engineering_area_id: % compared to previous: %', new.engineering_area_id, old.engineering_area_id;
+        else
+            raise debug 'setting location id and council district to null';
+            -- nullify position column
+            new.position = null;
+            -- reset location id
+            new.location_id = null;
+            -- use city id to determine full purpose jurisdiction
+            new.in_austin_full_purpose = coalesce(new.rpt_city_id = 22, false);
+            raise debug 'setting in_austin_full_purpose based on city id: %', new.in_austin_full_purpose;
+            -- reset council district
+            new.council_district = null;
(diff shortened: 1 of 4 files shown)


database/migrations/default/1729616317798_set_more_spatial_attributes/up.sql
```diff
@@ -0,0 +1,126 @@
...
+        --
+        new.engineering_area_id = (
+            select
+                engineering_area_id
+            from
+                geo.engineering_areas
+            where
+                st_contains(geometry, new.position)
+            limit 1);
+        raise debug 'engineering_area_id: % compared to previous: %', new.engineering_area_id, old.engineering_area_id;
```
johnclary, line 66: new trigger code starts here 👇

database/migrations/default/1729616317798_set_more_spatial_attributes/down.sql
```diff
@@ -0,0 +1,143 @@
+-- revwer to 1728581588450_create_geo_schema
```
roseeichelmann, line 1: refer? revert?

database/migrations/default/1729616317798_set_more_spatial_attributes/up.sql
```diff
@@ -0,0 +1,134 @@
...
+        --
+        -- get signal engineer area
+        --
+        new.signal_engineer_area_id = (
+            select
+                signal_engineer_area_id
+            from
+                geo.signal_engineer_areas
+            where
+                st_contains(geometry, new.position)
```
roseeichelmann, line 76: i was going to suggest to make this more efficient you could do 
`where (geometry && new.position and (st_contains(geometry, new.position))`
because ive been given that piece of advice in the past, but i just did a little digging and it seems like `st_contains` [automatically does a bounding box](https://postgis.net/docs/ST_Contains.html#:~:text=This%20function%20automatically%20includes%20a%20bounding%20box%20comparison%20that%20makes%20use%20of%20any%20spatial%20indexes%20that%20are%20available%20on%20the%20geometries.%20To%20avoid%20index%20use%2C%20use%20the%20function%20_ST_Contains.) `&&` check before moving on to the other part of the function, so adding an `&&` before calling `st_contains` would be [redundant](https://dba.stackexchange.com/questions/191666/vs-st-intersects-performance#:~:text=The%20reason%20why%20is%20simple%2C%20only%20the%20bounding%20box%20uses%20the%20index.%20That%20means%20it%27ll%20do%20a%20%26%26%20AND%20something%20else.%20And%20you%20can%20see%20that%20with%20%5CdfS%2B%20st_intersects). now im thinking of all the places in our code where we do this redundant double bounding box check. super curious if im understanding this correctly after doing my little rabbit hole dive

roseeichelmann (approved):
this all worked perfectly, i tested everything including the cris import. awesome to have these new gis layers being used in the crashes table now!! :rocket:

database/migrations/default/1729616317798_set_more_spatial_attributes/up.sql
```diff
@@ -0,0 +1,134 @@
...
+            -- reset engineering area
+            new.engineering_area_id = null;
+            -- reset signal eng area
+            new.signal_engineer_area_id = null;
+            -- reset zip code
+            new.zipcode = null;
+            -- reset apd_sector
+            new.apd_sector_id = null;
+            -- reset is_non_coa_roadway
+            new.is_non_coa_roadway = true;
```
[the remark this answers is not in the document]
johnclary: yes. a crash is not on a COA roadway until proven otherwise.

i've gone back and forth on whether or not we should call this field `is_coa_roadway` instead, and default it to `false`. Xavier convinced me to go with `is_non_coa_roadway`, but coming back to this i think the double negative  is too confusing and we should go with `is_coa_roadway` instead.

database/migrations/default/1729616317798_set_more_spatial_attributes/up.sql
johnclary: yes, that's right—if you're already using spatial indexes than that's not necessary. i don't think there are too many places where `&&` is still around except perhaps in the old location triggers. we will clean them up eventually...

PR #1586, opened 2024-10-23
Update OCR narrative ETL to work without _edits layer

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/19534

## Testing

**URL to test:** 
Local

**Steps to test:**
1. Check out this branch, spin up your local db, i used a snapshot from a couple weeks ago. 
2. Run this command to check how many narrative OCRs there are to do:
```sql
select count(*) from view_crash_narratives_ocr_todo;
```
3. Now we will test the OCR script with the new mutation that now updates `crashes` instead of `crashes_edits`. Head to the ./etl/cris_import directory and rebuild the ETL docker image:
```shell
docker compose build
```
4. We need to set our environment to use the prod S3 bucket, because it contains all of the CR3 PDFs that we will process. The best way to do this is to save a copy of your existing .env file as prod.local.env, and set the BUCKET_ENV value to prod in this new file (this PR renames the ENV var to BUCKET_ENV). Make sure your Hasura endpoint is set to you local host.
5. Now it really is time to run the OCR script. Use this run command to run the OCR process with your new env file:
```shell
docker run -it  --rm --env-file prod.local.env --network host -v $PWD:/app cris_import_cris_import ./cr3_ocr_narrative.py --verbose
```
7. Now you can check the the `crashes_change_log_view` and see your OCR update changes there, I also went to crashes that had their OCR updated in the VZE and saw the change in the change log on the front end.

---

#### Ship list

- [ ] Code reviewed


johnclary (approved):
✌️ ~_edits~ 

🚢 🚢 🚢 🚢

PR #1593, opened 2024-11-01
The is_manual_geocode column is not correct when CRIS values are null

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/19723

## Testing

**URL to test:** 
Local

**Steps to test:**
1. Apply the migrations
2. See that this crash now has an is_manual_geocode value of true
```sql
SELECT
    crash.latitude lat_crash,
    crash.longitude lon_crash,
    cris.latitude lat_cris,
    cris.longitude lon_cris,
    crash.is_manual_geocode
FROM
    crashes_list_view crash
    LEFT JOIN crashes_cris cris ON crash.id = cris.id
WHERE
    crash.cris_crash_id = 16318132;
```
3. Test editing the coordinates of a crash and see that the is manual geocode column updates accordingly
4. Test the down migration

---

#### Ship list

- [ ] Code reviewed


database/migrations/default/1730482587632_fix_is_manual_geocode/up.sql
```diff
@@ -0,0 +1,101 @@
+-- public.crashes_list_view source
+
+-- recreating this view to correct the is manual geocode column
+CREATE OR REPLACE VIEW public.crashes_list_view
+AS WITH geocode_status AS (
+         SELECT cris.id,
+            unified.latitude IS NOT NULL AND unified.latitude IS DISTINCT FROM cris.latitude OR unified.longitude IS NOT NULL AND unified.longitude IS DISTINCT FROM cris.longitude AS is_manual_geocode
```
roseeichelmann, line 7: `<>` cannot treat `null` as a comparable value, `IS DISTINCT FROM` can treat `null` as as a comparable value, this was the root of the bug :)

johnclary, line 7: @roseeichelmann this seems to be working, but i think you may need to wrap each `AND` case in parens, just to be 100% sure we're testing what we want to?

```sql
( unified.latitude IS NOT NULL AND unified.latitude IS DISTINCT FROM cris.latitude) 
OR 
(unified.longitude IS NOT NULL AND unified.longitude IS DISTINCT)
```

roseeichelmann: done!

johnclary (approved):
🚢 🚢 🚢 🚢 🚢 🚢 🚢

PR #1597, opened 2024-11-06
Migrate the VZ User and CR3 API into our VZ ECS cluster with an API gateway setup

frankhereford:
## Associated issues

This PR is part of the solution intending to close https://github.com/cityofaustin/atd-data-tech/issues/18154 and it closes https://github.com/cityofaustin/atd-data-tech/issues/18663 as well.

## Moving Parts

This PR moves a fair bit of things around, both in production, staging, and local development. This:

* Deprecate the use of Zappa and λ functions more broadly for the VZ CR3 / User API. 🔥
* Support the use of ECS to host the above API.
* Add various headers to the outgoing responses in pursuit of ISO's request.
* Incorporate the flask app into the local development stack. Start it with `./vision-zero api-up`. It will be available on port `8085` on localhost.
* Add `local-stack-up` and `local-stack-down` to the `vision-zero` orchestration app, to start and stop the 1) DB, 2) `graphql-engine` instance, 3) user / cr3 API. This is intended to be the "normal" command we use to start and stop our local backend stack.
  * Don't forget to update your `zsh` completion functions with the `-s zsh` flag's output -- this makes the orchestration so much easier to use. This applies to `bash` and `tcsh` users as well.
* Add clarity to the environment file of the VZE to allow you to point at local, staging, and production API endpoints, as needed.
* Incorporate the flask app into a docker image, available on docker hub as `atddocker/vz-user-cr3-api`. 
* Include CI to build and restart these services as appropriate.
* Update the flask app to use modern versions of libraries.
* Write out the use of the flask internals to allow for those updates. 

## Testing

* Take a look at the staging and production VZ ECS clusters. 
  * Check out the new, second services running in them. Do you like the names? Anything you'd change?
  * Check out the task definitions for these two new services. Do you like the resource sizing? Naming?

* Spin up your local stack: 
  * Create a `.env` based on `.env.example` using the 1PW entry `Vision Zero zappa_settings` which is a JSON file. Grab the staging values, or reach out to me for a `.env` via `wormhole`. 
  * `./vision-zero local-stack-up` will get your local stack running, which is now 3 docker containers.
  * Fire up your local VZE, and see if it works. Can you list the VZ users?
  * Review the `editor/.env`. Do you see options on how to adjust what flask app your local VZE points at?
  * ‼️ Aim your local VZE at the various endpoints. Everything work? This is testing the actual flask app itself as installed in the new AWS infrastructure. This is the most important thing to test.

* Review the changes to the CI:
  * These are hard to test per se, so give them the eyeball test, and we can confirm they work at merger. 
  * Review the images that are on docker hub; they were built by the CI.
  * Do you approve of the contents of those images? You can install `dive` with homebrew to make it easy to inspect these file manifests.

* ISO's concerns
  * Review my [header choices](https://github.com/cityofaustin/vision-zero/blob/dockerize-api/api/server.py#L73-L103) related to the ISO's requests. Do you agree? Change anything?
  * Consider my wide-open CORS headers. OK with you?

## Documentation
  * This is forth-coming, and will be written in preparation for my skill share next week on the infra side of this work. We can spin it off into an issue if needed, but I want to call it out of scope on this particular issue.

---

#### Ship list

- [x] Check migrations for any conflicts with latest migrations in `main` branch
- [ ] Code reviewed


editor/.env
frankhereford, line 1: While these are not technically secrets, I am not a fan of this file being in the repo. I would support an issue to remote it to spin off this work.

.github/workflows/build_docker_images.yml
```diff
@@ -80,3 +81,39 @@ jobs:
...
+        run: |
+          if [ "${{ github.ref }}" == "refs/heads/production" ]; then
+            echo "ECS_CLUSTER=atd-vz-data-cluster-production" >> $GITHUB_ENV
+            echo "ECS_SERVICE=arn:aws:ecs:us-east-1:295525487728:service/atd-vz-data-cluster-production/cr3-and-user-mgmt-api" >> $GITHUB_ENV
+          else
+            echo "ECS_CLUSTER=atd-vz-data-cluster-staging" >> $GITHUB_ENV
+            echo "ECS_SERVICE=arn:aws:ecs:us-east-1:295525487728:service/atd-vz-data-cluster-staging/cr3-and-user-mgmt-api" >> $GITHUB_ENV
+          fi
+
+      - name: Restart ECS service
```
johnclary, line 106: it is very cool to see this CI  here!!

two minor things i noticed with the current ECS deployment:
1. It appears we have a test cluster, `vz-api-dev` that can be removed

Screenshot 2024-11-07 at 12 28 13 PM

2. The current task that's deployed does not appear to be using the `atddocker/vz-user-cr3-api` image
Screenshot 2024-11-07 at 12 30 17 PM

johnclary (changes requested):
This looks really, really great. I am so happy to see this API moving out of zappa/lambda and being readily available in the local development stack!

I am able to download CR3s through my local API, which is very awesome. I have not, however, been able to list users. I'm hitting the error:

```
{
    "error": "Bad Request",
    "errorCode": "Bearer",
    "message": "Bad HTTP authentication header format",
    "statusCode": 400
}
```

It's a bit surprising, because I can see my request going out with a `Bearer` token as expected. I assume this is some issue with my environment, but I wanted to double-check with you, Frank, if you have any idea where i might have goofed this.

Other than that, my feedback is mostly related to documentation and deployment. Thanks again, Frank!

api/docker-compose.yaml
```diff
@@ -0,0 +1,12 @@
...
+  api:
+    build:
+      context: .
+    ports:
+      - "5000:5000"
+    environment:
+      - FLASK_ENV=production
+    volumes:
+      - .:/usr/src/app
+    command: flask run --host=0.0.0.0 --port=5000
```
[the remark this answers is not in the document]
mddilley: I am also curious if we should take another look at our pattern for deploying Flask apps w/o Lambda wrapping them. This would affect citybase, the EAZ analysis tool, this service, and Moped API in the future.

Here is the specific warning that I saw in the EAZ pr `WARNING: This is a development server. Do not use it in a production deployment. Use a production WSGI server instead.`. I also suspect that Lambda covered us here in the past.

I didn't have time to spin up the local API to see if we see the same warning but it could be good to check and make sure we aren't missing something.

mddilley (commented):
Ran out of time today, but I will pick back up on this tomorrow. Code looks great so far! I just need to test the API with the app.

api/requirements.txt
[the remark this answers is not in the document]
frankhereford: Yea, good call. I like that -- no change to the behavior but a record of some known working state. c8db6f9d5a8a37d29ae2b3a3f501e3f7c6ec2bbb

api/requirements.txt
```diff
@@ -0,0 +1,8 @@
+# pinning of major versions, allowing patch and minor updates
+
+boto3~=1.35
+Flask~=3.0
+python-dotenv~=1.0
+requests~=2.32
+python-jose~=3.0
+Flask-Cors~=3.0
```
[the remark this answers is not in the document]
frankhereford: Not at all; I'm not sure how I missed that one. 😅 https://github.com/cityofaustin/vision-zero/pull/1597/commits/83cf1ea14f0111634cbb65884e8ae284fc889633

api/docker-compose.yaml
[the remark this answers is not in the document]
mddilley: As an update, I attached to the API container, and I see the same warning.

api/server.py
```diff
@@ -67,6 +68,49 @@ def get_api_token():
 APP = Flask(__name__)
 
 
+# Add the appropriate security headers to all responses
+@APP.after_request
+def add_custom_headers(response):
```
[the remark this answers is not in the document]
frankhereford: Sure, great question. Here are some resources: 

* [OWASP Secure Headers Project](https://owasp.org/www-project-secure-headers/): This provides opinions on what should be set.
* [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers#security): This provides background on each header and its options.
* [Security Headers](https://securityheaders.com/): Site scanner for headers. For example, [our current API](https://securityheaders.com/?q=https://atd-vz-api.austinmobility.io/user/list_users?page%3D0%26per_page%3D50&followRedirects=on) & my [proposal's header's](https://securityheaders.com/?q=https://vision-zero-cr3-user-api.austinmobility.io&followRedirects=on).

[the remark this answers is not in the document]
frankhereford: As an aside, I found that I had left my `frankinaustin` docker images in the task definitions while working this and I fixed that and switched over to the `atddocker` images. I also enhanced the health check page in a way that makes it real easy to tell if you're looking at a just-started instance or one that's long running. You can see the new message [here](https://vision-zero-cr3-user-api-staging.austinmobility.io/).

api/server.py
```diff
@@ -67,6 +68,49 @@ def get_api_token():
...
+
+    # X-XSS-Protection to prevent reflected XSS attacks (deprecated in modern browsers)
+    response.headers["X-XSS-Protection"] = "1; mode=block"
+
+    # Expect-CT to enforce Certificate Transparency
+    response.headers["Expect-CT"] = "max-age=86400, enforce"
+
+    # Access-Control-Allow-Origin for CORS
+    response.headers["Access-Control-Allow-Origin"] = "*"
+    response.headers["Access-Control-Allow-Methods"] = "GET, POST, PUT, DELETE, OPTIONS"
```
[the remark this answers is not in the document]
frankhereford: I don't think so, but I didn't remove any from the list because I didn't expect my browser to be `OPTIONS`ing this endpoint, but it did.  I will par it down to `GET`, `POST`, and `OPTIONS` -- your suggestion is smart. 65d5a69fdbb50039d6cf3a86ba6d39a1a0d38f5f

editor/.env
```diff
@@ -1,9 +1,20 @@
 # REACT_APP_HASURA_ENDPOINT="https://vzd-staging.austinmobility.io/v1/graphql"
 REACT_APP_HASURA_ENDPOINT="http://localhost:8084/v1/graphql"
 REACT_APP_AUTH0_DOMAIN=atd-datatech.auth0.com
```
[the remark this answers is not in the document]
frankhereford: You bet. LMK if I didn't get it where it should be. 9e5d46f2d6849c95407125e32b119622b955c680

vision-zero
```diff
@@ -329,6 +341,37 @@ def graphqlEngineUp(args):
...
+            "-f",
+            "docker-compose.yml",
+            "stop",
+            "cr3-user-api",
+        ]
+    )
+    subprocess.run(["docker", "ps"])
+    print("🚀  CR3 / User flask API stopped.")
+
+
```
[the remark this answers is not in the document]
frankhereford: Great idea, thank you. I've added both pairs of new commands. 443b0c84a32a8546e98440a8309227acd700b813

api/README.md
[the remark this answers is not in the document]
frankhereford: Yea, it was mega-out-of-date. I have blasted it away entirely and rewritten a few pertinent parts. I think this covers everything, even though it's a shorter document. The old README went down some rabbit holes.

35d3d055d06ddd9bc796236117a1996e6dc16abc

api/docker-compose.yaml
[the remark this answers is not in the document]
frankhereford: We should, 100%. In this case, we don't push enough traffic onto the service to care, but we really should rethink this practice. The warning is right there when you start it.

We are just waiting for someone to DDOS this endpoint into oblivion as it were. But, I do think the ECS micro instance we run this on will be what breaks way before the service, but the point still stands. I would be so excited to get an issue to sort these out and develop a deployment pattern to bring all of our flask apps up to snuff.

.github/workflows/build_docker_images.yml
frankhereford: Thanks John! I have zapped the dev cluster where I cut my teeth on all this, and I also have updated the task definitions to use the latest and greatest builds on the `atddocker` account.

frankhereford:
@johnclary wrote:

> I am able to download CR3s through my local API, which is very awesome. I have not, however, been able to list users. I'm hitting the error

John, I think you are bumping into the same head-scratcher that I was, and which @chiaberry ultimately rescued me from. Can you confirm that you `REACT_APP_AUTH0_CLIENT_ID` in your editor's `.env` file matches the `CLIENT_ID` in your API environment? These values are the client id's for staging and production as found in auth0 (but not the API app in auth0).

There are few phrases I find more annoying, but I guess it's true: This does work on my macbook, so I think it may be your configuration.

.github/workflows/build_docker_images.yml
```diff
@@ -21,7 +20,7 @@ on:
       - "etl/afd_ems_import/**"
       - "etl/cris_import/**"
       - "etl/socrata_export/**"
-  # Allows you to run this workflow manually from the Actions tab
+      - "api/**"
```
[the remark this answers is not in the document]
frankhereford: This action shouldn't run on pull requests! You're so exactly correct - thank you! 2ca1a56a5d5299840614ddbd68ce92d1e1df5294

mddilley (approved):
I tested the API locally 🚀, and I am able to fetch user data and get CR3s as well. Sooo many pain points allieviated with moving this away from the original deployment! 🎉 Just really nice to spin up all the core services of this app locally with one command and get going! Also helps our AWS tidying with the zappa-created resources cleaned up after this goes live too. 🧹✨

api/server.py
```diff
@@ -281,8 +281,35 @@ def decorated(*args, **kwargs):
...
+            0
+        ]  # Remove microseconds for a terse format
+
+    response = (
+        "CR3 Download API - Health Check - Available @ %s - System Uptime: %s - Process Uptime: %s"
+        % (
+            now.strftime("%Y-%m-%d %H:%M:%S"),
+            uptime_str,
+            proc_uptime_str,
+        )
```
johnclary, line 312: cool—i had no idea how one would go about showing "uptime". do we have any concern about exposing this method publicly? would it be less scary if the health check just printed the current time and said everything's ok? i guess i'm just double-checking that exposing system calls on a public route is fine 😅

api/.env.example
```diff
@@ -1,8 +1,10 @@
...
 CLIENT_ID="2qbqz2sf8L9idBOwn0d5YA9efNgQbL7c"
-AWS_DEFALUT_REGION="us-east-1"
+AWS_DEFAULT_REGION="us-east-1"
 AWS_S3_CR3_LOCATION="staging/path_to_files"
 AWS_S3_BUCKET="your_bucket_name_here"
 AWS_S3_KEY="your_s3_key_here"
-AWS_S3_SECRET="your_s3_secret_here"
\ No newline at end of file
+AWS_S3_SECRET="your_s3_secret_here"
+API_CLIENTID=
```
johnclary, line 9: i found my problem! `API_CLIENTID` is wrong. it should be `API_CLIENT_ID` here

johnclary (changes requested):
I have the use API working as well now—incredible!

Everything looks so good, I had just a few small pieces of feedback.

As an aside, I am going to open an issue to clean up all of the references to `zappa_settings` in this repo, since there are many bc we use the settings file for our migration helper.

api/.env.example
```diff
@@ -1,8 +1,10 @@
 API_ENVIRONMENT="STAGING"
 AUTH0_DOMAIN="atd-datatech.auth0.com"
 CLIENT_ID="2qbqz2sf8L9idBOwn0d5YA9efNgQbL7c"
-AWS_DEFALUT_REGION="us-east-1"
+AWS_DEFAULT_REGION="us-east-1"
```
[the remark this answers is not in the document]
chiaberry: same!!

docker-compose.yml
```diff
@@ -42,3 +42,13 @@ services:
       - ./database/snapshots:/snapshots
     ports:
       - 5432:5432
+
+  cr3-user-api:
+    build: api
+    # image: atddocker/vz-user-cr3-api
```
chiaberry, line 48: is this here as an option for us to uncomment when running locally? or will it be uncommented before merging

api/README.md
johnclary, line 1: Frank, would you mind mentioning the production and staging endpoints here, as well as the healthcheck?

To what extent should we mention the API gateway here? Seeing how it's a crucial piece in making the app work, I think it should at least get mention here, and we can link to deeper technical docs if those feel outside the scope of this particular flask app.

api/server.py
frankhereford: I don't think it's any harm in showing the uptime, but I'm also happy to take it out. It was very handing knowing when a new version of the docker container had been fired up, so I thought it might want to stay around.

There is not time when we believe our service to go from secure to insecure, so exposing the length of time doesn't seem to have any downside, but again, no harm in taking it out either.

api/README.md
frankhereford: Thanks for this great suggestion John! 4bced22ba3d6215ff27323fb2a89323755f43bc4

api/server.py
```diff
@@ -67,6 +68,49 @@ def get_api_token():
...
+    response.headers["Referrer-Policy"] = "no-referrer-when-downgrade"
+
+    # Permissions-Policy to restrict browser feature access
+    response.headers["Permissions-Policy"] = "geolocation=(), microphone=(), camera=()"
+
+    # X-XSS-Protection to prevent reflected XSS attacks (deprecated in modern browsers)
+    response.headers["X-XSS-Protection"] = "1; mode=block"
+
+    # Expect-CT to enforce Certificate Transparency
+    response.headers["Expect-CT"] = "max-age=86400, enforce"
```
chiaberry, line 103: I'm glad you included these comments for these headers, geolocation and microphone permissions never even crossed my mind

api/server.py
```diff
@@ -199,24 +243,24 @@ def decorated(*args, **kwargs):
...
                     },
                     401,
                 )
-            except Exception:
+            except Exception as e:
                 raise AuthError(
                     {
                         "code": "invalid_header",
-                        "description": "Unable to parse authentication" " token.",
+                        "description": "Unable to parse authentication token.",
```
chiaberry, line 258: what do you think about also logging what `e` is? In case it has useful details for debugging.

chiaberry: I guess I am wondering if there could be an Exception that isnt necessarily an AuthError, or related to the token

api/server.py
chiaberry: > when a new version of the docker container had been fired up,

Mike brought up a good point in the skillshare, if theres a way to include what version is running. Future enhancement for sure

chiaberry (approved):
I left a note about error message logging, and have to admit that I didnt run this locally. But approving based on skillshare demo

Looking forward to seeing the documentation in gitbook, thank you in advance for writing it all up!

api/server.py
frankhereford: Totally - that's a great suggestion to include the exception in the exception's error message! b475a90ba028d0f232438fabc017f1cedba2b475

api/server.py
```diff
@@ -199,24 +243,24 @@ def decorated(*args, **kwargs):
...
                     {"code": "token_expired", "description": "token is expired"}, 401
                 )
-            except jwt.JWTClaimsError:
+            except jwt.JWTClaimsError as e:
                 raise AuthError(
                     {
                         "code": "invalid_claims",
-                        "description": "incorrect claims,"
-                        " please check the audience and issuer",
+                        "description": "incorrect claims, please check the audience and issuer",
```
chiaberry, line 250: could this also include e, or is always this description when a JWTClaimsError is thrown?

frankhereford: I thought that one, which only is thrown on a `jwt.JWTClaimsError`exception, would be specific enough - but we can add it if you think it's helpful.

johnclary (approved):
🚢 🚢 🚢 🚢 🚢  thanks again, Frank!!!

api/server.py
chiaberry: I couldn't remember if this was the error we were debugging together, where the error message didnt quite line up exactly with this description

PR #1598, opened 2024-11-07
Rename `is_non_coa_roadway` to `is_coa_roadway`

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/19794

Once deployed, we will need to batch update crash records to apply this new logic.

## Testing

**URL to test:** Local

1. Start your stack
3. In the VZE, find a crash that is located on a collector street in Austin. Edit its location so that it is located on I-35 in central austin. Verify in the record history that the crash has been updated with `is_coa_roadway` = `false`.
4. Edit the same crash by moving to a collector street, and observe `is_coa_roadway` flip back to `true`.
5. Edit the same crash by setting the **Private drive** flag to **Yes**, and notice `is_coa_roadway` back to `false`.
6. Lastly, edit the crash so that it is NOT on a private drive and not inside the Austin Full Purpose jurisdiction. Then edit the "City" field to be something other than Austin and verify that `is_coa_roadway` is now `false`.
7. Apply the down migration

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/database/migrations/default/1731003209538_rename_is_coa_roadway/down.sql
+++ b/database/migrations/default/1731003209538_rename_is_coa_roadway/down.sql
@@ -1,0 +1,148 @@
+-- revert to migrations/default/1729529581000_set_more_spatial_attributes
+alter table crashes rename column is_coa_roadway to is_non_coa_roadway;
+alter table crashes alter column is_non_coa_roadway set default true;
+
+CREATE OR REPLACE FUNCTION public.crashes_set_spatial_attributes()
+ RETURNS trigger
+ LANGUAGE plpgsql
+AS $function$
+begin
+    if (new.latitude is not null and new.longitude is not null) then
+        -- save lat/lon into geometry col
+        new.position = st_setsrid(st_makepoint(new.longitude, new.latitude), 4326);
+        --
+        -- get location polygon id
+        --
+        if (new.rpt_road_part_id != 2 and upper(ltrim(new.rpt_hwy_num)) in ('35', '183','183A','1','290','71','360','620','45','130')) then
+            -- use level 5 polygon
+            new.location_id = (
+                select
+                    location_id
+                from
+                    public.atd_txdot_locations
+                where
+                    location_group = 2 -- level 5
+                    and st_contains(geometry, new.position)
+                limit 1);
+        else
+            -- use the other polygons
+            new.location_id = (
+                select
+                    location_id
+                from
+                    public.atd_txdot_locations
+                where
+                    location_group = 1 -- not level 5
+                    and st_contains(geometry, new.position)
+                limit 1);
+        end if;
+
+        raise debug 'found location: % compared to previous location: %', new.location_id, old.location_id;
+        --
+        -- check if in austin full purpose jurisdiction
+        --
+        new.in_austin_full_purpose =  st_contains((select geometry from geo.atd_jurisdictions where id = 5), new.position);
+        raise debug 'in austin full purpose: % compared to previous: %', new.in_austin_full_purpose, old.in_austin_full_purpose;
+        --
+        -- get council district
+        --
+        new.council_district = (
+            select
+                council_district
+            from
+                geo.council_districts
+            where
+                st_contains(geometry, new.position)
+            limit 1);
+        raise debug 'council_district: % compared to previous: %', new.council_district, old.council_district;
+        --
+        -- get engineering area
+        --
+        new.engineering_area_id = (
+            select
+                engineering_area_id
+            from
+                geo.engineering_areas
+            where
+                st_contains(geometry, new.position)
+            limit 1);
+        raise debug 'engineering_area_id: % compared to previous: %', new.engineering_area_id, old.engineering_area_id;
+        --
+        -- get signal engineer area
+        --
+        new.signal_engineer_area_id = (
+            select
+                signal_engineer_area_id
+            from
+                geo.signal_engineer_areas
+            where
+                st_contains(geometry, new.position)
+            limit 1);
+        raise debug 'signal_engineer_area_id: % compared to previous: %', new.signal_engineer_area_id, old.signal_engineer_area_id;
+        --
+        -- get zip code
+        --
+        new.zipcode = (
+            select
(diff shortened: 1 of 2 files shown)


database/migrations/default/1731003209538_rename_is_coa_roadway/up.sql
```diff
@@ -0,0 +1,151 @@
...
+            where
+                st_contains(geometry, new.position)
+            limit 1);
+        raise debug 'apd_sector_id: % compared to previous: %', new.apd_sector_id, old.apd_sector_id;
+        --
+        -- check if is_coa_roadway
+        --
+        if ((new.in_austin_full_purpose or new.rpt_city_id = 22) and not new.private_dr_fl) then
+            new.is_coa_roadway = not st_contains((select geometry from geo.non_coa_roadways), new.position);
+            raise debug 'is_coa_roadway: % compared to previous: %', new.is_coa_roadway, old.is_coa_roadway;
```
johnclary, line 112: here is the updated business logic

database/migrations/default/1731003209538_rename_is_coa_roadway/up.sql
```diff
@@ -0,0 +1,151 @@
...
+
+drop trigger if exists crashes_set_spatial_attributes_on_update on crashes;
+
+create trigger crashes_set_spatial_attributes_on_update before update on crashes for each row when
+(
+    new.latitude is distinct from old.latitude
+    or new.longitude is distinct from old.longitude
+    or new.rpt_road_part_id is distinct from old.rpt_road_part_id
+    or new.rpt_hwy_num is distinct from old.rpt_hwy_num
+    or new.private_dr_fl is distinct from old.private_dr_fl
```
johnclary, line 149: we now need to trigger this function when `private_dr_fl` is edited ✅

database/migrations/default/1731003209538_rename_is_coa_roadway/up.sql
mddilley: @johnclary I might be testing this incorrectly, but I'm getting hung up on what is supposed to happen when toggling the private drive flag from `false` to `true` since the code after `then` won't run when `new.private_dr_fl` is `true` here. Should this run in either case and recalculate `is_coa_roadway`?

Could be that my Friday brain is struggling. 😅

johnclary: great catch, Mike—thank you! i just pushed two commits:
- add the missing `else` case to this block of code
- add `rpt_city_id` to the conditions that cause this trigger function to run, so that editing the crash's city actually has an affect

mddilley: awesome - i am now seeing the trigger fire in step 5. 🚀

mddilley (approved):
Looks great! I followed the test steps, and I looked over that last couple commits. It all makes sense to me and works! 🛳️ 🚀

chiaberry (approved):
had fun toggling the flag back and forth, and is_coa_roadway = true is way easier to wrap my head around than is_non_coa_roadway = false 

This step definitely checked out 
> Lastly, edit the crash so that it is NOT on a private drive and not inside the Austin Full Purpose jurisdiction. Then edit the "City" field to be something other than Austin and verify that is_coa_roadway is now false.

but I was wondering how come moving the crash to outside the austin full purpose jurisdiction isnt enough to trigger is_coa_roadway to be false.

database/migrations/default/1731003209538_rename_is_coa_roadway/up.sql
```diff
@@ -0,0 +1,154 @@
+alter table crashes rename column is_non_coa_roadway to is_coa_roadway;
+alter table crashes alter column is_coa_roadway set default false;
+
+comment on column public.crashes.is_coa_roadway is 'If the crash location occured within the City of Austin, not on a private drive, and not intersecting with the non_coa_roadways layer. Set via trigger.';
```
roseeichelmann, line 4: good comment :100:

database/migrations/default/1731003209538_rename_is_coa_roadway/down.sql
```diff
@@ -0,0 +1,148 @@
...
+                st_contains(geometry, new.position)
+            limit 1);
+        raise debug 'apd_sector_id: % compared to previous: %', new.apd_sector_id, old.apd_sector_id;
+        --
+        -- check if is_non_coa_roadway
+        --
+        if (new.in_austin_full_purpose or new.rpt_city_id = 22) then
+            new.is_non_coa_roadway = st_contains((select geometry from geo.non_coa_roadways), new.position);
+            raise debug 'is_non_coa_roadway: % compared to previous: %', new.is_non_coa_roadway, old.is_non_coa_roadway;
+        end if;
```
roseeichelmann, line 112: so previously if the city was updated to be outside of austin then no updates were happening to `is_non_coa_roadway` bc there is no `else` here? seems like a bug

roseeichelmann (approved):
test steps worked great. do we still need to backfill/update this col next?

database/migrations/default/1731003209538_rename_is_coa_roadway/down.sql
johnclary: yep—it was!

johnclary:
> I was wondering how come moving the crash to outside the austin full purpose jurisdiction isnt enough to trigger is_coa_roadway to be false.

@chiaberry the trigger also checks the `city_id` column to determine if the crash is Austin. both the spatial location and city ID have to agree that the crash is outside of Austin in order for this column to be `false`.  i have an open question with Xavier to double-check that this is still the desired behavior 🙏

johnclary:
ok, yeah, we removed `rpt_city_id` from the `is_coa_roadway` logic. we're still using this field in the `else` case when we set `in_austin_full_purpose`—Xavier and I agreed that we also want to change this logic, but it's too drastic of a change to address now. I will open a backlog for it. we also discussed reviving our in-house geocoding, which i think is a pretty good idea.

PR #1599, opened 2024-11-11
Crash details page: enable users to collapse the record history card

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/19804

## Testing

**URL to test:** 

https://deploy-preview-1599--atd-vze-staging.netlify.app/#/crashes

**Steps to test:**

The first time you visit a crash page, the Record History card should be collapsed aka closed. 
Click the header to open the card. 
If you refresh or leave and go to a different crash, the card should remain open. 
Clicking again will close the card. 
If you refresh or leave and go to a different crash, the card should remain closed. 

* It wasn't explicitly stated in the issue, but I included the badge with a count of how many entries are in the record history card, to mimic what is in the Units, People and Charges cards
* I considered saving a more generalized "accordion state" object in local storage, with an entry for this Card so that in the future we could store the open/closed state of all the other cards. But I opted to keep it simple

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/src/views/Crashes/CrashChangeLog.js
+++ b/editor/src/views/Crashes/CrashChangeLog.js
@@ -71,16 +71,24 @@ const isNewRecordEvent = change => change.operation_type === "create";
 
 /**
  * The primary UI component which renders the change log with clickable rows
- * can be collapsed
+ * The Card defaults to being "collapsed", the open/closed state is stored in localStorage
  */
 export default function CrashChangeLog({ data }) {
   const [selectedChange, setSelectedChange] = useState(null);
-  const [isOpen, setIsOpen] = useState(false);
+  const [isOpen, setIsOpen] = useState(
+    localStorage.getItem("showHistory") === "true"
+  );
 
   const changes = useChangeLogData(data);
   if (changes.length === 0) {
     return <p>No change history found</p>;
   }
+
+  const toggleCollapseHistory = () => {
+    const nextIsOpen = !isOpen;
+    setIsOpen(nextIsOpen);
+    localStorage.setItem("showHistory", nextIsOpen);
+  };
 
   return (
     <Card>
@@ -89,7 +97,7 @@ export default function CrashChangeLog({ data }) {
           block
           color="link"
           className="text-left m-0 p-0"
-          onClick={() => setIsOpen(!isOpen)}
+          onClick={toggleCollapseHistory}
         >
           Record history
           <Badge color="secondary float-right">{changes.length}</Badge>
--- a/editor/src/views/Crashes/CrashChangeLog.js
+++ b/editor/src/views/Crashes/CrashChangeLog.js
@@ -1,5 +1,13 @@
 import React, { useMemo, useState } from "react";
-import { Card, CardBody, CardHeader, Table } from "reactstrap";
+import {
+  Badge,
+  Button,
+  Card,
+  CardBody,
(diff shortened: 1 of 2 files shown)


editor/src/views/Crashes/CrashChangeLog.js
```diff
@@ -63,57 +71,79 @@ const isNewRecordEvent = change => change.operation_type === "create";
 
 /**
  * The primary UI component which renders the change log with clickable rows
+ * The Card defaults to being "collapsed", the open/closed state is stored in localStorage
  */
 export default function CrashChangeLog({ data }) {
   const [selectedChange, setSelectedChange] = useState(null);
+  const [isOpen, setIsOpen] = useState(
+    localStorage.getItem("showHistory") === "true"
+  );
```
johnclary, line 80: this is great! for the item name, would you mind prefixing it with the `appCodeName` like we do elsewhere? and maybe be more specific with the name, like `showCrashHistory`?

editor/src/views/Crashes/CrashChangeLog.js
```diff
@@ -63,57 +71,79 @@ const isNewRecordEvent = change => change.operation_type === "create";
...
-                <td>{formatDateTimeString(change.created_at)}</td>
+      <CardHeader>
+        <Button
+          block
+          color="link"
+          className="text-left m-0 p-0"
+          onClick={toggleCollapseHistory}
+        >
+          Record history
+          <Badge color="secondary float-right">{changes.length}</Badge>
```
johnclary, line 103: very nice touch 🙏 

how about we go a step further and add an icon and the same font size as the other card headers? i think a clock icon (`fa-clock-o`) works unless you have have a different FA v4 icon you prefer? https://fontawesome.com/v4/icons/

johnclary (changes requested):
Looks great and works great—thanks for picking up this little enhancement! I have requested a few minor changes 🙏

editor/src/views/Crashes/CrashChangeLog.js
chiaberry: updated. 
~~every release, the app defaults to not showing the history card until the user clicks it?~~
never mind, I see that the app name needs to be manually updated, and its not pulled from the release notes

I am assuming then that that app name is only updated if there are breaking changes to, say, the grid filters. But that would mean that if there are breaking changes to that config, everything that is stored in local storage gets reset to default, regardless of its affected by other local storage changes?

editor/src/views/Crashes/CrashChangeLog.js
chiaberry: I decided instead of clock to use `fa-history`

roseeichelmann (approved):
looks and works so great, i like the extra touch of the record count

editor/src/views/Crashes/CrashChangeLog.js
johnclary: yes, that's right. we only change the app code name if we make a breaking change to storage, and it would have the effect of resetting everyone's storage.

editor/src/views/Crashes/CrashChangeLog.js
johnclary: 🎯

johnclary (approved):
Looks great!! 🚢 🚢 🚢 🚢 🚢 🚢

PR #1603, opened 2024-11-15
Upgrade VZE to NextJS

johnclary:
## Associated issues

- https://github.com/cityofaustin/atd-data-tech/issues/19576

Here is the working prototype of the VZE in NextJS! I am envisioning that folks spend time with the code and add their questions/comments via review, and from there we convene a meeting to discuss next steps and scope issues.

## Testing

1. Star your local stack and apply migrations and metadata
2. Start the nextjs app: see [the README](https://github.com/cityofaustin/vision-zero/blob/nextjs/app/README.md) for instructions.
3. If you've tested this app previously, you may need to remove the old version of `crashesListViewQueryConfig` from your `localStorage`
4. What to test: you should be able to use the app without breaking it! Moreover, the current list of working features is in the readme, [here](https://github.com/cityofaustin/vision-zero/blob/nextjs/app/README.md#working-features).
5. Code review: all feedback is welcome, and especially related to the libraries i've chosen and overall project structure.
6. I have quite a few doubts and questions about the designing of the types, and a very important todo (i think) is to add in Zod schema validation to generate our core types and wrap all our API requests. 

---

#### Ship list

- [ ] Code reviewed


app/app/layout.tsx
```diff
@@ -0,0 +1,25 @@
...
+ * 
+ * todo: metadata
+ */
+export default function RootLayout({
+  children,
+}: {
+  children: React.ReactNode;
+}) {
+  return (
+    // <Html lang="en" style={{fontSize: "14px"}} data-bs-theme="dark">
```
johnclary, line 16: i left this here bc the dark them is glorious to dev on. would be fun to spin off an issue to make a UI toggle for it 🌚

app/app/layout.tsx
```diff
@@ -0,0 +1,25 @@
...
+    <html lang="en" style={{ fontSize: "14px" }}>
```
johnclary, line 17: the default bootstrap font size is `16`, and i think the ACC designs go down to `12`, maybe smaller. it's an accessibility issue that needs more consideration 👀

app/components/ChangeDetailHeader.tsx
johnclary, line 1: Most of the change log was copied verbatim from VZE, with the exception that it now renders even if there are no change records at all (fixing a weird quirk of VZE local dev where the change log is hidden)

app/components/ChangeLog.tsx
```diff
@@ -0,0 +1,135 @@
...
+} from "@/types/types";
+
+const KEYS_TO_IGNORE = ["updated_at", "updated_by", "position"];
+
+/**
+ * Return an array of values that are different between the `old` object and `new` object
+ * Each object in the array takes the format
+ * { field: <property name>, old: <old value>, new: <new value> }
+ */
+const getDiffArray = <T extends Record<string, unknown>>(
```
johnclary, line 19: getting this typed correctly was correctly, even though the result looks simple. these kind of data-munging tasks are what i've found to be the most onerous to convert to typescript.

johnclary: also, this is the first of many declarations of `Record<string, unknown>>` because the `ColBaseDef` extends from it. 

i _think_ it's right to extend from this type, but because it is used so often we should probably give it a name and export it. something like `<RecordWithStringKeys>`. that's all it is: an object with keys that are always strings.

this will make more sense after reading about the `Record` utility type: https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type  :)

app/components/ChangeLog.tsx
```diff
@@ -0,0 +1,135 @@
...
+    a.field < b.field ? -1 : a.field > b.field ? 1 : 0
+  );
+
+  return diffArray;
+};
+/**
+ * Prettify the user name if it is `cris`
+ */
+const formatUserName = (userName: string): string =>
+  userName === "cris" ? "TxDOT CRIS" : userName;
```
johnclary, line 45: i just want to point out that this function has a return type defined (`string`). typescript will infer the return type based on what your code returns, but as a best practice i think we want to strive for explicitly naming the return type. 

there are some spots where i failed to do this and i'm keeping an eye out for them.

app/contexts/Auth.tsx
johnclary, line 1: Auth implemented fully client-side, like the existing VZE. 

basically, the flow is that a user logs into Auth0, the JWT is stored in localstorage, and the app uses that JWT to make calls to the Hasura API. Hasura validates the JWT and permissions. So our app is just a passthrough for the authentication that manages the task of getting a token and refreshing it.

Auth0 does have a server-side lib, [nextjs-auth0](https://github.com/auth0/nextjs-auth0), which is specific to the app-router. The big difference with that approach is that our own nextjs server (Netlify or Vercel) actually validates the JWT that is exchanged. To make that possible, we need to store our JWT secret in the app environment. (whereas the client-side setup gets by on the publicly-shareable env vars).

If we want to authenticate a server-side function, we'll need to adopt the `nextjs-auth0` package. For example, if we wanted to replace the CR3 API. But for our current needs there is no benefit to using the nextjs Auth0 package, and in fact it adds the overhead of needing to manage a very sensitive secret in our local setup.

That's my digest of the tradeoffs, at least.

app/contexts/Auth.tsx
```diff
@@ -0,0 +1,29 @@
+"use client";
+import { Auth0Provider } from "@auth0/auth0-react";
```
johnclary, line 2: React context is a client-only feature, which is where the change in mental model really hit home for me.

if want to use a 3rd-party context, you need to wrap it in "use client", unless the author does it for you. [Here are the nextjs docs](https://nextjs.org/docs/14/app/building-your-application/rendering/composition-patterns#using-third-party-packages-and-providers) that discuss that.

app/env_template
johnclary, line 1: our `env` files are no longer committed to the repo 🎉

app/next.config.mjs
```diff
@@ -0,0 +1,15 @@
...
+  reactStrictMode: true,
+  async redirects() {
+    return [
+      {
+        source: "/",
+        destination: "/crashes",
+        permanent: true,
+      },
+    ];
+  },
```
johnclary, line 12: this file was generated by `create next app` and i added this redirect.

app/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "private": true,
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "lint": "next lint"
+  },
+  "dependencies": {
+    "@auth0/auth0-react": "^2.2.4",
+    "@types/lodash": "^4.17.13",
```
johnclary, line 13: the only lodash function we're using at the moment is `isEqual`, and as far as i can tell the bundler is doing a great job tree-shaking it.

i don't know what state the `lodash` project is in. i think v5 has been in the works for a very long time...

app/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "date-fns": "^4.1.0",
+    "graphql": "^16.9.0",
+    "graphql-request": "^7.1.0",
+    "immer": "^10.1.1",
+    "lodash": "^4.17.21",
+    "mapbox-gl": "^3.7.0",
+    "next": "14.2.13",
+    "react": "^18",
+    "react-bootstrap": "^2.10.4",
+    "react-datepicker": "^7.5.0",
```
johnclary, line 25: this datepicker gets the job done, but it looks really dated. i have not found a framework-agnostic alternative but didn't look very hard.

app/styles/global.scss
johnclary, line 1: we should be able to accomplish most of our theming here. bootstrap docs are pretty good about how to accomplish this, but getting all of the theme overrides correct will take some work.

app/tsconfig.json
johnclary, line 1: i did not create this file 🤖

app/utils/formatters.ts
```diff
@@ -0,0 +1,24 @@
+import { format, parseISO } from "date-fns";
+
+// const formatCostToDollars = (cost: number | null) =>
```
johnclary, line 3: i left this here because we'll need to render comp costs on the location details pages

app/components/DataCard.tsx
johnclary, line 1: This component replaces the `DataTable` component in VZE.  The only significant difference is that it fetches lookup table values when a lookup field is edited, as opposed to fetching all lookup values when the crash details page loads.

The lookup queries are cached, and in general are quick to load, but we need to test this our power users, who may notices a slight performance degradation.

app/components/ChangeDetailHeader.tsx
johnclary, line 1: Most of the change log was copied verbatim from VZE, with the exception that it now renders even if there are no change records at all (fixing a weird quirk of VZE local dev where the change log is hidden)

app/env_template
johnclary, line 1: our `env` files are no longer committed to the repo 🎉

app/styles/global.scss
johnclary, line 1: we should be able to accomplish most of our theming here. bootstrap docs are pretty good about how to accomplish this, but getting all of the sass variable overrides correct will take some work.

app/tsconfig.json
johnclary, line 1: i did not create this file 🤖

app/components/DataCard.tsx
```diff
@@ -0,0 +1,127 @@
...
+  mutation: string;
+  isValidating: boolean;
+  title: string;
+  onSaveCallback: () => Promise<void>;
+}
+
+/**
+ * Generic component which renders editable fields in a Card
+ */
+export default function DataCard<T extends Record<string, unknown>>({
```
johnclary, line 27: This component replaces the `DataTable` component in VZE.  The only significant difference is that it fetches lookup table values when a lookup field is toggled into its edit state, as opposed to fetching all lookup values when the crash details page loads.

The lookup queries are cached, and in general are quick to load, but we need to test this with our power users, who may notice a slight performance degradation.

app/types/types.ts
```diff
@@ -0,0 +1,167 @@
...
+  label: string;
+}
+
+export type InputType = "text" | "number" | "yes_no" | "select";
+
+/**
+ * The base definition of a database column referenced by our app — where
+ * <T> should be one of our core record types (Crash, Person, etc)
+ */
+export type ColBaseDef<T extends Record<string, unknown>> = {
```
johnclary, line 21: `ColBaseDef` is the smallest building of our types coming out of the database, and we can use them to build up any time of record: crashes, units, etc.

Also, you can see that I have extended from `Record<string, unknown>>` type, and i do that in many places. I think that's fine and good, but because it is used so often across the codebase we should probably give it a name and export it. something like `<RecordWithStringKeys>`. that's all it is: an object with keys that are always strings.

this will make more sense after reading about the `Record` utility type: https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type  :)

app/types/types.ts
johnclary, line 1: i started collecting our most types here, but have also left component-specific types and interfaces near to their components. i don't feel strongly about how to organize these, but i do think we'll want to split out entity-specific types like crash, unit, etc. into their own files. 

I also think that bringing `zod` into the mix might significantly change how we write out these types, so i didn't want to overthink it.  See my note in `zodExample.ts`.

app/utils/graphql.ts
```diff
@@ -0,0 +1,194 @@
+import { useState, useCallback, useEffect, useMemo } from "react";
+import useSWR, { SWRConfiguration } from "swr";
```
johnclary, line 2: all of our data fetching is managed with [SWR](https://swr.vercel.app/docs/getting-started) and [graphql-request](https://github.com/graffle-js/graffle/tree/graphql-request).

SWR is a Vercel-managed library designed specifically for client-side data fetching, and I really like working with it. 

The Apollo [setup for nextjs app router](https://www.apollographql.com/blog/using-apollo-client-with-next-js-13-releasing-an-official-library-to-support-the-app-router) looks fairly involved, and i haven't messed with it. I'm open to it, but I stuck with SWR since it's familiar to me and is part of the nextjs ecosystem.

app/utils/graphql.ts
```diff
@@ -0,0 +1,194 @@
...
+const DEFAULT_SWR_OPTIONS: SWRConfiguration = {
+  /**
+   * Dont refetch when the page/tab is refocused
+   */
+  revalidateOnFocus: false,
+  /**
+   * Dont refetch on network recon
+   */
+  revalidateOnReconnect: false,
+};
```
johnclary, line 25: these are just my opinions to cut back on unnecessary API calls—SWR really loves to make API calls.  i encourage y'all to check out the various settings available 🧪

app/utils/graphql.ts
```diff
@@ -0,0 +1,194 @@
...
+     * use idToken.__raw to grab the actual JWT. it seems like our setup
+     * may be misconfigured, because accessing .__raw seems like a hack :/
+     *
+     * the typical setup would enable use to use the getAccessTokenSilently()
+     * method, but that doesn't work with the opaque tokens.
+     *
+     * dicussion here: https://community.auth0.com/t/getting-the-jwt-id-token-from-auth0-spa-js/28281/3
+     */
+    const idToken = await getIdTokenClaims();
+    const token = idToken?.__raw || "";
```
johnclary, line 78: this hacky-feeling code is in the old VZE as well. more research needed here.

app/utils/graphql.ts
```diff
@@ -0,0 +1,194 @@
...
+    // todo: <T> passed here is assigned to query results without validation - we need to use a schema validator
+    return fetcher<T>([query, variables, token, hasuraRoleName]);
+  };
+
+  // todo: document falsey query handling
+  const { data, error, isLoading, mutate, isValidating } = useSWR<T>(
+    query ? [query, variables] : null,
+    fetchWithAuth,
+    { ...DEFAULT_SWR_OPTIONS, ...(options || {}) }
+  );
```
johnclary, line 89: here's where we call the actual `useSWR` hook—and a few important things to note:
- the first argument is the value that SWR uses as the cache key. so each combination of query + variables gets cached.
- passing `null` to the first argument is like an off switch, and enables [conditional data fetching](https://swr.vercel.app/docs/conditional-fetching#conditional). this comes in super handy when you don't want to start a request until you have, e.g, a user's email address from the auth hook.

app/utils/queryBuilder.ts
johnclary, line 1: the `queryBuilder` replaces the `graphqlAbstract` in the VZE. basically, it takes a giant config object and spits out a graphql. it puts the work of modifying the query configuration in the hands of UI components.

app/utils/queryBuilder.ts
```diff
@@ -0,0 +1,378 @@
...
+  operator:
+    | "_gte"
+    | "_lte"
+    | "_gt"
+    | "_eq"
+    | "_neq"
+    | "_is_null"
+    | "_ilike"
+    | "_in"
+    | "_nin";
```
johnclary, line 41: hasura supports more operators, but i've only added those that we're actually using. we can add more as needed 👍

app/configs/zodExample.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Hasura api calls in schema validation.
+ *
+ * Basically, we define our core types (crashes, units, etc)
+ * in zod terms, and let zod generate (infer) the
+ * the types for us.
+ * 
+ * At runtime, we would use the zod schema to validate api calls,
+ * and during development we have typechecking in sync with
+ * what zod is validating.
+ */
```
johnclary, line 14: i don't know if the below code makes much sense, but these are my thoughts on how we might approach a tighter integration with schema validation☝️ 

i will find some links to other open source projects where i've seen this pattern.

johnclary: ☝️ this second example does not do runtime type checking at all. it just hopes that the type declarations match what the API returns.

Once we adopt typescript across our project, one of the most likely sources for errors is going to be situations where our API does not return what we're expecting it to (for example, because we forgot to add a field to a query or we changed a field's type from number to string).

The reason we would reach for Zod (or Yup, etc) would be to centralize the error handling in those cases. We can check types before data passes from our API to UI components, and throw helpful error messages when something goes wrong.

app/package.json
johnclary: ah yes, here we go: https://github.com/gpbl/react-day-picker this one is used by [shadcn/ui](https://ui.shadcn.com/docs/components/calendar).

johnclary:
Just pushed a few minor changes i had locally. Will branch from now on 👍

app/configs/zodExample.ts
johnclary: i created https://github.com/cityofaustin/atd-data-tech/issues/20024 for this and am going to do a little spike setting up a proof concept branch for how i think we want to structure this.

johnclary: here is a branch/PR with a working setup for zod-driven typings:
* https://github.com/cityofaustin/vision-zero/pull/1612

sorting this out correctly and figuring out exactly what kind of tooling we want feels like the biggest open question for this app at the moment. I will update that PR with some more specific questions i have. 

Once folks have a chance to catch up on this it will probably be worth scheduling a meeting to discuss :/

app/components/RelatedRecordTable.tsx
```diff
@@ -0,0 +1,56 @@
+import Card from "react-bootstrap/Card";
+import Table from "react-bootstrap/Table";
+import RelatedRecordTableRow from "./RelatedRecordTableRow";
+import { ColDataCardDef } from "@/types/types";
+
+interface RelatedRecordTableProps<T extends Record<string, unknown>> {
```
mddilley, line 6: since i mentioned prop type declarations in an earlier comment, did you decide to use an interface here because of the `Record` type or as a stylistic choice since there are more props in this component? Just curious to learn more.

app/configs/zodExample.ts
mddilley: next up, i plan to look at the zod PoC PR early next week. I know you mentioned [in Slack](https://austininnovation.slack.com/archives/CM9MK950S/p1732558785583419) that it may not be your favorite at the moment. I think another spike into the gql codegen could be cool like you suggested. Especially if we could find a way to be selective about the generated types and what is kept in this codebase. 

I guess we would also need to think about what updating types would look like too if we added another CRIS value in the crash table for example. 🤔

app/package.json
mddilley: as long as we don't go with the one that I picked for VZV! 🙈

app/utils/graphql.ts
mddilley: i think swr is great, and I don't see anything it doesn't cover that we got out of Apollo. imo Apollo client probably gives more than we need, and we also ran into some weird problems with fetching data for csv exports. Actually, lazy querying is something that we'll probably need, but maybe it is covered?

app/utils/graphql.ts
mddilley: I think this may have answered my lazy query questions. 👀

app/utils/queryBuilder.ts
mddilley: i said it in the last meeting, but i like this strategy to build what we need now and extend later if needed and NOT try to cover all of the functionality coming out of the Hasura API. 🙌 It is a nice iteration that builds of our knowledge gained through the first go-around and Moped refactor. Very cool stuff.

mddilley (commented):
This was a fun one to read through and add some thoughts. I'm looking forward to checking out more of these PRs next week. 😎

app/app/layout.tsx
johnclary: Ah, thanks pointing to MUI's approach—interesting that they default to 14px. 

My understanding of the way it works is that `1rem` = whatever the browser default is, unless specified in the `html` tag, in which case `1rem` becomes the specified value.

app/components/ChangeDetailHeader.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { formatDateTime } from "@/utils/formatters";
+
+/**
+ * Header component of the change details table
+ */
+const ChangeDetailHeader = ({
+  selectedChange,
+}: {
+  selectedChange: ChangeLogEntryEnriched;
+}) => {
```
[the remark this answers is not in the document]
johnclary: Ah yes, another place where I have not been consistent. Sometimes declaring a separate props type/interface, and sometimes doing these inline declarations 🙏

app/utils/graphql.ts
johnclary: yeah!

PR #1606, opened 2024-11-18
Add crash-components table to Socrata export (to main)

johnclary:
## Associated issues

This is a copy of https://github.com/cityofaustin/vision-zero/pull/1600, which I foolishly merged into Charlie's branch after Charlie's branch was already merged to main  👍 👍 👍 

* https://github.com/cityofaustin/atd-data-tech/issues/19850

## Testing

**URL to test:** Local

**Steps to test:**

0. Start your local VZ stack

1.  Follow the steps to populate the `moped_component_crashes` in https://github.com/cityofaustin/vision-zero/pull/1596

2. Update your `.env` file in this ETL to include the SOCRATA_DATASET_CRASH_COMPONENTS dataset:

```
SOCRATA_DATASET_CRASH_COMPONENTS=gb78-y2pd
```

3. If you've never run this ETL, build the container: 

```
docker compose build
```

4. Run this ETL with the new `--crash-components` flag

```shell
docker run -it  --rm --env-file .env --network host -v $PWD:/app socrata_export_socrata_export python socrata_export.py --crash-components
```

5. Check out the socrata dataset to verify you see data (linked to in the README—requires sign-in)

---

#### Ship list

- [ ] Check migrations for any conflicts with latest migrations in `main` branch
- [ ] Confirm Hasura role permissions for necessary access
- [ ] Code reviewed
- [ ] Product manager approved

PR #1607, opened 2024-11-18
v2.4.1

johnclary:
## Associated issues
- https://github.com/cityofaustin/vision-zero/pull/1606


chiaberry (approved):

etl/moped_projects/README.md
```diff
@@ -18,6 +18,6 @@ docker build . -t atddocker/vz-moped-join
 ```
 
 ```
-docker run -it --env-file env_file atddocker/vz-moped-join /bin/bash
+docker run -it --rm --network host --env-file env_file atddocker/vz-moped-join /bin/bash
```
frankhereford, line 21: Not a big deal here, but the defaulting to the use of `--network host` defeats the network isolation which is a feature of the docker ecosystem and very generally will lead to port conflicts unless care is taken to make sure none get reused, and this tracking of used ports is maintained as new docker containers are added to the host.

frankhereford (approved):
Looks great 🩹 🚀

etl/moped_projects/README.md
johnclary: the ETL needs to connect to local moped and hasura, so this  is a shortcut to specifying the docker network(s) this container needs to reach.

in this case, i guess the care being taken is the use of the `--rm` flag?

frankhereford: The `--rm` flag means to clean up the container when it exits instead of hanging around to be restarted or cleaned up by some a `--remove-orphans` command. That's how I think of it at least .. an 'autoremove' feature.

johnclary: thanks for pushing for the internal host—i made the readme and env template with this change ✨ 

https://github.com/cityofaustin/vision-zero/commit/0194b64d5dbe359c6bfd40a394b0cc1a99d19dbe

PR #1611, opened 2024-11-19
Crash details: crash injury widgets

johnclary:
## Associated issues

- https://github.com/cityofaustin/atd-data-tech/issues/19966

This creates the basic working injury widgets, taking inspo from the ACC designs. I kept styling super simple: black badges if there are `0` injuries, primary (blue) badges if there are injuries. I also played around with only showing badges that had a non-zero value, which I kinda liked but didn't want to overthink this feature.

Feedback on design is welcome, and I'll plan on opening a follow-up issue to further refine the look and feel of the badges and the crash header in general.

Screenshot 2024-11-20 at 12 04 46 PM

## Testing

Navigate to various crashes and check that the badges are rendering correctly. This would be easier to test if we had the people records on the screen, but that won't happen until we do https://github.com/cityofaustin/atd-data-tech/issues/19965.

---

#### Ship list

- [ ] Code reviewed


app/queries/crash.ts
```diff
@@ -80,12 +80,14 @@ export const GET_CRASH = gql`
...
+        nonincap_injry_count
+        poss_injry_count
+        unkn_injry_count
+        tot_injry_count
         years_of_life_lost
         est_comp_cost_crash_based
         est_total_person_comp_cost
         crash_injry_sev_id
-        nonincap_injry_count
-        sus_serious_injry_count
```
chiaberry, line 88: wondering why this was removed, but i see that this is just a duplicate entry

chiaberry:
I agree with your refinement of the ACC design, the letters instead of spelled out is hard for a not frequent user to remember. and it looked like subway lines. 

I noticed that years of life lost is gone from the header, is that in a todo item somewhere?

app/components/CrashInjuryIndicators.tsx
```diff
@@ -0,0 +1,64 @@
...
+  className,
+}: {
+  label: string;
+  value: number | null;
+  className?: string;
+}) => {
+  return (
+    <span className={`${className || ""} text-nowrap`}>
+      <span className="me-1">{label}</span>
+      {/* todo: decided if/how we want to assign colors to these badges */}
```
chiaberry, line 16: I like it the way you have it, dark if 0, primary if value

app/components/CrashInjuryIndicators.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <InjuryBadge
+        value={injuries.sus_serious_injry_count}
+        label="Serious injuries"
+        className="me-3"
+      />
+      <InjuryBadge
+        value={injuries.nonincap_injry_count}
+        label="Minor injuries"
+        className="me-3"
+      />
```
chiaberry, line 49: this may be too much for MVP, but what about tooltips that define what a minor injury is vs a serious injury?

chiaberry (approved):
✅  
I will go review the acc designs to see what they recommended to do with years of life lost. I was flipping back and forth between a tab with this and vze prod, and this is so much cleaner and more compact its great.

app/components/CrashInjuryIndicators.tsx
johnclary: i will add this idea to the refinment/follow-up issue 🙏

johnclary:
> I noticed that years of life lost is gone from the header, is that in a todo item somewhere?

Nice catch—it was removed in the ACC designs, but I don't know why. I'll follow-up on that.

PR #1612, opened 2024-11-21
Proof of concept for zod-driven types and validation

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/20024

Here is the proof of concept for adding Zod schema to codebase and generating our types from them. I have mixed feelings about this! I assume we need a meeting to discuss/refine this approach, but will give y'all time to review 👍 

**URL to test:** Local
1. Start your local DB and hasura, and start up the nextjs `/app`  as usual.
2. Depending on what branch you've used recently, you might need to delete the `crashesListViewQueryConfig` entry from localstorage
3.  Click around. Awesome!

---

#### Ship list

- [ ] Code reviewed


app/utils/graphql.ts
```diff
@@ -27,40 +29,74 @@ const DEFAULT_SWR_OPTIONS: SWRConfiguration = {
...
+  request<T>({
     url: ENDPOINT,
     document: query,
     variables,
     requestHeaders: {
       Authorization: `Bearer ${token}`,
       "x-hasura-role": hasuraRoleName,
     },
+  }).then((data) => {
+    return schema.parseAsync(data);
```
johnclary, line 48: here's where we actually run the schema validation. if this errors, the error will be available in the `error` property returned by the `useQuery()` hook

app/utils/graphql.ts
```diff
@@ -27,40 +29,74 @@ const DEFAULT_SWR_OPTIONS: SWRConfiguration = {
...
   query: RequestDocument | null;
   variables?: Variables;
   options?: SWRConfiguration;
-}) => {
+  typename: string;
+  schema: ZodSchema<T>;
+}): UseQueryResponse<T> => {
   const { getIdTokenClaims, user } = useAuth0();
 
+  const responseSchema = useApiResponseSchema(schema, typename);
```
johnclary, line 94: this is how we handle the actual shape of the Hasura response, which returns something that looks like 

```json
{
  "crashes_list_view": [
    {
      "record_locator": "20484424",
      "case_id": "243100443",
      "crash_timestamp": "2024-11-05T14:40:00+00:00",
      "address_primary": "4700 N 35 HWY",
      "collsn_desc": "SAME DIRECTION - BOTH GOING STRAIGHT-REAR END"
    }
  ]
}
```

app/utils/graphql.ts
```diff
@@ -27,40 +29,74 @@ const DEFAULT_SWR_OPTIONS: SWRConfiguration = {
...
+  isLoading: boolean;
+  isError: boolean;
+  refetch: () => Promise<any>;
+  isValidating: boolean;
+}
 /**
  * Hook which wraps `useSWR` and provides auth
+ *
+ * This hook is limited to querying just one type per query as
+ * will be defined by the typename
```
johnclary, line 77: what i'm trying to say is that this `useQuery()` hook only supports fetching one type of object (crashes, units, etc) at the root of the query. 

so this is fine:

```gql
 query {
    crashes {
      id
    }
  }
}
```

but this is not:

```gql
query {
    crashes {
      id
    }
      units {
        id
      }
  }
```

we could extend this to support multiple root types but i wasn't up for figuring out how we would infer types from this kind of shape.

app/configs/crashesColumns.tsx
```diff
@@ -11,11 +12,11 @@ export const crashesColumns: {
...
   },
   address_primary: {
-    name: "Primary address",
+    name: "address_primary",
     label: "Address",
   },
   address_secondary: {
-    name: "Secondary address",
+    name: "address_secondary",
     label: "Secondary address",
```
johnclary, line 20: little bugs that surfaced as a result of the new zod validation 👍

app/types/crashes.ts
```diff
@@ -0,0 +1,4 @@
+import { z } from "zod";
+import { crashSchema } from "@/schema/crashes";
+
+export type Crash = z.infer<typeof crashSchema>;
```
johnclary, line 4: here's where a zod schema becomes a type 🐛 ➡️ 🦋

app/queries/crash.ts
```diff
@@ -223,7 +213,7 @@ export const UPDATE_CRASH = gql`
     update_crashes(where: { id: { _eq: $id } }, _set: $updates) {
       affected_rows
       returning {
-        cris_crash_id
+        id
```
johnclary, line 216: this is a good place to point out that we still have this awkward hard-coding of columns that should be returned by this graphql query, and we also need to make sure these columns are defined in the zod schema, and we also need to define these fields in the `crashesColumns.tsx` config. 

for the `crashesListView`, we work around this because the `queryBuilder` generates the graphql query from the config object, which itself is pulling it's columns from the `crashesListViewColumns` config file.

app/app/crashes/page.tsx
```diff
@@ -17,6 +18,7 @@ export default function Crashes() {
             columns={crashesListViewColumns}
             initialQueryConfig={crashesListViewQueryConfig}
             localStorageKey={localStorageKey}
+            schema={crashesListSchema}
```
johnclary, line 21: here we are passing in our zod schema, whose type is also referenced by the `crashesListViewColumns` when it is defined in `configs/crashesListViewColumns.tsx`. so the zod-generated type is going to provide all the typing for this table component, including data fetching. See `TableWrapper` for more notes on this...

app/components/TableWrapper.tsx
```diff
@@ -43,17 +46,21 @@ export default function TableWrapper<T extends Record<string, unknown>>({
 
   const query = useQueryBuilder(queryConfig);
 
-  const { tableName } = queryConfig;
-
-  const { data, isLoading } = useQuery<Record<typeof tableName, T[]>>({
+  const { data, isLoading, error } = useQuery({
```
johnclary, line 49: if you mouseover `data` in your code editor, you will see that it is typed as `T[]`. It's an array of the type that is coming from our trust zod schema 🔗

app/components/TableWrapper.tsx
```diff
@@ -14,11 +14,13 @@ import TableAdvancedSearchFilterToggle from "@/components/TableAdvancedSearchFil
 import TablePaginationControls from "@/components/TablePaginationControls";
 import { useActiveSwitchFilterCount } from "@/components/TableAdvancedSearchFilterToggle";
 import TableResetFiltersToggle from "@/components/TableResetFiltersToggle";
+import { ZodSchema } from "zod";
 
 interface TableProps<T extends Record<string, unknown>> {
   columns: ColDataCardDef<T>[];
   initialQueryConfig: QueryConfig;
   localStorageKey: string;
+  schema: ZodSchema<T>;
```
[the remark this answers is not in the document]
johnclary: also—you can see that the type of `T` that we're expecting is `Record<string, unknown>`: that's an object with keys that are all strings. We might also be able to find zod generic type to use here, like `ZodRecord`. I don't know that there would be a real benefit to pursuing that change, but it might reveal itself as we spend more time working on components that consume data from the `useQuery` hook...

app/types/types.ts
mddilley, line 1: awesome, i see that the Zod schema is where we would update if there was a new field coming from the backend. One advantage that I see about this is that we get a Zod validation schema for free if we ever needed to feed to React Hook Form. iirc this isn't something we need at the moment, and maybe it feels like we are defining the schema more than once? (Hasura metadata and Zod schema) 

The same point could be made for hand-typing too I guess - updating in two places. It is making me think about the question [you posed about whether we need runtime validation](https://austininnovation.slack.com/archives/CM9MK950S/p1732562571335399?thread_ts=1732558785.583419&cid=CM9MK950S) too. Makes me wonder if we should back away if we don't find something that feels maintainable. We have a pretty solid history of covering this with test scripts (though we have had errors slip by in this project and others 😅).

I found another resource that I didn't have time to listen to that covers type safety with React + swr + Hasura + zod (https://www.devshows.dev/podcasts/frontend-first/type-safety-from-hasura-to-swr). Could be useful? I'm not familiar with this podcast.

app/schema/crashes.ts
```diff
@@ -0,0 +1,61 @@
+import { z } from "zod";
+import { lookupOptionSchema } from "./lookupTable";
+import { unitSchema } from "./units";
+import { changeLogEntrySchema } from "./changeLog";
+import { crashInjuryMetricsSchema } from "./crashInjuryMetrics";
+
+export const crashSchema = z.object({
```
mddilley, line 7: I was able to test this schema and get a console error when I modified one of these column names to something that doesn't exist. I wasn't able to get feedback when I removed a column completely which I though was unexpected but now I'm not sure. 

I think this could end up with us having silent errors. If you remove `case_id` from this Zod object, no error occurs but the value no longer passes through to the Crash page.

mddilley (commented):
Very cool stuff. I left some thoughts, and it is cool to see that zod can cover this use case! I think that I share your "on-the-fence"-ness, and I'll spend some more time thinking about it.

PR #1613, opened 2024-11-25
Create locations pages and map

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/19960
* https://github.com/cityofaustin/atd-data-tech/issues/19975
* https://github.com/cityofaustin/atd-data-tech/issues/19977
* https://github.com/cityofaustin/atd-data-tech/issues/19976

## Testing

**URL to test:** Local

1. You will need to delete the `crashesListViewQueryConfig` entry from your localstorage
2. Use the sidebar to navigate to the locations page
3. Search by location ID
4. search by the location name
5. toggle the advanced filter switches on/off
6. sort and paginate through results
7. Refresh the page to see that your search is persisted
8. Navigate to a location details page
9. Verify that the location map loads correctly and is zoomed on the location polygon
10. See that the card to the right of the map displays the location ID

---

#### Ship list

- [ ] Code reviewed


app/app/locations/[location_id]/page.tsx
```diff
@@ -0,0 +1,66 @@
...
+    schema: locationSchema,
+    typename,
+  });
+
+  if (error) {
+    console.error(error);
+  }
+
+  if (!data) {
+    // todo: loading spinner (would be nice to use a spinner inside cards)
```
Comment (johnclary) [inline app/app/locations/[location_id]/page.tsx:34]: * https://github.com/cityofaustin/atd-data-tech/issues/20081

app/components/SidebarLayout.tsx
```diff
@@ -110,7 +110,7 @@ export default function SidebarLayout({ children }: { children: ReactNode }) {
               <SideBarListItem
                 isCollapsed={isCollapsed}
                 isCurrentPage={segments.includes("locations")}
-                Icon={FaMap}
+                Icon={FaLocationDot}
```
johnclary, line 113: i changed the sidebar icon to the map marker, this feels a lot more intuitive to me than the folding map icon

app/configs/locationColumns.ts
```diff
@@ -0,0 +1,16 @@
...
+  location_id: {
+    name: "location_id",
+    label: "Location ID",
+  },
+  // todo: need to support non-lookup table relationships
+  //   cr3_crash_count: {
+  //     name: "cr3_crash_count",
+  //     relationshipName: "locations_list_view"
+  //   }
+};
```
johnclary, line 16: the current data cards will render lookup table relationships, but there is mechanism to, render card values through other relationships, like the `locations` <> `locations_list_view` one-to-one relationship. it felt like a big enough lift to leave separate. and probably not worth touching until we settle the question of if we want to stick with zod-generated types 🤔

app/configs/locationDataCard.ts
```diff
@@ -0,0 +1,3 @@
+import { locationColumns } from "./locationColumns";
+
+export const locationCardColumns = [locationColumns.location_id];
```
johnclary, line 3: this feels a bit silly but i'm matching the pattern i used to set up the crash data card column arrays 🤷

app/schema/geojsonSchema.ts
```diff
@@ -0,0 +1,14 @@
...
+export const multiPolygonSchema = z.object({
+  type: z.literal("MultiPolygon"),
+  crs: z
+    .object({
+      type: z.string(),
+      properties: z.object({}),
+    })
+    .optional(),
+  coordinates: z.array(z.array(z.array(coordinateSchema))),
+});
```
johnclary, line 14: if we were going to define more geojson types i would go searching for a library that has already solved this problem

app/utils/queryBuilder.ts
```diff
@@ -145,6 +153,10 @@ export interface QueryConfig {
    * filter when its value is not an empty string
    */
   searchFilter: Filter;
+  /**
+   * The search fields that are available to select from when searching
+   */
+  searchFields: SearchFilterField[];
```
johnclary, line 159: these used to be hardcoded specifically for crashes

PR #1617, opened 2024-12-01
Create users and user details pages

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/19969
* https://github.com/cityofaustin/atd-data-tech/issues/19972

Note that we have separate issues for user mutations, which are out of scope:

* https://github.com/cityofaustin/atd-data-tech/issues/20120
* https://github.com/cityofaustin/atd-data-tech/issues/19973
* https://github.com/cityofaustin/atd-data-tech/issues/19974
* https://github.com/cityofaustin/atd-data-tech/issues/20121

## Testing

**URL to test:** Local

**Steps to test:**

1. Log out and then back into app — even it appears that you are logged in
2. Visit the new `/users` page - a spinner will appear while the complete set of users is retrieved
3. Verify that ~105 users load, sorted by most recent login date
4. Click on a user in the users table to navigate to the user details page, where their details load in a small card.
---

#### Ship list

- [ ] Code reviewed


app/utils/graphql.ts
```diff
@@ -24,6 +24,10 @@ const DEFAULT_SWR_OPTIONS: SWRConfiguration = {
    * Dont refetch on network recon
    */
   revalidateOnReconnect: false,
+  /**
+   * keep the previoius data while refetching
+   */
+  keepPreviousData: true,
```
johnclary, line 30: turns out there was already an SWR option that does what `useDataCache` did:
https://swr.vercel.app/docs/advanced/understanding.en-US#return-previous-data-for-better-ux

👍

app/utils/graphql.ts
```diff
@@ -213,18 +217,3 @@ export const useLookupQuery = (lookupTableDef: LookupTableDef | undefined) =>
       typename,
     ];
   }, [lookupTableDef]);
-
-/**
- * Hook that persists queried data while new data is being fetched / revalidated.
- * this works as a complement to SWR by persisting data when fetch args
- * have changed
- **/
-export const useDataCache = <T>(currentData: T | null) => {
```
johnclary, line 222: turns out there was already an SWR option that does what `useDataCache` did:
https://swr.vercel.app/docs/advanced/understanding.en-US#return-previous-data-for-better-ux

👍

app/utils/users.ts
```diff
@@ -0,0 +1,94 @@
...
+      pages
+        ?.flatMap((page) => page.users)
+        .sort((a, b) => {
+          // sort by last_login, which is undefined if a user has never logged in
+          if (!a.last_login) return 1;
+          if (!b.last_login) return -1;
+          return b.last_login > a.last_login ? 1 : -1;
+        }) || [],
+    [pages]
+  );
```
johnclary, line 59: i am resisting the urge to build a fancy <Table> component that has a column sorting UI and whatnot. If we move our user table out of Auth0 and into the VZ database we can use our graphql table component.

app/app/users/[user_id]/page.tsx
```diff
@@ -0,0 +1,96 @@
...
+  },
+];
+
+export default function UserDetails({
+  params,
+}: {
+  params: { user_id: string };
+}) {
+  const token = useToken();
+  const [user, setUser] = useState<User | null>(null);
```
Comment (johnclary) [inline app/app/users/[user_id]/page.tsx:53]: with not much more effort, we could pass the user data into this page component from the `users` list. the `next/router` component supports passing state data. but this is such a rarely used corner of the app that i didn't bother.

app/components/AppBreadCrumb.tsx
```diff
@@ -29,7 +29,7 @@ const useCrumbs = (path: string): Crumb[] =>
     });
     if (parts.length > 1) {
       crumbs.push({
-        label: parts[1].split("?")[0],
+        label: decodeURI(parts[1].split("?")[0]),
```
johnclary, line 32: the need for decoding came up because the Auth0 user ID's all of the vertical bar `|` in them that was rendering as `%7C`

app/utils/users.ts
```diff
@@ -0,0 +1,95 @@
...
+      return null;
+    }
+    if (previousPageData && previousPageData.users.length < PAGE_SIZE) {
+      // end of data - stop trying to fetch
+      return null;
+    }
+    // return the next URL to fetch
+    return `${process.env.NEXT_PUBLIC_CR3_API_DOMAIN}/user/list_users?page=${pageIndex}&per_page=${PAGE_SIZE}`;
+  };
+  const { data: pages, isLoading } = useSWRInfinite<ListUsersPage>(
```
johnclary, line 38: You can see we're passing the `ListUsersPage` type into this hook, and assuming the API is honoring its part of the contract. We can set up zod validation here if we want to.

app/schema/changeLog.ts
```diff
@@ -9,7 +9,7 @@ export const changeLogEntrySchema = z.object({
   record_id: z.number(),
   record_type: z.string(),
   record_json: z.object({
-    old: z.record(z.string(), z.unknown()),
+    old: z.record(z.string(), z.unknown()).nullable(),
     new: z.record(z.string(), z.unknown()),
   }),
 });
```
johnclary, line 15: i ran into a Zod exception when i connected to the staging DB. we might want to bring this commit upstream.

app/utils/auth.ts
```diff
@@ -10,23 +12,58 @@ interface CustomUser extends User {
...
   } else if (roles.includes("vz-admin")) {
     // todo
     // If user has more than one role, it is because they have `itSupervisor` and `vz-admin`
-    // `vz-admin` is the Hasura rolename, `itSupervisor` is some legacy name from back when
-    // we used the hasura `admin` role instead of a named role for admins. John is pretty sure
-    // sure we stop using `itSupervisor`
+    // `vz-admin` is the Hasura rolename, `itSupervisor` is a legacy role that was previously
+    // used to enable a user to create and edit other admins. this app will do away with
+    // that behavior: any admin can create and modify other admins. A follow-up task is
+    // to clean up the Auth0 user database that references `itSupervisor`
```
johnclary, line 33: IIRC it was originally a request from Lewis to be able to have a super-admin role, but I don't think it's worth the technical overhead to maintain this. I think we can safely get by with three tiers:

* Read only
* Editor: has all editing access except create/edit users
* Admin: has all editing access and can create/edit users

app/utils/graphql.ts
```diff
@@ -24,6 +24,10 @@ const DEFAULT_SWR_OPTIONS: SWRConfiguration = {
    * Dont refetch on network recon
    */
   revalidateOnReconnect: false,
+  /**
+   * keep the previous data while refetching
+   */
+  keepPreviousData: true,
```
johnclary, line 30: this option replaces the need for `useDataCache`

app/contexts/Auth.tsx
```diff
@@ -21,7 +22,14 @@ export default function AuthProvider({
...
         redirect_uri: redirect_uri || "",
-        scope: "openid profile email",
+        /**
+         * Note that offline_access is requried for refresh tokens -
+         * the alternative is to use a custom hook that occasionally
+         * calls getAccessTokenSilently(), which will manually refresh
+         * the token
+         */
+        scope: "openid profile email offline_access",
+        audience: AUDIENCE,
```
johnclary, line 32: while testing this i realized that our tokens were not refreshing—this fixes that. i spent some time staring at the old VZE and couldn't quite figure out how we are refreshing tokens in that app.

app/app/users/page.tsx
```diff
@@ -0,0 +1,96 @@
...
+                    <FaCopy className="me-2" />
+                    <span>Copy user emails</span>
+                  </AlignedLabel>
+                </Button>
+              </>
+            )}
+          </div>
+          {/* todo: standardize the way we show error messages and use error boundary */}
+          {Boolean(error) && (
+            <Alert variant="dange">Something went wrong</Alert>
```
chiaberry, line 49: typo on danger

app/app/users/[user_id]/page.tsx
```diff
@@ -0,0 +1,87 @@
...
+import Spinner from "react-bootstrap/Spinner";
+import Table from "react-bootstrap/Table";
+import AppBreadCrumb from "@/components/AppBreadCrumb";
+import { useToken } from "@/utils/auth";
+import { useUser } from "@/utils/users";
+import { User } from "@/types/users";
+import { formatDateTime } from "@/utils/formatters";
+
+type UserColumn = {
+  name: keyof User;
```
Comment (chiaberry) [inline app/app/users/[user_id]/page.tsx:14]: I cant decide if I'd prefer this to be called `columnName` since there is a column named Name and I was getting confused, but maybe then it would be overkill with `user[col.columnName]`

app/app/users/page.tsx
```diff
@@ -0,0 +1,96 @@
...
+                  <AlignedLabel>
+                    <FaUserPlus className="me-2" />
+                    <span>Add user</span>
+                  </AlignedLabel>
+                </Button>
+                <Button>
+                  <AlignedLabel>
+                    <FaCopy className="me-2" />
+                    <span>Copy user emails</span>
+                  </AlignedLabel>
```
chiaberry, line 42: I know these are on your list of todos, and there are issues written for them, but i went down a mini side rabbit hole looking at extensions or settings for vscode to highlight TODOs 

and you have todos written elsewhere, but maybe you remove the todo after you write up an issue??

app/contexts/Auth.tsx
chiaberry: our tokens last for 10 hours, so before adding this, id have to sign in every 10 hours?
or the query wouldnt work but refreshing the app would fix it, because it would refresh the token without offline_access?

johnclary: i believe our tokens expire after 24 hours. before this change, your token would expire and all of your API calls would fail, but you would still be logged into the app. if you signed into the netlify deployment more than 24 hours ago you can see this for yourself: https://vision-zero-nextjs.netlify.app/crashes

to resolve it, you have to manually sign out and sign back in to the app.

so, yeah, as long as you have a session token in local storage, the Auth0 library just assumes it's valid and the `useAuth0` hook returns a normal-looking `user` object with `isAuthenticated = true`.

If we wanted to fix this without using refresh tokens, we could  use a `try...catch` block with  `getAccessTokenSilently()`, which would fail and then we could redirect users to the login screen.

With refresh tokens, the app is going to grab a new token for you whenever you refresh the page. I'm still not 100% if your token expires before you refresh it—we may still need to do the `try...catch` thing to make sure users get sent back to login.

johnclary: and more thing—the reason this isn't obvious on local dev is because we don't run the local stack in JWT mode. the endpoints are not pw protected at all. we might want to take this opportunity to change that.

app/app/users/[user_id]/page.tsx
Comment (johnclary) [reply app/app/users/[user_id]/page.tsx:14]: 👍 i prefer `name`, but am open to alternatives. `key` or `field` also come to mind. whatever we choose, we should use the same property name to describe a column name in our graphql queries. which is `name` at the moment: https://github.com/cityofaustin/vision-zero/blob/a64f5f72a87a7eac726daf916a63b305eb7c21f4/app/types/types.ts#L15

☝️ the typing on these column configs is a bit messy and i kinda want to take a second pass at it once we remove the zod stuff.

app/app/users/page.tsx
johnclary: sure, that sounds good. there are a handful that feel to vague for an issue just yet, so i was just planning to sweep through them periodically. but if that's feeling too messy i can keep them in a doc. i don't want the code to look/feel crufty!

app/contexts/Auth.tsx
johnclary: and answering part of your question that i missed: the `offline_access` scope is a special Auth0 security permission that must be togged on in order to allow refresh tokens—i did not bother to understand why. it's not just adding it here to the `scope` that was needed, i had to flip a toggle for it in our Auth0 app settings.

app/app/users/page.tsx
chiaberry: i like them being in the code! i know there is robust testing before releases, but having it in the code could help prevent broken buttons sneak in to main

johnclary: oh i see you're pointing out that i should put a todo here!

chiaberry: well i didnt actually point it out, now that im reading what i wrote

chiaberry (approved):
I see users!

app/schema/changeLog.ts
johnclary: i did this by cherry-picking `0d4611c65acc71b914d4b3ab1b178f38ef136e41` into `nextjs`

app/app/users/page.tsx
```diff
@@ -0,0 +1,97 @@
...
+                    <td>{user.name}</td>
+                    <td>{user.email}</td>
+                    <td>{new Date(user.created_at).toLocaleDateString()}</td>
+                    <td>
+                      {user.last_login
+                        ? new Date(user.last_login).toLocaleDateString()
+                        : ""}
+                    </td>
+                    <td>{user.logins_count}</td>
+                    <td>{user.app_metadata.roles[0]}</td>
```
frankhereford, line 87: This is putting like `readonly`, `vz-admin`, `editor` and `itSupervisor` which isn't wrong, but these values feel like `code` to me and not human role names. Do you think it would be helpful to make a lookup object to turn these names into "Viewer", "Administrator", "Editor" and "System Administrator"?

app/components/AppBreadCrumb.tsx
frankhereford, line 32: Do you think this `auth0|<long string of hex>` could be replaced with the user's name? I agree that this is highly accurate, but I think it might be clearer if we breadcrumb with something human readable.

app/app/users/[user_id]/page.tsx
```diff
@@ -0,0 +1,87 @@
...
+  label: string;
+  renderer?: (user: User) => string;
+};
+
+// todo: this can be extended and used for the user list table column sorting, etc
+const COLUMNS: UserColumn[] = [
```
Comment (frankhereford) [inline app/app/users/[user_id]/page.tsx:20]: UPPER_SNAKE_CASE for constants -- this is great. Thank you for helping me discover these conventions and for adopting them out of the gate.

app/app/users/page.tsx
```diff
@@ -0,0 +1,97 @@
...
+                    <FaCopy className="me-2" />
+                    <span>Copy user emails - todo</span>
+                  </AlignedLabel>
+                </Button>
+              </>
+            )}
+          </div>
+          {/* todo: standardize the way we show error messages and use error boundary */}
+          {Boolean(error) && (
+            <Alert variant="dange">Something went wrong</Alert>
```
frankhereford, line 50: What would you think about putting the actual error in here? Maybe something like:


Maybe something like this?

```
{Boolean(error) && (
  <Alert variant="danger">Something went wrong: {error.message}</Alert>
)}
```

app/utils/auth.ts
frankhereford: Full agreement. We have access to the whole stack, top to bottom, so there's no reason that I can conceive of that we need to build special capabilities into the front-end UI that we can't do with our backend access.

app/components/SidebarLayout.tsx
```diff
@@ -12,6 +12,7 @@ import {
   FaGaugeHigh,
   FaLocationDot,
   FaAngleRight,
+  FaUserGroup,
```
frankhereford, line 15: This is outside the scope of this PR (literally), but a couple of lines down is a typo for `Container` that would be awesome if you'd consider tagging along in this work.

app/contexts/Auth.tsx
```diff
@@ -21,7 +22,14 @@ export default function AuthProvider({
       clientId={CLIENT_ID}
       authorizationParams={{
         redirect_uri: redirect_uri || "",
-        scope: "openid profile email",
+        /**
+         * Note that offline_access is requried for refresh tokens -
```
frankhereford, line 26: Spelling on 'required'

app/utils/auth.ts
```diff
@@ -10,23 +12,58 @@ interface CustomUser extends User {
...
+export const getHasuraRoleName = (roles?: string[]): string => {
   if (!roles) {
     return "";
   } else if (roles.includes("vz-admin")) {
     // todo
     // If user has more than one role, it is because they have `itSupervisor` and `vz-admin`
-    // `vz-admin` is the Hasura rolename, `itSupervisor` is some legacy name from back when
-    // we used the hasura `admin` role instead of a named role for admins. John is pretty sure
-    // sure we stop using `itSupervisor`
+    // `vz-admin` is the Hasura rolename, `itSupervisor` is a legacy role that was previously
```
frankhereford, line 30: Did you intend for a space in `rolename` or is that an intended literal?

app/utils/auth.ts
```diff
@@ -10,23 +12,58 @@ interface CustomUser extends User {
...
+ * Todo: our Auth0 app is configured to return idTokens, which are
+ * "opaque" proprietary Auth0 tokens, seemingly because this enables us
+ * to interact with the User Management API. as a result, we must
+ * use idToken.__raw to grab the actual JWT. it seems like our setup
+ * may be misconfigured, because accessing .__raw seems like a hack :/
+ *
+ * the typical setup would enable use to use the getAccessTokenSilently()
+ * method, but that doesn't work with the opaque tokens.
+ *
+ * dicussion here: https://community.auth0.com/t/getting-the-jwt-id-token-from-auth0-spa-js/28281/3
```
frankhereford, line 53: Discussion typo

frankhereford (changes requested):
You write really clean code John, this is premium stuff. I have a few questions, a few observations and a few micro typos.

The main reason I'm requesting changes is around the breadcrumb with the auth0 string and the VZ role names, but these are not terribly strongly held opinions. I'm eager to discuss -- thanks!

app/app/users/[user_id]/page.tsx
Comment (roseeichelmann) [inline app/app/users/[user_id]/page.tsx:1]: this new Next app routing is cool i like it

app/app/users/[user_id]/page.tsx
Comment (roseeichelmann) [reply app/app/users/[user_id]/page.tsx:14]: i think i prefer `field` so as not to get confused with something like chias example

app/app/users/[user_id]/page.tsx
```diff
@@ -0,0 +1,87 @@
...
+import AppBreadCrumb from "@/components/AppBreadCrumb";
+import { useToken } from "@/utils/auth";
+import { useUser } from "@/utils/users";
+import { User } from "@/types/users";
+import { formatDateTime } from "@/utils/formatters";
+
+type UserColumn = {
+  name: keyof User;
+  label: string;
+  renderer?: (user: User) => string;
```
Comment (roseeichelmann) [inline app/app/users/[user_id]/page.tsx:16]: can you explain to me this typescript code like im a baby/tell me if im on the right track

the `UseColumn` type has a property `renderer` which is a function that takes in a property `user` of type `User` than spits out/returns a string?

app/app/users/[user_id]/page.tsx
```diff
@@ -0,0 +1,87 @@
...
+    renderer: (user) => formatDateTime(user.last_login),
+  },
+  {
+    name: "updated_at",
+    label: "Updated at",
+    renderer: (user) => formatDateTime(user.updated_at),
+  },
+];
+
+export default function UserDetails({
```
Comment (roseeichelmann) [inline app/app/users/[user_id]/page.tsx:47]: it looks like you are mostly using function declarations now across the app which i like, but there are a few components where you still use an arrow function like `CrashMap`. should we develop some kinda of standard for when to use function vs arrow or keep leaving that choice up to the dev

app/utils/users.ts
roseeichelmann: mui datagrid?

app/types/users.ts
roseeichelmann, line 1: when do we put types in this folder vs just declare them in the file we are using them in? only when the type is being used in multiple diff files?

roseeichelmann (approved):
this looks super duper great and amazing and also i learned more typescript while reviewing this code. its all written super concise and clean, thanks for spearheading this nextjs & typescript movement. i see many users :sparkles:

app/app/users/[user_id]/page.tsx
Comment (frankhereford) [reply app/app/users/[user_id]/page.tsx:16]: Take me with a grain of salt, I think your read is word for word correct. You could add the word "optional" too if you wanted too: 

> the UseColumn type has an _optional_ property ....

app/app/users/[user_id]/page.tsx
Comment (johnclary) [reply app/app/users/[user_id]/page.tsx:47]: yes, great point. i'd love to try and be consistent with this. i will run through the rest of the code and rework the declarations in a separate PR 🙏

app/app/users/page.tsx
johnclary: (The error message in this case is just "Failed to fetch")

app/app/users/page.tsx
johnclary: Absolutely—I will make a little utility function for that and push up that change 🙏

app/components/AppBreadCrumb.tsx
johnclary: I do—i'll open a follow-up issue to refine that breadcrumb component. At the moment it's just rendering data from the `pathName` and we'll need to get fancy in order to customize this per page route 👍

app/components/SidebarLayout.tsx
johnclary: gruesome stuff—and now fixed. thank you!

app/utils/users.ts
johnclary: I was referring to our fancy table component that we use for the crashes and locations list 👀

frankhereford (approved):
I am 100% digging where this is going -- thanks for checking out my feedback!

app/contexts/Auth.tsx
chiaberry: oh I dont know if this is has been mentioned elsewhere, but I did get that alert from auth0 that they were requesting the new offline_access scope, could be nice to alert our users when the nextjs upgrade gets rolled out to production, to expect to grant new access. because they will see the same alert, right?

johnclary: Thanks for that reminder. Yes and I created https://github.com/cityofaustin/atd-data-tech/issues/20311 to remember things like this 🙏

johnclary:
>A few points that would be cool to discuss more in backlog sessions:
> * EIL5 overview of SRW and how its configured
> * elaborate on the convo of fancy Table component vs an existing React Table library

@mateoclarke sounds great—i will add these to our discussion list 🙏

johnclary:
I created https://github.com/cityofaustin/atd-data-tech/issues/20352 to document our architectural design decisions. Thanks, @mateoclarke — TIL!

PR #1622, opened 2024-12-04
Crash details: show banner when crash is outside of AFP or on a private drive

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/20007

I also replaced the nextjs favicon with the vze one but I can remove that if it was premature

## Testing

**URL to test:** 
local

**Steps to test:**

You will need to delete the crashesListViewQueryConfig entry from your localstorage if you have not already been testing Johns other PRs. 
Go to any crash page. If the crash is not on a private drive and in Austin full purpose, you will not see any location banner at the top. 
Toggle the crash to be on a private drive. The banner appears! 
Toggle the crash to not be on a private drive. Move the location using the map to be outside of Austin.  Banner appears again. 

Screenshot 2024-12-05 at 8 17 21 AM

---

#### Ship list

- [ ] Code reviewed


app/components/CrashLocationBanner.tsx
```diff
@@ -0,0 +1,21 @@
+import Alert from "react-bootstrap/Alert";
+
+interface CrashLocationBannerProps {
+  privateDriveFlag: boolean | null;
+}
```
chiaberry, line 5: Initially was passing the entire Crash object and thus had this as 
```typescript
interface CrashHeaderProps {
  crash: Crash;
}
```

but then switched to just the private drive flag. I considered then just getting rid of this entire interface and having the types in the function definition, but this felt easier to read. Lemme know if yall prefer the other way. This also feels easier to extend in the future, but I dont anticipate us passing more props into this component.

frankhereford: I am a huge fan of passing in what you need and not everything else. I endorse this for clarity and efficiency -- this feels like the way to me.

app/app/crashes/[record_locator]/page.tsx
```diff
@@ -59,6 +60,12 @@ export default function CrashDetailsPage({
     <>
       <AppBreadCrumb />
       <CrashHeader crash={crash} />
+      {
+        // show alert if crash on private drive or outside of Austin full purpose
+        (crash.private_dr_fl || !crash.in_austin_full_purpose) && (
```
Comment (frankhereford) [inline app/app/crashes/[record_locator]/page.tsx:65]: `|| !crash.in_austin_full_purpose` → "or not in austin full purpose"

That field renaming makes your code just jump off the page in legibility.

app/components/CrashLocationBanner.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Crash Location Banner is rendered if private flag drive is true or austin full purpose flag is false.
+ */
+export default function CrashLocationBanner({
+  privateDriveFlag,
+}: CrashLocationBannerProps) {
+  return (
+    <Alert variant="warning">
```
frankhereford, line 14: Like, this just doesn't matter, but there's something awkward to me about how you have to put that bit of whitespace in the interpolated bit of JSX, but I can't think of anything that's clearly better. Putting it inside the quoted conditional string is not great, maybe this? Again, this couldn't be a smaller detail and ignore it pls if you want.

```
<Alert variant="warning">
  {`This crash is not included in Vision Zero statistical reporting because ${
    privateDriveFlag 
      ? "it occurred on a private drive" 
      : "it is located outside of the Austin full purpose jurisdiction"
  }`}
</Alert>
```

app/components/CrashLocationBanner.tsx
```diff
@@ -0,0 +1,21 @@
...
+      This crash is not included in Vision Zero statistical reporting because{" "}
+      {privateDriveFlag
+        ? "it occurred on a private drive"
+        : "it is located outside of the Austin full purpose jurisdiction"}
```
frankhereford, line 18: Do you mind period at the ends of this sentence?

frankhereford (changes requested):
The only comment that is an actual request for changes is the period / full-stop one -- the rest are observations and ideas for consideration. This tested up great and is an awesome addition - thank you Chia!

app/public/favicon.ico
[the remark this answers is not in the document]
johnclary: cool—I think i can get on board with that! and thanks for the reminder that we need to replicate the environment banner as well. i will open the issues 🙏

app/components/CrashLocationBanner.tsx
chiaberry: I took the copy directly from the existing banner https://github.com/cityofaustin/vision-zero/blob/main/editor/src/views/Crashes/Crash.js#L151 and I wasn't sure if the lack of `.` was intentional. I can add it!

johnclary: It was! And I ran off to find some blurbage that aligns with my arbitrary opinion:
https://ux.stackexchange.com/questions/122147/periods-in-ux-ui-microcopy

I actually I think this is a convention that Amenity introduced and may even be written in a style guide of ours somewhere...

app/public/favicon.ico
johnclary, line 1: thank you for updating this, @chiaberry! 

⭐ bonus points awarded ⭐

app/components/CrashLocationBanner.tsx
```diff
@@ -0,0 +1,21 @@
+import Alert from "react-bootstrap/Alert";
+
+interface CrashLocationBannerProps {
+  privateDriveFlag: boolean | null;
```
johnclary, line 4: would you add a comment here so that it shows up in component's type hints? E.g. "If the crash occurred on a private drive".

johnclary (approved):
Looks great! I'm very happy to have you contributing to the new codebase! 🙌 🎉 🚢

app/components/CrashLocationBanner.tsx
johnclary: thanks for kicking off this convo. i agree with your approach and started a Slack discussion about some other considerations for how we want to manage our types, prop types, and comments: https://austininnovation.slack.com/archives/CHZE6BC6L/p1733420020334539

app/components/CrashLocationBanner.tsx
frankhereford: That's cool - I'm totally good if that's a style convention we want to adopt, but given a vote, I'd still be of the mind that a cogent collection of words that includes a subject and predicate deserves a period, but I do get it - and I think I'm in the minority -- I am super cool either way!!

frankhereford (approved):
Looks so good!

app/components/CrashLocationBanner.tsx
chiaberry: done! also I learned its very particular of it being a comment with `/**` to start, `/*` wont cut it

chiaberry: basically it has to be a JSDoc annotation and not a comment

app/public/favicon.ico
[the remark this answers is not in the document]
johnclary: https://github.com/cityofaustin/atd-data-tech/issues/20180

app/components/CrashLocationBanner.tsx
johnclary: Thanks, Frank—I definitely appreciate your attention to the details. It would be kinda nice to put these decisions into a VZ style guide that might live in the repo. Maybe something that could be done as part of https://github.com/cityofaustin/atd-data-tech/issues/20013.

PR #1624, opened 2024-12-09
Define prettier configuration and a handful of spelling corrections

frankhereford:
## Associated issues

_none_; I needed a PR to test some CI configurations, and I've had these commits waiting around for a PR to tag along with, so here we go.

## Testing

Nothing really to test, except to see if you like the prettier configuration in your editor of choice.

### DIFF
--- a/app/.prettierrc
+++ b/app/.prettierrc
@@ -1,0 +1,6 @@
+{
+  "trailingComma": "es5",
+  "tabWidth": 2,
+  "semi": true,
+  "singleQuote": false
+}
--- a/database/migrations/default/1731003209538_rename_is_coa_roadway/up.sql
+++ b/database/migrations/default/1731003209538_rename_is_coa_roadway/up.sql
@@ -1,7 +1,7 @@
 alter table crashes rename column is_non_coa_roadway to is_coa_roadway;
 alter table crashes alter column is_coa_roadway set default false;
 
-comment on column public.crashes.is_coa_roadway is 'If the crash location occured within the City of Austin, not on a private drive, and not intersecting with the non_coa_roadways layer. Set via trigger.';
+comment on column public.crashes.is_coa_roadway is 'If the crash location occurred within the City of Austin, not on a private drive, and not intersecting with the non_coa_roadways layer. Set via trigger.';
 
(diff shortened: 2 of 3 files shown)


app/.prettierrc
frankhereford, line 1: I picked these settings because they match what the code is currently exhibiting. I'm personally a fan of turning off the semicolons at the end of lines, but that's neither here nor there. Do you have a preference, out of curiosity?

johnclary: looks good to me—this matches my settings. thanks for creating this 🙏

database/migrations/default/1731003209538_rename_is_coa_roadway/up.sql
```diff
@@ -1,7 +1,7 @@
 alter table crashes rename column is_non_coa_roadway to is_coa_roadway;
 alter table crashes alter column is_coa_roadway set default false;
 
-comment on column public.crashes.is_coa_roadway is 'If the crash location occured within the City of Austin, not on a private drive, and not intersecting with the non_coa_roadways layer. Set via trigger.';
+comment on column public.crashes.is_coa_roadway is 'If the crash location occurred within the City of Austin, not on a private drive, and not intersecting with the non_coa_roadways layer. Set via trigger.';
```
johnclary, line 4: i know this is harmless, but it feels odd making this change since it won't be applied to the db. sadly, we'd need to create a new migration to fix my dumb typo 🫠

johnclary (approved):

database/migrations/default/1731003209538_rename_is_coa_roadway/up.sql
frankhereford: yea, it is strange changing it now, but the rational in making the change here is not so much to fix it in the DB, but more to keep my code editor from flagging the file as having an issue. it's not often that i do have this file open though, so i'll revert this so that we have a shot at finding it again down the road. thanks!

mddilley (approved):
🚢 💅 ✨

chiaberry (approved):
this is what my editor inferred the prettier settings were too 💅🏼

PR #1625, opened 2024-12-09
Remove zod-driven types + validation

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/20214

## Testing

**URL to test:** Local 

1. Really the fact that we're building successfully should cover it, but if you'd like to click through every working feature (as documented in the readme) that'd be great.

---

#### Ship list

- [ ] Code reviewed


app/types/crashesList.ts
```diff
@@ -1,4 +1,7 @@
-import { z } from "zod";
-import { crashesListSchema } from "@/schema/crashesList";
-
-export type CrashesListCrash = z.infer<typeof crashesListSchema>;
+export type CrashesListRow = {
```
johnclary, line 1: `CrashesListRow` seemed less awkward than `CrashesListRow` 🤷

app/utils/graphql.ts
```diff
@@ -48,24 +46,8 @@ const fetcher = <T>([query, variables, token, hasuraRoleName, schema]: [
       Authorization: `Bearer ${token}`,
       "x-hasura-role": hasuraRoleName,
     },
-  }).then((data) => {
-    return schema.parseAsync(data);
```
johnclary, line 52: 👋 goodbye runtime validation. I dunno! now i kind of miss that we know 100% that are types match our graphql queries. Last chance to change our minds! ;)

chiaberry: what would be the cause of our types not matching our graphql queries?

johnclary: if we changed a field name without updating our types, or if we add a field to a graphql query without adding it to our type. and vice versa. just these marginal edge cases.

johnclary: in the same way that it's our responsibility to update the Hasura metadata when we add a column to a db table, we also need to make sure we add the new column to our queries and our types.

frankhereford (approved):
This is 100% on the money. Thanks for removing zod; I think this is the right path.

[General Zod was a pretty uncool dude, as well.](https://youtu.be/_opgbovQqg4?si=XGyFqgV5T5HvmwIk&t=1)

app/utils/graphql.ts
johnclary: I created https://github.com/cityofaustin/atd-data-tech/issues/20272 to look into the code-generated types—and i think if we get to that point where we know 100% that our types match our API we should be able to rest pretty easy. Will be fun to pursue this!

app/types/crashesList.ts
johnclary: I like that a lot—thinking in terms of what the purpose of this view is rather than describing how the data is presented. I'll add a todo to reconsider this naming convention 🙏

PR #1626, opened 2024-12-09
Ability to create and update users

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/20120
* https://github.com/cityofaustin/atd-data-tech/issues/19973

## Testing

**URL to test:** Local

Please be aware of these limitations:
* ⚠️ Staging and production have a shared user pool—do not edit user accounts other than your own
* I intend to deprecate the `itSupervisor` role, and odd things may happen if you try to edit an `itSupervisor` user through the nextjs app
* As I mention [here](https://github.com/cityofaustin/vision-zero/pull/1626/files/94966f4c35911fa3f40126236ba1f05cfe136335#r1876858244), I'd like to present the user admin with a "reset password" URL that they can send to new users. Or something like that—the new user UX  needs refinment.

1. Start your local api. From `/api` run `docker compose up`. Let me know if you need help getting this set up.

2. Update your nextjs VZE environment (`/app/.env.local`) to use the local API:

```
NEXT_PUBLIC_CR3_API_DOMAIN=http://localhost:8085
```

3. Start your local db, hasura, and of course the nextjs app itself.

4. You will need to be logged in as an admin to test. There is test admin account or 1pass titled **Vision Zero Editor (VZE) Test User: VZ Admin**.

5. Navigate to the users page (`/users`). Notice that users are sorted by last login, descending.

6. Use the **Add user** button to add a new **read only** user. When creating a user email, use `<first>.<name>+test-xyz@ausintexas.gov`.

7. Verify that the new user is created and you are redirected to the user details page. Now use the **Update user** button to edit the user's name and role. Verify the changes are saved.

8. Navigate back to the users list (`/users`) and verify your new user is listed.

9. Try to create another user by using an email address that is already used by an existing user. Verify that the form returns a helpful error message.

10. Great—now head to [VZE prod](https://visionzero.austin.gov/editor/#/users) and delete the user(s) you created

11. Log out of the app and log back in with the test editor account, available in the 1pass item titled **Vision Zero Editor (VZE) Test User: Editor**. 

12. Notice that the users list doesn't load. Try creating a user and notice that you receive an error message.

---

#### Ship list

- [ ] Code reviewed


app/components/NewUserForm.tsx
```diff
@@ -0,0 +1,113 @@
...
+};
+
+export default function NewUserForm() {
+  const token = useToken();
+  const {
+    register,
+    handleSubmit,
+    control,
+    formState: { errors },
+  } = useForm<NewUserInputs>({
```
johnclary, line 18: react-hook-form with typescript is super easy—just pass in the type and suddenly the whole form is typesafe 🔥

app/components/NewUserForm.tsx
```diff
@@ -0,0 +1,113 @@
...
+        />
+        <Form.Control.Feedback type="invalid">
+          Email is required
+        </Form.Control.Feedback>
+      </Form.Group>
+      <Form.Group className="mb-3" controlId="userRole">
+        <Form.Label>Role</Form.Label>
+        <div>
+          <Controller
+            name="app_metadata.roles.0"
```
johnclary, line 75: very cool TIL that you can assign nested values with this syntax. i'm pretty sure this will be familiar to lodash users as an accessor pattern ✨

app/components/UserModal.tsx
```diff
@@ -0,0 +1,180 @@
...
+}: UserModalProps) {
+  const [isSubmitting, setIsSubmitting] = useState(false);
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
+  const token = useToken();
+  const {
+    register,
+    handleSubmit,
+    control,
+    formState: { errors, isDirty },
+  } = useForm<UserInputs>({
```
johnclary, line 44: react-hook-form with typescript is super easy—just pass in the type and suddenly the whole form is typesafe 🔥

app/utils/users.ts
```diff
@@ -14,10 +14,7 @@ const INITIAL_PAGE_LIMIT = 10;
 const fetcher = async (url: string, token: string) =>
   fetch(url, {
     headers: { Authorization: `Bearer ${token}` },
-  }).then((res) => {
-    if (!res.ok) throw new Error("Failed to fetch");
-    return res.json();
-  });
+  }).then((res) => res.json());
```
johnclary, line 17: this reworking of error message is to better surface the errors from our API rather than throwing hte generic "Failed to fetch"

api/docker-compose.yaml
```diff
@@ -3,10 +3,10 @@ services:
...
       context: .
     ports:
-      - "5000:5000"
+      - "8085:5000"
     environment:
       - FLASK_ENV=production
     volumes:
       - .:/usr/src/app
-    command: flask run --host=0.0.0.0 --port=5000
+    command: flask run --host=0.0.0.0 --port=5000 --debug
```
johnclary, line 11: `--debug` mode enables the code change detection + refresh behavior 👍

api/server.py
```diff
@@ -42,6 +45,32 @@
...
+    while True:
+        password = "".join(secrets.choice(alphabet) for i in range(num_chars))
+        if (
+            any(c.islower() for c in password)
+            and any(c.isupper() for c in password)
+            and any(c.isdigit() for c in password)
+            and any(c in special_chars for c in password)
+        ):
+            break
+    return password
```
johnclary, line 72: i tested this to see how many times the `while` block ran, and average it was just once. very rarely as many as three times.

api/server.py
```diff
@@ -63,9 +92,18 @@ def get_api_token():
...
+    return (
+        jsonify(
+            {
+                "error": "Forbidden",
+                "message": "You do not have permission to access this resource.",
+                "statusCode": 403,
+            }
+        ),
+        403,
+    )
```
johnclary, line 106: before this change, our API was not returning JSON on `403` errors. doing so makes it easier to write the frontend response handler.

api/server.py
```diff
@@ -104,7 +142,7 @@ def add_custom_headers(response):
 
     # Access-Control-Allow-Origin for CORS
     response.headers["Access-Control-Allow-Origin"] = "*"
-    response.headers["Access-Control-Allow-Methods"] = "GET, POST, OPTIONS"
+    response.headers["Access-Control-Allow-Methods"] = "GET, POST, PUT, OPTIONS"
```
johnclary, line 145: i spent a long time debugging this, and i could not figure out why the `update_user` API is not running into a CORS issue in prod. if you want to go down the same rabbit hole I did, undo this change and compare the CORS error you get locally vs prod 😵‍💫

api/server.py
```diff
@@ -315,8 +353,14 @@ def healthcheck():
...
-@cross_origin(headers=["Content-Type", "Authorization"])
-@cross_origin(headers=["Access-Control-Allow-Origin", CORS_URL])
+@cross_origin(
+    headers=[
+        "Content-Type",
+        "Authorization",
+        "Access-Control-Allow-Origin",
+        CORS_URL,
+    ],
+)
```
johnclary, line 363: it seems like a best practice not to repeat the `cross_origin` decorator, but it's probably NBD 🤷

api/server.py
```diff
@@ -407,28 +463,34 @@ def user_list_users():
             + page
             + "&per_page="
             + per_page
-            + "&include_totals=true"
+            + "&include_totals=true&sort=last_login:-1"
```
johnclary, line 466: this sorts the user list by the last login date—something we were previously doing with frontend javascript ✅

app/utils/users.ts
```diff
@@ -74,7 +60,7 @@ export function useUsersInfinite(token: string) {
  */
 export function useUser(userId?: string, token?: string) {
   const url = `${process.env.NEXT_PUBLIC_CR3_API_DOMAIN}/user/get_user/${userId}`;
-  return useSWR<User>(
+  return useSWR<User | UserAPIError>(
```
johnclary, line 63: this is one of those spots where we just gotta take it on faith that the API is going to return this types.

api/server.py
```diff
@@ -42,6 +45,32 @@
 
 ADMIN_ROLE_NAME = "vz-admin"
 
+CORS_URL = "*"
```
frankhereford, line 48: We want this wide-open CORS, right? I don't have a good intuition about this. I always end up putting down a wildcard when asked, but it's always to make it work for me, not make it correctly configured.

api/server.py
frankhereford, line 72: It's A+. The whole model of not letting the account creator know the initial password of the account receiver is a step above. It's really fine, thank you!

api/server.py
frankhereford: Yech, this is on me. 

I touched on it in the API gateway skill share but only for a few seconds. The API gateway is rewriting the CORS headers, and I have it set to wide open. 

Like I said in a comment above (or below, you never know with GH) -- I default to the wildcard, but I am pretty sure it should be the list of URL hostnames who are allowed to access the resource. I never quite get it, because I figure you can defeat it by changing how your browser works -- I don't understand CORS, beyond the most basic thing, if that.

frankhereford (approved):
Wow John, this is awesome. You brought the API app a long way with a bunch of improvements, and the front end code looks impeccable. 

I want to call out one of my comments -- it's the API gateway that is rewriting the CORS headers and that I have every reason to think that my current configuration is naïve and should be refined. 

Thanks for this!

app/utils/users.ts
frankhereford: I don't know to write this without sounding like a lunatic or bitter -- I'm not --  I am entirely impressed and on board with the stack currently! OK - that said

This API response typing is one of those things that I was so impressed with in the T3 stack with regard to how it used tRPC. The functions on the server are defined as normal ol' typescript and when you call them, almost exactly like you call a normal function in the frontend, client code, you get the result of the function just like normal (even thought it ran on the server), and the whole transaction over the wire is type safe with no extra work. You only have to define types like normal, and then only if they can't be inferred. It's kinda crazy how smooth it is.

app/components/UserModal.tsx
```diff
@@ -0,0 +1,180 @@
...
+            />
+            <Form.Control.Feedback type="invalid">
+              Email is required
+            </Form.Control.Feedback>
+          </Form.Group>
+          <Form.Group className="mb-3" controlId="userRole">
+            <Form.Label>Role</Form.Label>
+            <div>
+              <Controller
+                name="app_metadata.roles.0"
```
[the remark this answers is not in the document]
frankhereford: `_` is great, it's got the best swiss army knife of oddly useful functions. This usage is above my head right now, but I'm looking forward to learning it.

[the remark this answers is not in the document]
johnclary: what's happening here is pretty awesome. react-hook-form is managing an object with all our form data, and we want this object to be shaped like the Auth0 user object. to do that, the user role needs to be stored in a nested property in an array inside of an `app_metadata` prop. the whole object looks like this, where `readonly` is the value from the **User role** radio button:

```javascript
{
  name: "John",
  email: "some@email",
  app_metadata: { roles: ["readonly"] },
}
```

By specifying the field name as `"app_metadata.roles.0"`, RHF is smart enough to update that nested array element with the value of this input. So we don't have to fuss with manually transforming the form payload object back into the shape that our API is expecting ✨

api/server.py
johnclary: Ah, of course! I completely forgot about the gateway. I'm so glad this is making sense now. 

Hmm, yes, I'm with you—I think wildcard is definitely fine for verbs. Maybe we should go ahead and whitelist our referrers?

In terms of the headers that this API is sending, I guess we should try to make these match the gateway exactly, and I'll add a code comment that mentions that these may be overwritten in the production deployment?

app/utils/users.ts
johnclary: I would love—love!—to move this API into the nextjs app potentially adding an tRPC. It will have to live in the backlog, but I'll open an issue for it.

api/server.py
johnclary: I was tending to agree that wildcard is fine, but like I mention elsewhere, maybe we should tighten up our game and specify our referrer hostnames?

app/components/UserModal.tsx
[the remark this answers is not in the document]
frankhereford: John, thanks for the explanation!! That is really slick -- doing that transformation would be brittle and have to be kept in sync with props changing and whatnot. React hook form sounds like a real champ. 

Async anything when it involves a remote resource is like timezones -- here's a short story from my past.

Me @ hour 1: How hard can it be, I'll just fetch() ...
Me @ hour 5: What has my life become, everything is happening out of order and I just found another edge case ...

api/server.py
frankhereford: That's 100% cool, and for the CORS headers -- you also can just flat ignore them. Maybe leave a comment that there is an API gateway in use and it handles it.

I'm fairlyh strongly confident that nothing the app does will change what CORS behavior we exhibit on the other side of the gateway.

johnclary: > I'm fairly strongly confident that nothing the app does will change what CORS behavior we exhibit on the other side of the gateway.

Well, we're stuck in this awkward spot where we need to local API to work without the gateway, so those allowed methods definitely need to be right 🙃 

I dunno—maybe it would be better to dedupe the header setting and put them all in our flask app? i don't think it's a big deal to manage them in two places, so long as we add comments here as a warning...

api/server.py
```diff
@@ -439,6 +464,12 @@ def user_create_user():
     user_dict = current_user._get_current_object()
     if isValidUser(user_dict) and hasUserRole(ADMIN_ROLE_NAME, user_dict):
         json_data = request.json
+        # set the user's password - user will have to reset it for access
+        json_data["password"] = get_secure_password()
```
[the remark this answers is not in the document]
roseeichelmann: thanks for fixing thissss

api/README.md
```diff
@@ -14,7 +14,20 @@ These flask apps are deployed as long-running tasks in ECS and are reverse proxy
...
+```
+NEXT_PUBLIC_CR3_API_DOMAIN=http://localhost:8085
+```
+
+You can start the API using either the project wide `docker compose` file with `docker compose up cr3-user-api` or if you prefer, you can use the `docker compose` stack that is concerned only with part of the stack as found in the `api` directory. Use whichever is best for your development needs. 
+
+Both docker compose files enable local development by:
+
+- mounting your local `/api` directory into the container
+- use the `--debug` command so that the web server restarts when it detetcs code changes
```
frankhereford, line 28: I saw this after I submitted my review: typo on 'detects'

api/README.md
```diff
@@ -14,7 +14,20 @@ These flask apps are deployed as long-running tasks in ECS and are reverse proxy
 
 ## Local usage
```
roseeichelmann, line 16: nbd but i would prefer if the config and secrets sections lived above local usage, first things first in a new app directory i wanna get the env secrets sorted out :shrug:

app/app/users/[user_id]/page.tsx
```diff
@@ -51,17 +57,52 @@ export default function UserDetails({
 }) {
   const token = useToken();
   const userId = params.user_id;
-  const { data: user } = useUser(userId, token);
+  const { data: user, mutate } = useUser(userId, token);
+  const [showNewUserModal, setShowNewUserModal] = useState(false);
+  const onCloseModal = () => setShowNewUserModal(false);
+
+  const onUpdateUsercallback = useCallback(async () => {
```
Comment (roseeichelmann) [inline app/app/users/[user_id]/page.tsx:64]: should the capitalization here be -> `onUpdateUserCallback`

app/components/UserModal.tsx
```diff
@@ -0,0 +1,197 @@
...
+import Alert from "react-bootstrap/Alert";
+import Button from "react-bootstrap/Button";
+import Form from "react-bootstrap/Form";
+import Modal from "react-bootstrap/Modal";
+import Spinner from "react-bootstrap/Spinner";
+import { useForm, SubmitHandler, Controller } from "react-hook-form";
+import { User, UserAPIError } from "@/types/users";
+import { useToken } from "@/utils/auth";
+
+interface UserModalProps {
```
roseeichelmann, line 11: loving this interface with the comments

app/components/UserModal.tsx
```diff
@@ -0,0 +1,197 @@
...
+          "Content-Type": "application/json",
+        },
+        method,
+        body: JSON.stringify(data),
+      });
+      const responseJson: User | UserAPIError = await response.json();
+      if ("user_id" in responseJson) {
+        await onSubmitCallback(responseJson);
+      } else if ("message" in responseJson) {
+        setErrorMessage(responseJson.message);
```
roseeichelmann, line 90: why are we checking for the word "message" here?

app/app/users/[user_id]/page.tsx
```diff
@@ -51,17 +57,52 @@ export default function UserDetails({
...
+    // refetch the user details and close
+    await mutate();
```
Comment (roseeichelmann) [inline app/app/users/[user_id]/page.tsx:66]: so `mutate` functions as a refetch?

roseeichelmann (approved):
john this looks great, i combed through the code and im getting excited to write my own typescript code soon with yours as a reference :sparkles: also super great that you ironed out the way the api app works, like fixing the password handling :))
a couple things either to address now or maybe create future issues for --
- i feel like read only/editor users just shouldnt have access to the users tab at all (it shouldnt appear in navbar) if they arent gonna be able to have any functionality on the page.
- also i think the error message could be more helpful for when you type in an email address wrong: 
Screenshot from 2024-12-12 12-57-49
Like maybe saying it must end with @austintexas.gov
- also i found it annoying that I would try to highlight and copy my email address while on the Users list page and when i would click and drag to highlight i would just immediately get rerouted to my user page. I like how currently in our users page you have to click on the name to reroute to that user

api/server.py
frankhereford: Oh! Of course - I fully hadn't considered that it's got to run locally too. 🙃 

Yes - everything you're saying makes perfect sense now and I am entirely cool with however you feel best to go forward -- 100%.

frankhereford:
> * Like maybe saying it must end with @austintexas.gov

I don't think it's too far fetched to have an outside consultant or industry expert getting access to the app, so I'd suggest we run any restrictions on domain names in the email addresses by Joel. 

I do agree with @roseeichelmann that the error message there is for programmers not users, but I'm not sure how easy it would be to get one that's more user-friendly. It's a pretty quiet, rarely used functionality in the app, so i wouldn't want to see a ton of work go into it, however.

I like Rose's idea also about about not showing read-only accounts the users tab at all, but I think i'd suggest we draw the line at the "Add User" button instead. A read only account may want to check to see if one of their team mates has access, before they sent them a link to a crash maybe. I think it'd be right to hide that button from anyone who can't use it (and of course, keeping the permission validation in the api handler too).

api/server.py
johnclary: Cool—I added code comments to remind us that the headers may be overwritten by our gateway 🙏

app/app/users/[user_id]/page.tsx
Comment (johnclary) [reply app/app/users/[user_id]/page.tsx:64]: yep—thanks!

app/app/users/[user_id]/page.tsx
Comment (johnclary) [reply app/app/users/[user_id]/page.tsx:66]: yes. with SWR, you can call `mutate()` to refetch, or "revalidate" in their terms. [Here](https://swr.vercel.app/docs/mutation#revalidation) are the docs.

they use term `mutate` because you can also use this function to update the in-memory data cache.

app/components/UserModal.tsx
johnclary: this is a hack for checking if we have an ok response or an error response, since only an error response will have a `message` property. I wrote this before I fixed the problem of the API returning a `200` status with errors.  I reworked this to use the `response.status` code 🙏

johnclary:
I just pushed an improvement to the error handling, so that we catch that invalid email address error and push the error into the form state as well 👇 

Screenshot 2024-12-15 at 1 54 44 PM

I will update follow-up issues to:
- do away with the row click behavior and use a url in the name column instead. i agree that we should preserve the user's ability to select/copy from this table
- add permission-based visibility to this part of the app. this is on the todo list (and exists in the old VZE) but i haven't opened an issue for it yet 🙏

PR #1628, opened 2024-12-10
Crash details: charges related records table

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/19964

## Testing

**URL to test:** 

http://localhost:3000/crashes/20371838
here is a crash with a charge, also crash number 20392352

**Steps to test:**

There is now a charges table on the crash page. 
Clicking on the rows does nothing, they are view only. 

Because these rows are read only, there is no mutation. I am passing an empty string as the mutation. I tried specifying that the mutation string is optional, but that created an error because `useMutation` (https://github.com/cityofaustin/vision-zero/blob/john/20214-remove-zod/app/components/RelatedRecordTableRow.tsx#L41) does not accept undefined. 

---

#### Ship list

- [ ] Code reviewed


johnclary (approved):
I see charges! Very cool to have this part of the crash details page filled in. I will come back with issue numbers for two follow-up tasks:

1. Show a message if there no rows in the related records table. "No charges records", "No unit records", etc.
2. Enable the `mutation` prop on the `RelatedRecordsTable` to be null

🚢 🚢 🚢 🚢 🚢

frankhereford (approved):
Another slam dunk of a PR; wouldn't change a character. 

Thank you Chia! Your code is ✨ and the testing was 👍.

PR #1632, opened 2024-12-16
Fix ability to clear table search and refine layout 

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/20310

Fixes a bug with the search input not being clearable. And some light refinements to the UI 👇 

Screenshot 2024-12-16 at 2 18 26 PM

## Testing

**URL to test:** Local 

**Steps to test:**

1. Navigate to the crashes page and put some text into the search box, but don't hit the **Search** button
2. Use your keyboard to clear the search input. Observe that it can be cleared

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/app/components/TableSearch.tsx
+++ b/app/components/TableSearch.tsx
@@ -33,7 +33,7 @@ export default function TableSearch({
         <FaMagnifyingGlass />
       </InputGroup.Text>
       <Form.Control
-        placeholder="Find a crash..."
+        placeholder="Search..."
         aria-label="Crash search"
         aria-describedby="search-icon"
         onChange={(e) => {
@@ -50,10 +50,9 @@ export default function TableSearch({
               return newQueryConfig;
             });
             setQueryConfig(newQueryConfig);
-          } else {
-            searchSettings.searchString = e.target.value;
-            setSearchSettings({ ...searchSettings });
           }
+          searchSettings.searchString = e.target.value;
+          setSearchSettings({ ...searchSettings });
         }}
         value={searchSettings.searchString}
         type="search"
--- a/app/components/TableSearchFieldSelector.tsx
+++ b/app/components/TableSearchFieldSelector.tsx
@@ -6,7 +6,6 @@ export default function TableSearchFieldSelector({
   setSearchSettings,
   queryConfig,
 }: TableSearchProps) {
-  // todo: bug here where changing the search field clears the search input if it hasn't been searched
   return (
     <>
       <Form.Label className="fw-bold me-2 my-0">Search by </Form.Label>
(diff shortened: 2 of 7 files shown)


frankhereford (approved):
Pinch me, this looks so good. 

Having the list update when you remove the last character out of the search box but not before that feels great. The UI is polishing up real fine ✨ and the code is primo.


Here's an idea beyond the scope of this work 👇
 
What would you think about debouncing keyboard input events on that search box and executing the search automatically, say, after 250 milliseconds elapse since the last key press? So you get that "live searching / search as you type" feel?

app/components/TableSearchFieldSelector.tsx
```diff
@@ -9,7 +9,7 @@ export default function TableSearchFieldSelector({
   // todo: bug here where changing the search field clears the search input if it hasn't been searched
```
chiaberry, line 9: is the bug still there? maybe i dont understand how to get the bug to happen, but i couldnt reproduce

chiaberry (approved):
i like the style tweaks. Couldnt tell if it felt weird that number of filters applied badge isnt square/symmetrical but badges arent supposed to be symmetrical. 

I tried to come up with a way to make the filters button visually change if the filters were open (like a visual cue that clicking it will close the filter dropdown) but i cant come up with anything.

app/components/TableSearchFieldSelector.tsx
johnclary: looks like it's fixed! 🙏

app/components/TableSearch.tsx
```diff
@@ -50,10 +50,9 @@ export default function TableSearch({
               return newQueryConfig;
             });
             setQueryConfig(newQueryConfig);
-          } else {
-            searchSettings.searchString = e.target.value;
-            setSearchSettings({ ...searchSettings });
           }
+          searchSettings.searchString = e.target.value;
+          setSearchSettings({ ...searchSettings });
```
johnclary, line 55: @mateoclarke the fix was up here. the bug was that we were skipping the update to `searchSettings` when the input was cleared. 

btw, `searchSettings` is holding the local state of the search string and search field, and clicking the **Search** button or clearing the input completely is what triggers the `queryConfig` state to be updated and fire off a new Hasura API call.

johnclary:
@frankhereford thanks for that suggestion—I think that would be a really nice enhancement. I spent a few minutes trying to implement it but realized the config settings would need a bigger reworking than was reasonable to ship in this PR. I opened https://github.com/cityofaustin/atd-data-tech/issues/20353.

frankhereford:
@johnclary wrote:

> need a bigger reworking than was reasonable to ship in this PR

I totally get that John, thanks for looking it, and double thank you for opening a backlog issue for it. 🙏🙏🙏

johnclary:
> I tried to come up with a way to make the filters button visually change if the filters were open (like a visual cue that clicking it will close the filter dropdown) but i cant come up with anything.

@chiaberry i see what you mean. we could change the filters menu to a down arrow. or label the button "Show filters" or "Hide filters"

Maybe this full-width accordion UI design is just doomed. It does't look great on large monitors! A modal might be better?

PR #1634, opened 2024-12-18
Add summary statistics to location details and rework column metadata types

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/20354
* https://github.com/cityofaustin/atd-data-tech/issues/20316

Screenshot 2024-12-19 at 1 10 51 PM

This PR restructures the typing and config structure for managing editable columns in the app. It makes our config + components more abstract, so that they can better handle arbitrary relationships rather than the hardcoded "lookup table" relationship.

One additional change that I would like to make, but have not in the interest of the code diff, is to stop using the word "column" where we really mean "field". This came up in some PR feedback from Chia and Rose recently. 

Thank you for reviewing! Since this is so large, I'd be happy to set up a time to pair or review as a group interactively.

## Testing

**URL to test:** [Netlify](https://deploy-preview-1634--vision-zero-nextjs.netlify.app/crashes)

**Steps to test:**

1. Navigate to a location's details page and verify that the **Details** card is rendering normally.
2. Head over to a crash details page, and try editing a variety of fields, including fields that use a relationship (e.g. **Collision Type**), a yes/no field, and fields in the **Units** table.

---

#### Ship list

- [ ] Code reviewed


app/app/locations/[location_id]/page.tsx
```diff
@@ -46,10 +46,10 @@ export default function LocationDetailsPage({
       <AppBreadCrumb />
       <span className="fs-2">{location.description}</span>
       <Row>
-        <Col sm={12} md={8} className="mb-3">
+        <Col sm={12} md={6} lg={7} className="mb-3">
           <LocationMapCard location={location} />
         </Col>
-        <Col sm={12} md={4} className="mb-3">
+        <Col sm={12} md={6} lg={5} className="mb-3">
```
Comment (johnclary) [inline app/app/locations/[location_id]/page.tsx:52]: I like making the location map a bit wider than in the old VZE—just trying to strike a balance to prevent the card form sidescrolling

app/components/DataCard.tsx
```diff
@@ -37,7 +37,12 @@ export default function DataCard<T extends Record<string, unknown>>({
   // todo: handling of null/undefined values in select input
   const [editColumn, setEditColumn] = useState<ColDataCardDef<T> | null>(null);
   const { mutate, loading: isMutating } = useMutation(mutation);
-  const [query, typename] = useLookupQuery(editColumn?.lookupTable);
+  const [query, typename] = useLookupQuery(
+    editColumn?.editable && editColumn?.relationship
+      ? editColumn.relationship
+      : undefined
+  );
```
johnclary, line 44: we only need to use the lookup query when a column is editable and it has relationship metadata

app/components/DataCard.tsx
```diff
@@ -47,10 +52,18 @@ export default function DataCard<T extends Record<string, unknown>>({
...
 
   const onSave = async (value: unknown) => {
+    if (!editColumn) {
+      // not possible
+      return;
+    }
+    // Save the value to the foreign key column, if exists
+    const saveColumnName = editColumn.relationship?.foreignKey
+      ? editColumn.relationship?.foreignKey
+      : editColumn.path;
```
johnclary, line 62: this hints at the awkward dance we need to do between _rendering_ a value from an object relationship. E.g. `crash. collsn.label` vs. editing that value, which requires us to update the foreign key at `crash.fhe_collsn_id`.

This PR is trying to iron this out by moving the necessary metadata into the column `relationship` config, whereas before we had some special lookup table logic hardcoded into various functions. 

One important thing that these changes do is open up future iterations where we are editing relationships that are not just plain old lookup tables. E.g., if we wanted a user to select an EMS record from a list so that it can be linked to a crash.

app/configs/chargeColumns.tsx
```diff
@@ -1,21 +1,21 @@
...
+    path: "charge",
     label: "Charge",
   },
   citation_nbr: {
-    name: "citation_nbr",
+    path: "citation_nbr",
     label: "Citation Number",
   },
-};
+} satisfies Record<string, ColDataCardDef<Charge>>;
```
johnclary, line 21: [here are the docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html) about Typescript satisfies. it seems like it was designed for this exact scenario, it lets TypeScript infer this object's keys while still checking that column config matches the `ColDataCardDef`

this change fixes a type bug down in `crashDataCard` 👇

app/configs/crashDataCard.tsx
```diff
@@ -1,13 +1,11 @@
...
-
-/** Construct card-specific arrays of columns */
 export const crashDataCards = {
   summary: [
     crashesColumns.case_id,
     crashesColumns.crash_timestamp,
-    crashesColumns.fhe_collsn_id,
-    crashesColumns.rpt_city_id,
+    crashesColumns.collsn,
+    crashesColumns.city,
```
johnclary, line 8: before adding that `satisfies` magic to the column indexes (see my other comment), there was no type checking on these column names. You could have added a column to the summary section called `crashColumns.pizza` and typescript would not have blinked.

app/configs/crashesColumns.tsx
```diff
@@ -2,255 +2,259 @@ import { ColDataCardDef } from "@/types/types";
...
     editable: true,
     inputType: "select",
-    relationshipName: "collsn",
-    lookupTable: {
+    relationship: {
       tableSchema: "lookups",
       tableName: "collsn",
+      idColumnName: "id",
+      labelColumnName: "label",
+      foreignKey: "fhe_collsn_id",
```
johnclary, line 51: here is the new `Path` type in action, along with `relationship` metadata. the very cool thing here is that the `path` property is type-checked against the `Crash` type. so `"collsn.label"` is valid but `collsn.pizza` would not be. 

so this new `Path` type unlocks adding column metadata for nested properties in a typesafe way 🌈

app/configs/locationColumns.ts
```diff
@@ -1,16 +1,28 @@
...
   },
-  // todo: need to support non-lookup table relationships
-  //   cr3_crash_count: {
-  //     name: "cr3_crash_count",
-  //     relationshipName: "locations_list_view"
-  //   }
+  cr3_crash_count: {
+    path: "locations_list_view.cr3_crash_count",
+    label: "CR3 crashes (5 years)",
+  },
```
johnclary, line 14: again, here you can see the path magic in action, so we can now render this nested value on the location details card. not how the `relationship` property is absent—we don't need it because this field is not editable.

app/types/relationships.ts
```diff
@@ -0,0 +1,33 @@
...
+  /**
+   * The column to be used to label each lookup option
+   */
+  labelColumnName: string;
+}
+
+export type LookupTableOption = {
+  id: string | number;
+  label: string | number;
+};
```
johnclary, line 33: so now we have the generic `Relationship` type which covers lookup tables and any other object relationship that a record might have.

and we still have the `LookupTableOption` to describe the `id`/`label` structure of our conventional lookup tables.

app/types/utils.ts
johnclary, line 1: yeah, this new `Path` type is definitely typescript hard mode. I link to a 3rd-party lib that does something similar, which I wouldn't mind reaching for but it does have more functionality than we need.

app/utils/formHelpers.ts
```diff
@@ -75,47 +110,45 @@ const renderYesNoString = (value: unknown): string => {
 };
 
 /**
- * Return the `label` column from a related lookup table
- */
-export const renderLookupLabel = <T extends Record<string, unknown>>(
```
johnclary, line 80: here's one of those functions that was hardcoded to deal with with our lookup tables in the `lookups` schema, but not for other types of relationships.

app/utils/graphql.ts
```diff
@@ -153,31 +153,38 @@ export const useMutation = (mutation: RequestDocument) => {
...
 
     return [
       gql`
         query LookupTableQuery {
-          ${typename}(order_by: {label: asc}) {
-            id
-            label
+          ${typename}(order_by: {${relationship.labelColumnName}: asc}) {
+            id: ${relationship.idColumnName}
+            label: ${relationship.labelColumnName}
```
johnclary, line 184: like i mention in the JSdoc for this hook—we are aliasing the `id` and `label` properties of this query builder, so that any lookup table we load will have the same shape. right now, we're only using lookups in the `lookups` schema in our db, which all of the same shape, but if we wanted to get lookup values to an EMS record, we could do that as well and still use the same UI form components.

app/configs/chargeColumns.tsx
johnclary, line 1: i am also looking for feedback on the naming conventions and file organization around these configs. we have a few different kinds of configs:
- configs that organize column metadata (the `ColDataCardDef<T>` objects)
- configs that consume the column configs ☝️ and define which UI components render which columns (cards and related tables)
- the `queryConfig` objects that plug into the graphql table component for our crashes and locations list views

app/components/DataCard.tsx
johnclary: ah, right—one other wrinkle that is not new to this PR is that in order to be able to edit lookup values our graphql queries need to include both the foreign key column and the related record itself.

For example, in the `Crash` query, you can see [here](https://github.com/cityofaustin/vision-zero/blob/7e95be2e3264d6bae0f3aeacdab0f9a9bf701bd2/app/queries/crash.ts#L20-L24) that we have

```graphql
     # this is the foreign key to the lookups.collsn table
      fhe_collsn_id
      # this is the related collision type lookup value
      collsn {
        id
        label
      }
```

If you were to omit the foreign key from the query, the `select` input in the UI would never correctly show the currently selected lookup option. So it would be fairly obvious that something was wrong. This is not documented anywhere ☹️

app/components/DataCard.tsx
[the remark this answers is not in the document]
chiaberry: keeping it separated in a different PR would be good, so its not confused with any other enhancements in a PR (basically so **I** am not confused)

mddilley (commented):
This looks great, and I learned a lot. I like the new abstraction too! The location table shows the statistics just as expected, but I ran into one question about the light condition field in the crash view that I wasn't sure about. Other than that, it looks 🚢!

app/configs/chargeColumns.tsx
johnclary: i think that's a great idea. i added a todo for this  🙏

app/configs/crashesColumns.tsx
```diff
@@ -2,255 +2,259 @@ import { ColDataCardDef } from "@/types/types";
...
     inputType: "number",
   },
   law_enforcement_ytd_fatality_num: {
-    name: "law_enforcement_ytd_fatality_num",
+    path: "law_enforcement_ytd_fatality_num",
     label: "Law Enforcement YTD Fatal Crash",
     editable: true,
     inputType: "text",
   },
   light_cond_id: {
```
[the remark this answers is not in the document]
johnclary: @mddilley thanks for flagging this. I noticed two address fields that have the same issue—and i think that all of these should be editable. I opened a follow-up snackoo: https://github.com/cityofaustin/atd-data-tech/issues/20430

🙏

mddilley (approved):

app/components/DataCard.tsx
chiaberry: "fairly obvious" 😳

chiaberry (approved):
really nice to learn about the satisfies magic! and the level up with nested paths  
💖

app/configs/chargeColumns.tsx
johnclary: Nice post 🙏

PR #1635, opened 2024-12-20
Add zoom & rotate functionality to crash diagrams

frankhereford:
## Associated issues

This PR hopes to close https://github.com/cityofaustin/atd-data-tech/issues/19967.

## Ask forgiveness, not permission

Here's the guiding image:

Image 394877044 924x1134

Building out from the image mentally --  I took some liberties, but have done my best to stay true to the new nextjs visual theme and aesthetic. I'm very 👍 on the VZE style and hope we perfect it.

But to be specific, here's some of the things that are big enough changes from the image to point out. I'm really leaning into only including the elements indicated by the arrows in the guiding image.

* Recolor the button controls in the `primary` color like the 'Edit' button
* Increase the size of the iconography in the buttons.
* Replace the "do again" symbol with the "undo" symbol, but I did have to grab an icon from another font, not `bootstrap`. I think it matches really well.
* Drop the redundant "Reset" button from near the rotation control. I wired up the top-right "undo" button to reset both zoom and rotation. 
* Drop the instructive label on the rotation handle. Users are repeat users for us, and any curious person will teach themselves what this harmless handle does real quick, and we can avoid the clutter.

## Testing

* Grab the deploy preview link from below in the 🤖 comment and check out some crash diagrams. 
* Do all the things, see if you can break them.
* Consider the stylistic choices I have made. 
  * Does this belong on this page? 
  * Are you confused about what anything means...
    *  ... after you play with it for a few seconds?
* Visit a temp crash page to see what error messages look like. I found these crashes by loading local dev with production data.
  * Look good? 
  * Are you cool that you can slide them around a little like an image, but they snap back?
    * I kinda dig how it's a harmless fidget toy in a way, and it's kinda hard to disable that functionality, I think. I do hide the  image manipulation buttons though.

## PS: DX experience

I absolutely felt a pang of joy when I could click a link in the CI PR comment and get a build log from Netlify that clearly indicated what the build problem was, like only typescript will do. It was so easy to find and fix where I went sideways. Typescript rocks! 🐐 

---

#### Ship list

- [ ] Code reviewed


app/components/CrashDiagramCard.tsx
```diff
@@ -2,63 +2,165 @@ import { useState } from "react";
 import Alert from "react-bootstrap/Alert";
 import Card from "react-bootstrap/Card";
 import Image from "react-bootstrap/Image";
+import Button from "react-bootstrap/Button";
+import ButtonGroup from "react-bootstrap/ButtonGroup";
+import { BsZoomIn, BsZoomOut } from "react-icons/bs";
```
frankhereford, line 7: I added `react-icons` which has bootstrap and a jillion other ones available. I like how you import them here, so you only get the overhead of the buttons you want and not the whole menagerie.

app/components/CrashDiagramCard.tsx
```diff
@@ -2,63 +2,165 @@ import { useState } from "react";
...
+import { SlActionUndo } from "react-icons/sl";
+import {
+  TransformWrapper,
+  TransformComponent,
+  useControls,
+} from "react-zoom-pan-pinch";
```
frankhereford, line 13: @tillyw's choice on this library 4 years ago was on point; it's still the go-to.

app/components/CrashDiagramCard.tsx
```diff
@@ -2,63 +2,165 @@ import { useState } from "react";
...
 import { Crash } from "@/types/crashes";
 
-// import axios from "axios";
```
frankhereford, line 7: I zapped this axios library comment; I don't think we need it. Right?

app/components/CrashDiagramCard.tsx
```diff
@@ -2,63 +2,165 @@ import { useState } from "react";
...
-// import { TransformWrapper, TransformComponent } from "react-zoom-pan-pinch";
-
 const CR3_DIAGRAM_BASE_URL = process.env.NEXT_PUBLIC_CR3_DIAGRAM_BASE_URL!;
```
frankhereford, line 17: So, here's a style question, I have no idea. Is it good style to include these "sub-components", if you will, in this file? They are only going to be used here, so it seems to me it's good to keep them here where in the context of using them. Why pull them out if they will never be used anywhere else?

Obviously -- I will refactor this in a heartbeat based on y'all's advice.

app/components/CrashDiagramCard.tsx
```diff
@@ -2,63 +2,166 @@ import { useState } from "react";
...
+                  onError={() => {
+                    console.error("Error loading CR3 diagram image");
+                    setDiagramError(true);
+                  }}
+                />
+              )}
+              {diagramError && crash.is_temp_record && (
+                <Alert variant="info" style={{ marginTop: "50px" }}>
+                  <i className="fa fa-info-circle" />
+                  Crash diagrams are not available for temporary crash records
```
frankhereford, line 136: I changed this language. It felt accusatory that this wouldn't work because it was "user-made." This feels more like a statement of fact using commonly understood terms.

app/components/CrashDiagramCard.tsx
```diff
@@ -2,63 +2,166 @@ import { useState } from "react";
...
+                </Alert>
+              )}
+              {diagramError && !crash.is_temp_record && (
+                <Alert variant="danger" style={{ marginTop: "20px" }}>
+                  <p>The crash diagram is not available.</p>
```
frankhereford, line 141: I shortened this language as well. The extra sentence .. 

>Typically, this indicates there was an error when processing this crash&aposs CR3 PDF.

.. is unusual because it is speculating on an error instead of reporting it.

app/components/CrashDiagramCard.tsx
```diff
@@ -2,63 +2,166 @@ import { useState } from "react";
...
-              >
-                report a bug
-                <i className="fa fa-external-link"></i>
-              </a>
-              .
-            </p>
-          </Alert>
-        )}
+        <TransformWrapper>
+          {!diagramError && <ZoomResetControls setRotation={setRotation} />}
```
frankhereford, line 113: Withhold image manipulation tooling if an error is being shown.

app/package-lock.json
frankhereford, line 1: I'm pretty sure, but it's correct to include the lock file when you add a new library right?

app/package.json
```diff
@@ -26,8 +26,9 @@
     "react-datepicker": "^7.5.0",
     "react-dom": "^18",
     "react-hook-form": "^7.53.2",
-    "react-icons": "^5.3.0",
+    "react-icons": "^5.4.0",
```
frankhereford, line 29: I just "installed" this blindly with `npm install` and it looks like I have updated it instead. It's a minor bump, so I'm going to leave it, but I do want to flag it because it's not impossible i regressed something somewhere. Feels like a safe enough bet to me though.

app/components/CrashDiagramCard.tsx
chiaberry: Could there be the case though that the crash diagram is not related for a non-temporary crash? Like if there is a processing error or some other error?

chiaberry: ah now I scrolled down and see that you included a different message. I should scroll more before i speak

app/components/CrashDiagramCard.tsx
```diff
@@ -2,63 +2,164 @@ import { useState } from "react";
...
+    <div
+      className="tools"
+      style={{
+        display: "flex",
+        justifyContent: "space-between",
+        width: "100%",
+        marginBottom: "5px",
+      }}
+    >
+      <div>
```
chiaberry, line 47: is this div necessary?

chiaberry (commented):
My comments are about the UX and not the code, the code is 👍🏼 

I was playing with it in existing prod and noticed that the reset and undo buttons do two different things, resetting the rotation and the undo button resets the zoom. I do not know if users would like to be able to keep that functionality separate. Like if you rotated a specific way, zoomed out and then wanted to keep the rotation but reset the zoom. 

Speaking of zooming out, I can't zoom out on the preview link unless ive already zoomed in. I can zoom out on prod. 

The issue doesn't explicitly say to include the download cr3 button in this same card, I went to look for it on the ACC figma designs and I couldnt find it at first. Because they changed it into a little save icon at the bottom, which at first glance I don't like. They also combine the reset buttons into one like you did Frank, so maybe my desire for more fine tuned controls isn't what our users want anyway.

app/components/CrashDiagramCard.tsx
frankhereford: It's what I use to slide the zoom buttons over to the left while leaving the undo button on the right.

frankhereford:
@chiaberry wrote:

> I went to look for it on the ACC figma designs

I thought about trying to find them, but I didn't know where to start on that, and the issue itself doesn't mention using them -- it doesn't mention anything. So I fell back to doing just what the issue said (3 arrows), and I liked where it was going, so I stuck with that.

frankhereford:
@chiaberry wrote:

> Speaking of zooming out, I can't zoom out on the preview link unless ive already zoomed in. I can zoom out on prod.

I spent a good little chunk of time working on the image sizing, mainly focusing on making sure the entire image would be visible on first load, and by the nature of that, it's at its most meaningful "zoomed out" state on load. I think being able to zoom out to make the image smaller and less detailed than what it shows on-load an anti-feature. Just my two cents; I can probably figure out a way to make it zoom-out-able by starting with it zoomed in at 200% and then using CSS to scale that down to fit in the element -- something like that. Seems counter productive to me though.

app/components/CrashDiagramCard.tsx
johnclary: yes! great package—thanks for bumping us up a version ;)

For the time being, would you mind switching to the `fa6` icons? I kinda want to be consistent with font awesome v6 until proven otherwise. (Partially because I am considering requesting that we purchase the whole icon set so we can have the much nicer outline and "light" styles).

Here are those imports!

```javascript
import { FaMagnifyingGlassPlus } from "react-icons/fa6";
import { FaMagnifyingGlassMinus } from "react-icons/fa6";
```

app/components/CrashDiagramCard.tsx
johnclary: great to hear that!

app/components/CrashDiagramCard.tsx
johnclary: 💯 🙏

app/components/CrashDiagramCard.tsx
johnclary: personally, i don't mind having them in here. if they grew in size or if other's prefer not, we can stash things in `./utils` and `./configs`. 

i know there has been a lot of though put into file org in Moped recently too—i am not married to this convention if we want to switch things up 🙏

app/components/CrashDiagramCard.tsx
```diff
@@ -2,63 +2,162 @@ import { useState } from "react";
...
+    resetTransform();
+    setRotation(0);
+  };
+
+  // pixel pushing
+  const iconStyle = {
+    position: "relative" as const,
+    top: "-1px",
+    fontSize: "1.3em",
+  };
```
johnclary, line 35: would you mind taking a look at our other icon + button components and matching their style? you should be able to get away with using zero `style` props here vs the utility classes—and the `AlignedLabel` component might be helpful too.

And ff you feel like we need a generic "IconButton" component—by all means go for it.

And if we're still unhappy with look + feel without pixel pushing, I am thinking we can add these kinds of todos in https://github.com/cityofaustin/atd-data-tech/issues/20013—so that we can hopefully solve them with SCSS in a centralized file 💅

app/components/CrashDiagramCard.tsx
```diff
@@ -2,63 +2,162 @@ import { useState } from "react";
...
+
+  return (
+    <div
+      className="tools"
```
johnclary, line 39: is this `className` needed?

app/components/CrashDiagramCard.tsx
```diff
@@ -2,63 +2,162 @@ import { useState } from "react";
...
+      style={{
+        display: "flex",
+        justifyContent: "space-between",
+        width: "100%",
+        marginBottom: "5px",
+      }}
```
johnclary, line 45: you would mind replacing these styles with the bootstrap utility classes? you've picked the three i've found myself reaching for the most (flex, justify, and margin). Something like this should get you there: `"d-flex justify-content-between mb-2"` 🚀

app/components/CrashDiagramCard.tsx
```diff
@@ -2,63 +2,162 @@ import { useState } from "react";
...
+    >
+      <ButtonGroup>
```
johnclary, line 47: Button group looks great 🚀

app/components/CrashDiagramCard.tsx
```diff
@@ -2,63 +2,162 @@ import { useState } from "react";
...
+const RotateControls = ({
+  rotation,
+  setRotation,
+}: {
+  rotation: number;
+  setRotation: (value: number) => void;
+}) => {
+  const rotate = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setRotation(Number(event.target.value));
+  };
```
johnclary, line 81: lovely ts here ✨

app/components/CrashDiagramCard.tsx
```diff
@@ -2,63 +2,162 @@ import { useState } from "react";
...
+
+  return (
+    <div className="rotate-controls" style={{ marginTop: "10px" }}>
```
johnclary, line 84: ditto here—`rotate-controls` can possibly be removed, and swap out the styles for the margin utility class 🔧

app/components/CrashDiagramCard.tsx
```diff
@@ -2,63 +2,162 @@ import { useState } from "react";
...
+        type="range"
+        min="-180"
+        max="180"
+        value={rotation}
+        className="form-control-range"
+        id="formControlRange"
+        onChange={rotate}
+        style={{ width: "70%" }}
+        title="Rotate Diagram"
+      />
```
johnclary, line 95: TIL we actually have a `Form.Range` [component](https://react-bootstrap.netlify.app/docs/forms/range) from react-boostrap. Would you mind swapping that in here?

app/components/CrashDiagramCard.tsx
```diff
@@ -2,63 +2,162 @@ import { useState } from "react";
...
+  useControls,
+} from "react-zoom-pan-pinch";
 import { Crash } from "@/types/crashes";
 
-// import axios from "axios";
-// import { TransformWrapper, TransformComponent } from "react-zoom-pan-pinch";
-
 const CR3_DIAGRAM_BASE_URL = process.env.NEXT_PUBLIC_CR3_DIAGRAM_BASE_URL!;
 
+const ZoomResetControls = ({
```
johnclary, line 18: very cool seeing how this functionality works!

app/components/CrashDiagramCard.tsx
johnclary: I like this change—thank you!

app/components/CrashDiagramCard.tsx
```diff
@@ -2,63 +2,162 @@ import { useState } from "react";
...
+                  src={`${CR3_DIAGRAM_BASE_URL}/${crash.record_locator}.jpeg`}
+                  alt="crash diagram"
+                  onError={() => {
+                    console.error("Error loading CR3 diagram image");
+                    setDiagramError(true);
+                  }}
+                />
+              )}
+              {diagramError && crash.is_temp_record && (
+                <Alert variant="info" style={{ marginTop: "50px" }}>
```
johnclary, line 130: nice—i agree that moving this alert toward the vertical center of the card is a better look, but i would prefer to use flexbox utilities to center this vertically. we have a few uses of `align-items-center` if you want to check that out 👀

app/components/CrashDiagramCard.tsx
johnclary: I'm with you—shorter is sweeter. One thing I was trying to convey with the old message is that we should have 100% CR3 PDF coverage for all crashes except temp crashes or crashes that occurred 10+ years ago. And otherwise we kind of do want a user to let us know if a PDF is missing. 

I guess we could over-engineer this and check the `crash_timestamp` to further refine these errors?

app/components/CrashDiagramCard.tsx
```diff
@@ -2,63 +2,162 @@ import { useState } from "react";
...
+                    </a>
+                    .
+                  </p>
+                </Alert>
+              )}
+            </div>
+          </TransformComponent>
+        </TransformWrapper>
       </Card.Body>
-      <Card.Footer>Something else here</Card.Footer>
```
johnclary, line 61: 🎉

johnclary (changes requested):
Frank—it's so awesome to have this feature come to life. It works like a charm. And it's awesome to have you shipping on VZE!!

> Consider the stylistic choices I have made.

They're excellent—thanks for the subtle UX improvements — this looks sharp and intuitive.


> Are you cool that you can slide them around a little like an image, but they snap back?

Yep—totally fine!

One minor detail is that I'm getting a vertical scrollbar inside the card body. Can you track down the source of that unwanted pixel push?

Screenshot 2024-12-20 at 12 13 45 PM

The rest of my feedback is really just asking you to reach for the bootstrap utility classes instead of hardcoded styles. Thanks again! 🙏

app/components/CrashDiagramCard.tsx
frankhereford: Oh yea, for sure. Thanks for these. They look good. Do you dig the "undo" button, or do you have a recommendation for that?

app/components/CrashDiagramCard.tsx
frankhereford: OK cool - let's leave them in here so this functionality remains contained in this file and no one has to hunt for anything. If we change our mind at any time, no big, and we can shuffle it around then.

app/components/CrashDiagramCard.tsx
frankhereford: I'm game with whatever you feel is best, but I am still in support of the short'n'sweet message. If you do want to define some edge cases, we can probably step back a little and pull this erroring out into a component or something to tighten up this chunk of JSX.

frankhereford:
@johnclary wrote:

> One minor detail is that I'm getting a vertical scrollbar inside the card body. Can you track down the source of that unwanted pixel push?

Good call! I had my viewport set at 110% zoom for some reason which was masking this for me. https://github.com/cityofaustin/vision-zero/pull/1635/commits/ae3eaa246f717afc64b87c434119d5ec1e624cbc

app/components/CrashDiagramCard.tsx
frankhereford: Yea for sure! I didn't pick the hard-coded styles over the bootstrap utility on purpose -- i didn't even think of the bootstrap utility existing. I still have very little intuition on bootstrap since it's been every bit of 5ish years since I used it meaningfully. I'm super thankful for these helpful hints while I build that back.

https://github.com/cityofaustin/vision-zero/pull/1635/commits/6f8488555333f4999835232a55a49f929620395b

app/components/CrashDiagramCard.tsx
frankhereford: TIL x 2. Thanks!

https://github.com/cityofaustin/vision-zero/pull/1635/commits/1e533c6f275f1c822a4c2a77f8ec30ee8a89e376

frankhereford: You bet - TIL x 2. https://github.com/cityofaustin/vision-zero/pull/1635/commits/1e533c6f275f1c822a4c2a77f8ec30ee8a89e376 

Thanks!

app/components/CrashDiagramCard.tsx
frankhereford: Thanks @johnclary! 

Can you clarify -- you noted that you prefer it scooted up to the top as a better look, but ask for `align-items-center`. The scooted-up version would be using `align-items-start` i think, and if i use `align-items-center`, it's centered at the midpoint of the vertical space. 

Also, I don't think bootstrap has a utility that let's me define a custom width of 490px, so I can center properly. Any ideas?

PS: This is like my tRPC comment -- I fully dig how we're doing it, but here's a thing I loved about tailwind. They have a solution for when you really need a hardcoded value, like here, you'd be able to use the class `w-[490px]` and it would end up making the custom width utility for you out of that.

app/components/CrashDiagramCard.tsx
johnclary: ah, right...I can't find one that i prefer over what you've got. let's keep it as an exception to the font-awwesome preference 🙏

app/package-lock.json
johnclary: for sure!

app/components/CrashDiagramCard.tsx
frankhereford: https://github.com/cityofaustin/vision-zero/pull/1635/commits/584d84f57f28e37f2be8a8aaea1dc4e632fd2be2

Thanks John -- I looked at buttons from around the app by searching for <Button> and I found plenty of examples where we had an pictograph + text, and because you mentioned `AligntedLabel`, I'm taking it to mean that you want some labels on these buttons. Right?

How's this look?

app/components/CrashDiagramCard.tsx
frankhereford: Worked real fine - thank you!

app/components/CrashDiagramCard.tsx
```diff
@@ -2,63 +2,161 @@ import { useState } from "react";
...
-                <i className="fa fa-external-link"></i>
-              </a>
-              .
-            </p>
-          </Alert>
-        )}
+        <TransformWrapper initialScale={1}>
+          {!diagramError && <ZoomResetControls setRotation={setRotation} />}
+          <TransformComponent>
+            <div style={{ height: "330px", width: "100%", overflow: "hidden" }}>
```
johnclary, line 105: Frank, this is looking better, but I think we can get it to a place where we're not hardcoding any `div` sizes. 

Looking at the html markup, I think we can probably nix the below div completely, sine the `TransformWrapper` and `TransformComponent` are adding their own divs.

Here's a quick demo of leaning into the flex utilities for this. By making the `Card.Body` a flex-column, the rest of the divs should stack up without overflowing. I think you can also strip out the height/width on the `Image` component as well—since the `fluid` prop is going to let it fill the parent container.

```jsx
   <Card.Body className="crash-header-card-body text-center d-flex flex-column">
        <TransformWrapper initialScale={1}>
          {!diagramError && <ZoomResetControls setRotation={setRotation} />}
          <TransformComponent>
            {!diagramError && (
              <Image
                fluid
                style={{
                  transform: `rotate(${rotation}deg)`,
                }}
                src={`${CR3_DIAGRAM_BASE_URL}/${crash.record_locator}.jpeg`}
                alt="crash diagram"
                onError={() => {
                  console.error("Error loading CR3 diagram image");
                  setDiagramError(true);
                }}
              />
            )}
```

The more time I have been spending with bootstrap, the more I find that flexboxes are usually going to solve most layout problems, especially for growing to fit and managing vertical and horizontal alignment. It's always a bunch of trial and error, but the [bootstrap docs](https://getbootstrap.com/docs/5.3/utilities/flex/) have some nice visual helpers.

roseeichelmann (changes requested):
This is looking great Frank! Love seeing everyone's typescript and i learn a little from each PR i review :sparkles: 
 I like your UX choices to cut down on the number of reset buttons we have, i think having just one that resets the image completely is perfect.

I noticed that when I click on crashes with an error banner, the zoom/reset tools and the words "crash diagram" will be on the screen for awhile before the banner takes over. Sometimes its for long enough that i can even move around and zoom on the words and then some buggy behavior starts happening. If you could prevent those tools and those words from popping up before the banner shows that would be awesome! I'm also getting the pesky horizontal scroll on the temporary banner when im at 80% page zoom.
[Screencast from 01-06-2025 01:10:09 PM.webm](https://github.com/user-attachments/assets/03bf25ad-5143-4563-ba04-69e7bc6582ab)

On the other banner, if you could center it horizontally that would scratch a great itch for me too heh
image

roseeichelmann (changes requested):
This is looking great Frank! Love seeing everyone's typescript and i learn a little from each PR i review :sparkles: 
 I like your UX choices to cut down on the number of reset buttons we have, i think having just one that resets the image completely is perfect.

I noticed that when I click on crashes with an error banner, the zoom/reset tools and the words "crash diagram" will be on the screen for awhile before the banner takes over. Sometimes its for long enough that i can even move around and zoom on the words and then some buggy behavior starts happening. If you could prevent those tools and those words from popping up before the banner shows that would be awesome! I'm also getting the pesky horizontal scroll on the temporary banner when im at 80% page zoom.
[Screencast from 01-06-2025 01:10:09 PM.webm](https://github.com/user-attachments/assets/03bf25ad-5143-4563-ba04-69e7bc6582ab)

On the other banner, if you could center it horizontally that would scratch a great itch for me too heh
image

roseeichelmann:
also i noticed you added helper text to the zoom/reset buttons but i think those could be left as just symbols and if anything its the slide bar that could use some labeling, or those icons like @mateoclarke suggested look great!

johnclary:
Thanks everyone for the design feedback. @frankhereford here's a summary of where i'd like to go with this feature in scope of this PR.

1. Replace the CSS sizing with flexbox, like I mentioned in this [comment](https://github.com/cityofaustin/vision-zero/pull/1635#discussion_r1894476743). I'm very happy to pair on this—I know these can be a bit tricky
2. Let's go ahead and pull the Alert banners out of the part of the transform component so they don't wiggle around. So conditionally render either the transform component or the Alert banners.
3. Remove labels from zoom buttons (sorry for my miscommunication on that!) and add a rotate icon to the card footer. Here's a little design i drew up for this:

Screenshot 2025-01-07 at 1 01 14 PM

And here are a few bits of useful JSX i used for the design.

First, the card footer, which has the `FaRotate` icon in primary next to the `RotateControl`

```javascript
        <Card.Footer className="text-center">
          <div className="d-flex align-items-center">
            <div className="me-3 text-primary fs-5">
              <AlignedLabel>
                <FaRotate />
              </AlignedLabel>
            </div>
            <RotateControls rotation={rotation} setRotation={setRotation} />
          </div>
        </Card.Footer>
```

And a small change to the RotateControl itself—adding `w-100` to the outer div and removing the `style` entirely from the `Form.Range` component:


```jsx
    <div className="mt-2 w-100">
      <Form.Range
        min="-180"
        max="180"
        value={rotation}
        id="formControlRange"
        onChange={rotate}
        title="Rotate Diagram"
      />
    </div>
```

Thanks for bearing with all the feedback on this one!

frankhereford:
@johnclary I think this is ready to be handed back to you. I believe I have incorporated the changes you requested, and I am super grateful for the helpful examples. I have used them verbatim where I was able and made only minimal changes where I was not. That was very kind of you to put that much effort into the comment like you did. 🙏 

@roseeichelmann I've also asked for your review again because I believe the input you were asking had been incorporated, where applicable, into John's revised request. If you think I have overlooked something you want to see done, please let me know, and I'll be all over it. Thanks so much for all your thoughtful feedback.

chiaberry:
I know @johnclary requested the rotate icon to be in primary, but that made me think it was a button and I tried to click it, thinking it would reset or rotate the image by 90 degrees. 

> I think being able to zoom out to make the image smaller and less detailed than what it shows on-load an anti-feature. 
I have noticed that if you rotate the image, the edges fall out of the viewport, which could be why theres the option to zoom out in existing production. Of course, I don't know if VZE users would even miss this change in functionality.

@johnclary I looked in the readme todo's and github issues for the download cr3 button, I can't remember if you said it was left off the ACC designs intentionally or not, that was so last year 😬. Just don't want it to fall through the cracks

johnclary (changes requested):
@frankhereford this is looking great and is so so close. for some reason the diagram image is still overflowing the transform component on my display size, causing it to be partly obscured. i spent a few minutes trying to diagnose this and couldn't sort it out.

i am happy to open a follow-up for this if it feels like you want a break from this. The goal being to have the full diagram image displayed on init without having to zoom out.

Here's a screenshot showing the image pushing under the rotate control for some reason 🫠 

Screenshot 2025-01-15 at 2 10 35 PM

app/components/CrashDiagramCard.tsx
```diff
@@ -2,63 +2,179 @@ import { useState } from "react";
...
-              maxWidth: "100%",
-            }}
-            src={`${CR3_DIAGRAM_BASE_URL}/${crash.record_locator}.jpeg`}
-            alt="crash diagram"
-            onError={() => {
-              console.error("Error loading CR3 diagram image");
-              setDiagramError(true);
-            }}
-          />
+          <TransformWrapper initialScale={1}>
```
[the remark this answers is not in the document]
frankhereford: You bet - thank you! https://github.com/cityofaustin/vision-zero/pull/1635/commits/fcb216e1259c2bccd946772ccd0b8dae79fe3d3d

app/components/CrashDiagramCard.tsx
```diff
@@ -2,63 +2,179 @@ import { useState } from "react";
...
+              text: "report a bug",
+            }}
+          />
         )}
       </Card.Body>
-      <Card.Footer>Something else here</Card.Footer>
+      {!diagramError && (
+        <Card.Footer className="text-center">
+          <div className="d-flex align-items-center w-100">
+            <div className="me-3 text-primary fs-5">
```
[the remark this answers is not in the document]
frankhereford: I think her suggestion was 100% on point. Thank you @chiaberry. https://github.com/cityofaustin/vision-zero/pull/1635/commits/996f43de70bc4df8b84b3fda6bafc68b301cadf6

app/components/CrashDiagramCard.tsx
```diff
@@ -2,63 +2,179 @@ import { useState } from "react";
...
+    setRotation(0);
+  };
+
+  return (
+    <div className="d-flex justify-content-between w-100 mb-1">
+      <ButtonGroup>
+        <Button
+          size="sm"
+          variant="primary"
+          onClick={() => zoomIn()}
```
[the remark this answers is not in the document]
frankhereford: Absolutely -- this feels smoother to me. Thank you! https://github.com/cityofaustin/vision-zero/pull/1635/commits/43ef5407ecc13282436ecd9646ec3126b4def144

frankhereford:
@johnclary and everyone -- thank you all for your feedback on this. I believe I have incorporated all the requests, and I have been liberal in my use of the `Resolve Conversation` feature here in GitHub to try to pare down this PR for approachability, but I do not believe I have used it without addressing feedback.

The most notable change is now we have a `zoomToImage()` function which will zoom to the image, based on the dimensions of the component and the image itself. I'm calling this `onLoad` of the `Image` component -- thank you NextJS for that handy callback opportunity. This same function is used in the "Reset" button now as well.

I can't wait to hear what you think -- and thank you again for the chance work on my JS and contribute to the app.

app/components/CrashDiagramCard.tsx
```diff
@@ -118,19 +122,42 @@ export default function CrashDiagramCard({ crash }: { crash: Crash }) {
   const [diagramError, setDiagramError] = useState(false);
   const [rotation, setRotation] = useState(0);
 
+  const transformComponentRef = useRef<ReactZoomPanPinchRef | null>(null);
+
+  const zoomToImage = () => {
+    if (transformComponentRef.current) {
+      const { zoomToElement } = transformComponentRef.current;
+      zoomToElement("crashDiagramImage");
```
johnclary, line 130: wow—you found the magic bullet! this works perfectly.

one tiny change to request, which is setting the animation time down to `1` so as to disable the animation on load.

```
zoomToElement("crashDiagramImage", undefined, 1);
```

that seemed to do the trick unless you find another way to disable that initial load animation?

johnclary (approved):
boom! you did it! thank you, Frank!!!

i requested the tiniest change so as to nix the onload animation. but i'm approving it now 🚢 🚢 🚢 🚢 🚢

app/components/CrashDiagramCard.tsx
frankhereford: You bet - much cleaner. Thanks John!

https://github.com/cityofaustin/vision-zero/pull/1635/commits/c976f68b2935b9ccfcb1fae5d9d8384fe26262a9

PR #1638, opened 2024-12-21
Rework token hook to use access token instead of ID token

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/20391

## Testing

**URL to test:** [Netlify](https://deploy-preview-1638--vision-zero-nextjs.netlify.app/)

The best way to test this is to let your auth token expire and verify that the app refreshes it automatically. This takes 24 hours! Here's how to go about it:

1. Login to the deploy preview
2. Inspect your network requests and to copy your bearer token from the graphql request payload
3. Paste your token in `https://jwt.io/` and check the `exp` claim. It should be 24 hours from the time you logged in.
4. Have a great day (optional).
5. 25 hours later, open the deploy preview again. You should be logged in and the crashes page should load normally. Inspect token again, and the `exp` should now be 24 hours in the future. You can inspect the `iat` (Issued At) timestamp to verify your token was just issued.
6. Great, the tokens are refreshing as expected 🚢 

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/app/configs/locationsListViewColumns.tsx
+++ b/app/configs/locationsListViewColumns.tsx
@@ -10,7 +10,7 @@ export const locationsListViewColumns: ColDataCardDef<LocationsListRow>[] =
       label: "Location ID",
       sortable: true,
       valueRenderer: (record: LocationsListRow) => (
-        <Link href={`/locations/${record.location_id}`}>
+        <Link href={`/locations/${record.location_id}`} prefetch={false}>
           {record.location_id}
         </Link>
       ),
--- a/app/utils/auth.ts
+++ b/app/utils/auth.ts
@@ -1,4 +1,5 @@
 import { useState, useEffect } from "react";
+import { usePathname } from "next/navigation";
 import { useAuth0 } from "@auth0/auth0-react";
 import { User } from "@auth0/auth0-react";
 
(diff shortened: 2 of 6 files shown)


app/utils/auth.ts
```diff
@@ -53,32 +54,33 @@ export const formatRoleName = (role: string): string => {
...
 
 /**
- * Hook which retrieves the user's bearer token from the logged-in
- * user's IdToken claims.
- *
- * Todo: our Auth0 app is configured to return idTokens, which are
- * "opaque" proprietary Auth0 tokens, seemingly because this enables us
- * to interact with the User Management API. as a result, we must
- * use idToken.__raw to grab the actual JWT. it seems like our setup
- * may be misconfigured, because accessing .__raw seems like a hack :/
```
johnclary, line 63: This PR is accompanied by a change to our Auth0 post-login action, which I describe [here](https://github.com/cityofaustin/atd-data-tech/issues/20391#issuecomment-2558161548). Fixing the Auth0 configuration enables us to use the normal `getAccessTokenSilently()` and we can do away with `getIdTokenClaims()`

chiaberry: where did you update post-login action, is that configured in auth0 itself?

app/configs/locationsListViewColumns.tsx
```diff
@@ -10,7 +10,7 @@ export const locationsListViewColumns: ColDataCardDef<LocationsListRow>[] =
       label: "Location ID",
       sortable: true,
       valueRenderer: (record: LocationsListRow) => (
-        <Link href={`/locations/${record.location_id}`}>
+        <Link href={`/locations/${record.location_id}`} prefetch={false}>
```
chiaberry, line 13: does prefetching break if the token is expired?

chiaberry (approved):
followed all the test steps (even the optional one!) and i am still logged in 👍🏼

app/utils/auth.ts
johnclary: Yes. [Here is a direct link](https://manage.auth0.com/dashboard/us/atd-datatech/actions/library/details/3ea62805-ea37-465e-b0a5-1b1ef2bd26c8) to our **Set Hasura Claims**  post-login action. I also added this link to the my last comment on the issue.

app/configs/locationsListViewColumns.tsx
johnclary: ah—thanks for catching this. i meant to stash this in a different PR. i noticed this prefetching behavior on deploy previews (prefetching doesn't run locally), and my first reaction was to disable it because our pages don't have anything to prefetch 🤷

johnclary: i'm going to rebase this commit out of this PR

mddilley (approved):
I tested this without the optional part, but I will trust Chia's review for that! 🚀 Thanks for catching the bug in the Auth0 setup with the id vs. access token, and it is very nice that this removed the extra lines needed to document and work around that quirk. 🙌

app/configs/locationsListViewColumns.tsx
johnclary: I got scared of messing up the commit history with so many PRs open. I'll make a todo to revisit prefetching.

PR #1639, opened 2024-12-22
Crash details: recommendations

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/19963

Note that the old VZE has two features which are not implemented here:
- only show the recommendation form for fatal crashes
- only show the recommendation for admin users

## Testing

**URL to test:** Local

1. Start your local stack — remember, the app is on `http://localhost:3002` now
2. Navigate to any crash details and test out the new `Fatality Review Board Recommendations` card. Fill in the recommendation form. Edit one. Add partners, remove partners. Test the cancel button.
3. [Here](https://austininnovation.slack.com/archives/CHZE6BC6L/p1734971940631429) is a Slack thread in which I am looking for design feedback.

---

#### Ship list

- [ ] Code reviewed


app/app/crashes/[record_locator]/page.tsx
```diff
@@ -165,6 +166,15 @@ export default function CrashDetailsPage({
           />
         </Col>
       </Row>
+      <Row>
+        <Col sm={12} md={6} className="mb-3">
```
Comment (johnclary) [inline app/app/crashes/[record_locator]/page.tsx:170]: i went half-width with this card, thinking that we might put the crash notes component next to it. this would be a change from the old VZE where both the recommendations and notes sections are full-width on all display sizes.

app/components/CrashRecommendationCard.tsx
```diff
@@ -0,0 +1,294 @@
...
+/**
+ * Compares the old vs new RecommendationPartner arrays and returns
+ * new arrays of partner data that can be used in graphql mutation
+ * @param recommendationId - the pk of the current recommendation, if we are editing
+ * @param partnersOld - array of RecommendationPartner[] objs
+ * @param partnersNew - array of RecommendationPartner's from the form data. it is
+ * expected that these objects may only have a `partner_id` property
+ * @returns [addPartners, deletePartnerPks]
+ */
+const getPartnerChanges = (
```
johnclary, line 33: Should move this to a file in `/utils`...?

app/components/CrashRecommendationCard.tsx
```diff
@@ -0,0 +1,297 @@
...
+            {isEditing && (
+              <Form.Control
+                {...register("rec_text", {
+                  // coerce empty fields to null
+                  setValueAs: (v) => v?.trim() || null,
+                })}
+                as="textarea"
+                rows={6}
+                autoFocus={true}
+              />
```
johnclary, line 188: I passed on making this a reusable react-hook-form component, but [here's](https://github.com/cityofaustin/atd-moped/blob/main/moped-editor/src/components/forms/ControlledTextInput.js) a Moped example of one. I think we'll probably want to do something like this when we move on to the crash notes form and/or temp crash record form.

app/types/recommendation.ts
```diff
@@ -0,0 +1,61 @@
...
+  id: number;
+  created_at: string;
+  created_by: string;
+  crash_pk: number;
+  rec_text: string | null;
+  rec_update: string | null;
+  recommendation_status_id: number | null;
+  atd__recommendation_status_lkp: RecommendationStatus | null;
+  recommendations_partners: RecommendationPartner[] | null;
+};
```
johnclary, line 15: We have some database tech debt that would be nice to fix sooner than later:

* Move `atd__recommendation_status_lkp` into the `lookups` schema
* Add `updated_at` and `updated_by` fields, and display them in the Recommendations card in the VZE UI

app/components/CrashRecommendationCard.tsx
mddilley: I'm torn between moving it out of the component to reduce the file line count and not moving it because it probably won't be reused anywhere. I'm leaning towards moving to utils because I've been liking moving custom hook into a `hooks` folder to keep components tidier.

app/components/CrashRecommendationCard.tsx
```diff
@@ -0,0 +1,297 @@
...
+          record: payload,
+          id: recommendation?.id,
+          addPartners,
+          deletePartnerPks,
+        };
+      } else {
+        payload.recommendations_partners = { data: addPartners };
+        variables = { record: payload };
+      }
+      await mutate(variables, { skip_updated_by_setter: true });
```
mddilley, line 164: cool, i haven't seen this mutate option to skip setting who updated a record in the code yet. It is nice that it defaults to capture this data in the mutation. So, then we'll just remove this when `updated_by` is added to recommendations table in the future. 👍

app/components/CrashRecommendationCard.tsx
```diff
@@ -0,0 +1,297 @@
...
+                  ?.map(
+                    (x) => x.atd__coordination_partners_lkp?.coord_partner_desc
+                  )
+                  .join(", ") || ""}
+              </p>
+            )}
+            {isEditing && isLoadingPartners && <Spinner size="sm" />}
+            {isEditing && partners && (
+              <div style={{ height: "200px", overflowY: "scroll" }}>
+                <CrashRecommendationParters
```
mddilley, line 220: small type here - `Parters`

app/components/CrashRecommendationCard.tsx
```diff
@@ -0,0 +1,297 @@
...
+                })}
+                as="textarea"
+                rows={6}
+                autoFocus={true}
+              />
+            )}
+            {!isEditing && <p>{recommendation?.rec_text || ""}</p>}
+          </Form.Group>
+          <Form.Group className="mb-3">
+            <Form.Label className="fw-bold">Update</Form.Label>
```
mddilley, line 193: There is a small difference, and I don't think it will affect the current power users, but this is titled "Updates" in production since they seem to use input this to capture updates over time.

app/components/CrashRecommendationCard.tsx
```diff
@@ -0,0 +1,297 @@
...
+              <Form.Control
+                {...register("rec_update", {
+                  // coerce empty fields to null
+                  setValueAs: (v) => v?.trim() || null,
+                })}
+                as="textarea"
+                rows={6}
+              />
+            )}
+            {!isEditing && <p>{recommendation?.rec_update || ""}</p>}
```
mddilley, line 204: @johnclary I tested this on a really long collection of updates like in crash `17343328`, and it creates a really tall column. It isn't far from how it looks in production, but the half-width makes it more noticeable. I'm not keen on scrolling since I think the VZ team would want to see the entire contents on screen, but that is only my assumption. I'll think about it some more since I don't have any suggestions on this.

app/components/CrashRecommendationPartners.tsx
mddilley, line 1: this is a cool component to contain the partners form logic! 😎

mddilley (approved):
This works great! The types + docs in the code capture this feature really well - I really love the direction this codebase is going! ✨ I added a couple comments, and one about how longer recommendations and updates look in this view that could be addressed later.

app/components/CrashRecommendationCard.tsx
```diff
@@ -0,0 +1,297 @@
...
+ * expected that these objects may only have a `partner_id` property
+ * @returns [addPartners, deletePartnerPks]
+ */
+const getPartnerChanges = (
+  recommendationId: number | null,
+  partnersOld: Partial<RecommendationPartner>[],
+  partnersNew: Partial<RecommendationPartner>[]
+) => {
+  const addPartners: Partial<RecommendationPartner>[] = [];
+  const deletePartnerPks: number[] = [];
```
roseeichelmann, line 39: these feel like function names, i think `partnersToAdd` and `partnersToDelete` or something like that would be more obvious that these are just arrays containing the partners to add or delete

app/components/CrashRecommendationCard.tsx
roseeichelmann: by the end of reviews this PR the word partner looks weird and has lost all meaning to me :face_with_spiral_eyes:

app/components/CrashRecommendationCard.tsx
```diff
@@ -0,0 +1,297 @@
...
+            )}
+            {!isEditing && <p>{recommendation?.rec_update || ""}</p>}
+          </Form.Group>
+          <Form.Group className="mb-3">
+            <Form.Label className="fw-bold">Partners</Form.Label>
+            {!isEditing && (
+              <p>
+                {recommendation?.recommendations_partners
+                  ?.map(
+                    (x) => x.atd__coordination_partners_lkp?.coord_partner_desc
```
roseeichelmann, line 212: im kinda anti single letter variables, even though i guess this should be an exception to that rule bc its just a one liner loop?

app/components/CrashRecommendationCard.tsx
roseeichelmann: i think i lean towards leaving this function here because like mike said it wont be reused elsewhere and i dont think its ridiculously long? if we move this it starts feeling like where do we draw the line on moving functions to the utils folders and are we just making things more complicated and hard to follow at that point?

roseeichelmann (approved):
This looks great, I'm a huge fan of how its now one big form that you edit all at once, and some of the previous bugginess is gone now and its way more simple. I have one _recommendation_, ha, which is to move the partners and status sections to the top of the card like how it is in production. Wonderfully done and you are a great typescript teacher!

app/components/CrashRecommendationCard.tsx
johnclary: thank you 🤠

app/components/CrashRecommendationCard.tsx
johnclary: Nice catch—thanks!

app/components/CrashRecommendationCard.tsx
johnclary: whoa—i assumed i had tested a long update, but that's 5k+ characters. thanks for this example. i agree that moving the status + partners field above up top makes sense.

Maybe we should go back to full-width. I think we'd want to set a max container width in that case. After looking more closely at the notes feature, i don't think this layout will work as half-width anyway.

This example also makes me realize that "updates" should actually behave more like notes. It would be cool to take a Moped-style approach and have a note type column that would enable us to re-use the existing notes feature for FRB updates. 

I will get some feedback from Mary on this layout and figure out the best way to move forward in the short term 🙏

app/components/CrashRecommendationCard.tsx
johnclary: ditto 🙏

app/components/CrashRecommendationCard.tsx
johnclary: I moved status + partners to the top of the form. Will open a follow-up issue for further layout tweaks ✅

app/components/CrashRecommendationCard.tsx
johnclary: yeah. i agree that potential future reusability is a good test of when to move into the general `utils` folder. and i am coming around on the idea of every component living in a directory with a standard pattern for local utils, hooks, const files...

PR #1640, opened 2024-12-23
Location details: crashes list

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/19979
* https://github.com/cityofaustin/atd-data-tech/issues/19980

This PR adds the crashes list to the location details page. I decided to combine the CR3 and non-cr3 crashes into a single table. This feels like a nice improvement over the dual table experience in the old VZE. What do you think?

Screenshot 2024-12-23 at 3 17 05 PM

## Testing

**URL to test:** Local

**Steps to test:**

1. Start your local stack - you'll need to apply migrations and metadata for this one

```shell
# from /database
hasura migrate apply
hasura metadata apply
```

2. Navigation to a location details page and check out that new crashes list. try searching and sorting. Nice!

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/app/app/locations/[location_id]/page.tsx
+++ b/app/app/locations/[location_id]/page.tsx
@@ -19,10 +19,9 @@ import { locationCrashesQueryConfig } from "@/configs/locationCrashesTable";
 const typename = "atd_txdot_locations";
 
 /**
- * Hook which returns builds a queryBuilder Filter array with the
- * `location_id` param - this can be passed as a `contextFilter`
- * to the TableWrapper so that locationCrashesList is filtered
- * by location
+ * Hook which returns builds a Filter array with the `location_id` param.
+ * This can be passed as a `contextFilter` to the TableWrapper so that the
+ * location's crashes table is filtered by location.
  * @param {string} locationId - the location ID string from the page route
  * @returns {Filter[]} the Filter array with the single location filter
  */
--- a/app/app/locations/[location_id]/page.tsx
+++ b/app/app/locations/[location_id]/page.tsx
@@ -1,16 +1,43 @@
 "use client";
+import { useMemo } from "react";
 import { notFound } from "next/navigation";
-import Row from "react-bootstrap/Row";
 import Col from "react-bootstrap/Col";
+import Card from "react-bootstrap/Card";
+import Row from "react-bootstrap/Row";
 import AppBreadCrumb from "@/components/AppBreadCrumb";
 import DataCard from "@/components/DataCard";
 import LocationMapCard from "@/components/LocationMapCard";
+import TableWrapper from "@/components/TableWrapper";
 import { useQuery } from "@/utils/graphql";
 import { GET_LOCATION } from "@/queries/location";
 import { Location } from "@/types/locations";
+import { Filter } from "@/utils/queryBuilder";
 import { locationCardColumns } from "@/configs/locationDataCard";
+import { locationCrashesColumns } from "@/configs/locationCrashesColumns";
+import { locationCrashesQueryConfig } from "@/configs/locationCrashesTable";
 
 const typename = "atd_txdot_locations";
+
+/**
+ * Hook which returns builds a queryBuilder Filter array with the
+ * `location_id` param - this can be passed as a `contextFilter`
+ * to the TableWrapper so that locationCrashesList is filtered
+ * by location
+ * @param {string} locationId - the location ID string from the page route
+ * @returns {Filter[]} the Filter array with the single location filter
+ */
+const useLocationIdFilter = (locationId: string): Filter[] =>
+  useMemo(
+    () => [
+      {
+        id: "location_id",
+        value: locationId,
+        column: "location_id",
+        operator: "_eq",
+      },
+    ],
+    [locationId]
+  );
 
 export default function LocationDetailsPage({
   params,
@@ -19,6 +46,7 @@ export default function LocationDetailsPage({
 }) {
   const locationId = params.location_id;
 
+  const locationIdFilter = useLocationIdFilter(locationId);
   const { data, error } = useQuery<Location>({
     query: locationId ? GET_LOCATION : null,
     variables: { locationId },
@@ -58,6 +86,21 @@ export default function LocationDetailsPage({
             onSaveCallback={() => Promise.resolve()}
             record={location}
           />
+        </Col>
+      </Row>
+      <Row>
+        <Col>
+          <Card>
+            <Card.Header>Crashes</Card.Header>
+            <Card.Body>
+              <TableWrapper
+                columns={locationCrashesColumns}
+                initialQueryConfig={locationCrashesQueryConfig}
+                localStorageKey="locationCrashesQueryConfig"
+                contextFilters={locationIdFilter}
+              />
+            </Card.Body>
+          </Card>
         </Col>
       </Row>
     </>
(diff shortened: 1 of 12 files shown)


app/app/locations/[location_id]/page.tsx
```diff
@@ -60,6 +87,21 @@ export default function LocationDetailsPage({
...
+      <Row>
+        <Col>
+          <Card>
+            <Card.Header>Crashes</Card.Header>
+            <Card.Body>
+              <TableWrapper
+                columns={locationCrashesColumns}
+                initialQueryConfig={locationCrashesQueryConfig}
+                localStorageKey="locationCrashesQueryConfig"
+                contextFilters={locationIdFilter}
```
Comment (johnclary) [inline app/app/locations/[location_id]/page.tsx:99]: new `contextFilters` prop that let's us pass in this filter from local storage. The main challenge here is that the primary `queryConfig` gets saved in local storage, and we don't to include the `location_id` in the local storage config, because it's a dynamic value.

I thought about a few a different ways to approach, but this seemed simple enough for now. I believe this is the only place in the app where we'll need to use this new feature.

app/components/TableAdvancedSearchFilterMenu.tsx
```diff
@@ -53,7 +53,7 @@ export default function TableAdvancedSearchFilterMenu({
     <Row className="p-2 bg-light">
       {queryConfig.filterCards.map((filterCard) => {
         return (
-          <Col key={filterCard.id}>
+          <Col key={filterCard.id} xs={12} md={3}>
```
johnclary, line 56: keeps those filter cards from growing too wide

app/configs/locationCrashesTable.ts
```diff
@@ -0,0 +1,81 @@
...
+import { QueryConfig, FilterGroup } from "@/utils/queryBuilder";
+import { DEFAULT_QUERY_LIMIT } from "@/utils/constants";
+import {
+  getStartOfYearDate,
+  makeDateFilters,
+} from "@/components/TableDateSelector";
+
+const columns = locationCrashesColumns.map((col) => String(col.path));
+
+const locationCrashesFiltercards: FilterGroup[] = [
```
johnclary, line 11: the old VZE has a lot of advanced filters on the cr3 crashes table. i'm a bit skeptical we need all of them. i will check with Xavier.


i'd also like to check in with some of our non-VZ users who access this page. we have a few power users identified from the ACC research.

database/migrations/default/1734981892414_location_crashes_record_locator/up.sql
johnclary, line 1: Adding a few columns to this view:
* `record_locator` - so that we can link to the crash details page
* `crash_timestamp` - so that we have a UTC timestamp to operate on

mddilley, line 1: i tested the down migration, saw the expected error in the graphql response from the missing record location and timestamp columns, and then ran the up again. All of this works! 💯

mddilley (approved):
I tested this and everything worked as expected, and I like the combined table! I think the filters cover any case where someone would want to see only CR3 on non-CR3. I often used to forget to scroll all the way down, and I remember that causing a regression or two in the past. Feels like an upgrade to me! 🙌 🚀

app/components/TableAdvancedSearchFilterMenu.tsx
johnclary: ah yes—there's some questionable CSS working on this filter menu that i hoped you wouldn't notice. [here's](https://github.com/cityofaustin/vision-zero/pull/1632#issuecomment-2554832495) another short thread about this menu, too. i will open a follow-up issue to get this UI straightened out 🙏

app/app/locations/[location_id]/page.tsx
```diff
@@ -1,24 +1,51 @@
...
 import { Location } from "@/types/locations";
+import { Filter } from "@/utils/queryBuilder";
 import { locationCardColumns } from "@/configs/locationDataCard";
+import { locationCrashesColumns } from "@/configs/locationCrashesColumns";
+import { locationCrashesQueryConfig } from "@/configs/locationCrashesTable";
 
 const typename = "atd_txdot_locations";
 
+/**
+ * Hook which returns builds a Filter array with the `location_id` param.
```
Comment (roseeichelmann) [inline app/app/locations/[location_id]/page.tsx:22]: typo here, maybe just "Hook which returns a filter array.."

app/app/locations/[location_id]/page.tsx
Comment (roseeichelmann) [reply app/app/locations/[location_id]/page.tsx:99]: can you mention in your comment the need for the `location_id` filter to be dynamic unlike the query config thats saved in local storage? i feel like reading your comment here was super helpful but i might have been confused if i just was looking at the code

app/components/TableWrapper.tsx
```diff
@@ -20,6 +20,7 @@ interface TableProps<T extends Record<string, unknown>> {
   columns: ColDataCardDef<T>[];
   initialQueryConfig: QueryConfig;
   localStorageKey: string;
+  contextFilters?: Filter[] 
 }
 
 /**
```
roseeichelmann, line 26: what is the general opinion of docstrings in our new VZ typescript codebase? i think it would be helpful to have descriptions for these params here. here are the official [typescript doc specifications](https://tsdoc.org/)

app/utils/queryBuilder.ts
```diff
@@ -382,10 +397,18 @@ const buildQuery = ({
...
  */
-export const useQueryBuilder = (queryConfig: QueryConfig): string =>
+export const useQueryBuilder = (
+  queryConfig: QueryConfig,
+  contextFilters?: Filter[]
+): string =>
   useMemo(() => {
-    return buildQuery(queryConfig);
+    return buildQuery(queryConfig, contextFilters);
   }, [queryConfig]);
```
roseeichelmann, line 414: i see the eslint error here about `contextFilters` being missing in the dependency array

roseeichelmann (approved):
I reallyyyy like how you combined the two tables and simplified the filters for now. Looks and works so good!

app/components/TableWrapper.tsx
johnclary: you're totally right. these have definitely grown to the point of needing docs. i'm all for adding docstrings wherever and whenever—especially when the typings are not adding enough context on their own.

app/utils/queryBuilder.ts
johnclary: whoa—i am not seeing a build error or runtime error about this 😵 

i fixed it and will do a bit of digging as to why i'm missing this. thank you!

johnclary:
I caught a little bug [here](https://github.com/cityofaustin/vision-zero/pull/1640/commits/5561b0787c30f351388310f2093d793c4b326ef4) that surfaced thanks to the year rolling over. I will be merging this as soon as conflicts are resolved. Thank you!

app/components/TableWrapper.tsx
johnclary: here's the commit with updated docs: https://github.com/cityofaustin/vision-zero/pull/1603/commits/79428de532b9e78fc5d8e81c18e7d8a8ee72066d

app/app/locations/[location_id]/page.tsx
Comment (johnclary) [reply app/app/locations/[location_id]/page.tsx:99]: sorry—forgot to do this before merge. done here: https://github.com/cityofaustin/vision-zero/pull/1603/commits/852acdb22904d7180fb3349e65bc4adf86142e28 🙏

PR #1641, opened 2024-12-26
Upload Non-CR3 file

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/19984

This PR adds the ability to upload a CSV of non-CR3 crashes.

Screenshot 2024-12-26 at 2 56 33 PM

## Testing

**URL to test:** Local

1. Use the new sidebar item to navigate to `./upload-non-cr3`
2. Use the **Download CSV Template** link to download the CSV template
3. Try importing the template you downloaded - it has errors!
4. Correct your template by deleting the problematic rows—or fixing them.
5. Upload the valid CSV.
6. Modify your CSV file by duplicating one or more rows. Try to upload it, and verify that it fails to pass validation.
7. Bonus points: use your SQL client to verify your CSV records have been created in the db.

I think that's it!

---
#### Ship list

- [ ] Code reviewed


app/types/nonCr3.ts
```diff
@@ -0,0 +1,50 @@
+import { z } from "zod";
+
+/**
+ * Validation schema for a non-CR3 record uploaded through the UI
+ */
+export const nonCr3UploadSchema = z.object({
```
johnclary, line 6: Zod is back! and it's making quick work of a [bunch of validation code](https://github.com/cityofaustin/vision-zero/blob/e00391d9040bf6a4154759535a7bef2487665465/editor/src/views/UploadNonCr3/UploadNonCR3.js#L297) we used in the old editor.

app/app/upload-non-cr3/page.tsx
```diff
@@ -0,0 +1,257 @@
...
+  const [validationErrors, setValidationErrors] = useState<
+    NonCr3ValidationError[] | null
+  >();
+  const [uploadError, setUploadError] = useState(false);
+  const [success, setSuccess] = useState(false);
+  const { mutate, loading: isMutating } = useMutation(INSERT_NON_CR3);
+
+  const onSelectFile = useCallback(
+    (file: File) => {
+      Papa.parse<NonCr3Upload>(file, {
```
johnclary, line 44: the old version of this feature used [react-csv-reader](https://github.com/nzambello/react-csv-reader), which is just a wrapper around `papaparse`.

app/app/upload-non-cr3/page.tsx
```diff
@@ -0,0 +1,257 @@
...
+                >
+                  <AlignedLabel>
+                    <FaTriangleExclamation className="me-2" />
+                    <span>
+                      There was an error uploading your file - please try again.
+                    </span>
+                  </AlignedLabel>
+                </Alert>
+              )}
+              {validationErrors && (
```
johnclary, line 222: the old version of this feature used a [fancy table](https://handsontable.com/) component that enabled the user to edit the CSV and re-submit it through the UI.

I opted to keep things simpler. AFAIK Xavier uses this feature four times per year.

app/types/nonCr3.ts
```diff
@@ -0,0 +1,65 @@
+import { z } from "zod";
+
+/**
+ * This is a 4x4 decimal degree bounding box centered on the tx state
+ * capitol building
+ * */
```
johnclary, line 6: coped from the old VZE

app/types/nonCr3.ts
```diff
@@ -0,0 +1,86 @@
+import { z } from "zod";
```
johnclary, line 1: Zod is back! and it's making quick work of a [bunch of validation code](https://github.com/cityofaustin/vision-zero/blob/e00391d9040bf6a4154759535a7bef2487665465/editor/src/views/UploadNonCr3/UploadNonCR3.js#L297) we used in the old editor.

app/types/nonCr3.ts
```diff
@@ -0,0 +1,86 @@
...
+    return false;
+  });
+};
+
+/**
+ * Adds a dupe case ID test to the NonCr3Upload schema
+ */
+export const NonCr3UploadDedupedSchema = z
+  .array(NonCr3UploadSchema)
+  .refine(noDuplicateCaseIds, "Duplicate case IDs (case_id) detected");
```
johnclary, line 77: I don't know how common this problem is, but it was in the old VZE so i went ahead and added the test here. Without this check, Hasura is going to bomb with a very unhelpful error.

app/app/upload-non-cr3/page.tsx
```diff
@@ -0,0 +1,257 @@
...
+                        setParsing(true);
+                        onSelectFile(e.target?.files[0]);
+                      }
+                    }}
+                  ></Form.Control>
+                </Form>
+              </Col>
+              <Col className="my-auto">
+                <Link
+                  href="/files/non_cr3_template.csv"
```
mddilley, line 130: @johnclary this file is missing from the public folder but I was able to test with the template from staging. 🥼

edit: I just realized that .csvs are gitignored at the root project level so that makes sense why it snuck out of this diff. I was so confused when I was seeing a change locally. 😵‍💫

mddilley (changes requested):
I tested this new iteration, and I think it covers all the bases. 🙌 🚀 I requested changes because the template file is missing from the diff so it is a small fix. I also had one question about the validation behavior.

app/app/upload-non-cr3/page.tsx
johnclary: thanks for catching this! [fixed](https://github.com/cityofaustin/vision-zero/pull/1641/commits/01e469df86de6f5fd83d17307bafda787f56459e)

app/app/upload-non-cr3/page.tsx
```diff
@@ -0,0 +1,257 @@
...
+        <Card.Body>
+          {!data && (
+            <Row>
+              <Col xs={12} md={6} lg={3}>
+                <Form onSubmit={(e) => e.preventDefault()}>
+                  <Form.Control
+                    type="file"
+                    name="file"
+                    accept=".csv"
+                    onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
```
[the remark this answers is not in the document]
johnclary: thanks for catching that! TIL that uploading the same file name + path does not trigger `onChange`. i addressed this by clearing the file input after you select one. see `e.target.value = "";` down below. this fixes the problem, although i don't love that it wipes out the file name from the input.

i think the more diligent way to handle this would be to had a state var for the file and present users with a button to "Start over" or what have you.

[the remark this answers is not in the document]
mddilley: that is a TIL for me too! I think this fix is great in the context of how often it is used like you mentioned. I retested and it works great. 😎

mddilley (approved):
Less dependencies and less code too! Can't get better than that 🙌 🚀

chiaberry (approved):
went through the test steps and ✔️ 

I like how zod has returned!

app/queries/nonCr3s.ts
```diff
@@ -0,0 +1,15 @@
+import { gql } from "graphql-request";
+
+export const INSERT_NON_CR3 = gql`
+  mutation mutationInsertNonCR3($objects: [atd_apd_blueform_insert_input!]!) {
+    insert_atd_apd_blueform(
+      objects: $objects
+      on_conflict: {
+        constraint: atd_apd_blueform_pk
+        update_columns: [date, case_id, address, longitude, latitude, hour]
```
roseeichelmann, line 9: should we add `created_at`, `created_by`, etc fields to this table?

app/app/upload-non-cr3/page.tsx
roseeichelmann: i like your simpler version much better esp for the use case :sparkles:

app/queries/nonCr3s.ts
johnclary: yeah! i just opened https://github.com/cityofaustin/atd-data-tech/issues/20470 🙏

PR #1642, opened 2024-12-26
Crash details: people related records table

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/19965

> Show the the related people records on the crash details page. The same fields should be editable as in the old VZE.

## Testing

Test on local

Navigate to an individual crash page and see related people records. The primary people table and other people table have been combined into one table. 

You should be able to edit injury severity, age, sex, ethnicity and suspected unhoused. All of the inputs are drop downs except for age. 

A future PR will address the name editing, for now you should be able to see a full name when there is a fatality.

---

#### Ship list

- [ ] Code reviewed


app/types/person.ts
```diff
@@ -0,0 +1,23 @@
+import { LookupTableOption } from "./relationships";
+
+export type Person = {
+  crash_pk: number;
+  drvr_city_name: string;
+  drvr_ethncty: LookupTableOption;
+  prsn_ethnicity_id: number;
+  drvr_zip: string;
+  gndr: LookupTableOption;
+  prsn_gndr_id: number;
```
chiaberry, line 10: i am not sure if i should add a comment to specify if the key is a foreignkey or leave it as an exercise to the reader?

app/queries/crash.ts
```diff
@@ -129,7 +129,7 @@ export const GET_CRASH = gql`
           sus_serious_injry_count
         }
       }
-      people_list_view {
+      people_list_view (order_by: { unit_nbr: asc }) {
```
johnclary, line 132: i noticed the sort order of the rows jumping around while editing. would you add sorting on `prs_nbr` ascending as well? so that the records are sorted by unit # then person #?

app/types/person.ts
johnclary: i think it's obvious enough 👍

app/queries/crash.ts
johnclary: it also looks like you need to add the foreign key columns for all the editable fields with lookup values, e.g., `prsn_injry_sev_id`. 

making this change will ensure that the correct option is initially selected when you begin editing a lookup field. otherwise it defaults to "Select..." 👇 

Screenshot 2024-12-31 at 11 11 34 AM


And [here](https://github.com/cityofaustin/vision-zero/pull/1634#discussion_r1894241301) is a comment where i mention that this is not well documented 😇 any ideas?

app/app/crashes/[record_locator]/page.tsx
```diff
@@ -165,6 +167,18 @@ export default function CrashDetailsPage({
           />
         </Col>
       </Row>
+      <Row>
```
Comment (johnclary) [inline app/app/crashes/[record_locator]/page.tsx:170]: would you move the People table so that it is above the Charged table?

app/configs/peopleColumns.tsx
```diff
@@ -0,0 +1,87 @@
...
+    // todo: the special combined first/middle/last name field edit component
+    valueFormatter: (value, record, column) => {
+      const nameFields = [
+        record.prsn_first_name,
+        record.prsn_mid_name,
+        record.prsn_last_name,
+      ];
+      // filter out null fields then join into a string
+      const displayName = nameFields.filter((n) => n).join(" ");
+      return displayName;
```
johnclary, line 84: looks good 🙏

johnclary (changes requested):
Looking good, Chia! It's really nice to have this piece of the crash page in place. Just a few small changes needed 🙌

app/components/DataCardInput.tsx
```diff
@@ -16,7 +16,7 @@ interface DataCardInputProps {
   /**
    * Controls the type of input to be rendered
    */
-  inputType?: InputType;
+  inputType?: InputType; // wondering if this shoudl be required and not optional
```
chiaberry, line 19: if you dont specify an inputType, the form on the data card input component only shows the save and cancel buttons, there is no default input option. Which makes me think this should be required, or do we have a fallback input?

app/queries/crash.ts
chiaberry: ha, i read the comment but somehow completely missed that it had to be added to query, and not just the config. 
We could add a comment at the top of the crash query, but its such a long query, any comments at the top will be easily missed.

app/components/DataCardInput.tsx
johnclary: I see what you mean :/ 

I think ideally the type definition would require `inputType` if `editable` is `true`. Are you up for reworking the `ColDataCardDef` in this PR?

chiaberry: I will look at reworking the `ColDataCardDef`, but DataCardInput doesn't render unless a column is editable, right?

roseeichelmann (changes requested):
this all looks great! one thing that diverges from current prod though is that i tested updating a fatal injury from fatal to something else and i can still see their full name. do we want to keep the current behavior where we only display names if they are fatalities?

i also noticed i can save negative numbers into the ages, maybe we should prevent that?

app/configs/peopleColumns.tsx
```diff
@@ -0,0 +1,87 @@
...
+    path: "prsn_exp_homelessness",
+    label: "Suspected unhoused",
+    inputType: "yes_no",
+    editable: true,
+  },
+  prsn_last_name: {
+    path: "prsn_last_name",
+    label: "Name",
+    // todo: the special combined first/middle/last name field edit component
+    valueFormatter: (value, record, column) => {
```
roseeichelmann, line 76: i guess in this config somewhere we would need to specify to only render if `injry_sev` is fatality?

johnclary: thanks for pointing this out. let's revisit this when we build the special name edit component. i'll open an issue for that momentarily.

johnclary:
>  also noticed i can save negative numbers into the ages, maybe we should prevent that?

@roseeichelmann thanks for flagging this. i'll open an issue for us to add better/more flexible input validation. it's a problem here and on the create crash record form (like you mentioned) 🙏

johnclary (approved):
Looks great! 🙌

johnclary:
@chiaberry i'm going to square he conflicts here and merge—thank you!

PR #1643, opened 2024-12-27
Create crash records

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/19985
* https://github.com/cityofaustin/atd-data-tech/issues/19986

Screenshot 2025-01-03 at 11 04 43 AM

Note these two related features which are out of scope:

* Ability to delete temporary crash records - https://github.com/cityofaustin/vision-zero/pull/1644
* Hide this page from read-only users - https://github.com/cityofaustin/atd-data-tech/issues/20302

## Testing

**URL to test:** Local

1. Use the **Create crash** sidebar link to navigate to this new page
2. Use the **Create crash record** button to open the form. Create a few crash records while also testing the required fields validation. The required fields are Case ID, Crash date, primary address, and all three fields on each unit card.
3. Try adding and removing additional units while you're using the form.
4. Verify that the table refreshes after you submit the form.
5. Visit the crash details page and verify that the number and types of units created are correct, and that a person record is created for each injury.
6. On the temp crashers table, test the search and filter functionality.
7. I'd like to hide the filters altogether by adding a new prop called `staticFilters` (or something like that). I am holding off on that because there are other branches out there with changes to the table component's props.

---
#### Ship list

- [ ] Code reviewed

### DIFF
--- a/app/components/CreateCrashRecordModal.tsx
+++ b/app/components/CreateCrashRecordModal.tsx
@@ -1,4 +1,3 @@
-import Alert from "react-bootstrap/Alert";
 import Button from "react-bootstrap/Button";
 import Card from "react-bootstrap/Card";
 import Form from "react-bootstrap/Form";
@@ -22,9 +21,9 @@ interface CreateCrashModalProps {
    */
   onClose: () => void;
   /**
-   * An async callback fired after the user form is successfully submitted
+   *  Callback fired after the user form is successfully submitted
    */
-  onSubmitCallback: (crash: Crash) => Promise<void>;
+  onSubmitCallback: () => void;
   /**
    * If the modal should be visible or hidden
    */
@@ -76,22 +75,21 @@ const DEFAULT_FORM_VALUES: CrashInputs = {
  * that it can be included in a crash insert mutation
  */
 const makeUnitRecord = (unit: UnitInputs, index: number, userEmail: string) => {
-  const unitRecord = {
-    unit_nbr: index + 1,
+  const unitNumber = index + 1;
+
+  const unitRecord: Record<string, unknown> = {
+    unit_nbr: unitNumber,
     unit_desc_id: unit.unit_desc_id,
     cris_schema_version: "2023",
     updated_by: userEmail,
     created_by: userEmail,
-    people_cris: {
-      data: [],
-    },
   };
 
-  const unitNumber = index + 1;
+  const people = [];
 
   // create fatal injuries
   for (let i = 0; i < Number(unit.fatality_count); i++) {
-    unitRecord.people_cris.data.push({
+    const personRecord = {
       prsn_nbr: i + 1,
       unit_nbr: unitNumber,
       prsn_injry_sev_id: 4,
@@ -99,12 +97,14 @@ const makeUnitRecord = (unit: UnitInputs, index: number, userEmail: string) => {
       is_primary_person: true,
       updated_by: userEmail,
       created_by: userEmail,
-    });
+    };
+
+    people.push(personRecord);
   }
 
   // create other injuries
   for (let i = 0; i < Number(unit.injury_count); i++) {
-    unitRecord.people_cris.data.push({
+    people.push({
       prsn_nbr: unit.fatality_count + i + 1,
       unit_nbr: unitNumber,
       prsn_injry_sev_id: 1,
@@ -114,7 +114,8 @@ const makeUnitRecord = (unit: UnitInputs, index: number, userEmail: string) => {
       created_by: userEmail,
     });
   }
-  return unitRecord;
+
+  return { ...unitRecord, people_cris: { data: people } };
 };
 
 /**
@@ -165,7 +166,7 @@ export default function CreateCrashRecordModal({
     }>({ crash });
 
     if (responseData && responseData.insert_crashes_cris) {
-      await onSubmitCallback(responseData.insert_crashes_cris?.returning[0]);
+    onSubmitCallback();
     }
     reset();
   };
--- a/app/components/CreateCrashRecordModal.tsx
+++ b/app/components/CreateCrashRecordModal.tsx
@@ -1,0 +1,371 @@
+import Alert from "react-bootstrap/Alert";
+import Button from "react-bootstrap/Button";
+import Card from "react-bootstrap/Card";
+import Form from "react-bootstrap/Form";
+import Modal from "react-bootstrap/Modal";
+import Spinner from "react-bootstrap/Spinner";
+import AlignedLabel from "@/components/AlignedLabel";
+import FormControlDatePicker from "@/components/FormControlDatePicker";
+import { FaCirclePlus, FaCircleMinus } from "react-icons/fa6";
+import { useForm, SubmitHandler, useFieldArray } from "react-hook-form";
+import { useQuery, useMutation } from "@/utils/graphql";
+import { UNIT_TYPES_QUERY } from "@/queries/unit";
+import { CREATE_CRASH } from "@/queries/crash";
+import { LookupTableOption } from "@/types/relationships";
+import { Crash } from "@/types/crashes";
+import { useAuth0 } from "@auth0/auth0-react";
+
+interface CreateCrashModalProps {
+  /**
+   * A callback fired when either the modal backdrop is clicked, or the
+   * escape key is pressed
+   */
+  onClose: () => void;
+  /**
+   * An async callback fired after the user form is successfully submitted
+   */
+  onSubmitCallback: (crash: Crash) => Promise<void>;
+  /**
+   * If the modal should be visible or hidden
+   */
+  show: boolean;
+}
+
+type UnitInputs = {
+  unit_desc_id: string;
+  fatality_count: number;
+  injury_count: number;
+};
+
+type CrashInputs = {
+  case_id: string;
+  crash_timestamp: string;
+  created_by: string;
+  cris_schema_version: "2023";
+  is_temp_record: boolean;
+  private_dr_fl: boolean;
+  rpt_city_id: number;
+  rpt_sec_street_name: string;
+  rpt_street_name: string;
+  units_cris: UnitInputs[];
+  updated_by: string;
+};
+
+const DEFAULT_UNIT: UnitInputs = {
+  unit_desc_id: "",
+  fatality_count: 0,
+  injury_count: 0,
+};
+
+const DEFAULT_FORM_VALUES: CrashInputs = {
+  case_id: "",
+  crash_timestamp: "",
+  created_by: "",
+  cris_schema_version: "2023",
+  is_temp_record: true,
+  private_dr_fl: false,
+  rpt_city_id: 22,
+  rpt_sec_street_name: "",
+  rpt_street_name: "",
+  units_cris: [{ ...DEFAULT_UNIT }],
+  updated_by: "",
+};
+
+/**
+ * Construct a unit record with people records such
+ * that it can be included in a crash insert mutation
+ */
+const makeUnitRecord = (unit: UnitInputs, index: number, userEmail: string) => {
+  const unitRecord = {
+    unit_nbr: index + 1,
+    unit_desc_id: unit.unit_desc_id,
+    cris_schema_version: "2023",
+    updated_by: userEmail,
+    created_by: userEmail,
+    people_cris: {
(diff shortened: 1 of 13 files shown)


app/components/CreateCrashRecordModal.tsx
```diff
@@ -0,0 +1,372 @@
...
+};
+
+/**
+ * Construct a unit record with people records such
+ * that it can be included in a crash insert mutation
+ */
+const makeUnitRecord = (unit: UnitInputs, index: number, userEmail: string) => {
+  const unitNumber = index + 1;
+
+  const unitRecord: Record<string, unknown> = {
```
johnclary, line 80: similar to the recommendations form, I left the typings loose here because this data is heading straight to the graphql API, so I don't see a benefit in coding out the exact object shape. this seems sufficiently typesafe.

app/components/SidebarLayout.tsx
```diff
@@ -13,6 +13,7 @@ import {
   FaLocationDot,
   FaAngleRight,
   FaUserGroup,
+  FaCirclePlus,
```
johnclary, line 16: I don't like this icon and am looking forward to polishing the side nav sometime in the near future...

app/components/TableWrapper.tsx
johnclary, line 1: we've got another change to this component coming in via https://github.com/cityofaustin/vision-zero/pull/1640

app/components/TableWrapper.tsx
```diff
@@ -103,6 +105,12 @@ export default function TableWrapper<T extends Record<string, unknown>>({
     setAreFiltersDirty(!isEqual(queryConfig, initialQueryConfig));
   }, [queryConfig, initialQueryConfig]);
 
+  /**
+   * Hook to trigger refetch
+   */
+  useEffect(() => {
+    refetch();
+  }, [_refetch, refetch]);
```
johnclary, line 113: It would be cool to have some mec

app/configs/tempCrashesTable.ts
```diff
@@ -0,0 +1,57 @@
+import { tempCrashesListViewColumns } from "./tempCrashesListViewColumns";
+import { QueryConfig, FilterGroup } from "@/utils/queryBuilder";
+import { DEFAULT_QUERY_LIMIT } from "@/utils/constants";
+
+const columns = tempCrashesListViewColumns.map((col) => String(col.path));
+
+const tempCrashesListViewfilterCards: FilterGroup[] = [
```
johnclary, line 7: this is a bit of a hack, because we don't need to give user's the ability to switch **Temp records** on/off. i have a few ideas for this:

1. Add a `staticFilters` prop to the query config so that you can essentially hardcode extra filters that are not controlled through the UI
2. Move this whole feature into the main `/crashes` list and add a **Temporary records** filter to that page. Do we really need a separate page route just for creating these temp crashes?

app/components/FormControlDatePicker.tsx
johnclary, line 1: Hoping to establish a pattern here for custom react-hook-form components

app/components/CreateCrashRecordModal.tsx
```diff
@@ -0,0 +1,377 @@
...
+      ...DEFAULT_FORM_VALUES,
+      created_by: user?.email || "",
+      updated_by: user?.email || "",
+    },
+  });
+
+  const { fields, append, remove } = useFieldArray({
+    control,
+    name: "units_cris",
+  });
```
johnclary, line 154: this is a neat react-hook-form feature that enables repeatable arrays of form groups—exactly what we need for the **Units** section of this form

app/components/CreateCrashRecordModal.tsx
```diff
@@ -0,0 +1,377 @@
...
+                      </AlignedLabel>
+                    </Button>
+                  )}
+                </Card.Header>
+                <Card.Body>
+                  <Form.Group className="mb-3">
+                    <Form.Label>Unit type</Form.Label>
+                    {unitTypes && (
+                      <Form.Select
+                        {...register(`units_cris.${index}.unit_desc_id`, {
```
johnclary, line 278: here you can see the path accessor pattern for managing these groups of inputs—pretty cool!

it's somehow typesafe, too 😵‍💫

app/components/CreateCrashRecordModal.tsx
```diff
@@ -0,0 +1,377 @@
...
+  show: boolean;
+}
+
+type UnitInputs = {
+  unit_desc_id: string;
+  fatality_count: number;
+  injury_count: number;
+};
+
+type CrashInputs = {
```
mddilley, line 39: so nice to have all the expected fields for a temp crash in this type and the defaults below ✨

app/components/CreateCrashRecordModal.tsx
mddilley: oooh this is a TIL! I was wondering how this was going to work when I was testing. 🚀

app/components/CreateCrashRecordModal.tsx
```diff
@@ -0,0 +1,377 @@
...
+      </Modal.Header>
+      <Modal.Body>
+        <Form onSubmit={handleSubmit(onSubmit)} id="userForm">
+            {/* Case Id */}
+          <Form.Group className="mb-3">
+            <Form.Label>Case ID</Form.Label>
+            <Form.Control
+              {...register("case_id", { required: true })}
+              autoComplete="off"
+              data-1p-ignore
```
mddilley, line 195: 🙏

app/components/CreateCrashRecordModal.tsx
```diff
@@ -0,0 +1,377 @@
...
+          <Form.Group className="mb-3">
+            <Form.Label>Secondary address</Form.Label>
+            <Form.Control
+              {...register("rpt_sec_street_name", {
+                setValueAs: (v) => v?.trim() || null,
+              })}
+              autoComplete="off"
+              data-1p-ignore
+              placeholder="ex: 100 S CONGRESS AVE"
+              isInvalid={Boolean(errors.rpt_street_name)}
```
mddilley, line 251: @johnclary i saw red highlighting on this when I tested and had an invalid primary address, and I think this just need to be updated to `rpt_sec_street_name`.

app/components/CreateCrashRecordModal.tsx
```diff
@@ -0,0 +1,377 @@
...
+                    <Form.Label>Unit type</Form.Label>
+                    {unitTypes && (
+                      <Form.Select
+                        {...register(`units_cris.${index}.unit_desc_id`, {
+                          // coerce to number or null
+                          required: true,
+                          setValueAs: (v) => Number(v) || null,
+                        })}
+                        isInvalid={Boolean(
+                          errors.units_cris?.[index]?.unit_desc_id
```
mddilley, line 284: this is real cool to see how the index comes into play with fields array 😎

app/components/CreateCrashRecordModal.tsx
```diff
@@ -0,0 +1,377 @@
...
+                        ))}
+                      </Form.Select>
+                    )}
+                    <Form.Control.Feedback type="invalid">
+                      Unit type is required
+                    </Form.Control.Feedback>
+                    {isLoadingUnitTypes && <Spinner size="sm" />}
+                  </Form.Group>
+                  <Form.Group className="mb-3">
+                    <Form.Label>Fatlity count</Form.Label>
```
mddilley, line 301: tiny typo here - Fatality

app/components/FormControlDatePicker.tsx
```diff
@@ -0,0 +1,63 @@
...
+   * Optional invalid state while will cause the `is-invalid` css class
+   * to be applied
+   */
+  isInvalid?: boolean;
+}
+
+/**
+ * Date picker component for react-hook-form
+ */
+export default function FormControlDatePickerr<
```
mddilley, line 34: another small typo - Picker

app/configs/tempCrashesTable.ts
mddilley: Yeah, this is interesting to think about since the hardcoded filters would be hidden from users which could be confusing if used in other places. But, I also see that being able to see non-temp crashes isn't really needed or useful in this view.

I like the idea of merging the two views/tables since this is essentially a duplicate table. Especially since one of the outcomes of the data model work was rolling the temp crashes into the crashes table and introducing the `is_temp_record` column. It would also subtly communicate that all totals are coming from CRIS crashes + "temp" crashes.

mddilley (changes requested):
Looks and works great! There is cool stuff in this one, and I learned a new RHF trick. 🪄

I mentioned you in one comment to fix a field name that is crossing the wires in the form validation highlighting, but it is a small fix. Happy to see this form revamped! 🚀

app/components/CreateCrashRecordModal.tsx
johnclary: thank you, Mike! ✅

mddilley (approved):
Looked at the new code, and it looks amazing! So great to have RHF in this code. 😎 🚢

app/components/CreateCrashRecordModal.tsx
```diff
@@ -0,0 +1,377 @@
...
+              Crash date is required
+            </Form.Control.Feedback>
+          </Form.Group>
+          {/* Primary address */}
+          <Form.Group className="mb-3">
+            <Form.Label>Primary address</Form.Label>
+            <Form.Control
+              {...register("rpt_street_name", {
+                required: true,
+                setValueAs: (v) => v?.trim() || null,
```
roseeichelmann, line 230: could we maybe auto-capitalize this input to match how all the addresses look in our db? same with secondary address

app/configs/tempCrashesTable.ts
roseeichelmann: i think im with mike, im more interested in the idea of merging the two views. i dont think we need a separate page route

roseeichelmann (approved):
This looks sooo good! i like the idea of merging the create crash feature with the current crash page, i don't think we need to keep them separate. At first I thought person records weren't being created when fatalities/injuries were added but I realized that they are being added in the back end and the person related records table just isn't in this PR yet. 

I think there could be some more type validation also, like i should get errors when i try to save letters into the fatalities/injuries text boxes. But other than that this looks good to go to me :ship:

app/components/SidebarLayout.tsx
```diff
@@ -163,6 +164,13 @@ export default function SidebarLayout({ children }: { children: ReactNode }) {
                 label="Locations"
                 href="/locations"
               />
+              <SideBarListItem
+                isCollapsed={isCollapsed}
+                isCurrentPage={segments.includes("create-crash-record")}
+                Icon={FaCirclePlus}
+                label="Create crash"
```
chiaberry, line 171: I agree with Mike and Rose about combining the temporary crash records page with the crash records page, but if this does stay separate, I prefer "Create crash record" over "Create crash"

chiaberry (approved):
🚀

app/components/CreateCrashRecordModal.tsx
johnclary: yes—thanks. I updated this function to add uppercasing as well. the user will not see this while they're typing into the form, but it will be saved as upper case ✅

app/components/SidebarLayout.tsx
johnclary: thank you! i went with `FaFileCirclePlus`

app/components/SidebarLayout.tsx
johnclary: Now that you mention it, I prefer that as well. That label forces the sidebar to be a lot wider, and i think we definitely need to shrink the sidebar font size, but since we are going to move this into the crashes page anyway I will leave it as-is for now 🙏

app/configs/tempCrashesTable.ts
johnclary: thanks, y'all—here's the issue to nix to merge the two views: https://github.com/cityofaustin/atd-data-tech/issues/20469

johnclary:
Here's a look at the improved input validation. We still have a backlog issue to fix this on the crashes page. It's an easier fix here thanks to this being a one-off form versus the abstracted DataCardInput. 

Screenshot 2025-01-05 at 12 24 13 PM

PR #1644, opened 2024-12-31
Delete crash records

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/20433

Screenshot 2024-12-31 at 12 29 32 PM

I switched this up a bit from the old VZE, which had the delete button in the temp crashes list. Now the delete button is embedded in an alert banner at the top of the crash details page. A future enhancement will hide this button from non-editors.

I am making this change for these reasons:

1. It avoids the need to add a custom row action feature to our grid table
2. I assume it's customary to view the crash details before you delete the crash
4. Like i mentioned [here](https://github.com/cityofaustin/vision-zero/pull/1643#discussion_r1898711553), I am still hoping we can do away with the "Create crash record" page and use the normal crashes list.

## Testing

**URL to test:** Local

1. Create a crash record, as described [here](https://github.com/cityofaustin/vision-zero/pull/1643)
2. Navigate to the crash details page and delete your crash. Observe you are redirected to the `/create-crash-record` page.
3. Navigate back to the crash details page for the crash you deleted. You should be promptly redirected to the `404` page
4. Visit the crash details for a non-temporary crash. Verify that the banner + delete button are not visible.

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/app/components/CrashIsTemporaryBanner.tsx
+++ b/app/components/CrashIsTemporaryBanner.tsx
@@ -8,13 +8,8 @@ import { useMutation } from "@/utils/graphql";
 import { DELETE_CRIS_CRASH } from "@/queries/crash";
 
 interface CrashIsTemporaryBannerProps {
-  /**
-   * If the crash occurred on a private drive
-   */
   crashId: number;
 }
-
-// const onDelete = (crashId) = {}}
 
 /**
  * Banner that alerts the user when viewing a "temporary" crash record, aka
--- a/app/components/CrashIsTemporaryBanner.tsx
+++ b/app/components/CrashIsTemporaryBanner.tsx
@@ -1,4 +1,3 @@
-import { useState } from "react";
 import { useRouter } from "next/navigation";
 import Alert from "react-bootstrap/Alert";
 import Button from "react-bootstrap/Button";
--- a/app/components/CrashIsTemporaryBanner.tsx
+++ b/app/components/CrashIsTemporaryBanner.tsx
@@ -1,4 +1,5 @@
 import { useRouter } from "next/navigation";
+import { useAuth0 } from "@auth0/auth0-react";
 import Alert from "react-bootstrap/Alert";
 import Button from "react-bootstrap/Button";
 import Spinner from "react-bootstrap/Spinner";
@@ -18,6 +19,7 @@ interface CrashIsTemporaryBannerProps {
 export default function CrashIsTemporaryBanner({
   crashId,
 }: CrashIsTemporaryBannerProps) {
+  const { user } = useAuth0();
   const { mutate, loading: isMutating } = useMutation(DELETE_CRIS_CRASH);
   const router = useRouter();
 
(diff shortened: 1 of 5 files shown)


app/components/CrashLocationBanner.tsx
```diff
@@ -15,10 +17,16 @@ export default function CrashLocationBanner({
 }: CrashLocationBannerProps) {
   return (
     <Alert variant="warning">
-      This crash is not included in Vision Zero statistical reporting because{" "}
-      {privateDriveFlag
-        ? "it occurred on a private drive"
-        : "it is located outside of the Austin full purpose jurisdiction"}
+      <AlignedLabel>
+        <FaTriangleExclamation className="me-2 d-none d-lg-inline" />
```
johnclary, line 21: i added icons to both of these banners. it gets really squished on smaller displays, so i added the responsive display class so that it is hidden below `lg`.

app/queries/crash.ts
```diff
@@ -235,3 +235,17 @@ export const CREATE_CRASH = gql`
     }
   }
 `;
+
+export const DELETE_CRIS_CRASH = gql`
+  mutation DeleteCrisCrash($id: Int!) {
+    update_crashes_cris(
+      where: { id: { _eq: $id }, is_temp_record: { _eq: true } }
```
johnclary, line 242: using `is_temp_record: { _eq: true }` ensures that we cannot soft-delete anything but temp records 😅

app/components/CrashIsTemporaryBanner.tsx
```diff
@@ -0,0 +1,66 @@
...
+}
+
+/**
+ * Banner that alerts the user when viewing a "temporary" crash record, aka
+ * a crash record that was manually created through the UI
+ */
+export default function CrashIsTemporaryBanner({
+  crashId,
+}: CrashIsTemporaryBannerProps) {
+  const { mutate, loading: isMutating } = useMutation(DELETE_CRIS_CRASH);
```
mddilley, line 21: @johnclary Does this mutation need to soft delete the units too? I can't remember exactly how the totals are aggregated, but I am seeing `is_deleted = false` on the temp units after deleting the temp crash.

mddilley (commented):
I tested this on multiple temp crashes, and I could delete them successfully. I think the delete button being on the crash page is great, and I like the banner text that makes the purpose and status of a temp crash really clear to me.  🚀 I left one question about the temp units.

app/components/CrashIsTemporaryBanner.tsx
roseeichelmann: i was wondering the same thing too, and for any persons created as well

roseeichelmann (commented):
i think adding the delete button to the details page is a good solution, im curious if the VZ team will find it annoying to have to click thru if they have a lot to delete in a session though. not approving yet bc of the stray temp units/persons records that stick around after deleting like mike pointed out

johnclary:
thank you for catching that i overlook soft deletes of units and people—y'all are the best! fixed with https://github.com/cityofaustin/vision-zero/pull/1644/commits/7e598bee4a3850cc7b521ce39b3016e00708b5c4.

yes, I am also curious if the VZ users will find this change annoying. nbd if we need to revist 🙏 

I'm going to go ahead and merge this up.

PR #1645, opened 2025-01-02
Add total record count to tables

johnclary:
## Associated issues
* https://github.com/cityofaustin/atd-data-tech/issues/20202

This PR adds a record count display to the table pagination. This feature is being revived thanks to Rose fixing our database and making it fast enough to deal with aggregation queries 🙌 

 How can we make this look better?

Current (gray record count, primary pagination controls):

Screenshot 2025-01-02 at 2 06 03 PM

Gray record count and gray page number:

Screenshot 2025-01-02 at 2 05 45 PM

Primary everything with outline:

Screenshot 2025-01-02 at 2 04 29 PM

I also played around with re-labeling the page # to `Page 1 / 11` and short and sweet like  `1 / 11`. But i find that i don't really care how many pages there are so long as i know the total record count 🤷 

Here's the ACC design. Maybe I'll just leave as-is and revisit when we add record exporting?

Screenshot 2025-01-02 at 4 43 48 PM

## Testing

**URL to test:** Local or [Netlify](https://deploy-preview-1645--vision-zero-nextjs.netlify.app/)

Test out the crashes and locations list and verify that record counts are displaying correctly. Visually, do you hate it or love it?

---

#### Ship list

- [ ] Code reviewed


app/types/graphql.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+export interface HasuraAggregateData {
+  aggregate: {
+    count?: number;
+    sum: Record<string, number>;
+    avg: Record<string, number>;
+    max: Record<string, number>;
+    min: Record<string, number>;
+  };
+}
```
johnclary, line 15: We're only using `count` at the moment, but i went ahead and added these common aggregates 🤷

app/utils/graphql.ts
```diff
@@ -67,33 +97,31 @@ export const useQuery = <T extends Record<string, unknown>>({
...
+  hasAggregates,
+}: UseQueryProps): UseQueryResponse<T> => {
   const { user } = useAuth0();
   const token = useToken();
 
   const fetchWithAuth = async ([query, variables]: [
     RequestDocument,
     Variables
-  ]): Promise<{ [K in typeof typename]: T[] }> => {
+  ]): Promise<HasuraGraphQLResponse<T, typeof typename>> => {
```
johnclary, line 108: Before this change, we were telling SWR that our query always returned an object with a single property that matches the given `typename` and array of whatever Type (`T`), e.g.

```javascript
{ crashes : []}
```

Now the type looks like:

```javascript
{ crashes : [], crashes_aggregate: {} }
```

app/types/graphql.ts
```diff
@@ -0,0 +1,28 @@
...
+ * and optionally including aggregates.
+ *
+ * `T` is the type of object being returned
+ * `TTypeName` is the literal name of the query root
+ */
+export type HasuraGraphQLResponse<T, TTypeName extends string> = {
+  [K in TTypeName]: T[];
+} & {
+  [K in `${TTypeName}_aggregate`]?: HasuraAggregateData;
+};
```
johnclary, line 28: This generic is shaping our Hasura return object, which would look something like this:

```javascript
{ crashes : [], crashes_aggregate: {} }
```

roseeichelmann (commented):
honestly i think i like how the pagination worked more before this PR bc the next/prev buttons were obviously disabled when you reached the last page or if you were already on the first page. but now with just the little arrows its hard to tell when they are disabled so i found myself trying to click next when i was already on the last page. 

if the arrows turned grey when disabled that would be easier to see than turning a lighter blue color, i just tested that. so that would be my recommendation. other than that, i think this looks good

app/components/TablePaginationControls.tsx
```diff
@@ -47,8 +58,7 @@ export default function TablePaginationControls({
           }}
         >
           <AlignedLabel>
-            <FaCircleArrowLeft />
-            <span className="ms-2">Prev</span>
+            <FaAngleLeft />
```
frankhereford, line 61: This is a response to some of the questions which you pose in your PR's top-line comment. I'm with you about not caring about how many pages there are generally, but I do care about where I am in relation to the book-ends of those pages. To wit, I would suggest we add some `<<` and `>>` buttons on the outside of the `<` and `>` ones that beam to you to the first or last page respectively. This gives me a way to get to the end of the set if I do happen to care about the number of pages, which does happen sometimes.

For example, here's a little non-fiction drawn from personal experience: If one gets the task of QA/QC'ing all the fatal and serious injury crashes in the year 2017, knowing how many pages you have left is really helpful.

app/components/TableWrapper.tsx
frankhereford, line 1: Entirely unrelated to this PR, but since you may have this file open -- can you pick off a couple of typos?

Line 33: `kepts` → `kept`
Line 68: `dont` → `don't`

app/components/TablePaginationControls.tsx
```diff
@@ -6,28 +6,39 @@ import ButtonToolbar from "react-bootstrap/ButtonToolbar";
...
   queryConfig,
   setQueryConfig,
   recordCount,
   isLoading,
 }: PaginationControlProps) {
   const currentPageNum = queryConfig.offset / queryConfig.limit + 1;
+  const totalRecords = aggregateData?.aggregate?.count || 0;
+
   return (
     <ButtonToolbar>
```
frankhereford, line 35: To your questions about visual style, I'm a fan of it just the way you have it.

The interactive bits are blue (the most classic link-color) and they are using a single character to represent the action, in a light stroke. It's so direct, it's so correct. 

 As many labels, borders, colors, decorations, nicknacks, and general clutter that can be left behind -- the better.

app/components/TablePaginationControls.tsx
```diff
@@ -47,8 +58,7 @@ export default function TablePaginationControls({
           }}
         >
           <AlignedLabel>
-            <FaCircleArrowLeft />
-            <span className="ms-2">Prev</span>
+            <FaAngleLeft />
           </AlignedLabel>
         </Button>
         <Button
```
frankhereford, line 64: I'm not entirely sure why this label is wrapped in a button, but I can dig it if it's what makes this look like it does inside this `<AlignedLabel>`. It feels like it wants to be a div or a span maybe?

But anyway -- the real thing I want to propose here is likely beyond the scope of this issue, but may be worth spinning off into the backlog:

What would you think about a little "hidden functionality" where if you click on the `Page _n_` label, you get a drop-down from which you can select which page you want to jump to. The functionality is non-critical to the use of the page, so I feel that it's entirely fine if we don't indicate that it's possible, and just let our power-users discover it in their own time and way.

frankhereford (approved):
This is ready to ship! I have a couple of ideas that I want to throw out there, but if this is how this ends up, it's a total winner.

chiaberry (commented):
I don't like the page number having the same style as clickable buttons, if it is not itself a clickable button. And being semantically tagged a button you can select it through tabbing, which is confusing that nothing happens when you hit enter.

frankhereford:
@chiaberry wrote:

> And being semantically tagged a button you can select it through tabbing

This is a super good point. I am totally convinced by this that the it's worth investigating how to get a different mark-up in there for this element other than `<Button>`. Thanks for this observation Chia.

johnclary:
Thank you for all the great feedback. I have a preference for the angle icons versus the arrows, so here's what I did:

- show page count in gray/secondary color
- show button in secondary color if disabled
- use a `<span>` for the page number indictor instead of a `<button>`
- remove unnecessary `AlignedLabel` element 

Please let me know if that's agreeable—and if you still have feeling strong feelings about changes to this. We do have the generic app theme issue ([#20013](https://github.com/cityofaustin/atd-data-tech/issues/20013)) where I am happy to stash additional design todos.

Lastly—thank you Frank for the input on the more sophisticated pagination controls, such as jumping to the end, showing the  total page count, and jumping to a specific page. I'm going to put those in a follow-up enhancement since they are squarely new VZE functionality but I agree they'd be quite nice to have for our power users 🙏

chiaberry (approved):
I approve of the changes!

app/components/TablePaginationControls.tsx
```diff
@@ -31,6 +30,10 @@ export default function TablePaginationControls({
   const currentPageNum = queryConfig.offset / queryConfig.limit + 1;
   const totalRecords = aggregateData?.aggregate?.count || 0;
 
+  const pageLeftButtonDisabled =
+    recordCount === 0 || queryConfig.offset === 0 || isLoading;
+  const pageRightButtonDisabled = recordCount < queryConfig.limit || isLoading;
```
frankhereford, line 35: This change, although it may seem small, transmits a lot of information to the user in a subtle way. Thanks so much for adding this.

frankhereford (approved):
I dig the new changes a ton; thank you for them, and I am 100% cool with where you've drawn the line between what should go in this PR vs future enhancements. 🚢🚢🚢🚢 Thanks John!

johnclary:
I am merging but feedback is still welcome 🙏

PR #1648, opened 2025-01-06
Fix year rollover time of day chart bug

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/20471

## Testing

https://deploy-preview-1648--atd-vzv-staging.netlify.app/

**Steps to test:**

Scroll down to the "By Time of Day" chart and click through the different tabs. 2023 and 2024 have different heatmaps. Compare that to staging (https://visionzero.austin.gov/viewer/) where they do not. 

The code was pulling the current year from today's date and not the current year from the data. so it was off by a year on the tabs. 

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/viewer/package-lock.json
+++ b/viewer/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "viewer",
-  "version": "1.44.1",
+  "version": "2.4.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "viewer",
-      "version": "1.44.1",
+      "version": "2.4.1",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
--- a/viewer/src/views/summary/CrashesByTimeOfDay.js
+++ b/viewer/src/views/summary/CrashesByTimeOfDay.js
@@ -94,7 +94,8 @@ const calculateHourBlockTotals = (records, crashType) => {
  * @returns {String} The query url for the Socrata query
  */
 const getFatalitiesByYearsAgoUrl = (activeTab, crashType) => {
-  const yearsAgoDate = format(sub(new Date(), { years: activeTab }), "yyyy");
+  // subtract years ago (based on activeTab) from current year
+  const yearsAgoDate = format(sub(parseISO(summaryCurrentYearStartDate), { years: activeTab }), "yyyy");
   let queryUrl =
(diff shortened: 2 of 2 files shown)


viewer/package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "viewer",
-  "version": "1.44.1",
+  "version": "2.4.1",
```
frankhereford, line 3: I think it's fine to include this version bump in the lock file, but I would normally think this would happen in response to an updated library or some other change in the `package.json` file. Is there a corollary change to that file you want to include, or maybe not include this? 

And also - leaving it in is fine too, since it's a legit reflection of the current version of the viewer.

frankhereford (approved):
This looks great Chia - thank you for fixing this. I hope this is is it for any new-year fire drill fixes!

roseeichelmann (approved):
thanks for the quick fix chia! :zap:

viewer/package-lock.json
roseeichelmann: she could have ran `npm install` which will auto update the version, i think its fine to include since it reflects the current version :shrug:

johnclary: yep, the lock file version is being brought back into sync with `package.json`. i guess as a best practice we should be running `npm install` after bumping our own app version...

johnclary (approved):
looks great—thank you for tracking this down so quickly!!!!!

i asked a question in the issue:

> A related question: when should we expect the VZV to show year 2025 data?

It looks like we've [hardcoded the whole app](https://github.com/cityofaustin/vision-zero/blob/717d31045247ca1c698a1738dc1b729f1f65513c/viewer/src/constants/time.js#L31) to work on a 14-day shift?

mddilley (approved):
Thanks for fixing! 🙌

frankhereford:
@johnclary wrote:

> It looks like we've [hardcoded the whole app](https://github.com/cityofaustin/vision-zero/blob/717d31045247ca1c698a1738dc1b729f1f65513c/viewer/src/constants/time.js#L31) to work on a 14-day lag?

That's how I've understood it to work from even back when I was on the VZ team. The lag let's the VZ crew QA/QC stuff and to keep the numbers from "shifting around" as they do that QA/QC.

PR #1650, opened 2025-01-06
patch: Fix year rollover time of day chart bug

chiaberry:
## Associated issues

https://github.com/cityofaustin/vision-zero/pull/1648

Should I bump the editor version too? I just realized I didnt bump any versions in my PR to main

### DIFF
--- a/viewer/package.json
+++ b/viewer/package.json
@@ -1,6 +1,6 @@
 {
   "name": "viewer",
-  "version": "2.4.2",
+  "version": "2.4.3",
   "homepage": "/viewer",
   "description": "Vision Zero Viewer",
   "author": "Austin Transportation & Public Works - Data & Technology Services",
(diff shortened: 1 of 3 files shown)


viewer/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "viewer",
-  "version": "2.4.2",
+  "version": "2.4.3",
```
frankhereford, line 3: The "from" number here is different than the PR from earlier -- is that cool?

frankhereford (approved):
🚢🚢🚢

viewer/package.json
chiaberry: good point, i didnt touch the package.json in the last PR and just assumed the package-lock was matching the package.json, i should have updated the version to main too, shouldnt i have?

johnclary (approved):
thank you!

viewer/package.json
johnclary: yes please— both of these commits need to make there way into main

PR #1651, opened 2025-01-06
Add jurisdiction layer to helper

johnclary:
### Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/19441
* This PR unblocks https://github.com/cityofaustin/atd-data-tech/issues/20461, which is why I prioritized it 👍 

This PR brings the jurisdiction layer into our normal geo layer maintenance process. It replaces the old `atd_jurisdictions` layer that was created through an ad-hoc process that merged all features by jurisdiction label. Going forward, the table will resemble the authoritative CTM layer, where there may be multiple rows (aka features) for a single jurisdiction type. 

Also, whereas the old DB table had rows for the various ETJs and surrounding jurisdictions, I elected to exclude these since we don't currently make any use of them.

#### Before

Screenshot 2025-01-06 at 4 34 51 PM

#### After

Screenshot 2025-01-06 at 4 34 30 PM

## Testing

1. Start your local setup
2. Apply the DB changes: `hasura migrate apply` and `hasura metadata apply`
3. Run the helper script to load the jurisdiction layer from AGOL:

```shell
 node --env-file=local.env src/load_agol_layer.js --layer jurisdictions
```

4. Start the ole VZE. Move some crashes into and out of the AFP jurisdiction. Moving a crash out of AFP will cause the alert banner to appear above the crash map.

5. The VZ team pointed out that `20271128` should not be assigned to AFP. You can test that by moving it 1 pixel and observing that `in_austin_full_purpose` changes to `false`

6. I have prepared [these sql commands](https://github.com/cityofaustin/atd-data-tech/issues/20461#issuecomment-2576082208) to backfill the various AFP flags that we have for crashes, EMS, and AFD records. If you would please review and run each of those and confirm that they look good 🙏 

---

#### Ship list

- [ ] Code reviewed


database/migrations/default/1736197283308_jurisdiction_refactor/up.sql
```diff
@@ -0,0 +1,143 @@
...
+
+        raise debug 'found location: % compared to previous location: %', new.location_id, old.location_id;
+        --
+        -- check if in austin full purpose jurisdiction
+        --
+        new.in_austin_full_purpose = EXISTS (
+            select 1 
+            from geo.atd_jurisdictions 
+            where JURISDICTION_LABEL = 'AUSTIN FULL PURPOSE' AND  ST_Intersects(geometry, new.position)
+        );
```
johnclary, line 46: here's the updated code. we're filtering on the jurisdiction label rather than the arbitrary ID sequence

toolbox/load_agol_layer/src/settings.js
```diff
@@ -53,6 +53,13 @@ const LAYERS = {
     fields: ["ENGINEERING_AREA_ID", "ATD_ENGINEER_AREAS"],
     tableName: "engineering_areas",
   },
+  jurisdictions: {
+    service_name: "BOUNDARIES_jurisdictions_planning",
+    layer_id: 0,
+    query_params: { ...DEFAULT_ESRI_QUERY_PARAMS, where: "JURISDICTION_LABEL = 'AUSTIN FULL PURPOSE'" },
```
johnclary, line 59: we're filtering out all features that are not AFP. there are thousands of features in this layer, and a single jurisdiction label can have multiple rows.  E.g. there are currently two `AUSTIN FULL PURPOSE` features which, when combined, represent the entire AFP.

toolbox/load_agol_layer/README.md
johnclary, line 1: before this change, our [geospatial docs](https://github.com/cityofaustin/vision-zero/tree/main/database#geospatial-layers) were a lie. 


The `atd_jurisdiction` layer was not an "ArcGIS Online authoritative layer owned by CTM GIS". It was a custom layer. Now it is true.

database/migrations/default/1736197283308_jurisdiction_refactor/up.sql
```diff
@@ -0,0 +1,143 @@
+alter table geo.atd_jurisdictions add column jurisdictions_id integer;
```
frankhereford, line 1: Is there any way at all we can drop the `atd_` out of this table name? `coa_` is appropriate here, or my true preference: just `jurisdictions`. 

`atd` is no longer, and while `tpw` is marginally better, these jurisdictions belong to the larger city. This similar thing happened with the new VZ S3 bucket name, and it'd be great to avoid this usage here because they are harder to change the longer they remain in place.

database/migrations/default/1736197283308_jurisdiction_refactor/up.sql
frankhereford: 👋 goodbye magic number - thank you!

database/migrations/default/1736197283308_jurisdiction_refactor/up.sql
```diff
@@ -0,0 +1,143 @@
...
+        -- check if is_coa_roadway
+        --
+        if (new.in_austin_full_purpose and not new.private_dr_fl) then
+            new.is_coa_roadway = not st_contains((select geometry from geo.non_coa_roadways), new.position);
+            raise debug 'is_coa_roadway: % compared to previous: %', new.is_coa_roadway, old.is_coa_roadway;
+        else
+            new.is_coa_roadway = false;
+        end if;
+        else
+            raise debug 'reseting spatial attributes due to null latitude and/or longitude values';
```
frankhereford, line 118: typo of `resetting` as `reseting`

database/migrations/default/1736197283308_jurisdiction_refactor/down.sql
```diff
@@ -0,0 +1,137 @@
...
+        -- check if is_coa_roadway
+        --
+        if (new.in_austin_full_purpose and not new.private_dr_fl) then
+            new.is_coa_roadway = not st_contains((select geometry from geo.non_coa_roadways), new.position);
+            raise debug 'is_coa_roadway: % compared to previous: %', new.is_coa_roadway, old.is_coa_roadway;
+        else
+            new.is_coa_roadway = false;
+        end if;
+        else
+            raise debug 'reseting spatial attributes due to null latitude and/or longitude values';
```
frankhereford, line 114: typo on `resetting` as `reseting`

database/migrations/default/1736197283308_jurisdiction_refactor/up.sql
```diff
@@ -0,0 +1,143 @@
...
+        end if;
+
+        raise debug 'found location: % compared to previous location: %', new.location_id, old.location_id;
+        --
+        -- check if in austin full purpose jurisdiction
+        --
+        new.in_austin_full_purpose = EXISTS (
+            select 1 
+            from geo.atd_jurisdictions 
+            where JURISDICTION_LABEL = 'AUSTIN FULL PURPOSE' AND  ST_Intersects(geometry, new.position)
```
frankhereford, line 45: I think you can get some meaningful speed-up on this where clause by doing:

```
where 
JURISDICTION_LABEL = 'AUSTIN FULL PURPOSE' 
AND geometry && new.position
AND  ST_Intersects(geometry, new.position)
```

`&&` is the "has intersecting bounding boxes" operator, and it's a zillion times faster to see if a point and a box overlap than a point and an arbitrary polygon. By doing this blazing fast operation first, you can skip over checking the `st_intersect()` on all the pairs of geometries that don't have overlapping bounding boxes, knowing that they couldn't intersect no matter what.

Also PG's GIST indexing on geometry is aimed at exactly this -- it stores the bounding boxes of geometries as its index data, so `&&` can get really lightning fast.

toolbox/load_agol_layer/README.md
frankhereford: I just want to point out that as [the author](https://github.com/cityofaustin/vision-zero/commit/1758bfdeb3cf4bee17fdf723bbce02079397e028) of those docs, you have all the authority one could have to call out that they were in error, but for all other readers, I want to push back on calling these a lie. That feels a little too strong and doesn't capture the nuance here.

While this work brings them fully into compliance as written in the README, they most certainly were previously derived from the CTM data and used in good faith. This data is much better in its provenance, but they were far from fabricated or intentionally misleading previously.

frankhereford (changes requested):
Just to confirm -- you do want this into `main` and not the collector branch? That sounds right to me, but I wanted to double check that you're not maybe wanting to rebase this.

The testing worked out great, and I think this is ship-shape for the big moving parts. I do have a pair of observations that I'd like to put forward enough to request changes, but I am happy to switch this to a ✅ after some consideration of them.

Of my observations, I'm most interested in not using the deprecated `atd_` acronym, and also short-circuiting out that `st_intersects()` by the query engine where possible via the usage of `&&`.

When testing the SQL backfill commands, I get these number of updated rows, in order of the queries presented:

```
Updated Rows	29
Updated Rows	86
Updated Rows	9
Updated Rows	18
Updated Rows	29
Updated Rows	36
```

chiaberry (approved):
I got the same updated rows counts as Frank did when running the backfill commands, ✅ 

I really appreciate the comments in the migration, it helps me understand what is going on.

toolbox/load_agol_layer/src/settings.js
mddilley: thanks for pointing this out because I was curious about the 2 rows in the DB. Makes sense! 🙏

mddilley (approved):
I ran through the test steps and see the in and out of AFP calculating as expected when using the editor. I also think nixing the `atd_` in the table would be nice like Frank mentioned (probably because everything is already nice and tidy in here ✨). 

Cool to see yet another AGOL layer covered with this tooling! 😎 🙌

database/migrations/default/1736197283308_jurisdiction_refactor/up.sql
johnclary: thank you for nudging me on this! i was a bit worried about blowing up the scope of this, but you're totally right: might as well get this cleaned up once and for all now. i checked with Milad and Xavier and they are ok with this change. i'll push it up!

database/migrations/default/1736197283308_jurisdiction_refactor/up.sql
johnclary: i ran those two statements on staging and the results are consistent: removing the `&&` check results in better performance by nearly 50% 😵

frankhereford: John - What a twist! Running those two queries on my machine come up with a less dramatic difference in timing 765 ms vs 515 ms, but the extra clause I'm suggesting really does seem to be the difference that is slowing it down.  I'm fully confused about why it's doing it this way, but the numbers don't like -- thank you for testing this out and providing such a clear cut set of data. 🙏

frankhereford (approved):
Thank you John for picking off that `atd_` prefix and also for digging into my misbelief that the bounding box operator would speed things up! Can't wait to see this go live!

roseeichelmann (approved):
this all looks great john! i reviewed these [update statements](https://github.com/cityofaustin/atd-data-tech/issues/20461#issuecomment-2576082208) and want to make sure you also update `afd_incidents` which are now AFP as well if thats necessary cuz i didnt see that command there. i agree on dropping `atd_`, i dont see that update pushed up yet but i think thats a great change to clean things up! :broom: :sparkles:

toolbox/load_agol_layer/README.md
johnclary: thank you, Frank—"lie" was a bit harsh but i was a bit surprised to find that I hadn't correctly documented the provenance of that layer. glad it is square now 😊

johnclary:
folks, i requested review because the table name change impacted quite a lot of code, including dropping a dormant function from the pre-LDM days 🙏

frankhereford (approved):
John - this looks good! I have run the DB up through the migrations and metadata presented in this PR as well as gone over the code. This is on top of my previous review, so I am feeling really great about this.

## Thank you

Thanks for taking on the entirely out-of-scope request to rename that field and drop the depreciated department name. That was no small thing, and I am really grateful to you. 🙏

roseeichelmann (approved):
retested, it looks great!! :ship:

johnclary:
headed to `main`— thank you!

database/migrations/default/1736197283308_jurisdiction_refactor/up.sql
johnclary: really truly a twist! i'd love to spend some more time understanding what's going on here. i'd love to re-test this scenario against a table with more rows, like `atd_txdot_locations`....

PR #1653, opened 2025-01-07
Crash details: Ability to swap addresses

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/19961

## Testing

**URL to test:** 
Local or [netlify](https://deploy-preview-1653--vision-zero-nextjs.netlify.app/crashes)

**Steps to test:**
1. Go to a crash details page, scroll down to the primary address data card and see the new swap address button. 
2. Test clicking it, see the window that pops up, confirm that the address fields swap correctly on confirm.
3. Is it weird that you can swap while editing a field?
4. Scroll down and see your changes in the record history

---

#### Ship list

- [ ] Code reviewed


app/app/users/[user_id]/page.tsx
```diff
@@ -132,7 +132,7 @@ export default function UserDetails({
                       onClick={() => {
                         if (
                           window.confirm(
-                            "Are you user you want to delete this user?"
+                            "Are you sure you want to delete this user?"
```
Comment (roseeichelmann) [inline app/app/users/[user_id]/page.tsx:135]: wittle typo

app/components/SwapAddressButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  return (
+    <div>
+      <Button
+        size="sm"
+        variant="primary"
+        disabled={isMutating}
+        onClick={async () => {
+          if (
+            window.confirm(
```
roseeichelmann, line 49: john and i talked about this and in the future we may create a reuseable confirm/cancel modal to use across the app but for now we are going with window.confirm

roseeichelmann:
should we do more auto uppercase on user input here/other places in the app?
image

chiaberry (approved):
It does feel weird that if you are editing and then click the swap addresses button, the input stays open with the previous value. i was trying to come up with a scenario where someone would accidentally duplicate the primary address instead of swapping, but that feels far fetched and the values are in the change log, so nothing would really be lost. 

Visually i want the lines to line up better, but its like this in prod now anyway, so mvp mvp 🚢 
Screenshot 2025-01-09 at 4 50 54 PM

chiaberry:
I just saw the comment about auto all caps -- and if its thats how it coming from CRIS, then I think so yes, we should make it match the rest of the data. But I can be convinced otherwise!

frankhereford (commented):
I'm on the fence between request-changes and comment because I think that you're 100% on to something with your question:

> Is it weird that you can swap while editing a field?

I think it's very worth it to either disable the Swap Addresses button when you're editing or to exit out of edit mode without warning and without saving if the user hits Swap Addresses when they are editing. 

The two features do not work well together at all, and it's verrrry easy to end up with the same value in both of the addresses if you hit swap while you're editing. 

But - aside from that one thing that you've already seen coming, this is good to go, and I will give this a ✅ after some discussion and possible changes to that editing-while-swapping behavior.

app/app/users/[user_id]/page.tsx
Comment (johnclary) [reply app/app/users/[user_id]/page.tsx:135]: 🙈

johnclary (changes requested):
@roseeichelmann it is so great to see this feature come to live in the new VZE!

my main feedback is about making the `DataCard` component more flexible and doing away with the using the card title to conditionally render this component.

i agree with others that it would be nice to reset the field editing stating when you swap addresses, although sorting out how to do that in combination with a more abstract header component looks to me, at first glance, to be fairly challenging and something I'd prefer to leave for a follow-up issue.

frankhereford (approved):
@johnclary wrote:

> i agree with others that it would be nice to reset the field editing stating when you swap addresses, although sorting out how to do that in combination with a more abstract header component looks to me, at first glance, to be fairly challenging and something I'd prefer to leave for a follow-up issue.

I think that is perfectly reasonable to move that work into a future issue, and given that, this PR is ready to roll. Thank you so much @roseeichelmann!

app/components/DataCard.tsx
```diff
@@ -3,6 +3,7 @@ import Card from "react-bootstrap/Card";
 import Spinner from "react-bootstrap/Spinner";
 import Table from "react-bootstrap/Table";
 import DataCardInput from "./DataCardInput";
+import SwapAddressButton from "./SwapAddressButton";
```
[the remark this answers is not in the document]
roseeichelmann: im absolutely for the absolute movement (ha) i just reconfigured my code editor to auto complete imports to be absolute so hopefully yall wont have to keep catching this in my PRs :p

roseeichelmann:
@johnclary this is ready for re-review, and im happy to address the bugginess of the editing state in a follow up issue!

johnclary (approved):
really great work here—thanks for that refactor! 🚢 🚢 🚢 🚢 🚢

PR #1654, opened 2025-01-09
Move temporary records functionality to crashes list

roseeichelmann:
## Testing

**URL to test:** 

**Steps to test:**

---

#### Ship list

- [ ] Code reviewed


app/configs/crashesListViewTable.ts
```diff
@@ -203,7 +203,8 @@ const crashesListViewfilterCards: FilterGroup[] = [
   {
     id: "internal_filters",
     label: "Internal",
-    groupOperator: "_or",
+    // because one of the filters is enabled and inverted we need to join using the "_and" operator
+    groupOperator: "_and",
```
johnclary, line 207: 👍 🙏

johnclary (changes requested):
@roseeichelmann thanks for making such quick work of this—it looks fantastic! One small thing: when you delete a temp crash, the `CrashIsTemporaryBanner` is still routing to the old `/create-crash-record` page. that can be updated to `/crashes`.

roseeichelmann:
@johnclary such a great catch, just fixed this!

johnclary (approved):
🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

app/components/SidebarLayout.tsx
frankhereford, line 1: This is totally outside the scope of this PR, but while I was reviewing it, a typo jumped out at me in this file. Would you mind adding a space to the comment `  /** Check local storage for initialsidebar state */` on line 42? No big if not!!

app/configs/crashesListViewTable.ts
```diff
@@ -220,6 +221,20 @@ const crashesListViewfilterCards: FilterGroup[] = [
           },
         ],
       },
+      {
+        id: "is_temp_record",
+        label: "Temporary records",
```
frankhereford, line 226: What would you think about adding the word "Only" to the label here, so it reads "Only temporary records?" The idea being that the two internal filters have a very different interaction to the existing set. 

"Include private drive crashes" adds to the existing set by allowing in more crashes, and "Temporary records" is unclear if it's going to limit to just temporary records or allow the temporary records to be added to the existing set being shown. 

Because it ends up doing a different operation (limit as opposed to add-to) as the previous option, I think adding a verb to the label can help the user get the right expectation.

I feel like I've done a really poor job in explaining my rational - LMK if i can help explain any of this better. 

-------

Edit: I was reading more of this as I reviewed, and you pointed out how the one above is `_and` and this one is still an `_or`. They are the right operations, of course -- we just need to broadcast that nuance in the labels to the user.

frankhereford (approved):
This fulfills the request on the issue perfectly -- 💯. Code looks great too -- thank you for this Rose!

My only suggestion is micro, and is related only to the label of the new filter and not the functionality. As is, though, this is plenty shippable. 🚢🚢🚢🚢

app/configs/crashesListViewTable.ts
johnclary: Great point! Rose, would you rename the switch label to `"Temporary records only"`? (I prefer putting "only" at the end bc it makes it easier to visually scan the filter labels)

🙏

PR #1655, opened 2025-01-10
Crash details: make change log collapsible

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/19968

## Testing

**URL to test:** 
Local or netlify

**Steps to test:**
1. Go to a crash details page, scroll down to the change log. It should be collapsed on default first load.
2. Now expand it and refresh the page. The new state should have saved in local storage so on refresh it should still be expand it.
3. Collapse it and refresh again, making sure your state is saved. Go to other crashes and make sure the saved state is persisting.

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/app/components/ChangeLog.tsx
+++ b/app/components/ChangeLog.tsx
@@ -1,6 +1,6 @@
 import { useMemo, useState } from "react";
-import Card from "react-bootstrap/Card";
 import Table from "react-bootstrap/Table";
+import Accordion from "react-bootstrap/Accordion";
 import { formatDateTime } from "@/utils/formatters";
 import ChangeLogDetails from "./ChangeLogDetails";
 import {
@@ -8,6 +8,9 @@ import {
   ChangeLogDiff,
   ChangeLogEntryEnriched,
 } from "@/types/changeLog";
+
+// used to track the accordion expanded state of the change log
+const localStorageKey = "expandedItem";
 
 const KEYS_TO_IGNORE = ["updated_at", "updated_by", "position"];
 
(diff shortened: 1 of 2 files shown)


app/components/ChangeLog.tsx
```diff
@@ -77,59 +80,72 @@ const isNewRecordEvent = (change: ChangeLogEntryEnriched) =>
...
-            </tr>
-          </thead>
-          <tbody className="font-monospace">
-            {changes.length === 0 && (
-              // this should only happen in local dev where change log is not downloaded from replica
+    <Accordion
+      defaultActiveKey={expandedItem}
+      // on accordion click save new expanded item state to local storage
+      onSelect={(e) => {
+        const localStorageValue = e !== null ? "0" : JSON.stringify(null); // local storage value must be a string type
```
roseeichelmann, line 94: i was getting a typescript linting error when just setting to null "Type 'null' is not assignable to type 'string'." and this is just never something i would have been super aware before programming in a strongly typed language. its weird but interesting having to take into account the parameter typing of external methods that we dont control, like `setItem`,

mddilley: i'm glad you shared about this in dev sync because it is a TIL for me too! Purely in opinion zone here but I think it would be okay to use the String constructor to make null into a string to work around this too. 🤷

I do think the `e` could be change to `eventKey` after I found myself wondering how these Bootstrap accordions work. The [docs](https://react-bootstrap.netlify.app/docs/components/accordion/#api) show that the first parameter in the callback is a string and the second is an event object.

mddilley (commented):
Looks great and I'm happy to see this feature carrying over into the new app. I left one suggestion for the code, but this works just as expected! 💯

app/components/ChangeLog.tsx
```diff
@@ -1,6 +1,6 @@
 import { useMemo, useState } from "react";
-import Card from "react-bootstrap/Card";
 import Table from "react-bootstrap/Table";
+import Accordion from "react-bootstrap/Accordion";
 import { formatDateTime } from "@/utils/formatters";
 import ChangeLogDetails from "./ChangeLogDetails";
```
johnclary, line 5: could convert to an absolute import 😇

app/components/ChangeLog.tsx
johnclary: > I think it would be okay to use the String constructor to make null into a string to work around this to

i like this suggestion a lot. my inclination would be to restrict `expandedItem` to being a string only, and deal with conversion to a string with the `String` constructor. And that way we can always store whatever `String(e)` into local storage without caring about it's value beyond that.

also, for what it's worth we don't have to use `"0"` as the key value. we could use something that's more obviously a string, like `"recordHistory"`. and it might to be good to stash that value as `const` var 🤷 

One other detail @roseeichelmann—I think that instead of the `defaultActiveKey` prop, the `activeKey` prop is a better choice to use instead. i see this kind of like using the `value` prop instead of `defaultValue` prop on a controlled form input.

app/components/ChangeLog.tsx
```diff
@@ -9,6 +9,9 @@ import {
   ChangeLogEntryEnriched,
 } from "@/types/changeLog";
 
+// used to track the accordion expanded state of the change log
+const localStorageKey = "expandedItem";
```
johnclary, line 13: would you mind making this a little more descriptive and specific instead of `"expandedItem"`, like `"crashHistoryCardExpandedItem"`?

johnclary: btw, i created https://github.com/cityofaustin/atd-data-tech/issues/20575 to add in support for version local storage across the app 👍

johnclary (changes requested):
Rose—this is looking really good. I requested to changes to tighten up some typing and naming.

I also want to remark that I'd like to nix the blue background on the accordion header when it's expanded. but i am happy to stash that in a follow-up issue. as part of that follow-up work it would be nice toto make the header look the same as the bootstrap `Card` header for consistency with the other cards on the page, i think?

app/components/ChangeLog.tsx
```diff
@@ -77,59 +80,72 @@ const isNewRecordEvent = (change: ChangeLogEntryEnriched) =>
...
+      defaultActiveKey={expandedItem}
+      // on accordion click save new expanded item state to local storage
+      onSelect={(e) => {
+        const localStorageValue = e !== null ? "0" : JSON.stringify(null); // local storage value must be a string type
+        localStorage.setItem(localStorageKey, localStorageValue);
+        setExpandedItem(localStorageValue);
+      }}
+    >
+      <Accordion.Item eventKey="0">
+        <Accordion.Header>Record history</Accordion.Header>
```
frankhereford, line 100: This header gets a baby blue color when it's expanded, and it's the only header row in on the page to get that color. Can it be `rgba(33, 37, 41, 0.03)` like the rest of of the card headers?

frankhereford (approved):
This looks super Rose. Really well done. I think the comments from other reviewers all make sense and get support from me, and I added one little thing concerning a color choice, but that's hardly enough to stop this from shipping. Send it out to sea!

🚢🚢🚢🚢🚢

app/components/ChangeLog.tsx
roseeichelmann: these are all great comments, thanks yall! just pushed these changes

app/components/ChangeLog.tsx
```diff
@@ -88,15 +90,16 @@ export default function ChangeLog({ logs }: { logs: ChangeLogEntry[] }) {
...
   return (
     <Accordion
-      defaultActiveKey={expandedItem}
+      activeKey={expandedItem}
       // on accordion click save new expanded item state to local storage
-      onSelect={(e) => {
-        const localStorageValue = e !== null ? "0" : JSON.stringify(null); // local storage value must be a string type
+      onSelect={(eventKey) => {
+        const localStorageValue =
+          eventKey !== null ? recordHistoryItemName : String(null); // local storage value must be a string type
```
johnclary, line 97: nbd, but you can do away with the ternary and use  `String(eventKey)` as the `localStorageValue` since it will always be null or the `recordHistoryItemName`.

app/components/ChangeLog.tsx
johnclary: agreed—and i created https://github.com/cityofaustin/atd-data-tech/issues/20606 to follow-up on this 🙏

mddilley (approved):
Thanks for making these adjustments! I retested, and I can still save my preferred layout. 🙌 🚀 🚢

PR #1657, opened 2025-01-13
People records: ability to edit victim name

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/20467

This PR adds the ability to edit a person's name. It is not a straightforward an edit as the other columns because we need to edit the first, middle and last name all at once. I took advantage of react hook form already being in the app, and made a custom component for the person columns.

I tried to set it up so if we need to create other custom edit components, we can follow this pattern. 

Let me know if you think I should name it something different, or if the docstrings could be improved, I welcome any feedback. 

Out of scope is any permissions based viewing or injury severity rendering. 

## Testing

**URL to test:**

https://deploy-preview-1657--vision-zero-nextjs.netlify.app/crashes/20339647

**Steps to test:**

Navigate to a crash page. Scroll down to the Peoples card. The rightmost column is the name column. 
Clicking the column brings it into edit state. Here you can add or edit a person's first, middle and last name. 
Try editing and hitting cancel, editing and saving changes. You can't hit save unless the fields have been edited.

Your edits should appear after hitting save, no need to refresh. 

---

#### Ship list

- [ ] Code reviewed


app/utils/formHelpers.ts
```diff
@@ -1,6 +1,5 @@
 import { ReactNode } from "react";
 import { ColDataCardDef, InputType } from "@/types/types";
-import { LookupTableOption } from "@/types/relationships";
```
chiaberry, line 3: this import isnt being used in the file, no message about it in the terminal either :(

app/README.md
```diff
@@ -71,9 +71,11 @@ npm run dev
     - Edit recommendation
   - Related records - units: click field to edit it
   - Related records - charges
-  - Change log: change log works normally with details modal. But it is not collapseable (yet).
+  - Related records - people
+      - name edit component
```
chiaberry, line 75: move this to the working features section and fixed some typos below

app/components/PersonNameField.tsx
```diff
@@ -0,0 +1,154 @@
...
+  /**
+   * This column's edit state
+   */
+  isEditingThisColumn: boolean;
+}
+
+type PersonNameFormInputs = {
+  first_name: string | null;
+  middle_name: string | null;
+  last_name: string | null;
```
johnclary, line 35: might as well prefix all these fieldnames with `prsn_` so that you don't have to modify them when you do the mutation?

i guess also because we're going with the snake casing here i am expecting these fields to match a db fieldname exactly 🪞

app/components/PersonNameField.tsx
```diff
@@ -0,0 +1,154 @@
+import { useCallback } from "react";
+import Button from "react-bootstrap/Button";
+import Form from "react-bootstrap/Form";
+import { useForm } from "react-hook-form";
+import { Person } from "@/types/person";
+import { useMutation } from "@/utils/graphql";
```
johnclary, line 6: nice `@`bsolute import — note to self to keep working on this habit 👍

app/components/RelatedRecordTable.tsx
```diff
@@ -27,7 +27,7 @@ export default function RelatedRecordTable<T extends Record<string, unknown>>({
     <Card>
       <Card.Header>{title}</Card.Header>
       <Card.Body>
-        <Table striped hover>
+        <Table striped hover responsive>
```
johnclary, line 30: This fixes the table overflowing its card on smaller displays 🙏 

Screenshot 2025-01-13 at 1 40 29 PM

app/components/PersonNameField.tsx
```diff
@@ -0,0 +1,154 @@
...
+          prsn_first_name: data.first_name,
+          prsn_mid_name: data.middle_name,
+          prsn_last_name: data.last_name,
+        },
+      });
+      await onSaveCallback();
+      // onCancel resets the current edit column to null
+      onCancel();
+    },
+    [mutate, onSaveCallback, onCancel]
```
johnclary, line 77: it looks like `record` is missing from your dep array 🫠

app/components/PersonNameField.tsx
```diff
@@ -0,0 +1,154 @@
...
+    reset,
+    handleSubmit,
+    formState: { isDirty },
+  } = useForm<PersonNameFormInputs>({
+    defaultValues: {
+      first_name: record.prsn_first_name,
+      middle_name: record.prsn_mid_name,
+      last_name: record.prsn_last_name,
+    },
+  });
```
johnclary, line 61: it's awesome that you reached for RFH—this looks so good!

app/types/types.ts
```diff
@@ -41,6 +41,17 @@ export interface ColDataCardDef<T extends Record<string, unknown>> {
...
+   * Function that returns a custom component, used for display and/or editing
+   * when the column is dependent on other columns of the same record
+   */
+  customComponent?: (
+    record: T,
+    onCancel: () => void,
+    mutation: string,
+    onSaveCallback: () => Promise<void>,
+    isEditingThisColumn: boolean,
+  ) => ReactNode;
```
johnclary, line 54: This is great ✨  I _think_ we want this property to be title cased like a react component, i.e. `CustomComponent` ?

app/components/RelatedRecordTableRow.tsx
```diff
@@ -92,8 +92,20 @@ export default function RelatedRecordTableRow<
...
+            {
+              col.customComponent &&
+                col.customComponent(
+                  record,
+                  onCancel,
+                  mutation,
+                  onSaveCallback,
+                  isEditingThisColumn
+                )
+            }
```
johnclary, line 104: i love how you solved this—the `customComponent` looks great. i want to propose a slight change to this approach, which is to:

1. Change this to be a custom _edit_ component, and move it down into the `DataCardInput` component
2. Use the existing `valueFormatter` option for rendering the column value when not editing

My thinking for this change is that it feels more consistent with the existing API / component structure we've built out for the edit components. Because the `DataCardInput` (which probably needs a better name) is used by both the `DataCard` and the `RelatedRecordTableRow` components, you will enable the custom component functionality for both the card- and table-based editing. This also narrows the scope of concerns for the custom component by focusing on editing only, which seems preferable.

johnclary (changes requested):
Chia—this is looking excellent and works perfectly! I'm curious what you think about the moving the custom edit component further down in the hierarchy 🙏

app/components/RelatedRecordTableRow.tsx
chiaberry: I like this idea i will work on it! moving the PR back to WIP

mddilley (approved):
Tested this locally and updated the names all ways that I could. Works great, and I agree that it so cool to see RHF making its way into these forms! 😎 🚀

app/components/PersonNameField.tsx
```diff
@@ -0,0 +1,144 @@
...
+  mutation: string;
+  /**
+   * The function to call on cancel button click, which is expected
+   * to set the parent component's edit state
+   */
+  onCancel: () => void;
+  /**
+   * Function that is an async wrapper around data refetch
+   */
+  onSaveCallback: () => Promise<void>;
```
roseeichelmann, line 25: greattt comments!! makes me think i should have done comments in my interface for the swap address component but it was only 3 self explanatory props, but maybe we should just make this a standard in our code to always docstring the interface props :shrug:

roseeichelmann (approved):
this looks and works so great! i love your solution, and love that you set up a pathway for future custom field edit components. once againnnn i hate to keep bringing it up bc i know @johnclary is in denial about cris data being all caps but should we auto caps this input too? like we currently do in VZ prod?

app/components/PersonNameField.tsx
```diff
@@ -0,0 +1,144 @@
...
+type PersonNameFormInputs = {
+  prsn_first_name: string | null;
+  prsn_mid_name: string | null;
+  prsn_last_name: string | null;
+};
+
+/**
+ * Component that manages the form UI for editing a Person's name, first, middle and last
+ */
+const PersonNameField = ({
```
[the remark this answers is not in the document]
johnclary: ok, yes, let's change this to a functino declaration instead of an arrow function for consistency. i'll make a snackoo to cleanup the the rest of the `/components` folder

johnclary:
wrt capitalization, i'll open a follow-up issue for it. i really don't like upper casing people's names. names are case sensitive!

app/components/PersonNameField.tsx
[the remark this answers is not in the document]
johnclary: * https://github.com/cityofaustin/atd-data-tech/issues/20704

PR #1658, opened 2025-01-14
Add role-based feature visibility

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/20302

Here's the first iteration of restricting access to VZE features based on user role. In the interest of avoiding code conflicts I have excluded the crash details page from this work.

## Testing

**URL to test:** Local

1. Start your local VZE, and also start the user API following the readme in `/api`. Once you have your env set up you should be able to run `docker compose up` and be on your way. Let me know if you need help.

2. You're going to test the VZE with three user accounts whose 1pass entries are named as follows:
* Vision Zero Editor (VZE) Test User: VZ Admin
* Vision Zero Editor (VZE) Test User: Editor
* Vision Zero Editor (VZE) Test User: Read-only

Please test all of the pages/features listed below and verify that the app matches this matrix 👇 

feature | admin | edtior | read-only
-- | -- | -- | --
Upload Non-CR3 page | visible | visible | hidden
Users page - user list | visible | visible | visible
Users page - add user button | visible | hidden | hidden
User details page | visible | visible | visible
User details page - edit and delete user buttons | visible | hidden | hidden
Crashes page - create crash record button | visible | visible | hidden

To summarize these rules: Read-only users can edit nothing. Editors can edit everything except users. Admins can edit everything.

That's it! nice!

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/api/server.py
+++ b/api/server.py
@@ -391,33 +391,6 @@ def download_crash_id(crash_id):
     return jsonify(message=url)
 
 
-def isValidUser(user_dict):
-    valid_fields = [
-        "email",
-        "name",
-        "https://hasura.io/jwt/claims",
-        "email_verified",
-        "aud",
-    ]
-
-    user_email = user_dict.get("email", None)
-
-    # Check for valid fields
-    for field in valid_fields:
-        if user_dict.get(field, False) == False:
-            return False
-
-    # Check for verified email
-    if user_dict["email_verified"] != True:
-        return False
-
-    # Check email for austintexas.gov
-    if not str(user_email).endswith("@austintexas.gov"):
-        return False
-
-    return True
-
-
 def hasUserRole(role, user_dict):
     claims = user_dict.get("https://hasura.io/jwt/claims", False)
     if claims != False:
@@ -438,11 +411,7 @@ def hasUserRole(role, user_dict):
 )
 @requires_auth
 def user_test():
-    user_dict = current_user._get_current_object()
-    if isValidUser(user_dict):
-        return jsonify(message=current_user._get_current_object())
-    else:
-        return notAuthorizedError()
+    return jsonify(message=current_user._get_current_object())
 
 
 @APP.route("/user/list_users")
@@ -458,20 +427,18 @@ def user_test():
 def user_list_users():
     user_dict = current_user._get_current_object()
     page = request.args.get("page")
-    per_page = request.args.get("per_page")
-    if isValidUser(user_dict) and hasUserRole(ADMIN_ROLE_NAME, user_dict):
-        endpoint = (
-            f"https://{AUTH0_DOMAIN}/api/v2/users?page="
-            + page
-            + "&per_page="
-            + per_page
-            + "&include_totals=true&sort=last_login:-1"
-        )
-        headers = {"Authorization": f"Bearer {get_api_token()}"}
-        response = requests.get(endpoint, headers=headers)
-        return jsonify(response.json()), response.status_code
-    else:
-        return notAuthorizedError()
+    per_page = request.args.get("per_page")    
+    endpoint = (
+        f"https://{AUTH0_DOMAIN}/api/v2/users?page="
+        + page
+        + "&per_page="
+        + per_page
+        + "&include_totals=true&sort=last_login:-1"
+    )
+    headers = {"Authorization": f"Bearer {get_api_token()}"}
+    response = requests.get(endpoint, headers=headers)
+    return jsonify(response.json()), response.status_code
+
 
 
 @APP.route("/user/get_user/<id>")
@@ -485,14 +452,10 @@ def user_list_users():
 )
 @requires_auth
 def user_get_user(id):
-    user_dict = current_user._get_current_object()
-    if isValidUser(user_dict) and hasUserRole(ADMIN_ROLE_NAME, user_dict):
-        endpoint = f"https://{AUTH0_DOMAIN}/api/v2/users/" + id
-        headers = {"Authorization": f"Bearer {get_api_token()}"}
-        response = requests.get(endpoint, headers=headers)
-        return jsonify(response.json()), response.status_code
-    else:
-        return notAuthorizedError()
+    endpoint = f"https://{AUTH0_DOMAIN}/api/v2/users/" + id
+    headers = {"Authorization": f"Bearer {get_api_token()}"}
+    response = requests.get(endpoint, headers=headers)
+    return jsonify(response.json()), response.status_code
 
 
 @APP.route("/user/create_user", methods=["POST"])
@@ -501,7 +464,7 @@ def user_get_user(id):
 @requires_auth
 def user_create_user():
     user_dict = current_user._get_current_object()
-    if isValidUser(user_dict) and hasUserRole(ADMIN_ROLE_NAME, user_dict):
+    if hasUserRole(ADMIN_ROLE_NAME, user_dict):
         json_data = request.json
         # set the user's password - user will have to reset it for access
         json_data["password"] = get_secure_password()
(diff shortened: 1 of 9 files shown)


api/server.py
```diff
@@ -391,33 +391,6 @@ def download_crash_id(crash_id):
     return jsonify(message=url)
 
 
-def isValidUser(user_dict):
```
johnclary, line 394: per our discussion [in slack](https://austininnovation.slack.com/archives/CM9MK950S/p1736890649856289) and this [issue comment](https://github.com/cityofaustin/atd-data-tech/issues/20391#issuecomment-2558161548). we're no longer validating these fields which existed on the Auth0 `idToken`.

api/server.py
```diff
@@ -458,20 +427,18 @@ def user_test():
 def user_list_users():
     user_dict = current_user._get_current_object()
     page = request.args.get("page")
-    per_page = request.args.get("per_page")
-    if isValidUser(user_dict) and hasUserRole(ADMIN_ROLE_NAME, user_dict):
```
johnclary, line 462: i removed the `hasUserRole` check for `/list_users` and `/get_user` — thereby enabling any authenticated users to access these methods. see [this slack thread](https://austininnovation.slack.com/archives/CM9MK950S/p1736956508498339).

app/components/PermissionsRequired.tsx
```diff
@@ -0,0 +1,24 @@
...
+ * Component wrapper that renders `null` if the user does not have one of the
+ * allwed roles. If the allowedRoles array is undefined then the permission
+ * check is skipped
+ */
+export default function PermissionsRequired({
+  children,
+  allowedRoles,
+}: PermissionsRequiredProps) {
+  const { user } = useAuth0();
+  if (!allowedRoles || (user && hasRole(allowedRoles, user))) {
```
johnclary, line 20: it might be worth memoizing this but i wasn't sure 🤷

api/server.py
roseeichelmann: im not super convinced read only users need to be able to see user details like IP addresses and such but :shrug: feels more like a feature for admins. but i dont feel strongly about this

app/app/users/[user_id]/page.tsx
```diff
@@ -11,11 +11,14 @@ import { FaUserEdit, FaUserAltSlash } from "react-icons/fa";
 import AlignedLabel from "@/components/AlignedLabel";
 import AppBreadCrumb from "@/components/AppBreadCrumb";
 import UserModal from "@/components/UserModal";
+import PermissionsRequired from "@/components/PermissionsRequired";
 import { useToken, formatRoleName } from "@/utils/auth";
 import { useUser } from "@/utils/users";
 import { User } from "@/types/users";
 import { formatDateTime } from "@/utils/formatters";
 
+const allowedRoles = ["vz-admin"];
```
Comment (roseeichelmann) [inline app/app/users/[user_id]/page.tsx:20]: maybe `editUserRoles`

app/components/PermissionsRequired.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { hasRole } from "@/utils/auth";
+
+interface PermissionsRequiredProps {
+  children: ReactNode;
+  allowedRoles?: string[];
+}
+
+/**
+ * Component wrapper that renders `null` if the user does not have one of the
+ * allwed roles. If the allowedRoles array is undefined then the permission
```
roseeichelmann, line 12: wittle typo "allwed"

app/components/SidebarLayout.tsx
```diff
@@ -95,41 +91,20 @@ export default function SidebarLayout({ children }: { children: ReactNode }) {
...
-                href="/upload-non-cr3"
-              />
-              <SideBarListItem
-                isCollapsed={isCollapsed}
-                isCurrentPage={segments.includes("users")}
-                Icon={FaUserGroup}
-                label="Users"
-                href="/users"
-              />
+              {routes.map((route) => (
```
roseeichelmann, line 94: super snazzy use of mapping thu the routes i like it

app/configs/routes.ts
```diff
@@ -0,0 +1,44 @@
+import {
+  FaShieldHeart,
+  FaGaugeHigh,
+  FaLocationDot,
+  FaUserGroup,
+  FaCloudArrowUp,
+} from "react-icons/fa6";
+import { IconType } from "react-icons";
+
+type Route = {
```
roseeichelmann, line 10: still trying to figure out when to use types vs interfaces. from these [typescript docs](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces):

>  "For the most part, you can choose based on personal preference, and TypeScript will tell you if it needs something to be the other kind of declaration. If you would like a heuristic, use interface until you need to use features from type."

app/utils/auth.ts
```diff
@@ -38,6 +38,15 @@ export const getHasuraRoleName = (roles?: string[]): string => {
   }
 };
 
+/**
+ * Check if a user has any of the provided role names
+ * @param {string[]} roles - an array of roles to check for
+ * @param {CustomUser} user - the user object
```
roseeichelmann, line 44: i read that for typescript u can leave out type declarations in the docstrings bc it can be repetitive since they are also listed in the param declarations below but nbd

chiaberry:
I am running through the test steps and when I log out of vze locally, it redirects me to localhost:3000 and 404 not found because it should be localhost:3002. its a setting in auth0 dashboard right, the sign out redirect?

app/app/users/[user_id]/page.tsx
Comment (chiaberry) [reply app/app/users/[user_id]/page.tsx:20]: I agree with Rose!

app/app/crashes/page.tsx
```diff
@@ -8,9 +8,12 @@ import CreateCrashRecordModal from "@/components/CreateCrashRecordModal";
 import { FaCirclePlus } from "react-icons/fa6";
 import { crashesListViewColumns } from "@/configs/crashesListViewColumns";
 import { crashesListViewQueryConfig } from "@/configs/crashesListViewTable";
+import PermissionsRequired from "@/components/PermissionsRequired";
 import TableWrapper from "@/components/TableWrapper";
 const localStorageKey = "crashesListViewQueryConfig";
 
+const allowedRoles = ["vz-admin", "editor"];
```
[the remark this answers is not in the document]
chiaberry: agree here too!

chiaberry (approved):
I agree with Rose's comments about the variable names. 

test steps check out 🚢

johnclary:
wrt the logout issues, i tested this some more and see now that we need to pass the logout url to the `logout()`function. i will patch it up in this PR!

app/configs/routes.ts
johnclary: yes, i think that's a good enough rule. i changed it to an interface🙏

app/utils/auth.ts
johnclary: removed 👍

app/components/AppNavBar.tsx
```diff
@@ -52,7 +52,12 @@ export default function AppNavBar({ user, logout }: NavBarProps) {
                 Account
               </AlignedLabel>
             </Dropdown.Item>
-            <Dropdown.Item eventKey="2" onClick={() => logout()}>
+            <Dropdown.Item
+              eventKey="2"
+              onClick={() =>
+                logout({ logoutParams: { returnTo: window.location.origin } })
```
johnclary, line 58: here's the fix for the busted logout url 👍

mddilley (approved):
I tested this with all of the user roles, and I see the appropriate UI pieces hidden. Code looks great too. 🙏 🚢 🚀

api/server.py
johnclary: hmm—yes I wonder if we should altogether scrap the IPs from the  app. if we need to do that kind of forensics on an ad-hoc basis we can grab all the info we need and more from the Auth0 dashboard

johnclary: I created https://github.com/cityofaustin/atd-data-tech/issues/20751 to remove IP addresses from the API and VZE

johnclary:
Thank you for your reviews!

PR #1659, opened 2025-01-15
Crash notes

mateoclarke:
## Associated issues

Closes https://github.com/cityofaustin/atd-data-tech/issues/19962

**Notes table w/ a column in edit mode**
Screenshot 2025-01-30 at 1 05 34 AM

**Add new note modal (w/ field validation warning)**
Screenshot 2025-01-30 at 1 05 59 AM

## Testing

**URL to test:** 
Local bc of migrations

**Steps to test:**

**First test the migrations from [John's PR](https://github.com/cityofaustin/vision-zero/pull/1680)**

1. Start your db and  apply migrations and metadata
2. Query the notes tables and make sure they look right, with the `created_at` and `updated_at` reflecting dates in the past rather than a few seconds ago.

```sql
select * from crash_notes;

select * from location_notes;
```

4. Test the down migration

```
hasura migrate apply --down 1
```

**Now test the FE:**

1. Run migrations (if you haven't already)
1. Go to a crash page on in your local environment, ideally with exisiting notes
2. Confirm the date is format as expected (yyyy-mm-dd) and each note has a user.
3. Try to edit a note you didn't author. This should not be possible. You should only be able to edit the note if the user matches the currentUserEmail.
4. Add a new note. (Here is where you could give more opinions if the modal is ok for this PR and/or should we add inline edit functionality)
5. Edit the note you just created.
~6. Delete a note you created. (Again, you should only be able to do this for your authored notes)~  (After some discussion, deleting a note is out of scope but I'll probably add that as a quick follow on.)

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/app/components/CrashNotesCard.tsx
+++ b/app/components/CrashNotesCard.tsx
@@ -7,7 +7,6 @@ import { UPDATE_CRASH_NOTE } from "@/queries/notes";
 import RelatedRecordTable from "./RelatedRecordTable";
 import { ColDataCardDef } from "@/types/types";
 import { formatDate } from "@/utils/formatters";
-import { useAuth0 } from "@auth0/auth0-react";
 import AlignedLabel from "./AlignedLabel";
 
 interface CrashNotesCardProps {
@@ -54,7 +53,6 @@ const CrashNotesCard = ({
   crashPk,
   onSaveCallback,
 }: CrashNotesCardProps) => {
-  const { user } = useAuth0();
   const [showModal, setShowModal] = useState(false);
   const [isValidating] = useState(false);
   const handleClose = () => setShowModal(false);
@@ -70,16 +68,6 @@ const CrashNotesCard = ({
         records={notes}
         columns={notesColumns}
         mutation={UPDATE_CRASH_NOTE}
-        mutationVariables={(variables: {
-          id: number;
-          updates: Record<string, unknown>;
-        }) => ({
-          id: variables.id,
-          updates: {
-            text: variables.updates.text,
-            updated_by: user?.email,
-          },
-        })}
         isValidating={isValidating}
         title="Notes"
         onSaveCallback={onSaveCallback}
--- a/app/components/RelatedRecordTable.tsx
+++ b/app/components/RelatedRecordTable.tsx
@@ -11,10 +11,6 @@ interface RelatedRecordTableProps<T extends Record<string, unknown>> {
   title: string;
   headerActionButton?: React.ReactNode;
   onSaveCallback: () => Promise<void>;
-  mutationVariables?: (variables: {
-    id: number;
-    updates: Record<string, unknown>;
-  }) => { id: number; updates: Record<string, unknown> };
 }
 
 /**
@@ -28,7 +24,6 @@ export default function RelatedRecordTable<T extends Record<string, unknown>>({
   title,
   headerActionButton,
   onSaveCallback,
-  mutationVariables,
 }: RelatedRecordTableProps<T>) {
   return (
     <Card>
@@ -66,7 +61,6 @@ export default function RelatedRecordTable<T extends Record<string, unknown>>({
                   onSaveCallback={onSaveCallback}
                   record={record}
                   mutation={mutation}
-                  mutationVariables={mutationVariables}
                 />
               ))
             )}
--- a/app/components/RelatedRecordTableRow.tsx
+++ b/app/components/RelatedRecordTableRow.tsx
@@ -17,10 +17,6 @@ interface RelatedRecordTableRowProps<T extends Record<string, unknown>> {
   mutation: string;
   isValidating: boolean;
   onSaveCallback: () => Promise<void>;
-  mutationVariables?: (variables: {
-    id: number;
-    updates: Record<string, unknown>;
-  }) => { id: number; updates: Record<string, unknown> };
 }
 
 /**


app/.eslintrc.json
```diff
@@ -1,3 +1,3 @@
 {
-  "extends": ["next/core-web-vitals", "next/typescript"]
+  "extends": ["next/core-web-vitals", "next/typescript", "prettier"]
```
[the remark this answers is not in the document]
chiaberry: 💅🏼

app/components/RelatedRecordTableRow.tsx
```diff
@@ -61,68 +81,141 @@ export default function RelatedRecordTableRow<
...
         [saveColumnName]: value,
       },
-    });
+    };
+
+    if (context?.type === "note") {
+      await mutate(
+        mutationVariables ? mutationVariables(variables) : variables,
+        { skip_updated_by_setter: true }
+      );
```
johnclary, line 96: if i'm following correctly, it looks like the only reason we need to add this`context` and `mutationVariables` override is to skip the updated by setter. maybe we should just pass that setting specifically as a prop? and i am happy to branch off right quick and add an `updated_by` field to the notes table so that it can live in harmony with our other related tables without this workaround.

app/components/RelatedRecordTable.tsx
```diff
@@ -53,12 +63,18 @@ export default function RelatedRecordTable<T extends Record<string, unknown>>({
                   onSaveCallback={onSaveCallback}
                   record={record}
                   mutation={mutation}
+                  deleteMutation={deleteMutation}
+                  mutationVariables={mutationVariables}
+                  currentUserEmail={currentUserEmail}
+                  quickEditColumn={quickEditColumn}
```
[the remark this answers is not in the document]
johnclary: It's cool to see how you solved for this. My preference would be that we stick with the existing functionality of using the click-to-edit functionality only, with no pencil button. i am a strong disliker of the pencil buttons!

but i hear you that we need to provide better visual feedback for editable fields. i will open a follow-up issue for that. i think a subtle border around editable fields so they look kinda-like form inputs might do the trick 🤷

app/components/NotesCard.tsx
```diff
@@ -0,0 +1,102 @@
...
+const AddNoteButton = (handleShow: () => void) => {
+  return (
+    <Button size="sm" variant="primary" onClick={handleShow}>
+      <FaCirclePlus className="me-2" />
+      Add Note
+    </Button>
+  );
+};
+
+const NotesCard = ({ notes, crashPk, onSaveCallback }: NotesCardProps) => {
```
johnclary, line 57: it might be best to name this `CrashNotesCard` to be consistent with how we've named other crash-related components.. ideally we could have this card do double-duty for the location notes feature, but i think we can make that decision when we get there. at first glance i don't think it would be too much refactoring to extend this work with the location notes page.

app/components/NotesModal.tsx
```diff
@@ -0,0 +1,77 @@
+import Modal from "react-bootstrap/Modal";
+import Button from "react-bootstrap/Button";
+import Form from "react-bootstrap/Form";
+import { useForm, SubmitHandler } from "react-hook-form";
+import { CrashNote } from "@/types/crashNote";
+import { INSERT_CRASH_NOTE } from "@/queries/notes";
+import { useMutation } from "@/utils/graphql";
+import { useAuth0 } from "@auth0/auth0-react";
+
+interface NotesModalProps {
```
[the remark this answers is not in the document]
johnclary: i'm cool with keeping this as a modal for now. it works fine and we can visit inline note adding in a separate scope of work 🙏

app/components/NotesModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+    }
+    reset();
+    onClose();
+  };
+
+  return (
+    <Modal show={show} onHide={onClose}>
+      <Form onSubmit={handleSubmit(onSubmit)} id="noteForm">
+        <Modal.Header closeButton>
+          <Modal.Title>Add Note</Modal.Title>
```
johnclary, line 56: Would you change this title to sentence casing: `Add note` - ?

app/components/RelatedRecordTableRow.tsx
```diff
@@ -61,68 +81,141 @@ export default function RelatedRecordTableRow<
...
-                  />
+    <>
+      <tr>
+        {columns.map((col) => {
+          const isEditingThisColumn = col.path === editColumn?.path;
+          const isEditable =
+            col.editable &&
+            (!col.editableCheck || col.editableCheck(record, currentUserEmail));
+
+          if (col.path === "actions") {
```
johnclary, line 123: seeing this col property hacking and also note-specific content here is illustrating to me that i think it'd be good to approach the delete functionality by abstracting it a bit—maybe by passing in a generic `rowAction` component into the related record table.

in Rose's [address swap PR](https://github.com/cityofaustin/vision-zero/pull/1653), the HeaderAction component takes a set of props that open up a fair amount of flexibility for how we might want to extend that functionality in the future, and I think it'd be good to take a similar approach here. for example, we might want to have a row action button that soft deletes instead of hard deletes (i'm thinking of EMS record linking features we have in the backlog).

so i would lean toward pushing the delete functionality out into a separate issue so we could scope that further. curious what others think.

app/components/DataCardInput.tsx
```diff
@@ -68,6 +68,15 @@ const DataCardInput = ({
             inputMode={inputType === "number" ? "numeric" : undefined}
           />
         )}
+        {inputType === "textarea" && (
```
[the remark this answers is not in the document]
johnclary: nice!

johnclary (changes requested):
@mateoclarke this is looking really great. and thank you for dealing with the gaps in scoping + design for this feature.

most of my feedback is about pulling the editable check and delete functionality out into future work. i think this is really close to shippable 🚀

app/components/NotesCard.tsx
```diff
@@ -0,0 +1,95 @@
...
+        mutationVariables={(variables: { id: number; updates: Record<string, unknown> }) => ({
+          id: variables.id,
+          updates: {
+            text: variables.updates.text,
+          },
+        })}
+        isValidating={isValidating}
+        title="Notes"
+        onSaveCallback={onSaveCallback}
+        footer={AddNoteButton(handleShow)}
```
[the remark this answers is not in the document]
chiaberry: I like the option of a footer section, but I am not sure if I like the add note button to be at the bottom of the card -- new notes appear at the top of the card, and if there are a lot of long notes (https://deploy-preview-1659--vision-zero-nextjs.netlify.app/crashes/20339647) you have to scroll down to see the add note button. 
the swap addresses button is in the header but the edit crash recommendations button is in the footer, and if there are only a few short notes, it works fine either way

app/components/RelatedRecordTableRow.tsx
```diff
@@ -15,30 +18,43 @@ interface RelatedRecordTableRowProps<T extends Record<string, unknown>> {
...
+  T extends Record<string, unknown>,
 >({
   record,
   columns,
   mutation,
+  deleteMutation,
   isValidating,
   onSaveCallback,
+  mutationVariables,
+  currentUserEmail,
```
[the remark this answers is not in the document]
chiaberry: if notes could be edited by anyone, the user column should be renamed to something like "Created by", and then actually the updated_by should come back into play, so we can track who edited someone else's note.

app/types/types.ts
```diff
@@ -41,6 +46,7 @@ export interface ColDataCardDef<T extends Record<string, unknown>> {
     column: ColDataCardDef<T>
   ) => string;
   valueRenderer?: (record: T, column: ColDataCardDef<T>) => ReactNode;
+  style?: React.CSSProperties;
```
johnclary, line 49: forgot to mention: this is slick! i am borrowing it in another PR 😇

app/components/NotesCard.tsx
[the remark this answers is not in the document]
johnclary: great call—i think we should be consistent across the app and put this button in the header using the component prop Rose set up 🙏

app/components/RelatedRecordTableRow.tsx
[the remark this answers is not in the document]
johnclary: @mateoclarke i have PR'd the `updated_by` and `created_by` fields for you [here](https://github.com/cityofaustin/vision-zero/pull/1680).

for the column rendering, let's use the following labels + fields

- **Updated at** (`updated_at`)
- **Updated by** (`updated_by`)
- **Note** (`text`)

app/components/NotesCard.tsx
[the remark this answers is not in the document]
johnclary: nope—i believe that capitalization was my bad. would be happy to see it gone 🙏

app/queries/crash.ts
```diff
@@ -224,14 +224,12 @@ export const GET_CRASH = gql`
       }
       crash_notes(
         where: { is_deleted: { _eq: false } }
-        order_by: { date: desc }
+        order_by: { updated_at: asc }
```
johnclary, line 227: would you sort these by `created_at` instead?

app/components/CrashNotesCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+    await onSaveCallback();
+  };
+
+  return (
+    <>
+      <RelatedRecordTable
+        records={notes}
+        columns={notesColumns}
+        mutation={UPDATE_CRASH_NOTE}
+        mutationVariables={(variables: {
```
johnclary, line 73: now that this table has an `updated_by` column, i think you can scrap this prop and let the built-in functionality do it's thing. the `RelatedRecordTableRow` will create that variable payload exactly as you have it here 🚀

app/components/DataCardInput.tsx
```diff
@@ -49,11 +49,13 @@ const DataCardInput = ({
   const [editValue, setEditValue] = useState<string>(initialValue);
 
   const isDirty = editValue !== initialValue;
+  const isValid = editValue.trim() !== "";
```
johnclary, line 52: hmm, i'm i see how we have unfinished work around input validation here :/ 

i like this, but it is going to break other fields that are nullable/clearable, like the address fields. i think we should break this off into a separate issue to add support for required fields? i know that means this will ship with an unhandled failure state :/

app/components/RelatedRecordTable.tsx
```diff
@@ -9,10 +9,12 @@ interface RelatedRecordTableProps<T extends Record<string, unknown>> {
   mutation: string;
   isValidating: boolean;
   title: string;
-  footer?: React.ReactNode;
+  headerActionButton?: React.ReactNode;
```
johnclary, line 12: nice!

johnclary (commented):
Nice!! This is looking so great—I ran into just a few minor things

app/components/RelatedRecordTable.tsx
```diff
@@ -21,11 +26,19 @@ export default function RelatedRecordTable<T extends Record<string, unknown>>({
...
+  mutationVariables,
 }: RelatedRecordTableProps<T>) {
   return (
     <Card>
-      <Card.Header>{title}</Card.Header>
+      <Card.Header>
+        <div className="d-flex justify-content-between">
+
+        {title}
+        {headerActionButton && headerActionButton}
```
chiaberry, line 39: this is checking if it exists and if so, render it? Trying to picture what would render if it was just `{headerActionButton}` or maybe it should be `{!!headerActionButton && headerActionButton}`?

chiaberry (commented):
I tested the new code and i can add and edit notes. I'll come back and hit approve on it after the input validation gets removed ( 😢 ) and the other changes requested are folded in

app/queries/crash.ts
johnclary: ah yes, that's terrible. let's show `created_at` in the header instead. that way the notes are sorted correctly, and when you edit one it's sort position doesn't jump. sorry for the back and forth!

app/components/DataCardInput.tsx
johnclary: 👍 
* https://github.com/cityofaustin/atd-data-tech/issues/20878

app/components/CrashNotesCard.tsx
johnclary: @mateoclarke — one last change needed right here ☝️

johnclary (approved):
🚢 🚢 🚢 🚢 🚢 🚢 🚢

PR #1660, opened 2025-01-15
Users page: copy user email addresses

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/20121

## Testing

**URL to test:** 
Local - you will need to run the users api and also update `isValidUser` in `api/server.py` to return true until john gets rid of that function

**Steps to test:**
1. Go to the users page. While the users are loading you should see a spinner and not see the buttons. Once the users load, you see the buttons.
2. Click on the copy emails button, check that they copied to your clipboard. Make sure test accounts arent being included.
3. Navigate to crashes and back to users. While `isValidating` is true there will be a spinner by the buttons and they should be disabled.

---

#### Ship list

- [ ] Code reviewed


app/app/users/page.tsx
```diff
@@ -48,21 +68,29 @@ export default function Users() {
                 <Button
                   className="me-2"
                   onClick={() => setShowNewUserModal(true)}
+                  disabled={isValidating}
```
roseeichelmann, line 71: disable the buttons while we are showing the `isValidating` spinner (different from the `isLoading` spinner which is for loading all the users in the table)

johnclary:
by the way, [here's the PR](https://github.com/cityofaustin/vision-zero/pull/1658) that fixes the user API bug.  you add `return True` right [here](https://github.com/cityofaustin/vision-zero/blob/2347efc65a4e52c2506981027ef24057bf675bb6/api/server.py#L395) to workaround this in the meantime.

chiaberry (approved):
copied everyone, confirmed no test users slipped in the list

mddilley (approved):
I tested this out, and I see the user emails copied to my clipboard. Nice detail to disable the interaction while the user list is revalidating in the background. 🚀

PR #1661, opened 2025-01-15
Fixes the VZV By Year & Month chart

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/20643

Thank you @chiaberry for flagging this and running down the problem. I went ahead and pushed up the fix i came up with based on our Slack convo 🙏 

## Testing

**URL to test:** Netlify

Test out that By Year & Month chart by toggling back and forth between the **Fatalities**, **Serious Injuries**, and **All** filters as well as the **Monthly** and **Cumulative** views.

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/viewer/src/views/summary/CrashesByYearAverage.js
+++ b/viewer/src/views/summary/CrashesByYearAverage.js
@@ -36,12 +36,8 @@ const CrashesByYearAverage = ({ avgData, currentYearData }) => {
       };
     };
 
-    const isDataFetched = !!avgData.length && !!currentYearData.length;
-
-    if (isDataFetched) {
-      const formattedData = formatChartData(avgData, currentYearData);
-      setChartData(formattedData);
-    }
+    const formattedData = formatChartData(avgData, currentYearData);
+    setChartData(formattedData);
   }, [avgData, currentYearData]);
 
   return (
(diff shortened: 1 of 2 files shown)


viewer/src/views/summary/CrashesByYearAverage.js
```diff
@@ -36,12 +36,8 @@ const CrashesByYearAverage = ({ avgData, currentYearData }) => {
       };
     };
 
-    const isDataFetched = !!avgData.length && !!currentYearData.length;
```
johnclary, line 39: in addition to testing it, i read through this code a few times and i don't see a reason to be concerned with `isDataFetched`. there could be some sneaky type error lurking here but it seems fine to me 🤷

mddilley (approved):
Thanks for fixing this bug. I see it in prod but not after this change. 🙏🙏 Definitely a fun trip down memory lane in this code 😅

I tried my best to break this locally, but I can't and agree that it feels safe to remove. 🚢 🚀

chiaberry (approved):
it was eagle eyed Charlie who noticed the bug! 

🚢

johnclary:
thank you! and thank you @Charlie-Henry 🔍 !

PR #1663, opened 2025-01-16
Create the dashboard page

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/19987

[Here](https://www.figma.com/proto/2acDnn8lRJ3FHnPEDGlk3A/Applied-Labs-Figma?type=design&node-id=811-60895&t=oEA7X290kfUxsKrT-0&scaling=scale-down-width&page-id=0%3A1&starting-point-node-id=811%3A60895&hotspot-hints=0&hide-ui=1) are the ACC design suggestion for the dashboard, which I took inspiration from but obviously did not fully implement.

Screenshot 2025-01-16 at 3 09 31 PM

## Testing

**URL to test:** Local (the deploy preview is pointing to prod VZV, which has the x overflow problem)

**Steps to test:**

1. In the nextjs app root (`/app`), add a new line to your `.env.local` file: 

```
NEXT_PUBLIC_VZV_ENDPOINT=http://localhost:3000/viewer
```

2. In `/viewer/package.json`, update the `start` script [here](https://github.com/cityofaustin/vision-zero/blob/2347efc65a4e52c2506981027ef24057bf675bb6/viewer/package.json#L57) by changing the `REACT_APP_VZV_ENVIRONMENT` from `STAGING` to `PRODUCTION`

3. Start your local nextjs VZE and your local VZV. You'll need to use `nvm` because these apps require different node versions

4. In the nextjs app, navigate to the `/dashboard` page and verify that the embedded VZV measures load correctly. Resize to a.medium-ish and small-ish window size and verify that no horizontal scrollbar appears. The measures will collapse vertically on a smaller size and require vertical scrolling. That's ok 👍

5. Test all of the link cards and verify that their targets load correctly for you

---

#### Ship list

- [ ] Code reviewed


viewer/public/index.html
```diff
@@ -52,7 +52,7 @@
     <!--VZV_HEADER_SNIPPETS-->
     <!-- DO NOT ALTER "ABOVE" THIS LINE FOR AUTOMATED ENVIRONMENT SNIPPETS -->
   </head>
-  <body style="background: #fff;">
+  <body style="background: #fff; overflow-x: hidden">
```
johnclary, line 55: The summary measures are overflowing their container and causing a horizontal scrollbar on certain display sizes. You can replicate this by removing this style property and setting your display size to 1366 x 1024. This is the rotated **iPad Pro** size on your chrome dev tools.

i spent quite a bit of time trying to fix the source of overflow in the DOM, but i ultimately gave up and decided to brute force it like this.

if we are uncomfortable apply this change globally, i initially used the `postMessage` API so that this change would only take place when the VZE loads the /measures view in an iFrame. Here is that code:
https://github.com/cityofaustin/vision-zero/pull/1663/commits/c9b8b2759456ea7fcbb15a8ca50d7f48f705f173

app/utils/formatters.ts
```diff
@@ -36,3 +36,20 @@ export const formatDate = (value: unknown): string => {
 export const formatFileTimestamp = (date: Date): string => {
   return format(date, "yyyy-MM-dd h.mm.ss a");
 };
+
+/**
+ * Return morning, afternoon, or evening
+ */
+export const formatGreetingTime = (
```
johnclary, line 43: This whole greeting thing is maybe a bit too cutesy for the subject matter 🤷

app/app/dashboard/page.tsx
```diff
@@ -1,4 +1,101 @@
...
-  }
-  
\ No newline at end of file
+  const { user } = useAuth0();
+  const firstName = formatFirstNameFromEmail(user?.email || "");
+
+  return (
+    <>
+      <Row className="mb-3">
+        <Col md={6}>
```
chiaberry, line 62: md={6} feels a little scrunched, i tried changing it to 9 and it feels less smushed over to the side.

app/app/dashboard/page.tsx
```diff
@@ -1,4 +1,101 @@
...
+            <span>{`Good ${formatGreetingTime(new Date())}, `}</span>
+            <span className="text-capitalize">{firstName}</span>!
+          </h6>
+          <p>
+            Welcome to the Vision Zero Editor, which provides access to the City
+            of Austin&apos;s traffic crash data. The below metrics reflect our&nbsp;
+            <a href={VZV_ENDPOINT} target="_blank" rel="noreferrer">
+              public crash data
+            </a>
+            , which excludes crashes which occurred within the last 14 days.
```
chiaberry, line 73: the greeting part of this feels fine, but the explainer of what the VZE is feels misplaced for someone who works in it every day. Also "which" appears here 3 times, which (ha) feels like maybe 2 times too many?

roseeichelmann: I agree about the which's, and also feel like the greeting doesnt need to be this long. ppl using the app probably already know what its for, i think you could shorten it to just explain where the metrics come from: "Welcome to the Vision Zero Editor. The below metrics reflect our [public crash data](http://localhost:3000/viewer), which excludes crashes that occurred within the last 14 days."

app/app/dashboard/page.tsx
```diff
@@ -1,4 +1,101 @@
...
+import { formatGreetingTime } from "@/utils/formatters";
+import { FaMap, FaChartPie } from "react-icons/fa6";
+import { formatFirstNameFromEmail } from "@/utils/auth";
+import DashboardLinkCard, {
+  DashboardLinkCardProps,
+} from "@/components/DashboardLinkCard";
+
+const VZV_ENDPOINT = process.env.NEXT_PUBLIC_VZV_ENDPOINT || "";
+
+const dashboardLinks: DashboardLinkCardProps[] = [
```
roseeichelmann, line 14: should we put all the power BI stuff on one side and the rest on the other like we currently do? its jumbled for me

app/utils/auth.ts
```diff
@@ -53,6 +55,17 @@ export const formatRoleName = (role: string): string => {
...
 
+/**
+ * Return the text from an email up to the first `.`
+ *
+ * @example
+ * // returns "John"
+ * formatFirstNameFromEmail("john.c@austintexas.gov")
+ */
+export const formatFirstNameFromEmail = (email: string): string => {
+  return email?.split(".")[0] || "";
```
roseeichelmann, line 66: :sparkles:

app/utils/formatters.ts
roseeichelmann: i kinda agree, i tested replacing it with "Welcome to the Vision Zero Editor" and then the text below just explains the metrics. but idk maybe they would like the personalization but to me its kinda weird

app/app/dashboard/page.tsx
johnclary: thanks for that copy—looks good to me 🙏

app/app/dashboard/page.tsx
johnclary: yes—fixed 🙏

app/app/dashboard/page.tsx
johnclary: i did this—thanks

johnclary:
interesting, the widgets were not wrapping on my laptop screen but decreasing my window by just a few pixels triggered the wrapping. I reduced their scale down to 75% which seems to be a lot better.

i also incorporated Chia and Rose's other feedback, although I left the cutesy greeting for now 🙏

app/app/dashboard/page.tsx
```diff
@@ -1,4 +1,102 @@
...
+            <span>{`Good ${formatGreetingTime(new Date())}, `}</span>
+            <span className="text-capitalize">{firstName}</span>!
+          </h6>
+          <p>
+            Welcome to the Vision Zero Editor. The below metrics reflect
+            our&nbsp;
+            <a href={VZV_ENDPOINT} target="_blank" rel="noreferrer">
+              public crash data
+            </a>
+            , which exclude crashes which occurred within the last 14 days.
```
chiaberry, line 73: I like Rose's version better 
>"Welcome to the Vision Zero Editor. The below metrics reflect our [public crash data](http://localhost:3000/viewer), which excludes crashes that occurred within the last 14 days."
With the copy you have I thought excludes would subject/verb agree better than exclude, but style guides say both work. its Friday and im tired,

chiaberry: I am apparently on a which hunt (ba dum ch ...sorry 😞 )

app/app/dashboard/page.tsx
```diff
@@ -1,4 +1,102 @@
...
+            src={`${VZV_ENDPOINT}/measures`}
+            title="Vision Zero Viewer Stats"
+            style={{
+              width: "100%",
+              height: "275px",
+              transformOrigin: "top left", // Anchor scaling to the top-left corner
+              border: "none",
+              overflowX: "hidden",
+              scale: ".75",
+            }}
```
frankhereford, line 89: This is interesting -- the element is computed at full width and then scaled down to 3/4 of its size. I like the smaller widgets, but I wish the shrunken element could still be full width. 

I was going to ask: Can you `width: 130%;` which is just a little under "the whole width?", but that makes for trouble in the vertical direction. I think what you have here maybe is the best compromise, but for me, any view port width < 1277px might be better of not rendering these, as the iframe starts to scroll. This is real tricky.

I see your comment below about how this was a rabbit hole and all that -- nothing new coming out of this review comment from me!

----

I dream of the day when the VZV is a page in the VZE so we can share these components easily between the apps and it becomes a `haproxy` or webserver config operation that serves the "VZV" out of the VZ web app on the VZV url.

app/utils/auth.ts
frankhereford: This is pretty cool, but I fear it's brittle. If this returns the empty string, this whole thing gets wonky on the page with a dangling `,!`. I think your comment below about the greeting thing being cutesy is the real answer, and I'd strike the whole salutation entirely. Just my two cents there.

app/utils/formatters.ts
frankhereford: I stated it above, but I agree pretty strongly. I think John's inclination and Rose's suggestion are the ticket.

app/app/dashboard/page.tsx
```diff
@@ -1,4 +1,102 @@
...
+  const { user } = useAuth0();
+  const firstName = formatFirstNameFromEmail(user?.email || "");
+
+  return (
+    <>
+      <Row className="mb-3">
+        <Col md={9}>
+          <h6 className="display-6">
+            <span>{`Good ${formatGreetingTime(new Date())}, `}</span>
+            <span className="text-capitalize">{firstName}</span>!
```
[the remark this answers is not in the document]
frankhereford: Another concern here is that if the `firstName` isn't correctly parsed (such as an external user of the VZE, which I believe is possible, right? did we make it `austintexas.gov` only?), this will have a dangling `,` followed by the `!`. I touch back on this later.

frankhereford (approved):
Bunch of tricky formatting in here, and I think you found a totally workable middle ground. I'm 👍 on this shipping, so ✅, but I think some more consideration about the content and tone are welcome too.

Thanks John!

app/app/dashboard/page.tsx
johnclary: yeah, this one is really pretty frustrating! tested your suggested. TIL you can go over 100% width! that does solve the problem in a new way but still leaves us with this wrapping problem :/

i also tried adding [responsive display classes](https://getbootstrap.com/docs/5.3/utilities/display/#how-it-works) to hide the metrics entirely on `md` and below, but i think we're betting off letting users deal with the scrolling container in case they really want to view these metrics on their iPad 🙃 

i think the right solution probably involves the `postMessage` api so that we can dynamically resize the container. or, like you said, bringing the VZV into the VZE codebase 😍

chiaberry (approved):
my which hunt worked
✔️

PR #1664, opened 2025-01-17
Ability to search by crash or case ID from the navigation bar

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/19997

## Testing

**URL to test:** 
Netlify or local

**Steps to test:**
1. Go to the crashes page, copy a crash id. Test searching for that crash id in the new navbar search, you should be rerouted to that page.
2. From there, test typing in random stuff, see the error message. When you start typing again the error message should go away.
3. Test going back and forth between searching for crash ids and case ids

---

#### Ship list

- [ ] Code reviewed


app/components/AppNavBar.tsx
```diff
@@ -1,18 +1,20 @@
+import { User } from "@auth0/auth0-react";
 import Container from "react-bootstrap/Container";
-import Navbar from "react-bootstrap/Navbar";
-import Image from "react-bootstrap/Image";
 import Dropdown from "react-bootstrap/Dropdown";
-import { User } from "@auth0/auth0-react";
+import Image from "react-bootstrap/Image";
+import Col from "react-bootstrap/Col";
+import Navbar from "react-bootstrap/Navbar";
```
roseeichelmann, line 6: these just got reorganized. til you can organize your imports from your code editor

app/components/AppNavBar.tsx
```diff
@@ -26,74 +28,79 @@ export default function AppNavBar({ user, logout }: NavBarProps) {
...
-                CR3 code sheet
-                <FaSquareArrowUpRight className="ms-3 text-muted" />
-              </AlignedLabel>
-            </Dropdown.Item>
-          </Dropdown.Menu>
-        </Dropdown>
+          </Navbar.Brand>
+        </Col>
+        <Col className="d-flex justify-content-end">
+          <NavBarSearch />
```
roseeichelmann, line 42: this little thing right here is the new navbarsearch

app/utils/queryBuilder.ts
roseeichelmann, line 1: all prettier linting

johnclary (changes requested):
@roseeichelmann this works perfectly and the code looks great 🙌 

Just a few minor details I am requesting:

- Change the button and input sizes to small (`sm`)
- Change button variants to `secondary` (or `light` — curious what you think)
- Adjust the vertical alignment so that the search component is vertically centered with the the VZ logo and the avatar

Otherwise this is good to go 🚀

johnclary:
@roseeichelmann one small thing i forgot to mention in my review—would you force the search button to be disabled when the input is empty?

i also wanted to mention that i'd like to switch this over to a dropdown menu for switching between crash ID and case ID, but that feels out of scope given that what you have mirrors VZE prod. i can open a follow-up for that

Screenshot 2025-01-27 at 10 39 25 AM

roseeichelmann:
@johnclary addressed your feedback, i dont think we should do a light variant bc its so hard to see 
Screenshot from 2025-01-27 11-27-58
i pushed up changes w the secondary variant but i will say i much prefer the blue buttons :shrug:

app/components/NavBarSearch.tsx
```diff
@@ -0,0 +1,94 @@
...
+            onChange={(e) => {
+              setSearchClicked(false);
+              setSearchValue(e.target.value.trim());
+            }}
+            type="search"
+            value={searchValue}
+            isInvalid={isSearchInvalid}
+          />
+          <Form.Control.Feedback type="invalid" tooltip>
+            Crash not found
```
frankhereford, line 80: Can this say "Case not found" when searching by case ID? 

image

I get that this is technically right because either a crash id or case id search is used to find crashes .. but not having this error message acknowledge that we did a case-id search makes me think it didn't work and my case ID was used to check against crashes. 

I'm not doing great at explaining this -- lmk if this doesn't make sense.

frankhereford (approved):
This is real fine - thank you @roseeichelmann! You're making this look easy; this is a top-shelf PR.

app/components/NavBarSearch.tsx
johnclary: I agree. Maybe the text should say "Case ID not found" and "Crash ID not found" to be super clear. would you mind making that change @roseeichelmann?

johnclary (approved):
> i much prefer the blue button

Fair. Let's leave it as secondary and revisit it very soon in the [app theme issue](https://github.com/cityofaustin/atd-data-tech/issues/20013). I added a note a there:

> Revisit the navigation crash search button variant (secondary is ugly)

If you would update the not found message text please, this is good to go!

🚢 🚢 🚢 🚢 🚢 🚢

app/components/NavBarSearch.tsx
roseeichelmann: yup great suggestion @frankhereford !

chiaberry (approved):
🚢 it!

I briefly looked to see how straightforward it would be to adjust the order of the icons in the form, or nudge one up or down and it doesnt look like they are directly in our control, so i dropped that investigation.

chiaberry:
oh and form I am talking about is 
Screenshot 2025-01-28 at 1 02 39 PM

PR #1665, opened 2025-01-20
Various follow-ups to record exporting

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/20481

This PR adds the remaining visible columns to the crashes list, and adds a bunch of extra columns to the CSV export.

It also reworks the way we inject column names in to the auto-generated graphql query, and introduces the ability to render related record values in the table component, which, as it turns out, was not necessary for this work and I am asking for feedback on whether or not I should shelve that part of the code.

The diff looks a bit worse than it is because I finally moved the queryBuilder typings into a separate file. 

## Testing

**URL to test:** Local or netlify 

⚠️ This PR makes breaking changes to the `queryConfig` in local storage - other branches may throw an error after you test this PR. The solution is to manually clear your localStorage when switching branches ⚠️

1. Visit the crashes page and use the **Reset** button to clear out the old query config.
2. Verify that you're seeing additional columns in the crashes list: fatality count, sus. serious injury count, and est. comp cost.
3. Use the export feature to see a whole lotta columns. Open the CSV in your fav spreadsheet editor to make sure it loads correctly.
4. Visit the locations page and make sure it loads correctly.
5. Visit the location details page and make sure the location crashes table loads correctly.

---

#### Ship list

- [ ] Code reviewed


app/configs/crashesListViewTable.ts
```diff
@@ -1,14 +1,10 @@
...
-  getStartOfYearDate,
   getYearsAgoDate,
   makeDateFilters,
 } from "@/components/TableDateSelector";
-import { crashesListViewColumns } from "@/configs/crashesListViewColumns";
-import { QueryConfig, FilterGroup } from "@/utils/queryBuilder";
+import { QueryConfig, FilterGroup } from "@/types/queryBuilder";
 import { DEFAULT_QUERY_LIMIT } from "@/utils/constants";
 
-const columns = crashesListViewColumns.map((col) => String(col.path));
```
johnclary, line 10: this PR removes the array of column names out of the `queryConfig` object and instead let's uses the column definitions that are passed into the `TableWrapper` component.

app/configs/tempCrashesListViewColumns.tsx
johnclary, line 1: i guess we overlooked these when Rose moved the nixed the temp crashes table 🙃

app/types/crashesList.ts
johnclary, line 1: as you can see, there is a lot of overlap here with the `Crash` type, and either in this PR or in a follow-up we should probably dedupe these typings 🤷

app/types/types.ts
```diff
@@ -25,6 +25,11 @@ export interface ColDataCardDef<T extends Record<string, unknown>> {
    * If the column is editable
    */
   editable?: boolean;
+  /**
+   * If the field should only be available in the exported table data -
+   * only affects when column is used in a table config
+   */
+  exportOnly?: boolean;
```
johnclary, line 32: new prop that enables columns to be hidden from the Table component

app/utils/queryBuilder.ts
```diff
@@ -299,6 +130,51 @@ const getWhereExp = (filterGroups: FilterGroup[]): string => {
...
+    let current = tree;
+    // split the path parts and save field names under their
+    // parent path part
+    path.split(".").forEach((part) => {
+      if (!current[part]) current[part] = {};
+      current = current[part];
+    });
+  });
+  return stringify(tree);
+};
```
johnclary, line 176: here's the new function that creates the field selections for the grahpql query string. this may be slightly over-engineered in that it enables the querying related fields through our dot-notated field paths.

we are not using any related fields in our tables today, but it does feel right to have this in place for future proofing🤷

app/utils/queryBuilder.ts
```diff
@@ -420,20 +300,21 @@ const buildQuery = (
  * URL query param
  * @returns {string} a graphql querry
  */
-export const useQueryBuilder = (
+export const useQueryBuilder = <T extends Record<string, unknown>>(
```
johnclary, line 303: and here we now need to declare carry along the generic `T` so that we can accept the column arrays as  `ColDataCardDef<T>`.

app/components/TableWrapper.tsx
```diff
@@ -69,8 +65,12 @@ export default function TableWrapper<T extends Record<string, unknown>>({
...
   });
 
-  const query = useQueryBuilder(queryConfig, contextFilters);
-  const exportQuery = useExportQuery(queryConfig, contextFilters);
+  const visibleColumns = useMemo(
+    () => columns.filter((col) => !col.exportOnly),
+    [columns]
+  );
+  const query = useQueryBuilder(queryConfig, visibleColumns, contextFilters);
+  const exportQuery = useExportQuery(queryConfig, columns, contextFilters);
```
johnclary, line 73: just pointing out that we now pass columns into the query builder so that it can generate the query string form them, whereas before we were grabbing an array of column name strings from the `queryConfig`.

app/utils/queryBuilder.ts
```diff
@@ -396,16 +280,29 @@ const buildQuery = (
...
-  )
+  /**
+   * If using aggregates, patch in the aggregation query
+   *  - must do this before we patch in the other template
+   * vars
+   */
+  let queryString = BASE_QUERY_STRING.replace(
+    "$aggregationQuery",
+    includeAggregates ? AGGREGATION_QUERY_STRING : ""
+  );
```
johnclary, line 291: here's the lil change that enables us to skip the aggregate data query when exporting. the aggregates add a lot of query time and are completely ignored for the purposes of generating the CSV download

app/types/types.ts
chiaberry: ✨

app/utils/queryBuilder.ts
johnclary: @frankhereford discovered that we do in fact need this functionality for the Fatalities view—so we will have a test of it in the very near future!

app/utils/queryBuilder.ts
```diff
@@ -317,9 +196,13 @@ const getWhereExp = (filterGroups: FilterGroup[]): string => {
...
  */
-const buildQuery = (
-  {
-    columns,
+const buildQuery = <T extends Record<string, unknown>>(
+  queryConfig: QueryConfig,
+  columns: ColDataCardDef<T>[],
+  includeAggregates: boolean,
+  contextFilters?: Filter[]
+): string => {
```
chiaberry, line 204: I need typescript 101 help here, what does adding `<T extends Record<string, unknown>>` add to our type?

app/utils/queryBuilder.ts
```diff
@@ -299,6 +133,51 @@ const getWhereExp = (filterGroups: FilterGroup[]): string => {
...
+    const isEmpty = Object.keys(value).length === 0;
+    return isEmpty ? key : `${key} ${stringify(value)}`;
+  });
+  return `{ ${fields.join(" ")} }`;
+}
+
+/**
+ * Given an array of dot-notaed column paths, generate
+ * a query string the is a graphql-compatible field
+ * selection set string
```
chiaberry, line 152: *dot-notated
and i think "query string that is"

app/utils/queryBuilder.ts
```diff
@@ -299,6 +133,51 @@ const getWhereExp = (filterGroups: FilterGroup[]): string => {
   return andExps.length > 0 ? `{ _and: [ ${andExps.join("\n")} ]}` : "{}";
 };
 
+/**
+ * Recursively stringify a field tree object into a nested graphql
+ *  field selection set
```
chiaberry, line 138: you spoiled me with the sample input and output in the function below's docstring, now i want it everywhere 😬

chiaberry (approved):
All the test steps check out and I read the code and follow the changes. I have some typescript 101 questions, still trying to get a sense of some of the syntax. 

As for the fancy nested query part, I think its fine to keep because it doesnt seem like too much code and it will be nice to have it on hand if and when we need those columns.

chiaberry:
I was working on my PR review and missed your comment that we do need the nested query builder, so I am glad that I said let's keep it

app/utils/queryBuilder.ts
johnclary: because we added `ColDataCardDef<T>` as a prop, and it requires a generic (`T`), we need to supply `T` in our function declaration. that's just the syntax for using generics.

generics themselves can be of a specific base type, and in our case the type for our `ColDataCardDef<T>` is  an extension of the `Record<string, unknown>` utility type. You can see that this specific type comes from our definition of  the `ColDataCardDef` interface in `types.ts.` 

So, to be compliant with the generic type `ColDataCardDef` is expecting, we need specify that this function specifically excepts `<T extends Record<string, unknown>>` rather than plain 'ole `T`.

I hope that helps! If you remove `extends Record<string, unknown>` from this function you will see the way typescript barks about it.

app/types/queryBuilder.ts
frankhereford, line 1: John, this file is absolute 🔥. This is peak documentation to me.  

At least 3 things:

* You're laying out how this data structure works in a human readable document here in the code, 
* you're injecting that same documentation **into my editor** that is consumable both as tool tips but also as LLM context, 
* and enabling typescript to enforce sanity and everything great that that entails.

I'm going to challenge myself to type like this in the future, this is just fantastic.

app/utils/queryBuilder.ts
chiaberry: this does help! thank you

app/configs/crashesListViewColumns.tsx
frankhereford, line 1: @johnclary -- sorry for the piecemeal feedback here 🙈 

Should this file be a `.ts` file? Totally non-rhetorical question; btw -- please forgive me if I'm misunderstanding. I **think** because it's not containing any JSX markup, we might be able to call this just plain typescript.

app/configs/crashesListViewColumns.tsx
```diff
@@ -35,4 +45,165 @@ export const crashesListViewColumns: ColDataCardDef<CrashesListRow>[] = [
...
     sortable: true,
   },
+  {
+    path: "vz_fatality_count",
+    label: "Fatalities",
+    sortable: true,
+  },
+  {
+    path: "sus_serious_injry_count",
+    label: "Sus serious injuries",
```
roseeichelmann, line 55: idk why this shortening of suspected to "sus" bothers me but its what we currently have in prod and it keeps it from being so long so i guess its fine :shrug:

app/types/queryBuilder.ts
```diff
@@ -0,0 +1,193 @@
...
+  /**
+   * If the filter should be wrapped with `%`
+   * --should only be used with string types
+   */
+  wildcard?: boolean;
+}
+
+interface FilterGroupBase {
+  /**
+   * The arbitrary ID must uniquely identifier of this group amongst
```
roseeichelmann, line 53: maybe "This arbitrary ID must uniquely identify this group amongst all other filter groups"?

app/types/queryBuilder.ts
```diff
@@ -0,0 +1,193 @@
...
+   * all other filter groups
+   */
+  id: string;
+  /**
+   * Optional label that can be used to label the filter cards
+   */
+  label?: string;
+  /**
+   * Applies to FilterCard's filter groups only—aka switches - it enables
+   * switch filtegroups to be present in the config and ignored
```
roseeichelmann, line 63: typo filtegroups

app/utils/queryBuilder.ts
roseeichelmann: definitely gonna be so handy to have for future work and also super coooolll thanks for adding this functionality!!

app/utils/queryBuilder.ts
roseeichelmann: so clever!!

roseeichelmann (approved):
all the steps work great! this is such high quality work, super snazzy i love the way you implemented and organized everything :sparkles:

app/types/queryBuilder.ts
johnclary: thank you—fixed here and in the filter interface dostring above it

app/configs/crashesListViewColumns.tsx
johnclary: yeah, "Sus" is not good. i removed it and fixed the casing (ty Chia). the [ACC designs](https://www.figma.com/proto/2acDnn8lRJ3FHnPEDGlk3A/Applied-Labs-Figma?type=design&node-id=811-60916&t=oEA7X290kfUxsKrT-0&scaling=scale-down-width&page-id=0%3A1&starting-point-node-id=811%3A60895&hotspot-hints=0&hide-ui=1) use the single-letter injury indicators which i could get on board with if we had a hover tooltip that explained what the letters meant

app/configs/crashesListViewColumns.tsx
johnclary: you're right, this does not need to be a `tsx` file, and it probably is named as such because i copied from the crashes list view config. it would indeed need to be a `tsx` file if we used a `valueFormatter` or `valueRenderer` that returned JSX. which kinda makes me want to leave it as-is for future proofing?

app/utils/queryBuilder.ts
johnclary: just want to mention that we have https://github.com/cityofaustin/atd-data-tech/issues/20790 to test out this new related record functionality. i added a note in that issue about sorting on related record columns, because i very much doubt that will work without more code changes 🙃

app/configs/crashesListViewColumns.tsx
frankhereford: This is just my two cents -- weakly held opinions, at that!

This is merged, and it's fine as merged, but I think that we should treat the "future proofing" question sort of like we do with commented out code. If it's not part of the present solution, it's probably best to be struck (or in this case, renamed) to make it correct-as-of-now as it's confusing without the immediate context of this particular thread we're having here in this PR.

johnclary: sounds good to me—i'll to that in the nextjs branch 👍

johnclary: actually, this file does have JSX in it, so probably best left as a `.tsx` file? we do have a few other configs that don't meet that requirement and i patched them here: https://github.com/cityofaustin/vision-zero/pull/1603/commits/5ef07f552447a287f13d49cd947aedc041fc10ea

PR #1670, opened 2025-01-21
Update AFD/EMS ETL WRT the CTM AFP jurisdiction format

frankhereford:
## Associated issues

_none_; this is the PR that came from [this conversation](https://austininnovation.slack.com/archives/CM9MK950S/p1737479396401139).

## Testing

* Replicate the production data to your local DB
* Apply this migration, both up and down (and then back up again)
* Try to do an insert which will cause the function to do its thing:

```
INSERT INTO afd__incidents (
    incident_number,
    unparsed_ems_incident_number,
    ems_incident_numbers,
    call_datetime,
    calendar_year,
    jurisdiction,
    address,
    problem,
    flagged_incs,
    geometry
) 
VALUES (
    8675309, 
    '-', 
    '{}', 
    '2025-01-15 03:00:00', 
    2025, 
    'AFD', 
    'CAD_ADDRESS_STRING_HERE', 
    'CAD_PROBLEM_STRING_HERE', 
    1, 
    ST_SetSRID(ST_Point(-97.753991, 30.262542), 4326)
) 
ON CONFLICT DO NOTHING;
```
* Since we're also updating the EMS trigger as well, make sure this insert works too:

```
INSERT INTO ems__incidents (
    pcr_key,
    incident_date_received,
    incident_time_received,
    incident_number,
    incident_location_address,
    incident_location_city,
    incident_location_state,
    incident_location_zip,
    incident_location_longitude,
    incident_location_latitude,
    incident_problem,
    incident_priority_number,
    pcr_cause_of_injury,
    pcr_patient_complaints,
    pcr_provider_impression_primary,
    pcr_provider_impression_secondary,
    pcr_outcome,
    pcr_transport_destination,
    pcr_patient_acuity_level,
    pcr_patient_acuity_level_reason,
    pcr_patient_age,
    pcr_patient_gender,
    pcr_patient_race,
    mvc_form_airbag_deployment,
    mvc_form_airbag_deployment_status,
    mvc_form_collision_indicators,
    mvc_form_damage_location,
    mvc_form_estimated_speed_kph,
    mvc_form_estimated_speed_mph,
    mvc_form_extrication_comments,
    mvc_form_extrication_datetime,
    mvc_form_extrication_required_flag,
    mvc_form_patient_injured_flag,
    mvc_form_position_in_vehicle,
    mvc_form_safety_devices,
    mvc_form_seat_row_number,
    mvc_form_vehicle_type,
    mvc_form_weather,
    pcr_additional_agencies,
    pcr_transport_priority,
    unparsed_apd_incident_numbers,
    pcr_patient_acuity_initial,
    pcr_patient_acuity_final,
    geometry,
    apd_incident_numbers
) VALUES (
    123456,
    '2025-01-01 00:00:00',
    '12:34:56.000000',
    'ABC123-0000',
    '123 Placeholder St',
    'Placeholder City',
    'XY',
    99999.0,
    -100.000000,
    40.000000,
    'Sample Problem Description',
    1,
    'Sample Cause of Injury',
    'Sample Complaint',
    'Primary Impression',
    'Secondary Impression',
    'Outcome Placeholder',
    'Transport Destination Placeholder',
    'High Acuity',
    'Reason Placeholder',
    30,
    'Other',
    'Unspecified Race',
    'Sample Airbag Deployment',
    'Sample Airbag Deployment Status',
    'Sample Collision Indicator',
    'Sample Damage Location',
    50.0,
    30.0,
    'Extrication Comment Placeholder',
    NULL,
    1,
    0,
    'Sample Vehicle Position',
    'Sample Safety Device',
    '1',
    'Vehicle Type Placeholder',
    'Weather Placeholder',
    NULL,
    NULL,
    NULL,
    'Initial Acuity Placeholder',
    'Final Acuity Placeholder',
    ST_SetSRID(ST_Point(-100.000000, 40.000000), 4326),
    '{}'
) ON CONFLICT DO NOTHING;
```

---

#### Ship list

- [ ] Code reviewed


etl/afd_ems_import/afd_incidents_trigger.sql
frankhereford, line 1: These two SQL files were out of date. Instead of fixing them to become in-date again, I am asserting that they live in a non-standard location. 

IMHO, if we want to keep SQL around of triggers and views and other things that are actually defined by the DB, not the code base, then we should incorporate some version of the moped CI that writes them on change into back into the code base for easy reference.

database/migrations/default/1737481505501_specify-schema-for-jurisdictions/up.sql
```diff
@@ -0,0 +1,70 @@
...
+create or replace function public.afd_incidents_trigger()
+returns trigger
+language plpgsql
+as $function$
+BEGIN
+  update afd__incidents set
+    austin_full_purpose = (
+      select ST_Contains(jurisdiction.geometry, incidents.geometry)
+      from afd__incidents incidents
+      left join geo.jurisdictions jurisdiction on (jurisdiction.jurisdiction_label = 'AUSTIN FULL PURPOSE')
```
frankhereford, line 17: This is the line that is changed, with the addition of `geo.`.

database/migrations/default/1737481505501_specify-schema-for-jurisdictions/up.sql
```diff
@@ -0,0 +1,78 @@
...
+create or replace function public.afd_incidents_trigger()
+returns trigger
+language plpgsql
+as $function$
+BEGIN
+  WITH jurisdiction_union AS (
+    SELECT ST_Union(geometry) AS geometry
+    FROM geo.jurisdictions
+    WHERE jurisdiction_label = 'AUSTIN FULL PURPOSE'
+  )
```
frankhereford, line 17: Unioning the two records we keep from CTM that reflect this single piece of geometry

database/migrations/default/1737481505501_specify-schema-for-jurisdictions/up.sql
```diff
@@ -0,0 +1,78 @@
...
+  UPDATE afd__incidents
+  SET
+    austin_full_purpose = (
+      SELECT ST_Contains(jurisdiction_union.geometry, incidents.geometry)
+      FROM afd__incidents incidents, jurisdiction_union
```
frankhereford, line 22: And using that ephemeral single piece of geometry instead of the table which now lives in `geo.` and no longer in `public.`, which is related to what brought this all to light.

etl/afd_ems_import/afd_incidents_trigger.sql
frankhereford, line 1: These two SQL files were out of date. Instead of fixing them to become in-date again, I am asserting that they live in a non-standard location and us humans are bad at keeping them up to date.

IMHO, if we want to keep SQL around of triggers and views and other things that are actually defined by the DB, not the code base, then we should incorporate some version of the moped CI that writes them on change into back into the code base for easy reference.

database/migrations/default/1737481505501_specify-schema-for-jurisdictions/up.sql
johnclary: Cool. The [`ems_incidents_trigger`](https://github.com/cityofaustin/vision-zero/blob/f9ac8dd7b609572b815b14d1e841777eb2831907/database/migrations/default/1736197283308_jurisdiction_refactor/up.sql#L157-L161) function does not do this union. does that look like a bug to you? 🤔

etl/afd_ems_import/afd_incidents_trigger.sql
johnclary: Totally agree! 🙏 
* https://github.com/cityofaustin/atd-data-tech/issues/18791

database/migrations/default/1737481505501_specify-schema-for-jurisdictions/up.sql
```diff
@@ -0,0 +1,78 @@
+drop trigger if exists afd_incidents_trigger_insert on public.afd__incidents;
```
johnclary, line 1: not a big deal, but i don't think it's necessary to drop and re-add the triggers in order to modify the trigger function—at least it's not something I did in the previous PR 🤷

database/migrations/default/1737481505501_specify-schema-for-jurisdictions/up.sql
```diff
@@ -0,0 +1,78 @@
...
+        or (
+            old.ems_incident_numbers is distinct
+            from
+            new.ems_incident_numbers
+        )
+        or (
+            old.call_datetime is distinct
+            from
+            new.call_datetime
+        )
```
johnclary, line 76: my understanding is that, because it is very common for us to upsert thousands of records which already exist in the db, these checks cut down on unnecessary running of the trigger function. does that sound right?

johnclary (approved):
looks fantastic to me! i did have a few questions.

thank you for tackling this so quickly 🙌

database/migrations/default/1737481505501_specify-schema-for-jurisdictions/up.sql
frankhereford: Oh, I see one reason at least why you were asking about if we get records back from all time. If memory serves, and that's a huge if, I think you're exactly correct. We would be processing this transaction upwards of a 100K times a day without this condition. I guess you could handle it here or you could handle it in the ETL, half dozen one way, six the other. 🤷

database/migrations/default/1737481505501_specify-schema-for-jurisdictions/up.sql
frankhereford: ah -- i had put my `drop trigger` line in first which caused me to have to drop the triggers, but I see what you're saying with the `or replace` part of the `create` statement. I'll update this for brevity. Thanks!

database/migrations/default/1737481505501_specify-schema-for-jurisdictions/down.sql
frankhereford, line 1: Wow, I have never seen `git` do this before. I have no idea why it thinks this file was renamed and moved. Sure, they had some similar content, but this is the furthest from what I actually did. If I caused this, I'd be so grateful to know how, if anyone has any idea.

database/migrations/default/1737481505501_specify-schema-for-jurisdictions/up.sql
frankhereford: I do think this is a bug, and it is valid SQL so it's not croaking. It is picking one portion of the geometry or other at random and saving the `austin_full_purpose` flag based on that portions of the geometry's containment of the incident or not. Do you want me to fold a fix into this PR @johnclary?

johnclary: yes, please, would you fix that function, Frank? I'm really glad you caught this!

good call on the backfill. we should be able to rerun [these update statements](https://github.com/cityofaustin/atd-data-tech/issues/20461#issuecomment-2576082208) for that 👍

frankhereford:
@johnclary I have updated the PR to include a fix for the EMS side of this as well. If you want to re-review, please feel free, otherwise, I'll look to merge this when it picks up another ✅. Thanks!

database/migrations/default/1737481505501_specify-schema-for-jurisdictions/down.sql
johnclary: yes, this is odd and surprising! 😵

johnclary (approved):
this looks great—thank you for stitching up that issue with the EMS incidents as well!!! 
🚢 🚢 🚢 🚢 🚢

PR #1671, opened 2025-01-21
Prettier

mateoclarke:
## Associated issues

All I did in this PR was run:
`npm install --save-dev prettier eslint-config-prettier`

And I added a `format` script to the package.json so I could run:
`npm run format`

## Testing

**URL to test:** 

**Steps to test:**

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/app/app/dashboard/page.tsx
+++ b/app/app/dashboard/page.tsx
@@ -50,7 +50,6 @@ const dashboardLinks: DashboardLinkCardProps[] = [
 ];
 
 export default function Dashboard() {
-
   return (
     <>
       <Row>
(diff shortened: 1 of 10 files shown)


frankhereford (approved):
👋 Thanks for bringing prettier into the club here. I guess I had been working with the prettier installed globally on my mac here -- I'm not sure. Anyway, this is clearly better, and also, thank you for the big linting pass. 🧹🧹🧹 

sbsp_clean

mddilley (commented):
Looks prettier! 🚀 The only thing that I wondered is if we would want to add a config file like we have in [the current VZE project](https://github.com/cityofaustin/vision-zero/blob/main/editor/.prettierrc). Similar to what Frank mentioned about his setup, I have my code editor set up to auto-format on save and use a config file if there is one in the project directory.

roseeichelmann (approved):
thanks for prettifying our codebase mateo!! :sparkles:

mddilley:
@mateoclarke Yep, that is exactly what i overlooked. 😅 thanks! 🙏

PR #1674, opened 2025-01-23
Add branding to the sign-in page and make it look nicer

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/20755

Screenshot 2025-01-24 at 2 17 25 PM

## Testing

**URL to test:** Local or netlify

1. Sign out
2. Look at the sign-in page

---

#### Ship list

- [ ] Code reviewed


app/styles/global.scss
johnclary, line 1: Also for consideration, these logos with the COA seal are also available but only in monotone white or asphalt. It would be nice to get the COA seal on this page but i think i'm partial to the non-stacked, two-tone logo

Screenshot 2025-01-23 at 4 09 51 PM

Screenshot 2025-01-23 at 4 11 03 PM

app/styles/global.scss
```diff
@@ -1,6 +1,24 @@
 /* see https://github.com/twbs/examples/blob/main/react-nextjs/src/styles/styles.scss */
 @import "bootstrap/scss/bootstrap.scss";
 
+$vz-gold: #edb41e;
+$vz-gold-darker: #b38715;
```
chiaberry, line 5: this is _technically_ not the VZ Sunshine color hex code from the branding guide but their own stated RGB doesn't even match their hex code and the two hexes are essentially the same color. 
Where did you get the darker version?

johnclary: thanks for pointing this out. i used a color picker to grab the gold code before checking the brand guide PDF.

i updated these to match the name and codes from the branding guide, and also added `peacock` and `sand` colors for the heck of it.

for the darker sunshine, i created it myself by dropping the light value on the color. feel free to suggest a different one! i went poking around in the boostrap css and they have a  function that applies the darkness/tint to their color system. i think we could probably import that and use it—perhaps as a todo in https://github.com/cityofaustin/atd-data-tech/issues/20013?

app/styles/global.scss
johnclary: here's another stacked one using `peacock` as the button color. i'm coming around to this, i just find the bottom line of the Vision Zero stack kinda hard to read, i think because the person icon is not being used as a letter but the car and bicycle are 🤷 

Screenshot 2025-01-24 at 2 18 05 PM

amenity: > the person icon is not being used as a letter but the car and bicycle are
Thanks, John... Now I can't unsee this weirdness. 🙈

I like both of the black + gold options. 🌚 🌞

app/components/LoginContainer.tsx
```diff
@@ -15,14 +17,31 @@ export default function LoginContainer({
     <Container
       fluid
       style={{ height: "100vh", overflow: "hidden" }}
-      className="bg-dark"
+      className=""
     >
       <div className="d-flex justify-content-center align-content-center h-100">
         <div className="align-self-center p-5 bg-white rounded text-center">
```
frankhereford, line 23: Not every day you see the web-dev challenge of "center this div horizontally and vertically" done in practice. ✨ very nice!

frankhereford (approved):
As is, this looks great. Huge improvement.

Of the various versions of this you put up, I'm most a fan of the black one because it sets out a reasonable border containing this bit of the functionality. Without the black, it's sort of floating out in a sea of white -- not bad, just not contained.

The COA seal is also a nice addition; it makes it "ours" and not one of the other cities' VZ systems. In terms of SEO, something on this page that says austin, even if its a filename or some sort of hinting would be pretty great, as this is the only part of the site google will ever see. 

After the black coloring, my next fav is peacock, followed up by sunshine as my third fav. Still -- this is a huge improvement.

roseeichelmann (approved):
this looks great, i like the non stacked version you have now, but i also like the stacked with the COA seal. i think i prefer the white background w/ peacock button but if we want to fit the style guide more i guess gold button :shrug:

johnclary:
thanks, y'all! i am going to bring this in as-is, but i've already started changing this in  the app theme PR: https://github.com/cityofaustin/vision-zero/pull/1676. i'm onboard with one of the stacked logos, so we have the seal, and i am going to propose a different color palette that works better for UI needs.

PR #1680, opened 2025-01-28
Add and clean up notes audit fields

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/20820

## Testing

**URL to test:** 

**Steps to test:**

---

#### Ship list

- [ ] Code reviewed


database/metadata/databases/default/tables/public_crash_notes.yaml
```diff
@@ -11,87 +11,81 @@ insert_permissions:
...
       columns:
         - crash_pk
-        - created_at
         - date
-        - id
         - is_deleted
         - text
-        - updated_at
-        - user_email
+        - updated_by
```
johnclary, line 88: in both tables i have revoked users ability to insert into the `id`, `created_at`, and `updated_at` columns and revoked the ability to modify those columns plus the `created_by` column 🔒

database/migrations/default/1738088180607_notes_audit_fields/up.sql
```diff
@@ -0,0 +1,26 @@
...
+--
+alter table crash_notes add column created_by text, add updated_by text;
+
+update crash_notes set created_by = user_email, updated_by = user_email;
+
+alter table crash_notes
+alter column created_by set not null,
+alter column updated_by set not null,
+drop column user_email,
+drop column date;
```
johnclary, line 13: this `date` column is lingering from pre-data model days. we already have `created_at` and `updated_at` working on this tables

johnclary:
thanks @mateoclarke for catching this! it looks like the `update` statements in the migration are triggering the `updated_at` trigger to fire. i will patch the migration so that those triggers are disabled when the updates execute.

database/migrations/default/1738088180607_notes_audit_fields/up.sql
```diff
@@ -0,0 +1,32 @@
+-- use replica mode to disable triggers
+set session_replication_role = replica;
```
johnclary, line 2: @mateoclarke here's the fix for the triggers. a trick we've [used in Moped](https://github.com/cityofaustin/atd-moped/blob/main/moped-database/migrations/1707160404494_clear_component_empty_strings/up.sql) a few times.

johnclary:
hmm, it just occurred to me is that this migrations is obviously not going to work on staging, because it will break the staging VZE. we need this migration to be applied to staging in order to have a deploy preview.

i am going to modify this PR once more so that it does not delete any column or revoke any hasura permissions. can we do that db cleanup post go-live.

johnclary:
OK—sorry for all the noise. I just patched this so that it is backwards compatible with staging. I tested this with the old VZE locally and it looks good. I am tracking the post-go live work in https://github.com/cityofaustin/atd-data-tech/issues/20753.

i squashed these commits to that's it's easier to bring into `main`.

PR #1681, opened 2025-01-29
Update docs about CRIS delivery config

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/20827

## Testing

Read the docs 👍 

---

#### Ship list

- [ ] Code reviewed


mddilley (approved):
Read through the docs, and they spell out this process nicely! ✨ 🚢

database/README.md
```diff
@@ -82,7 +82,30 @@ The credentials for our CRIS logins are in the password store, including a note
...
+You will need to follow these steps to create or update our delivery configuration, which would be necessary if we were using a new/re-activated CRIS user account, our S3 bucket structure changed, etc.
+
+1. Login to CRIS using the appropriate account (see above): https://cris.dot.state.tx.us/
+2. Use the account dropdown menu in the top-right corner of the screen and select **Delivery Options**
+
+![CRIS delivery config - dropdown](../docs/images/delivery_config_1.png)
+
+1. Select `AWS` as the delivery method and fill in the config form as pictured below. See the **CRIS Import** item in 1pass for the current credentials.
+
+2. Note that the **Key Name Prefix** must not begin with a slash: `prod/cris_extracts/inbox/`
```
chiaberry, line 100: It would be nice to include the key name prefixes in the 1pass item as well and not just here in the Note. I am assuming staging is just this with staging swapped but still would be nice to have it written out.

chiaberry (approved):
Do you think we need to specify that the delivery config needs to be set before attempting to create an extract request? If I make the extract request and then adjust the delivery configuration, I dont think it updates the existing request.

database/README.md
johnclary: Great idea—I just did that by adding a new section at the bottom of the item. I didn't want to make it see as those these secrets were depended on by the ETL or Airflow DAG

Screenshot 2025-01-30 at 10 11 53 AM

johnclary:
> If I make the extract request and then adjust the delivery configuration, I dont think it updates the existing request.
I'm not quite sure if this is true. If we changed our config today, I _think_ our recurring extract would use the updated AWS settings on tomorrow's delivery?

PR #1682, opened 2025-01-29
Link to user details page from account dropdown menu item

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/20739

## Testing

**URL to test:** 
Local

**Steps to test:**
Test using the account link in the dropdown, it should route to your user details page

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/app/components/AppNavBar.tsx
+++ b/app/components/AppNavBar.tsx
@@ -1,3 +1,4 @@
+import Link from "next/link";
 import Container from "react-bootstrap/Container";
 import Dropdown from "react-bootstrap/Dropdown";
 import { User, LogoutOptions } from "@auth0/auth0-react";
@@ -31,7 +32,7 @@ export default function AppNavBar({ user, logout }: NavBarProps) {
     <Navbar expand="lg" className="pe-3">
       <Container fluid>
         <Col className="d-flex justify-content-start">
-          <Navbar.Brand href="/crashes">
+          <Navbar.Brand href="/crashes" as={Link}>
             <Image
               src="/assets/img/brand/visionzerotext.png"
               alt="Vision Zero Logo"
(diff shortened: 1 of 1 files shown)


app/components/AppNavBar.tsx
```diff
@@ -52,7 +54,7 @@ export default function AppNavBar({ user, logout }: NavBarProps) {
             <Dropdown.Menu>
               <Dropdown.Header>{user.email}</Dropdown.Header>
               <Dropdown.Divider />
-              <Dropdown.Item eventKey="1">
+              <Dropdown.Item eventKey="1" href={`users/${userId}`}>
```
johnclary, line 57: this works, but i think we want to use the nextjs `Link` component here so that the internal app router handles this correctly and does fire off a full page reload. 

All you gotta do is set the `as` prop to be `as={Link}` , like we do [here](https://github.com/cityofaustin/vision-zero/blob/b3dc2b56faff636bb8cdbc9a68d92e526b56a42c/app/components/SideBarListItem.tsx#L30).

Browsing this component, I see that I missed this in the `Navbar.Brand` component, too. Would you mind updating that one as well?

roseeichelmann: updated!

chiaberry (commented):
if you are on your account page and then in the dropdown try to navigate to your account page again the url stacks up and you get a 404 ie `http://localhost:3002/users/users/auth0%7C61bcce76838a5d00706e8102`

app/components/AppNavBar.tsx
```diff
@@ -52,7 +55,7 @@ export default function AppNavBar({ user, logout }: NavBarProps) {
             <Dropdown.Menu>
               <Dropdown.Header>{user.email}</Dropdown.Header>
               <Dropdown.Divider />
-              <Dropdown.Item eventKey="1">
+              <Dropdown.Item eventKey="1" href={`users/${userId}`} as={Link}>
```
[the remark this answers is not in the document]
roseeichelmann: thanks for catching this @chiaberry , the leading slash fixed it!

johnclary (approved):
🚢

PR #1700, opened 2025-02-06
Add coordinate inputs to map card

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/20975

Screenshot 2025-02-07 at 11 02 25 AM

## Testing

**URL to test:** [Netlify](https://deploy-preview-1700--atd-vze-staging.netlify.app/)

1. Try editing the crash location by keying lat/lon
2. Notice that when you move the map, the lat/lon fields update
4. Notice that if you save an invalid lat/lon (has alpha characters or is outside the allowed range) you hit a validation error and cannot save
5. Hit the cancel button and then start editing again, observe that validation errors have been cleared
6. Fix the validation errors and save coordiantes

---

#### Ship list

- [ ] Code reviewed


editor/components/CrashMap.tsx
```diff
@@ -17,9 +17,13 @@ import "mapbox-gl/dist/mapbox-gl.css";
...
-  latitude: number | null;
-  longitude: number | null;
+/**
+ * We work with lat/lon as strings so that the map UI
+ * and form inputs can be linked while editing
+ */
+export interface LatLonString {
+  latitude: string;
+  longitude: string;
 }
```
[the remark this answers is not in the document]
frankhereford: Yech, no kidding when you say it's tricky. Thank you for this detailed explanation of what's going on here and why, that is very helpful to me. 🙏

editor/utils/formHelpers.ts
```diff
@@ -187,3 +187,28 @@ export const handleFormValueOutput = (
...
+  }
+  if (isLatitude && Math.abs(value) > 90) {
+    return null;
+  } else if (!isLatitude && Math.abs(value) > 180) {
+    return null;
+  }
+  return (
+    Math.round(value * Math.pow(10, MAP_COORDINATE_PRECISION)) /
+    Math.pow(10, MAP_COORDINATE_PRECISION)
+  );
```
frankhereford, line 213: I like how you did this with the `Math` built in library. I was really surprised to see that there isn't a rounding method that takes an argument to define precision. 

I'm not sure this is good advice, but when looking for something that did offer that precision argument, the `.toFixed()` ([docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed)) method on the the number class. (number prototype? I'm not sure what the fundamental are called here.) `.toFixed()` might offer slightly improved readability, but, again, this is indubitably correct how you've done it.

editor/components/CrashMap.tsx
```diff
@@ -102,8 +109,8 @@ export const CrashMap = ({
       )}
       {isEditing && (
         <Marker
-          latitude={editCoordinates.latitude || 0}
-          longitude={editCoordinates.longitude || 0}
+          latitude={Number(editCoordinates.latitude || 0)}
+          longitude={Number(editCoordinates.longitude || 0)}
```
frankhereford, line 113: Can we do something that runs this lat and lon prop through something like this before it's passed down into this component?


```
  let inputValue = ;

  // Allow empty input or a valid number as an intermediate state
  if (inputValue === "" || inputValue.endsWith(".")) {
    return inputValue;
  }

  let num = Number(inputValue);
  if (!isNaN(num)) {
    setState(num.toFixed(MAP_COORDINATE_PRECISION));
  }
```

I feel like my idea here is probably half banked in terms of where it should be plugged in, but essentially, I'm hoping to find a way to avoid showing the 20 decimal digits in the time between when a user moves the map in edit mode and when they save it. 

If this is cat-brained, i totally accept that and I view this as already a win. I guess I'm hung up on what the user sees, even if we do it all super correct on the backend.

A last thought, this is super minor and don't let it hold up anything important!

frankhereford (approved):
I think this is 100% an improvement and is shipable. 🚢 ✅ 

I feel like i'm offering up dubious advice, so please don't take me to literally, but I hope you'll see what i'm trying to convey, even if I'm communicating it poorly.

johnclary:
derp—this breaks if you enter an alpha character. i am going unrequest review and use RFH and zod 😖

editor/components/CrashMap.tsx
```diff
@@ -12,16 +12,39 @@ import MapGL, {
...
+    .number({ invalid_type_error: "Must be a number" })
+    .min(-90, { message: "Cannot be less than -90" })
+    .max(90, { message: "Cannot be greater than 90" }),
+  longitude: z.coerce
+    .number({ invalid_type_error: "Must be a number" })
+    .min(-180, { message: "Cannot be less than -180" })
+    .max(180, { message: "Cannot be greater than 180" }),
+});
+
+export type CoordinateValidationError = ZodFormattedError<LatLon>;
```
johnclary, line 46: new Zod schema to deal with validation

editor/components/CrashMapCard.tsx
```diff
@@ -42,11 +49,17 @@ export default function CrashMapCard({
...
-    longitude: 0,
+  const [isEditing, setIsEditing] = useState(false);
+  const [mapLatLon, setMapLatLon] = useState<LatLon>({
+    latitude: DEFAULT_MAP_PAN_ZOOM.latitude,
+    longitude: DEFAULT_MAP_PAN_ZOOM.longitude,
   });
+  const [formLatLon, setFormLatLon] = useState<LatLonString>({
+    latitude: String(DEFAULT_MAP_PAN_ZOOM.latitude),
+    longitude: String(DEFAULT_MAP_PAN_ZOOM.longitude),
+  });
```
johnclary, line 60: now we have a pair of state variables that manage the lat/lon coordinates. a numeric interface for the map, and a string-based on for the form inputs

editor/components/CrashMapCard.tsx
```diff
@@ -55,71 +68,108 @@ export default function CrashMapCard({
...
               <Button
                 size="sm"
                 variant="primary"
                 disabled={isMutating}
                 onClick={async () => {
-                  if (!isEditingCoordinates) {
-                    setIsEditingCoordinates(true);
+                  if (!isEditing) {
+                    setIsEditing(true);
                   } else {
```
johnclary, line 126: when we save, we check if the map coordinates match the form coordinates, if not, we validate the form coordinates and use them if they var valid

editor/components/CrashMap.tsx
```diff
@@ -54,27 +77,33 @@ export const CrashMap = ({
...
+  mapLatLon,
+  setMapLatLon,
 }: CrashMapProps) => {
   const onDrag = useCallback(
     (e: ViewStateChangeEvent) => {
-      const latitude = e.viewState.latitude;
-      const longitude = e.viewState.longitude;
-      setEditCoordinates({ latitude, longitude });
+      // truncate values to our preferred precision
+      const latitude = +e.viewState.latitude.toFixed(MAP_COORDINATE_PRECISION);
```
johnclary, line 86: thanks @frankhereford for the suggestion to use `toFixed` as a rounding hack — nice and readable 🙏

editor/components/CrashMapCoordinateForm.tsx
```diff
@@ -0,0 +1,95 @@
...
+                  ...formLatLon,
+                  latitude: e.target.value.trim(),
+                });
+              }}
+              isInvalid={Boolean(validationError?.latitude)}
+            />
+            <Form.Control.Feedback
+              type="invalid"
+              className="d-block"
+              style={{ minHeight: "1.2em" }}
```
johnclary, line 62: this `minHeight` makes room for the form validation error messages

editor/components/CrashMap.tsx
```diff
@@ -12,16 +12,39 @@ import MapGL, {
...
+export const LatLonSchema = z.object({
+  latitude: z.coerce
+    .number({ invalid_type_error: "Must be a number" })
+    .min(-90, { message: "Cannot be less than -90" })
+    .max(90, { message: "Cannot be greater than 90" }),
+  longitude: z.coerce
+    .number({ invalid_type_error: "Must be a number" })
+    .min(-180, { message: "Cannot be less than -180" })
+    .max(180, { message: "Cannot be greater than 180" }),
+});
```
chiaberry, line 44: what about restricting these to min and max that are closer to Austin? Just to see what it would say, I put 0 in as a the latitude, and it saves it successfully, but when i refresh the page, its just the original latitude but rounded down. I don't know why anyone would put 0, but its valid input

chiaberry: oh i can also save empty strings, it just thinks i'm saving 0s

johnclary: hmm, this is happening because of `coerce`, which transforms `null` and empty strings to `0` 😕 

good idea—i imported the bbox we use for the map limits and used it to set the min/max 🙏

chiaberry:
I can't save 0 anymore, 👍🏼

frankhereford (approved):
I'm floored John. For this feature, this is just a whole cut above everything that came before it. To me, this is a no-expenses-spared implementation of this component. 

This should be a published library. 🚢

chiaberry (approved):

editor/components/CrashMap.tsx
```diff
@@ -42,8 +66,8 @@ interface CrashMapProps {
   /**
    * The lat/lon coordinates that are saved while editing
    */
-  editCoordinates: LatLon;
-  setEditCoordinates: Dispatch<SetStateAction<LatLon>>;
+  mapLatLon: LatLon;
```
roseeichelmann, line 69: i like this var name update :+1:

editor/components/CrashMap.tsx
```diff
@@ -12,16 +12,40 @@ import MapGL, {
...
 
+export const LatLonSchema = z.object({
+  latitude: z.coerce
+    .number({ invalid_type_error: "Must be a number" })
+    .min(MAP_MAX_BOUNDS[0][1], { message: "Latitude is invalid" })
+    .max(MAP_MAX_BOUNDS[1][1], { message: "Latitude is invalid" }),
+  longitude: z.coerce
+    .number({ invalid_type_error: "Must be a number" })
+    .min(MAP_MAX_BOUNDS[0][0], { message: "Longitude is invalid" })
+    .max(MAP_MAX_BOUNDS[1][0], { message: "Longitude is invalid" }),
```
roseeichelmann, line 44: i wonderrr if this message would be more helpful if it said "Longitude out of bounds"

roseeichelmann (approved):
looks great john! :ship:

editor/components/CrashMap.tsx
johnclary: much better—i was really struggling to come up with a helpful message 🙏

johnclary:
> I'm floored John. For this feature, this is just a whole cut above everything that came before it. To me, this is a no-expenses-spared implementation of this component.

thank you, @frankhereford! i really appreciate your input nudging me to level this up a bit 🙏 

i just tried to search for Stacy but it seems like our nearmap imagery is down... i'll look into that momentarily :/

frankhereford:
@johnclary wrote:

> i just tried to search for Stacy but it seems like our nearmap imagery is down... i'll look into that momentarily :/

Good luck with that near map stuff -- if you end up reaching out to CTM, I'd be interested in what that looks like. I believe the interaction with them that got us that API key was before my time, and is a bit of a black box to me.

I was joking about actually seeing stacy, but I did enjoy seeing TSD on there. thanks again.

PR #1701, opened 2025-02-06
Prevent buttons from stacking and add cross street to title

chiaberry:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/20699
* https://github.com/cityofaustin/atd-data-tech/issues/20982

## Testing

**URL to test:** 

https://deploy-preview-1701--atd-vze-staging.netlify.app/editor/crashes/20338968

**Steps to test:**

Make your window smaller and open to edit a data card. Observe the buttons not stacking. 

Also confirm you see cross streets in the crash titles. 

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/components/CrashHeader.tsx
+++ b/editor/components/CrashHeader.tsx
@@ -12,7 +12,9 @@ export default function CrashHeader({ crash }: CrashHeaderProps) {
   return (
     <div className="d-flex justify-content-between mb-3">
       <span className="fs-3 fw-bold text-uppercase">
-        {crash.address_primary}
+        {`${crash.address_primary ? crash.address_primary : ""} ${
+          crash.address_secondary ? "& " + crash.address_secondary : ""
+        }`}
       </span>
       {crash.crash_injury_metrics_view && (
         <CrashInjuryIndicators injuries={crash.crash_injury_metrics_view} />
--- a/editor/components/DataCardInput.tsx
+++ b/editor/components/DataCardInput.tsx
@@ -105,7 +105,7 @@ const DataCardInput = ({
           </Form.Select>
         )}
       </div>
-      <div className="text-end">
+      <div className="text-end text-nowrap">
         <span className="me-2">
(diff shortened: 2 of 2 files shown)


editor/components/CrashHeader.tsx
```diff
@@ -12,7 +12,9 @@ export default function CrashHeader({ crash }: CrashHeaderProps) {
   return (
     <div className="d-flex justify-content-between mb-3">
       <span className="fs-3 fw-bold text-uppercase">
-        {crash.address_primary}
+        {`${crash.address_primary ? crash.address_primary : ""} ${
+          crash.address_secondary ? "& " + crash.address_secondary : ""
+        }`}
```
chiaberry, line 17: How likely is it that we wouldnt have an address_primary? I took this code from the old editor, this is how we rendered the title. 

but if we have a secondary address and no primary we'd end up having something like
"& Congress Ave"

editor/components/DataCardInput.tsx
```diff
@@ -105,7 +105,7 @@ const DataCardInput = ({
           </Form.Select>
         )}
       </div>
-      <div className="text-end">
+      <div className="text-end text-nowrap">
```
chiaberry, line 108: so they dont stack up

editor/components/CrashHeader.tsx
johnclary: unlikely, but we have many cases like this

Screenshot 2025-02-06 at 6 30 00 PM

which happens because the `rpt_street_name` is `NOT REPORTED`. The missing piece is that we don't include `rpt_hwy_num` in our address concatenation. In the example above it is `290`. 

so long winded way of saying that we need to revisit how we construct the address at the db level. what you have looks good to me 🙌

johnclary (approved):
Looks very great to me! 🚢 🚢 🚢 🚢

mddilley (approved):
I tried my best, but I could not get the buttons to stack. 🙌 I see the secondary street name in the header too. 🚢 🚀

PR #1702, opened 2025-02-07
Editable Field Input validation

mateoclarke:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/20878

I've included a few examples of field validation we'll probably want to delete. Mainly just wanted to get this under your review sooner than later John and team. Let me know if you think this is the right direction and appropriate scope.

## Testing

**URL to test:** 
https://deploy-preview-1702--atd-vze-staging.netlify.app/editor/crashes/20339647

**Steps to test:**
1. Go to a [crash page](https://deploy-preview-1702--atd-vze-staging.netlify.app/editor/crashes/20339647).
2. Try to edit Primary Address - Street Number. Try different values and note the validation warnings.
    - It should not be nullable because its been configured as required
    - It should only be a number.
3. Try to edit a note.
    - It should not be nullable because its been configured as required
    - It must be longer than 3 characters

I tried to keep this a simple abstraction but I'm super open to feedback on how to tighten this up for primetime.

---

#### Ship list

- [ ] Code reviewed


editor/components/EditableField.tsx
```diff
@@ -31,60 +37,91 @@ interface DataCardInputProps {
...
+interface FormValues {
+  value: string;
 }
 
 /**
  * Component that manages the form UI for a single
  * editable field
  */
-const DataCardInput = ({
+const EditableField = ({
```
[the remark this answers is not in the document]
roseeichelmann: this naming looks good to me!

editor/components/EditableField.tsx
```diff
@@ -1,10 +1,16 @@
...
 import Button from "react-bootstrap/Button";
 import Form from "react-bootstrap/Form";
 import { InputType } from "@/types/types";
 import { LookupTableOption } from "@/types/relationships";
 
-interface DataCardInputProps {
+interface ValidationRules {
+  required?: boolean;
+  pattern?: RegExp;
+  custom?: (value: string) => boolean | string;
```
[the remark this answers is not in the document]
roseeichelmann: just to add tho maybe we could make the comments single line like
`/** comment here /*`
so they dont take up quite as much room

editor/utils/formHelpers.ts
```diff
@@ -187,3 +187,16 @@ export const handleFormValueOutput = (
...
+/**
+ * Generates validation rules for EditableField based on column definition
+ */
+export const getValidationRules = (col: ColDataCardDef<any>) => {
+  return {
+    required: col.required,
+    ...(col.inputType === "number" && {
+      pattern: /^\d*\.?\d*$/,
+    }),
+    ...(col.validation && { custom: col.validation }),
```
roseeichelmann, line 200: it feels like validation rules are being created in two diff places - here, and then again as `validationRules` in the `editableField` component. I wonder if it can be simplified so that all the logic and params needed to be passed to react hook form `register` are returned here in this function, so you could do away with creating `validationRules` again in `editableField`.

for example:
```
  return {
    required: col.required && "This field is required",
    ...(col.inputType === "number" && {
      pattern: {
        value: /^\d*\.?\d*$/,
        message: "Please enter a valid number"
      }
    }),
    ...(col.validation && { validate: col.validation }),
  };
```

i think the above would return everything that `register` expects. lmk if i can explain this better!

johnclary: along the lines of Rose's comments, and thinking about organizing all these types of rules, where we have `required` and `pattern` and `validation`: what if we combined everything into an `inputOptions` property that mirrored the `RegisterOptions` type that [RFH uses](https://react-hook-form.com/docs/useform/register)?  You're already shaping an object into `RegisterOptions` down when you call `register()` on the input, so we might as well construct that option at the column-level.

quick example with the `rpt_block_num` field:

```typescript
// ColDataCardDef
interface ColDataCardDef {
  ...
  inputOptions?: RegisterOptions
 ...
}

// crashesColumns
{...
  rpt_block_num: {
      path: "rpt_block_num",
      label: "Street number",
      editable: true,
      inputType: "text",
      inputOptions: {
        required: true,
        validate: (value: string) => {
          const isNumber = /^\d+$/.test(value);
          if (!isNumber) {
            return "Street number must be a number";
          }
          return true;
        },
      },
    },
  ...
}
```

editor/configs/crashesColumns.ts
```diff
@@ -122,6 +122,14 @@ export const crashesColumns = {
     label: "Street number",
     editable: true,
     inputType: "text",
+    required: true,
+    validation: (value: string) => {
+      const isNumber = /^\d+$/.test(value);
+      if (!isNumber) {
+        return "Street number must be a number";
```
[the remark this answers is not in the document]
johnclary: 👍 the flexibility here is great. would be nice to centralize common validation rules somewhere, maybe in `formHelpers.ts`? To keep things generic, I think it would be ok to leave the field name out of the message 🤔

johnclary (changes requested):
@mateoclarke this is looking great. very cool to see react-hook-form doing it's thing. i suggested a way to consolidate the way the validations are making their way down to the form input. Let me know what you think 🙏

editor/utils/formHelpers.ts
roseeichelmann: love @johnclary 's idea!

johnclary:
no action needed, but just wanted to mention that i think this enhancement might open up the door for the transforms we need to do in https://github.com/cityofaustin/atd-data-tech/issues/20705

editor/utils/formHelpers.ts
```diff
@@ -187,3 +188,13 @@ export const handleFormValueOutput = (
...
+
+/**
+ * Generates validation rules for EditableField based on column definition
+ */
+export const commonValidations = {
+  isNumber: (value: string) => {
+    return /^\d+$/.test(value) || "This field must be a number";
+  },
+  required: "This field is required"
+};
```
chiaberry, line 200: ☝🏼

editor/components/DataCard.tsx
```diff
@@ -149,6 +150,7 @@ export default function DataCard<T extends Record<string, unknown>>({
                           inputType={col.inputType}
                           selectOptions={selectOptions}
                           isMutating={isMutating || isValidating}
+                          inputOptions={col.inputOptions as RegisterOptions<FormValues, "value">}
```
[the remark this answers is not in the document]
johnclary: yep, that's exactly right. it's a type "assertion" that tells compiler, "trust me, this is the type i say it is". 

it is something we want to avoid, and in this case i think Mateo just forgot to remove it, because it's no longer needed ✨

editor/configs/crashesColumns.ts
```diff
@@ -241,6 +246,13 @@ export const crashesColumns = {
     label: "Street name",
     editable: true,
     inputType: "text",
+    inputOptions: {
+      required: true,
+      validate: (value: string) => {
+        const isSesameStreet = value === "Sesame";
+        return isSesameStreet || "Street name must be Sesame";
```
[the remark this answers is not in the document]
roseeichelmann: awesome thanks for showing an example of how this works!

editor/types/types.ts
```diff
@@ -12,7 +13,7 @@ export type InputType = "text" | "number" | "yes_no" | "select" | "textarea";
  * These properties control how a column is rendered and edited
  * in various UI components
  */
-export interface ColDataCardDef<T extends Record<string, unknown>> {
+export interface ColDataCardDef<T extends Record<string, unknown>, V = any> {
```
roseeichelmann, line 16: what does this `V = any` do here?

editor/utils/formHelpers.ts
```diff
@@ -187,3 +188,13 @@ export const handleFormValueOutput = (
...
 };
+
+/**
+ * Generates validation rules for EditableField based on column definition
+ */
+export const commonValidations = {
+  isNumber: (value: string) => {
+    return /^\d+$/.test(value) || "This field must be a number";
+  },
+  required: "This field is required"
```
roseeichelmann, line 199: is this required string being put into action anywhere?

editor/configs/crashesColumns.ts
```diff
@@ -122,6 +123,10 @@ export const crashesColumns = {
     label: "Street number",
     editable: true,
     inputType: "text",
+    inputOptions: {
+      required: true,
+      validate: commonValidations.isNumber
```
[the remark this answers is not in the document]
johnclary: this is cool! 

if i remove the `required` setting, i cannot save this field when it's blank, i guess because `""` is not a number.  
i'm not sure what the best way to handle this is... maybe we have a separate `isNumberNullable` validation?

emptynumber

we might also want to specify that this is an integer in this case, since decimal values are not allowed:

Screenshot 2025-02-14 at 1 23 29 PM

editor/configs/crashesColumns.ts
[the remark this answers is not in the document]
johnclary: 🚀

editor/components/EditableField.tsx
```diff
@@ -31,61 +25,72 @@ interface DataCardInputProps {
...
-      onSubmit={async (e) => {
-        e.preventDefault();
-        await onSave(editValue);
-      }}
-    >
+    <Form onSubmit={handleSubmit(onSubmit)}>
       <div className="mb-2">
         {(inputType === "text" || inputType === "number") && (
           <Form.Control
+            {...register("value", inputOptions)}
```
johnclary, line 69: cool, i see how we're using the generic field name `"value"` here, and that matches the field name we established in the `FormValues` interface you created.

i got confused here briefly and wonder if a better name might be `"input_value"`?

editor/types/types.ts
johnclary, line 1: not saying we should do this in this PR, but i'd really like to rename the `ColDataCardDef` interface to `FieldDefinition`

editor/components/CrashNotesCard.tsx
```diff
@@ -37,6 +37,13 @@ const notesColumns: ColDataCardDef<CrashNote>[] = [
     label: "Note",
     editable: true,
     inputType: "textarea",
+    inputOptions: {
+      required: true,
```
johnclary, line 41: Would you add a message here, "Note is required" - ?

johnclary (changes requested):
great stuff, thanks @mateoclarke!

besides a few minor issues with the validation, this is pretty much ready for primetime. i guess you can remove the demo validations now, and i'll open a follow-up issue to do a sweep of all the editable inputs that could use validation.

editor/components/EditableField.tsx
```diff
@@ -1,25 +1,19 @@
...
 import { LookupTableOption } from "@/types/relationships";
 
-interface DataCardInputProps {
-  /**
-   * The initial value to populate the input
-   */
+
+
+interface EditableFieldProps {
+  /** The initial value to populate the input */
```
roseeichelmann, line 10: yay i think making the comments one line is great to save on space

editor/configs/crashesColumns.ts
```diff
@@ -1,6 +1,7 @@
 import { ColDataCardDef } from "@/types/types";
 import { Crash } from "@/types/crashes";
 import { formatDateTime } from "@/utils/formatters";
+import { commonValidations } from "@/utils/formHelpers";
```
roseeichelmann, line 4: this isnt being used in this file right?

roseeichelmann:
are we supposed to see the validation errors in this PR or is that gonna be in a future PR and this is just setting up the foundation for it? bc right now i can add numbers and null values to street number without any error messages

roseeichelmann (approved):
thanks for sticking with all the iterations of feedback, the code looks great!! excited to see the validations be put to use in a future PR, this will be a huge improvement :) :rocket:

editor/components/EditableField.tsx
johnclary: i like it!

editor/types/types.ts
```diff
@@ -70,3 +77,7 @@ export interface MutationVariables extends Variables {
     updated_by?: string;
   };
 }
+
+export interface FormValues {
+  value: string;
+}
```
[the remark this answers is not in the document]
johnclary: 🦾

editor/utils/formHelpers.ts
johnclary: looks like this file is missing a newline / linting

johnclary (commented):
🚢 LGTM! thanks for sticking with it.

I will repurpose https://github.com/cityofaustin/atd-data-tech/issues/20468 to add validations everywhere they are needed 🙏

johnclary (approved):

editor/utils/formHelpers.ts
chiaberry: thats what my pointer finger emoji was trying to point out

PR #1703, opened 2025-02-07
Update service definition name in deployment CI for flask API

frankhereford:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/20952

## Testing

😬 Tough one to test here. I think it's:

* Merge to main
* Observe the CI

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/.github/workflows/build_docker_images.yml
+++ b/.github/workflows/build_docker_images.yml
@@ -101,7 +101,7 @@ jobs:
             echo "ECS_SERVICE=arn:aws:ecs:us-east-1:295525487728:service/atd-vz-data-cluster-production/cr3-and-user-mgmt-api" >> $GITHUB_ENV
           else
             echo "ECS_CLUSTER=atd-vz-data-cluster-staging" >> $GITHUB_ENV
-            echo "ECS_SERVICE=arn:aws:ecs:us-east-1:295525487728:service/atd-vz-data-cluster-staging/cr3-and-user-mgmt-api" >> $GITHUB_ENV
+            echo "ECS_SERVICE=arn:aws:ecs:us-east-1:295525487728:service/atd-vz-data-cluster-staging/backend-flask-api-staging" >> $GITHUB_ENV
           fi
 
       - name: Restart ECS service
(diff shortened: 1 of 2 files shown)


.gitignore
frankhereford, line 1: I stuck this in here because my `node_modules` from the `app` folder in the `nextjs` branch leaves me with a jillion outstanding files when i checkout `main`.

.github/workflows/build_docker_images.yml
```diff
@@ -101,7 +101,7 @@ jobs:
             echo "ECS_SERVICE=arn:aws:ecs:us-east-1:295525487728:service/atd-vz-data-cluster-production/cr3-and-user-mgmt-api" >> $GITHUB_ENV
           else
             echo "ECS_CLUSTER=atd-vz-data-cluster-staging" >> $GITHUB_ENV
-            echo "ECS_SERVICE=arn:aws:ecs:us-east-1:295525487728:service/atd-vz-data-cluster-staging/cr3-and-user-mgmt-api" >> $GITHUB_ENV
+            echo "ECS_SERVICE=arn:aws:ecs:us-east-1:295525487728:service/atd-vz-data-cluster-staging/backend-flask-api-staging" >> $GITHUB_ENV
```
frankhereford, line 104: I know it seems like a lot of mucking about just to get this name, but underneath name, I've done a lot of experimentation and reorganization in ECS in particular with regards to the Service Connect capability that allows the API gateway to find the private service.

While I have a little bit more organization to do, I know what I want to do now and I am confident that this CI will remain functional. I'll follow up with a second PR for production in the very near future - Monday.

frankhereford: As a final thought, if/when we finally do the VPC reorganization, sooooo much of this will simplify. Imagine that we have a `production` vpc and a `staging` vpc. The "staging" designation goes on the VPC at the top of the "hierarchy of aws components," and we don't have to "repeat ourselves" as it were putting "-staging" and "-production" on all these pairs of resources that content for ports and names, etc. 

Those will be salad days!

PR #1704, opened 2025-02-07
Adds geocoder control to crash map

johnclary:
## Associated issues

Counting this as a Discovery day project 🧑‍🚀 

* https://github.com/cityofaustin/atd-data-tech/issues/15175

## Testing

**URL to test:** Local, because we've branched off a branch

On the crash details page, edit the crash location and use the geocoder tool. Search for an address and notice that when you click on a search result, the map navigates to the result and the map marker is centered on the address. Save your new map coordinates.

Notice how the search results are limited to Austin and its enviorns.

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/components/MapGeocoderControl.tsx
+++ b/editor/components/MapGeocoderControl.tsx
@@ -1,0 +1,124 @@
+import { useControl, ControlPosition } from "react-map-gl";
+import MapboxGeocoder, { GeocoderOptions } from "@mapbox/mapbox-gl-geocoder";
+import "@mapbox/mapbox-gl-geocoder/dist/mapbox-gl-geocoder.css";
+import { LatLon } from "@/components/CrashMap";
+
+const MAPBOX_TOKEN = process.env.NEXT_PUBLIC_MAPBOX_TOKEN;
+
+type GeocoderControlProps = Omit<
+  GeocoderOptions,
+  "accessToken" | "mapboxgl" | "marker"
+> & {
+  position: ControlPosition;
+  onResult?: (latLon: LatLon) => void;
+};
+
+const ATX_BOUNDING_BOX: [number, number, number, number] = [
+  -98.182, 29.987, -97.304, 30.663,
+];
+
+/**
+ * Geocode control
+ * See: https://github.com/visgl/react-map-gl/tree/7.0-release/examples/geocoder
+ */
+export default function MapGeocoderControl(props: GeocoderControlProps) {
+  const geocoder = useControl<any>( // eslint-disable-line @typescript-eslint/no-explicit-any
+    () => {
+      const ctrl = new MapboxGeocoder({
+        ...props,
+        marker: false,
+        accessToken: MAPBOX_TOKEN || "",
+        bbox: ATX_BOUNDING_BOX,
+        flyTo: { speed: 2.5 },
+      });
+      ctrl.on("result", (evt) => {
+        const { result } = evt;
+        const location =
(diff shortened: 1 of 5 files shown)


editor/components/MapGeocoderControl.tsx
```diff
@@ -0,0 +1,124 @@
...
+const ATX_BOUNDING_BOX: [number, number, number, number] = [
+  -98.182, 29.987, -97.304, 30.663,
+];
+
+/**
+ * Geocode control
+ * See: https://github.com/visgl/react-map-gl/tree/7.0-release/examples/geocoder
+ */
+export default function MapGeocoderControl(props: GeocoderControlProps) {
+  const geocoder = useControl<any>( // eslint-disable-line @typescript-eslint/no-explicit-any
```
johnclary, line 25: The typing on this needed some work after I copypasta'd the example. I couldn't get this bit to work and gave up. I think there is really low risk here given that this control clearly works...?

editor/components/MapGeocoderControl.tsx
johnclary, line 1: copy/pasta'd wall of code is not my favorite way of bringing in new features, but fwiw this is exactly what we've done in Moped, and I don't think we need to overthink this very minor piece of UI functionality.

editor/components/MapGeocoderControl.tsx
```diff
@@ -0,0 +1,124 @@
...
+    // }
+    // if (geocoder.getFuzzyMatch() !== props.fuzzyMatch && props.fuzzyMatch !== undefined) {
+    //   geocoder.setFuzzyMatch(props.fuzzyMatch);
+    // }
+    // if (geocoder.getRouting() !== props.routing && props.routing !== undefined) {
+    //   geocoder.setRouting(props.routing);
+    // }
+    // if (geocoder.getWorldview() !== props.worldview && props.worldview !== undefined) {
+    //   geocoder.setWorldview(props.worldview);
+    // }
```
chiaberry, line 121: are these commented out because they aren't part of the official type so we cant use them?

chiaberry (approved):
i zoomed around the map and did observe the results are confined to the austin area

editor/components/MapGeocoderControl.tsx
roseeichelmann: :spaghetti:

editor/components/MapGeocoderControl.tsx
johnclary: that's right—if we wanted to enable them i think we'd have to manually customize the type declaration

editor/components/MapGeocoderControl.tsx
roseeichelmann: ive been getting warnings about using the `any` type in a few places in our code. i think using it here is maybe an okay exception since it can be so hard to make types work w external libraries :shrug: 

curious though why doesnt `const geocoder = useControl<MapboxGeocoder>` work like in the copypasta'd example?

roseeichelmann (approved):
looks great, works beautifully! :rocket: 
it would be awesome to write an issue that would snap the map back to the lat/lon point when you click cancel after editing coordinates and moving the map around

editor/components/MapGeocoderControl.tsx
chiaberry: why keep them around?

johnclary: removed 🙏

PR #1708, opened 2025-02-09
Use materialized location crashes view

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/20999

This PR modifies the `location_crashes_view` by converting it to a `materialzed` view. It initially included two different options for refreshing this view. 

~1. Hasura cron tigger that executes a refresh command on an hourly basis. The upside of this approach is that there's not much code to maintain. The downside is that it will not be obvious to us if the refresh process fails, because failures do not trigger notifications of any kind.~

2. New Python ETL that can be deployed as an Airflow DAG. The upside here is that we can monitor this task like all our other tasks. The downside is that we have a bunch of new code and a whole new docker image for this—and i haven't yet updated our CI or created the Airflow DAG. <-- This is the option we're going with 👍 

## Testing

**URL to test:** Local

### Test the Python ETL

2. Navigate to `/etl/run_sql` and activate a python env that has `requests` installed.
3. You can set the sole env var at the same time as you run the script

```shell
HASURA_GRAPHQL_ENDPOINT=http://localhost:8084/v2/query python run_sql.py -c refresh_location_crashes
```
4. Cool—now that you have that working, use your local editor to navigate to location ID [`65243E7A83`](http://localhost:3002/editor/locations/65243E7A83). Take note of how many records are in the crashes table on this page. Now visit any of the location's crashes and modify its location so that it is no longer assigned to location ID `65243E7A83`. 
5. Navigate back to the location details page, then re-run the the python ETL. Refresh the page and noticed that the crash you edited is no longer included in the crashes table 👍 

### Test the Cron trigger

1. To test the cron trigger, you'll probably want to increase it's frequency to 1 minute. In the Hasura console, navigate to http://localhost:8084/console/events/cron/refresh_location_crashes_view/modify. Change the trigger's schedule to `* * * * *`, then scroll down to the bottom of the page and click **Update Cron trigger**
2. Repeat the same steps you followed to edit a crash's location and verify that the crashes table updates within 1 minute. You can inspect the [**Processed events**](http://localhost:8084/console/events/cron/refresh_location_crashes_view/processed) tab in the Hasura console to see the cron request and response.

---

#### Ship list

- [ ] Code reviewed


database/metadata/cron_triggers.yaml
```diff
@@ -1 +1,17 @@
-[]
+- name: refresh_location_crashes_view
+  webhook: http://localhost:8080/v2/query
+  schedule: '0 * * * *'
+  include_in_metadata: true
+  payload: {}
+  headers:
+    - name: x-hasura-admin-secret
+      value_from_env: CRON_TRIGGER_API_SECRET
```
johnclary, line 8: new env var `CRON_TRIGGER_API_SECRET` will need to be added to ECS configs and set to the same value as the admin secret. this value is actually ignored locally since we don't use an admin secret.

database/metadata/cron_triggers.yaml
johnclary, line 1: this [cron trigger](https://hasura.io/docs/2.0/scheduled-triggers/create-cron-trigger/) executes SQL against Hasura's [schema API](https://hasura.io/docs/2.0/api-reference/schema-api/run-sql/). 

This pattern of having Hasura call itself is the same we use in Moped to carry out the activity logging.

One consideration is that this will accumulate event invocation logs that should probably be purged. we could do that using this same pattern, or...create an ETL to do it? I can't recall how we're handing this in Moped...

database/metadata/cron_triggers.yaml
```diff
@@ -1 +1,17 @@
-[]
+- name: refresh_location_crashes_view
+  webhook: http://localhost:8080/v2/query
+  schedule: '0 * * * *'
```
johnclary, line 3: we could modify this so that it only runs during business hours. i think i'd prefer to have this run consistently so we don't have to try and remember what the table's schedule is 🤷

editor/app/locations/[location_id]/page.tsx
```diff
@@ -90,6 +92,14 @@ export default function LocationDetailsPage({
...
             <Card.Header>
               <Card.Title>Crashes</Card.Title>
+              <Card.Subtitle className="fw-light text-secondary">
+                <AlignedLabel>
+                  <FaCircleInfo className="me-2" />
+                  <span>
+                    The data in this table is refreshed on an hourly basis
+                  </span>
+                </AlignedLabel>
+              </Card.Subtitle>
```
Comment (johnclary) [inline editor/app/locations/[location_id]/page.tsx:102]: Screenshot 2025-02-09 at 12 08 54 PM

etl/run_sql/README.md
johnclary, line 1: This readme needs a bit more work, but I paused here to get feedback on this approach 👍

etl/run_sql/utils/commands.py
johnclary, line 1: we can add additional sql commands here, and they will be piped in as options in the CLI

etl/run_sql/run_sql.py
```diff
@@ -0,0 +1,54 @@
...
+
+import requests
+
+from utils.cli import get_cli_args
+from utils.commands import COMMANDS
+
+ENDPOINT = os.getenv("HASURA_GRAPHQL_ENDPOINT")
+ADMIN_SECRET = os.getenv("HASURA_GRAPHQL_ADMIN_SECRET")
+
+ENDPOINT="http://localhost:8084/v2/query"
```
frankhereford, line 15: This appears to overload this configuration such that it will only work locally with our non-standard port, correct?

etl/run_sql/README.md
```diff
@@ -0,0 +1,21 @@
+# Run SQL
+
+This tool enables us to run arbitrary SQL using the Hasura [schema API](https://hasura.io/docs/2.0/api-reference/schema-api/run-sql/). It is intended to be used to schedule
```
frankhereford, line 3: I think you may have an unintended sentence fragment here.

etl/run_sql/run_sql.py
```diff
@@ -0,0 +1,54 @@
...
+    """Run arbitrary SQL through the the Hasura schema API.
+
+    The ENDPOINT must provide the full path to the schema API, which
+    is currently `/v2/query`.
+
+    Args:
+        sql (Str): The sql to be executed, including a final semicolon.
+    """
+    if not ENDPOINT:
+        raise OSError("HASURA_GRAPHQL_ENDPOINT env var is required")
```
frankhereford, line 30: I think you may have two conflicting variables here.

frankhereford (changes requested):
I've identified a few things that I think you may want to look at it which you'll find in the comments.

I see now that this has an a docker image that is required, and is more than just a cron like entry in the `graphql-engine` platform. Given this minor, but non-zero additional complexity, and especially since this introduces a folder in our ETLs that breaks the pattern of being executed by airflow -- based on those things, I am more strongly in favor of not introducing this exception to our patterns. I think I'd like to see us move in the direction of using airflow here.

Additionally, airflow will give us some insight into errors if one occurs; I am not aware of anything we do to trap and store `graphql-engine` logs, at least for now.

johnclary:
@frankhereford thanks for your feedback. just to be clear, for the Hasura cron option does not require any of the code in the new `/etl/run_sql` directory. it is configured entirely through the Hasura metadata.

the new directory here is to hold Option #2, which is a python script that can be DAG-ified and run on a schedule.

but i hear your point about running this through our standard orchestration tool. so i will wok on the rest of the steps needed to make that happen 🙏

etl/run_sql/run_sql.py
johnclary: yep, that's right. i've just run into this situation before and found it a tad confounding to debug 😅

etl/run_sql/run_sql.py
johnclary: oops—thanks! leftover hardcoding from dev.

johnclary:
i have added Docker CI, and i will remove the Hasura cron component of this once Rose finishes here review.

I have also opened an Airflow PR, here: https://github.com/cityofaustin/atd-airflow/pull/269

roseeichelmann (approved):
tested, this works great! thanks for finding a quick solution to the lagging query times :racehorse:

chiaberry:
I'll come back and review when the cron trigger code is removed. I did test the airflow part of it, so i did test the python script part of it

chiaberry (approved):
tested via airflow!

PR #1716, opened 2025-02-12
Location details: notes

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/19978

## Testing

**URL to test:** 

**Steps to test:**

---

#### Ship list

- [ ] Code reviewed


editor/components/NotesCard.tsx
roseeichelmann, line 1: renamed this file from `CrashNotesCard`

editor/components/NotesModal.tsx
```diff
@@ -2,16 +2,15 @@ import Modal from "react-bootstrap/Modal";
...
 import { useMutation } from "@/utils/graphql";
 import { useAuth0 } from "@auth0/auth0-react";
 
 interface NotesModalProps {
   show: boolean;
-  onClose: () => void;
-  onSubmitCallback: (data: CrashNote) => void;
-  crashPk?: number;
+  handleCloseModal: () => void;
+  onSubmitCallback: () => void;
```
roseeichelmann, line 11: i think its okay that im updating this callback to take no arguments?

editor/components/NotesModal.tsx
```diff
@@ -31,26 +31,24 @@ export default function NotesModal({
...
       userEmail: user?.email,
     };
-    const responseData = await mutate<{
-      insert_crash_notes_one: { returning: CrashNote };
-    }>(noteData);
-    if (responseData && responseData.insert_crash_notes_one) {
-      onSubmitCallback(responseData.insert_crash_notes_one.returning);
+    const responseData = await mutate(noteData);
+    if (!!responseData) {
+      onSubmitCallback();
```
roseeichelmann, line 44: i realized this callback doesnt actually need arguments, so i was able to generalize this part of the function

editor/queries/crashNotes.ts
```diff
@@ -30,9 +30,9 @@ export const UPDATE_CRASH_NOTE = gql`
   }
 `;
 
-export const DELETE_CRASH_NOTE = gql`
-  mutation DeleteCrashNote($id: Int!) {
-    delete_crash_notes(where: { id: { _eq: $id } }) {
+export const SOFT_DELETE_CRASH_NOTE = gql`
+  mutation SoftDeleteCrashNote($id: Int!) {
+    update_crash_notes(where: { id: { _eq: $id } }, _set: {is_deleted: true}) {
```
roseeichelmann, line 35: bringing in soft-deletes in this PR :)

database/metadata/databases/default/tables/public_location_notes.yaml
```diff
@@ -18,8 +19,9 @@ insert_permissions:
       columns:
         - created_by
         - date
-        - text
+        - is_deleted
         - location_id
+        - text
         - updated_by
         - user_email
```
roseeichelmann, line 26: can we delete this column from the notes tables? @johnclary

roseeichelmann: update just found https://github.com/cityofaustin/atd-data-tech/issues/20753 and assigned it to myself, i will do this in a follow up PR :)

editor/queries/crashNotes.ts
```diff
@@ -2,12 +2,13 @@ import { gql } from "graphql-request";
 
 export const INSERT_CRASH_NOTE = gql`
   mutation InsertCrashNote(
-    $crashPk: Int!
+    $recordId: Int!
     $text: String!
     $userEmail: String!
+    $createdBy: String!
```
johnclary, line 8: i think this `createdBy` var can be removed. i am hitting a graphql error when i try to create a _crash_ note because it's missing.

editor/queries/crashNotes.ts
johnclary: hooray!!!!!!  thanks for doing this 😍 

this is more of a stylistic thing, but technically you could use the `UPDATE_CRASH_NOTE` mutation to do the same thing as the delete, since they are both update 🤔

johnclary: ah—one tricky thing here: the automatic `updated_by` setter only works if the mutation payload has a variable called `updates`. [here's](https://github.com/cityofaustin/vision-zero/blob/b36b06377b170e32a122f371bc587d2fc30e7a38/editor/utils/graphql.ts#L151-L158) that bit of code.

so when you delete a note, the `updated_by` value is not currently being changed. this is a bug carried over from my previous PR???

i really don't like the how hidden that whole functionality is. if we moved our user table into the db, we could set this automatically based on the jwt session token and not fuss with this at all 🤷

editor/app/locations/[location_id]/page.tsx
```diff
@@ -87,6 +99,20 @@ export default function LocationDetailsPage({
...
+            notesColumns={locationNotesColumns}
+            updateMutation={UPDATE_LOCATION_NOTE}
+            insertMutation={INSERT_LOCATION_NOTE}
+            deleteMutation={SOFT_DELETE_LOCATION_NOTE}
+            onSaveCallback={onSaveCallback}
+            recordId={location.location_id}
+            refetch={refetch}
+          />
+        </Col>
+      </Row>
```
Comment (johnclary) [inline editor/app/locations/[location_id]/page.tsx:115]: would you mind placing this below the crashes table? i know it's going to seem hidden, but i don't like that it pushes the crashes table below the fold, despite the fact that it's empty 99.9% of the time.

alternatively, we could make the location map half-width, and push the notes to be below the Details card?

editor/components/NotesModal.tsx
johnclary: seems like it!

johnclary (changes requested):
wow, really nice work abstracting the existing notes component! and thank youu for finally implementing soft deletes!

editor/components/DeleteNoteButton.tsx
```diff
@@ -4,25 +4,24 @@ import AlignedLabel from "@/components/AlignedLabel";
...
+  onSaveCallback,
+}: HeaderActionComponentProps<T>) {
+  const { mutate, loading: isMutating } = useMutation(mutation);
 
   return (
     <PermissionsRequired allowedRoles={allowedDeleteCrashNoteRoles}>
       <Button
         size="sm"
-        variant="secondary"
+        variant="outline-danger"
```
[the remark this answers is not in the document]
roseeichelmann: sounds good, i was just testing diff things :) i thought the outline was a nice in-between w/ having color and not having a whole bunch of bright red delete buttons that are overwhelming when they r all stacked. ill shelve this for later though!

editor/queries/crashNotes.ts
roseeichelmann: thanks i missed this!

editor/queries/crashNotes.ts
roseeichelmann: ahh thanks for pointing that out. i updated to remove the separate soft delete mutations and instead use `UPDATE_CRASH_NOTE`, and added that `updates` variable so that the setter will be used

editor/app/locations/[location_id]/page.tsx
Comment (roseeichelmann) [reply editor/app/locations/[location_id]/page.tsx]: i placed it below the crashes table, i tested moving it into the column under the Details card, but it looks weird as you add more notes and theres a big white space between the map card and the crashes table card

editor/components/DeleteNoteButton.tsx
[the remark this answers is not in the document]
johnclary: i hear ya and opened https://github.com/cityofaustin/atd-data-tech/issues/21176. feel free to add to it 🙏

editor/app/locations/[location_id]/page.tsx
Comment (johnclary) [reply editor/app/locations/[location_id]/page.tsx]: cool—thanks for that

johnclary (approved):
looks and works great! thanks, @roseeichelmann!

to answer your question:

> do we want to limit editing/deleting to people who actually created the note? I think thats how it worked in our old @johnclary

No, we do not need to restrict editing of notes based on author. It did work that way in the old VZE, but now that we're tracking `created_by` and `updated_by` it's a free for all 🤗 

🚢

database/metadata/databases/default/tables/public_location_notes.yaml
chiaberry: Thanks for linking to this issue, I was wondering user_email missing from my new notes was a bug

editor/components/NotesCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+export default function NotesCard<T extends Record<string, unknown>>({
+  notes,
+  notesColumns,
+  updateMutation,
+  insertMutation,
+  recordId,
+  onSaveCallback,
+}: NotesCardProps<T>) {
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  const [isValidating] = useState(false);
```
mddilley, line 48: might be missing something but this seems to be read-only. can `false` be passed directly to the prop below?

mddilley (approved):
Works so great! Cool to see the new notes component used in both places. And soft-deletes!! 🙌 🚀 🚢

editor/components/NotesCard.tsx
roseeichelmann: good catch! i didnt write that code, but it seems u are right :thinking:

PR #1718, opened 2025-02-19
Crash and location notes audit field follow-ups

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/20753

## Testing

**URL to test:** 
Local

**Steps to test:**
1. Apply migrations and metadata to your local db.
2. Test editing a note written by another user. In your hasura console make sure that your email is now in the `updated_by` col, and `created_by` remained the same. Make sure that `updated_at` was updated to the current time.
3. Now test deleting a note by another user and make sure `updated_by` and `updated_at` were set.
4. Also see the crufty columns are gone :heavy_check_mark: 
5. Test down migration

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/database/migrations/default/1739913612321_alter_notes_audit_fields/up.sql
+++ b/database/migrations/default/1739913612321_alter_notes_audit_fields/up.sql
@@ -1,4 +1,7 @@
 -- removing unused columns and setting audit fields to not nullable
+
+update crash_notes set created_by = updated_by where created_by is null;
+
 alter table crash_notes 
     drop column date,
     drop column user_email,
--- a/database/migrations/default/1739913612321_alter_notes_audit_fields/up.sql
+++ b/database/migrations/default/1739913612321_alter_notes_audit_fields/up.sql
@@ -16,3 +16,6 @@ alter table location_notes
     drop column user_email,
     alter column created_by set not null,
     alter column updated_by set not null;
+
+-- reset the session replication role, even though the session is about to end :)
+set session_replication_role = default;
--- a/database/migrations/default/1739913612321_alter_notes_audit_fields/up.sql
+++ b/database/migrations/default/1739913612321_alter_notes_audit_fields/up.sql
@@ -1,0 +1,12 @@
+-- removing unused columns and setting audit fields to not nullable
+alter table crash_notes 
+    drop column date,
+    drop column user_email,
+    alter column created_by set not null,
+    alter column updated_by set not null;
+    
+alter table location_notes 
+    drop column date,
+    drop column user_email,
+    alter column created_by set not null,
+    alter column updated_by set not null;
--- a/database/migrations/default/1739913612321_alter_notes_audit_fields/up.sql
+++ b/database/migrations/default/1739913612321_alter_notes_audit_fields/up.sql
@@ -1,3 +1,6 @@
+-- use replica mode to disable triggers
+set session_replication_role = replica;
+
 -- removing unused columns and setting audit fields to not nullable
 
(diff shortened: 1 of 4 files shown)


database/migrations/default/1739913612321_alter_notes_audit_fields/up.sql
```diff
@@ -0,0 +1,15 @@
+-- removing unused columns and setting audit fields to not nullable
+
+update crash_notes set created_by = updated_by where created_by is null;
```
roseeichelmann, line 3: needed to add this because with a copy of prod data from today, there are 5 crash notes that have `null` for `created_by`. all by xavier and with dates starting post release. we introduced a bug with the release (which is fixed in https://github.com/cityofaustin/vision-zero/pull/1716) that didnt add a `created_by` or a `user_email` when adding a note, just an `updated_by`. we shouldnt be losing any data when we run this statement bc it seems like these are all notes created by xavier and not updated by anyone else.

chiaberry (approved):
I tested editing others notes, deleting notes and it all works as expected. 
The backfill migration makes sense to me, but I dont know if it should be handled separately out of this migration, like a one time command when this is merged to prod. I am probably overthinking it

database/migrations/default/1739913612321_alter_notes_audit_fields/up.sql
johnclary: i'm really glad you're stitching this up! 🙌 

i think it'd be good to disable the `updated_at` trigger at the beginning of this migration, so that we preserve the `updated_at` timestamp of the last user update. 

you can see how we did that [here](https://github.com/cityofaustin/vision-zero/blob/2137e097940d79c218ff733dca2b57bb8ca94e54/database/migrations/default/1738088180607_notes_audit_fields/up.sql#L2) by setting the `session_replication_role` at the beginning of the migration.

johnclary (changes requested):
Everything looks good but for disabling that update trigger during the migration 🙏

database/migrations/default/1739913612321_alter_notes_audit_fields/up.sql
roseeichelmann: awesome thats a cool TIL, just pushed that up!

johnclary: one last thing—would you reset the session role to default at the end of the migration? 
https://github.com/cityofaustin/vision-zero/blob/2137e097940d79c218ff733dca2b57bb8ca94e54/database/migrations/default/1738088180607_notes_audit_fields/up.sql#L18C1-L19C40

roseeichelmann:
@johnclary this is ready for re review btw!

mddilley (approved):
Tested this, and all the steps check out. 😎 ✅ I also see the new not null constraints on the _by columns in my local DB. All looks good but +1 to [John's comment](https://github.com/cityofaustin/vision-zero/pull/1718#discussion_r1977630005) about unpausing the triggers at the end of the migration to get this ready for release.

database/migrations/default/1739913612321_alter_notes_audit_fields/up.sql
roseeichelmann: gahh i missed that and also somehow missed ur comment til mike pointed it out, thank u!

roseeichelmann:
@mddilley thanks for pointing out johns comment that i somehow missed!

PR #1719, opened 2025-02-20
Add keyboard shortcuts for navigating the crash details page

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/21080

## Testing

**URL to test:** 
Netlify

**Steps to test:**
1. Go to the crash details page. Test hitting Shift + A on your keyboard to scroll down to the address section
2. Shift + U to scroll to the Units
3. Shift + P to scroll to the People

Another way this could be implemented is maybe a sort of tabbing through the sections? So could go like Shift + ↓ or something like that and it like "tabs" thru the diff sections we have shortcuts to? The issue left a lot of room for interpretation so this is what i came up with

---

#### Ship list

- [ ] Code reviewed


editor/app/crashes/[record_locator]/page.tsx
```diff
@@ -124,7 +133,7 @@ export default function CrashDetailsPage({
           />
         </Col>
       </Row>
-      <Row>
+      <Row id="A">
```
Comment (roseeichelmann) [inline editor/app/crashes/[record_locator]/page.tsx:136]: is this a weird way to do this with the element id just being "A" "U" or "P"? :shrug:

editor/utils/shortcuts.ts
```diff
@@ -0,0 +1,30 @@
+import { useCallback, useLayoutEffect, useRef, useEffect } from "react";
+
+/**
+ * Custom hook for adding keyboard shortcuts
+ */
+export const useKeyboardShortcut = (
```
roseeichelmann, line 6: i created a custom hook that can be used elsewhere in the app we want to use keyboard shortcuts

editor/app/crashes/[record_locator]/page.tsx
```diff
@@ -34,6 +35,14 @@ export default function CrashDetailsPage({
 }) {
   const recordLocator = params.record_locator;
 
+  // Handles scrolling down to element on key press
+  const onKeyPress = (event: KeyboardEvent) => {
+    const element = document.getElementById(String(event.key));
+    element?.scrollIntoView({ behavior: "smooth" });
```
Comment (johnclary) [inline editor/app/crashes/[record_locator]/page.tsx:41]: i would typically prefer the `smooth` behavior, but because we're trying to enable super fast QA/QC, i think the default scroll behavior would be best. we can also use Xavier as a test subject...

johnclary (changes requested):
Rose, this is super cool. I love this new hook!

One little bug i ran into is that the keyboard shortcuts were triggering while i was writing a crash note :/

Unless we find a way to mute the triggers if the user is inputting text, maybe we need to add the meta key into the mix? curious what you think.

Thanks for working on this!

editor/utils/shortcuts.ts
```diff
@@ -0,0 +1,42 @@
...
+  const callbackRef = useRef(callback);
+  useLayoutEffect(() => {
+    callbackRef.current = callback;
+  });
+  const handleKeyPress = useCallback(
+    (event: KeyboardEvent) => {
+      // Type guard to access tagName property, see https://iifx.dev/en/articles/156175355
+      if (event.target instanceof Element) {
+        // Don't trigger callback if focused element is an input or textarea
+        if (event.target.tagName === "INPUT" || event.target.tagName === "TEXTAREA") {
```
roseeichelmann, line 25: this should cover all of our text input elements on the crash details page per my testing

editor/utils/shortcuts.ts
```diff
@@ -0,0 +1,42 @@
...
+  callback: (event: KeyboardEvent) => void
+) => {
+  const callbackRef = useRef(callback);
+  useLayoutEffect(() => {
+    callbackRef.current = callback;
+  });
+  const handleKeyPress = useCallback(
+    (event: KeyboardEvent) => {
+      // Type guard to access tagName property, see https://iifx.dev/en/articles/156175355
+      if (event.target instanceof Element) {
```
roseeichelmann, line 23: another typescript TIL, before doing this i couldnt access the `event.target.tagName` bc typescript couldnt be positive that `event.target` was of type `Element`  :roll_eyes: https://stackoverflow.com/questions/28900077/why-is-event-target-not-element-in-typescript

editor/app/crashes/[record_locator]/page.tsx
```diff
@@ -24,16 +24,36 @@ import { peopleRelatedRecordCols } from "@/configs/peopleRelatedRecordTable";
...
+  A: "address",
+  U: "units",
+  P: "people",
+};
+
+// Handles scrolling down to element on key press
+const onKeyPress = (event: KeyboardEvent) => {
+  const shortcutKey = event.key.toUpperCase();
+  const elementId =
+    shortcutKeyLookup[shortcutKey as keyof typeof shortcutKeyLookup];
```
Comment (roseeichelmann) [inline editor/app/crashes/[record_locator]/page.tsx:42]: this was a typescript TIL for me, just doing `const elementId = shortcutKeyLookup[shortcutKey]` was telling me "No index signature with a parameter of type 'string' was found on type"  https://bobbyhadz.com/blog/typescript-no-index-signature-with-parameter-of-type-string

editor/utils/shortcuts.ts
```diff
@@ -0,0 +1,42 @@
...
+        // Don't trigger callback if focused element is an input or textarea
+        if (event.target.tagName === "INPUT" || event.target.tagName === "TEXTAREA") {
+          return
+        }
+        // Check if key in shortcutKeys array and used in conjunction with shift key
+        if (shortcutKeys.some((key) => event.shiftKey === true && event.key.toUpperCase() === key.toUpperCase())) {
```
roseeichelmann, line 29: for future use cases elsewhere in app, we may want to do shortcut key combos that dont include the shiftKey in which case we could pass `shouldUseShift` or something as a param to this hook, but until then i think this is fine

editor/app/crashes/[record_locator]/page.tsx
Comment (johnclary) [reply editor/app/crashes/[record_locator]/page.tsx:42]: yep, these array accessors can be quite annoying in TS. although this works, it is still a type assertion that's shortcutting the type safety in the slightest way. it's totally tolerable here, because we know that `document.getElementById(elementId);` can handle when `elementId` is undefined (it returns `null`).

we _could_ add a proper type guard here to avoid the type assertion, but it would require a lot of code. 

TLDR, i like what you've got here, but maybe add a comment that acknowledges the type assertion?

editor/utils/shortcuts.ts
johnclary: TIL for me too—nice! i love how typescript leads you to learn more about the various types you don't usually need to think about. 

in contrast to that earlier TIL with the array accessor, this is a good ole fashion type guard that let's us narrow down the instance type we need to operate on 👍

editor/utils/shortcuts.ts
johnclary: agreed!

editor/utils/shortcuts.ts
johnclary: nice! big TIL for me here. this is great.

it might be good to add `SELECT` elements to this test as well. i know that Xavier uses the keyboard to navigate dropdown lists by pressing a letter to find the first option that starts with that letter. the `shift` key probably makes it a real edge case, but might as well be safe.

johnclary (approved):
the bug is gone. this is working so well! i made a few very small suggestions and will go ahead and approve now 🚢 🚢 🚢 🚢

editor/utils/shortcuts.ts
roseeichelmann: yeah its definitely such a great learning tool!

editor/app/crashes/[record_locator]/page.tsx
Comment (roseeichelmann) [reply editor/app/crashes/[record_locator]/page.tsx:44]: added a comment, thank you! yeah i have a hard time knowing when its okay/worth the tradeoff to shortcut type safety, glad you think this is alright and its a good thing to keep in mind

editor/utils/shortcuts.ts
roseeichelmann: great call, added that!

chiaberry (approved):
this is really cool! I am trying to keep up with the typescript learnings here, thank you for linking to some extra reading ❤️

mddilley (approved):
Awesome stuff! I thoroughly enjoyed testing out all the cool shortcuts. Xavier and all are going to zoom through the pages now! 🏎️

editor/utils/shortcuts.ts
```diff
@@ -0,0 +1,49 @@
...
+ * in order to fire the handleKeyPress callback.
+ */
+export const useKeyboardShortcut = (
+  // Array of keys that should trigger the callback when used alongside the shift key
+  shortcutKeys: string[],
+  // Callback that handles what happens when the keys are pressed
+  callback: (event: KeyboardEvent) => void
+) => {
+  const callbackRef = useRef(callback);
+  useLayoutEffect(() => {
```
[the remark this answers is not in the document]
roseeichelmann: i copied this pattern from elsewhere, but here are some articles describing why to use [this pattern](https://www.epicreact.dev/the-latest-ref-pattern-in-react) with [useLayoutEffect](https://kentcdodds.com/blog/useeffect-vs-uselayouteffect).

basically i think we are wanting to make sure that the ref is updated before the other code runs? fwiw when i update to useeffect i dont see any changes on my end :shrug:  but i kept finding this pattern everywhere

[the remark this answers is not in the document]
mddilley: thanks for the link, and that pattern is a TIL for me. Also made me take notice of the dependency array being left out intentionally. 🙏  very cool to know because we ran into issues with trying to keep refs up to date before other code ran in Moped iirc. Thanks!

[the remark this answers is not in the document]
johnclary: i totally looked over this—it's a TIL for me too 💡 

are we sure we don't want `callback` in a dependency array? i feel like this might be ESLint failing us again 🤔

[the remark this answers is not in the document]
roseeichelmann: the first article i linked above would lead me to believe we dont need it in a dependency array?

"And because we don't need or want a re-render when we update the callback to the latest value, it means we also don't need to (and really shouldn't) include it in a dependency array for useEffect, useCallback, or in our case useMemo. This is an important point, so I want to dive into it a bit."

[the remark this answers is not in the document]
johnclary: it's a really interesting pattern. i'm still trying to wrap my head around when it makes sense to reach for a ref instead of a state var. 

i think the article is talking about some case where your `callback` function is dynamic, or encloses variables which are dynamic. in our case, the callback is static—we don't need to worry about it becoming stale. 

by leaving out the dependency array, that hook is running on every render of the page component, which ends up being quite a bit more than necessary. any edit saved on the page triggers the hook. you can see this by logging inside the hook with and without the array:

```javascript
  useLayoutEffect(() => {
    console.log("Hello i fired my effect!");
    callbackRef.current = callback;
  }, [callback]);
```

as i look at this some more, i actually think we can do away with this ref entirely for the same reason. we can safely add `callback` to the `useCallback` dependency array and we're all set.

PR #1720, opened 2025-02-24
Set hours on date filter to be end of day

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/21206

John did the [triaging](https://github.com/cityofaustin/atd-data-tech/issues/21206) and found that the date filter is using the current timestamp. I added a line in the filters onChange to set the hours as the end of the day. 

## Testing

**URL to test:** 
https://deploy-preview-1720--atd-vze-staging.netlify.app/editor/crashes

**Steps to test:**

This may be a quirky one to test, just because its involving time. But if you go into [staging](https://visionzero-staging.austinmobility.io/editor/crashes) and filter down to crashes between 8/16/24 and 8/16/24, depending on the time of day that you are testing, you may get 0 results or 1 result. Because the current behavior is the timestamp will just set to whatever time it is now, for example when I check the network request I get `# date_end \n crash_timestamp: { _lte: \"2024-08-16T17:16:36.000Z\" } } ` . 

Running the same query on the deploy preview, the query says ` # date_end \n crash_timestamp: { _lte: \"2024-08-17T04:59:59.999Z\" } } ` which is central daylight time 8/16/24 11:59pm in UTC time. 

-- edited to add:
Another easy way to see the difference is in staging to filter crashes after and before 8/13/24, I just tested now and got only 11 records in the results, but on the preview branch there are 25

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/components/TableDateSelector.tsx
+++ b/editor/components/TableDateSelector.tsx
@@ -99,6 +99,8 @@ export default function TableDateSelector({
             className="form-control"
             onChange={(date) => {
               if (date) {
+                // set the hours to the beginning of the date selected
+                date.setHours(0, 0, 0, 0);
                 setCustomDateRange({
                   start: date,
                   end: customDateRange.end,
--- a/editor/components/TableDateSelector.tsx
+++ b/editor/components/TableDateSelector.tsx
@@ -117,6 +117,8 @@ export default function TableDateSelector({
             className="form-control"
             onChange={(date) => {
               if (date) {
+                // set the hours to the end of the date selected
+                date.setHours(23, 59, 59, 999);
                 setCustomDateRange({
                   start: customDateRange.start,
                   end: date,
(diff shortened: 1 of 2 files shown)


roseeichelmann (approved):
:ship:  thanks to you and john for the quick fix!

editor/components/TableDateSelector.tsx
```diff
@@ -117,6 +117,8 @@ export default function TableDateSelector({
             className="form-control"
             onChange={(date) => {
               if (date) {
+                // set the hours to the end of the date selected
+                date.setHours(23, 59, 59, 999);
```
mddilley, line 121: Do we know why the current time isn't being used in the picker for the start of the range too? I was looking through the code, and I'm probably just missing something!

mddilley (approved):
I checked the graphql queries, and I see the bug in staging and the fix in the deploy preview. 🚢 🙌 📆

editor/components/TableDateSelector.tsx
chiaberry: I dont know and it confused me because I thought I was going to have to move it to start of the day. 
This [line](https://github.com/cityofaustin/vision-zero/blob/main/editor/utils/dates.ts#L25) sets the target start date to midnight but i dont think that date affects the date picker at all.

johnclary (approved):
Looks great. Thanks for fixing this!

PR #1686, opened 2025-01-31
Refines the navbar and adds a footer 

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/20911

screencapture-deploy-preview-1686-vision-zero-nextjs-netlify-app-crashes-2025-02-04-14_10_33

## Testing

**URL to test:** [Netlify](https://deploy-preview-1686--vision-zero-nextjs.netlify.app/)

How does it look?

---

#### Ship list

- [ ] Check migrations for any conflicts with latest migrations in `main` branch
- [ ] Confirm Hasura role permissions for necessary access
- [ ] Code reviewed
- [ ] Product manager approved

PR #1725, opened 2025-03-05
Add environment warning banner

mateoclarke:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/21079

Screenshot 2025-03-05 at 12 38 43 PM

Screenshot 2025-03-05 at 12 54 30 PM

## Testing

1. [Open the app](https://deploy-preview-1725--atd-vze-staging.netlify.app/editor/crashes) in the Netlify/staging environment and confirm the banner is **blue** and says **staging**.
2. Open the app locally and confirm the banner is **yellow** and says **local**.

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/components/SidebarLayout.tsx
+++ b/editor/components/SidebarLayout.tsx
@@ -90,7 +90,7 @@ export default function SidebarLayout({ children }: { children: ReactNode }) {
       <div className="d-flex flex-grow-1 overflow-hidden">
         {/* Sidebar */}
         <div
-          className={`bg-dark app-sidebar d-flex flex-column overflow-hidden app-sidebar-${
+          className={`bg-dark app-sidebar d-flex flex-column app-sidebar-${
             isCollapsed ? "collapsed" : "expanded"
           }`}
         >


editor/components/EnvironmentBanner.tsx
```diff
@@ -0,0 +1,32 @@
...
+        ) : (
+          <FaCircleInfo className="me-2" />
+        )}
+        <span>
+          This is a <b>{environment}</b> environment for testing purposes.
+        </span>
+      </AlignedLabel>
+    </Alert>
+  );
+}
```
frankhereford, line 32: If it's not too late to add a linefeed here at the end of this file for POSIX compliance -- but not at the expense of shipping.

frankhereford (approved):
Looks stellar - ship ship ship! Thank you Mateo!

editor/app/layout.tsx
```diff
@@ -33,6 +34,7 @@ export default function RootLayout({
     <html lang="en" style={{ fontSize: "14px" }}>
       <AuthProvider>
         <body>
+          <EnvironmentBanner />
```
mddilley, line 37: i noticed that the extra height of the banner is causing a second scroll in the sidebar. more of a product question for @johnclary. wanted to mention since we have a similar layout shift in Moped from env banners that has caused some confusion at times.

https://github.com/user-attachments/assets/348dcba6-a25e-49b2-a595-5bcfd8ea4f00

editor/components/EnvironmentBanner.tsx
mddilley: makes me wonder if this is an option in Prettier. I somehow have my code editor set up to add new lines at end of files so i could look into it.

editor/app/layout.tsx
johnclary: hmm, yeah the double scrollbar is blech. @mateoclarke i think we need to get the banner into the flex layout somehow. if it's helpful, i made [this visual](https://github.com/cityofaustin/vision-zero/pull/1686#discussion_r1941742376) the last time i reworked the containers.

editor/components/EnvironmentBanner.tsx
mddilley: i didn't find my setting, but Prettier says [this about it](https://prettier.io/docs/rationale.html#empty-lines) so maybe my format on save is just doing that by default.

johnclary (changes requested):
The new banner component looks great. Requesting changes to get the layout issue fixed 🙏

editor/components/EnvironmentBanner.tsx
roseeichelmann: yeah, my code editor is also set up to add new lines at the end of files so i dont ever have to think about it

roseeichelmann (approved):
this looks so good and it will be awesome to have this feature back :sparkles:  approving on the assumption that you will address john/mikes feedback on the scrolling, it would be great to fix that in this PR if possible!

johnclary (approved):
This looks great—thanks for reworking the layouts and making adding even more flexboxes 🚀 

I re-requested review from the other devs just because the code change was prettry substantial.

roseeichelmann (approved):
updates look great to me, love the look of the banners in dark mode :heart_eyes:

PR #1726, opened 2025-03-05
Update ESLint rules to enforce `react-hooks/exhaustive-deps`

mateoclarke:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/20607

## Testing

This needs to be tested locally. 

1. Uncomment the code in the `editor/components/TestComponent.tsx` file. It has commented `exhaustive-deps` errors by design. (I'll delete this file before merging)
2. Run `npx eslint . --fix` and you should at least see the `/editor/components/TestComponent.tsx` cause 2 warnings.
3. Run `npm run build` to replicate lint warnings/errors and compile errors similar to what happens with Netlify deploy preview checks.

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/next.config.mjs
+++ b/editor/next.config.mjs
@@ -8,6 +8,20 @@ const BASE_PATH = process.env.NEXT_PUBLIC_BASE_PATH || "";
 const nextConfig = {
   reactStrictMode: true,
   basePath: BASE_PATH,
+  eslint: {
+    dirs: [
+      "app",
+      "components",
+      "configs",
+      "contexts",
+      "utils",
+      "queries",
+      "schema",
+      "types",
+      "utils",
+    ],
+    ignoreDuringBuilds: false,
+  },
   async redirects() {
     return [
       // handle when client navigates to the root


editor/types/types.ts
```diff
@@ -71,9 +71,9 @@ export interface ColDataCardDef<T extends Record<string, unknown>> {
 }
 
 export interface MutationVariables extends Variables {
-  [key: string]: any;
+  [key: string]: unknown;
```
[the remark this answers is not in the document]
johnclary: excellent 👍 🙏

editor/utils/auth.ts
```diff
@@ -76,6 +76,7 @@ export const useToken = (): string | null => {
         const accessToken = await getAccessTokenSilently();
         setToken(accessToken);
       } catch (err) {
+        console.error("Error getting access token:", err);
```
johnclary, line 79: 🙏

editor/.eslintrc.json
```diff
@@ -1,3 +1,10 @@
 {
-  "extends": ["next/core-web-vitals", "next/typescript", "prettier"]
+  "extends": ["next/core-web-vitals", "next/typescript", "prettier"],
+  "plugins": ["react-hooks"],
+  "rules": {
+    "react-hooks/rules-of-hooks": "error",
+    "react-hooks/exhaustive-deps": "warn",
+    "no-unused-vars": "off",
+    "@typescript-eslint/no-unused-vars": ["warn"]
```
roseeichelmann, line 8: im getting a warning ha about this expecting two args, i guess you can remove the brackets around it

editor/utils/formHelpers.ts
```diff
@@ -28,9 +27,9 @@ export function getFromPath(
  * Convert a record value to a string so that it can be used as the initial value
  * in a form input
  */
-export const valueToString = (
+export const valueToString = <T extends Record<string, unknown>>(
   value: unknown,
-  col: ColDataCardDef<any>
+  col: ColDataCardDef<T>
```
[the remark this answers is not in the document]
johnclary: all of these improvements are great. i'd love for this `any`  error to surface from `npm run build`.

 i don't think i understand why some rules surface from `eslint .` and not `npm run build`. is it that we need to add the rules directly to our `eslistrc.json` in order for them to show up during builds?

editor/.eslintrc.json
johnclary, line 8: how do folks feel about setting `exhaustive-deps`  and `no-unused-vars` to `error` instead of `warn`? i'd like our rules to be as strict as possible 😇

johnclary (commented):
@mateoclarke this is awesome, thank you! i'm glad the code fixes weren't too severe, either 😅 

i'm ready to approve but think it would be nice to enforce the `no-explicit-any` error for builds if possible 🙏

editor/.eslintrc.json
mddilley: yep, that sounds like a great plan imo. Failed builds would be a great way to avoid React pitfalls and keep the TS game in here as strong as it is right now. 💪

mddilley (commented):
Looks amazing! Thanks for adding and addressing some of the suggested fixes. I agree with John that upgrading from `warn` to `error` would be a nice touch to enforce these rules with failed builds.

Screenshot 2025-03-19 at 12 55 56 PM

editor/components/TestComponent.tsx
```diff
@@ -0,0 +1,29 @@
...
+// export default function TestComponent({ externalValue }: TestProps) {
+//   const [count, setCount] = useState(0);
+
+//   // Missing dependency: externalValue should be in the deps array
+//   useEffect(() => {
+//     console.log("External value is:", externalValue);
+//   }, []);
+
+//   // Missing dependency: setCount should be in the deps array
+//   const incrementLater = useCallback(() => {
```
roseeichelmann, line 17: weirdly i am not getting the warning about missing setCount dependency here, are yall?

editor/next.config.mjs
```diff
@@ -8,6 +8,20 @@ const BASE_PATH = process.env.NEXT_PUBLIC_BASE_PATH || "";
 const nextConfig = {
   reactStrictMode: true,
   basePath: BASE_PATH,
+  eslint: {
+    dirs: [
```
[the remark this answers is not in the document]
mddilley: `ignoreDuringBuilds` is a TIL for me! I wonder if removing the `dirs` altogether might cause it to scan everything by default? I am also assuming that there isn't anything that we would want to ignore.

mddilley (approved):
I also seeing the same errors when running the npm scripts to lint and to build. Awesome!! 🚀 🚢

chiaberry (approved):
thanks for bringing us closer to the best TS we can be

editor/next.config.mjs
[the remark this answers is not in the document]
johnclary: thanks for sorting this out. @mddilley [the docs](https://nextjs.org/docs/app/api-reference/config/eslint#linting-custom-directories-and-files) mention that next uses a default subset of directories if you don't explicitly specify them 🙃

editor/next.config.mjs
```diff
@@ -8,6 +8,20 @@ const BASE_PATH = process.env.NEXT_PUBLIC_BASE_PATH || "";
...
+      "app",
+      "components",
+      "configs",
+      "contexts",
+      "utils",
+      "queries",
+      "schema",
+      "types",
+      "utils",
```
johnclary, line 21: oops, `utils` is in here twice 🙈

johnclary (commented):
thanks for making our linting better 😄

PR #1717, opened 2025-02-14
Rework AFD and EMS imports

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/21123
* https://github.com/cityofaustin/atd-data-tech/issues/18693

This PR provides a new AFD + EMS import ETL with a few notable changes:

* it is designed to process files in the `atd-vision-zero` bucket and follows the `/inbox` + `/archive` pattern of the CRIS import ETL
* it uses the graphql API instead of a direct DB connection
* refactors the DB trigger functions to be more performant, and uses `BEFORE` instead of `AFTER` triggers

Related todo's:
* update AWS SES rules to route emails to `atd-vision-zero/<data-source>/inbox`
* open a follow-up issue to sort out why AFD incident reports [must go through](https://austininnovation.slack.com/archives/CM9MK950S/p1739892128298639) Frank's inbox rule
* Remove old S3 buckets: `atd-ems-incident-data` and `atd-afd-incident-data`
* Update Airflow DAG
* reprocess the Feb 14th EMS incident file, located [here](https://us-east-1.console.aws.amazon.com/s3/object/atd-ems-incident-data?region=us-east-1&bucketType=general&prefix=atd-ems/5hv840qkqlpund35c8k8tpk9aur1v4sk1d2tfq81), which contains a backfill of records dating to 2020.

## Testing

**URL to test:** Local

Follow the instructions in the readme to run the script locally. Remember to apply migrations and metadata.

Please test both the  `afd` and `ems` data sources. If you want to test the script without the `--skip-archive` directive, please be sure to move the email file back to its respective `/inbox`.

---
#### Ship list

- [ ] Check migrations for any conflicts with latest migrations in `main` branch
- [ ] Confirm Hasura role permissions for necessary access
- [ ] Code reviewed
- [ ] Product manager approved

PR #1731, opened 2025-03-07
Revisions to AFD + EMS incident updates

johnclary:
## Associated issues

This PR updates https://github.com/cityofaustin/vision-zero/pull/1717 by adding timezone awareness to a number of columns as well as dropping a few columns that are redundant and/or can easily be generated via SQL query.

## Testing

Follow the steps in https://github.com/cityofaustin/vision-zero/pull/1717.

---

#### Ship list

- [ ] Code reviewed


database/migrations/default/1739497581109_ems_trigger_updates/up.sql
johnclary, line 1: we could simplify both AFD and EMS ETLs by asking for changes to the source files we get from each department. that would be a nice follow-up, but i didn't want to deal with the back and forth of making edits to those files while also breaking our existing prod setup 🤷

chiaberry:
To fully test this PR, do you have to make sure to run the down migrations in PR 1717 and then run the migrations here (that is assuming you already tested 1717)? Because will hasura think these migrations have already been run, since they are edits to previous migrations?

johnclary:
@chiaberry it would best to to start from a fresh replication. i did not test running the old down migrations first, but should work in theory.

chiaberry (approved):
on a fresh db replication, followed steps in readme and afd resulted in `INFO:root:105,530 total records to upsert`, ems `INFO:root:37,554 total records to upsert`  and `INFO:root:7,704 total records to upsert` with audit fields!  ✨

frankhereford (approved):
I had very similar results to Chia here, and the code looks good. 🚢🚢🚢 Thank you John!

AFD: `INFO:root:105,530 total records to upsert.`
EMS: `INFO:root:37,554 total records to upsert.` and then a second set: `INFO:root:7,704 total records to upsert.`

database/migrations/default/1739497581109_ems_trigger_updates/up.sql
```diff
@@ -17,24 +17,35 @@ where id in (
...
+-- create new incident_received_datetime column and drop old columns
+alter table ems__incidents add column incident_received_datetime timestamptz;
+update ems__incidents set
+    incident_received_datetime = (incident_date_received || ' ' || incident_time_received)::timestamp at time zone 'america/chicago';
+alter table ems__incidents drop column incident_date_received, drop column incident_time_received;
+
+-- add audit fields and updated_at trigger;
+alter table ems__incidents add column created_at timestamptz default now();
+alter table ems__incidents add column updated_at timestamptz default now();
+create trigger set_public_ems__incidents_updated_at before update on public.ems__incidents for each row execute function public.set_current_timestamp_updated_at();
```
roseeichelmann, line 46: yay!

roseeichelmann (approved):
got the same results as frank and chia! im having a hard time wrapping my head around whether the audit fields are working as expected, all of the `created_at` fields are set to todays date at varying times, is that bc all of the rows underwent an upsert and it happened in 1000 row chunks?

database/migrations/default/1739497581109_ems_trigger_updates/up.sql
johnclary: definitely! thank you 🙏

johnclary:
>  all of the created_at fields are set to todays date at varying times

I believe what we're seeing is a large chunk of records with the same `created_at` timestamp, which reflects the time at which you run the `add column` migration. Then when the ETL runs a small percentage of the records are either created or updated from the ETL. in my local test, 85% of the EMS incidents still had the same `created_at` after I ran the ETL.

So I think everything is looking as expected. Good to check that! 🙏

PR #1732, opened 2025-03-10
Ability to edit unit year and person type columns

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/21371

## Testing

**URL to test:** 

**Steps to test:**

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/configs/peopleColumns.tsx
+++ b/editor/configs/peopleColumns.tsx
@@ -45,6 +45,15 @@ export const ALL_PEOPLE_COLUMNS = {
   prsn_type: {
     path: "prsn_type.label",
     label: "Type",
+    editable: true,
+    inputType: "select",
+    relationship: {
+      tableSchema: "lookups",
+      tableName: "prsn_type",
+      foreignKey: "prsn_type_id",
+      idColumnName: "id",
+      labelColumnName: "label",
+    },
   },
   prsn_age: {
     path: "prsn_age",
--- a/editor/configs/unitColumns.ts
+++ b/editor/configs/unitColumns.ts
@@ -1,5 +1,6 @@
 import { ColDataCardDef } from "@/types/types";
 import { Unit } from "@/types/unit";
+import { commonValidations } from "@/utils/formHelpers";
 
 export const ALL_UNIT_COLUMNS = {
   unit_nbr: {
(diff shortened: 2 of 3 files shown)


editor/configs/unitColumns.ts
```diff
@@ -35,6 +36,11 @@ export const ALL_UNIT_COLUMNS = {
   veh_mod_year: {
     path: "veh_mod_year",
     label: "Year",
+    editable: true,
+    inputType: "number",
+    inputOptions: {
+      validate: commonValidations.isNumber,
```
roseeichelmann, line 42: like john mentioned in [this comment](https://github.com/cityofaustin/vision-zero/pull/1702#discussion_r1956563098), now you cant save this field when its blank. do we need to fix that in a separate issue before rolling this out? it seemed out of scope to address that in this PR

editor/configs/peopleColumns.tsx
roseeichelmann, line 1: should we implement the number validation for person age too?

chiaberry: I dont see why not -- I tried updating someones age to be "no" and it saves but then wont show the update. Might as well prevent that, right?

chiaberry (approved):
I was able to edit the columns, ✅
I left feedback on your questions, basically just more questions

editor/configs/unitColumns.ts
mddilley: yea I guess if someone made a mistake and wanted to go back to null? seems edge case but idk

i wonder if some day it would make sense to hook in zod and be able chain validations but that could be over the top. definitely feels beyond the scope here to me!

mddilley (approved):
Works perfectly! Nice to see an example of what code it takes to go from readable to editable. 😎 🚢  🚀

editor/configs/unitColumns.ts
johnclary: adding in zod is a cool idea! now that Mateo built RHF into our inputs it should be straightforward to use the Zod adapater, i think 🤩 

i agree that we should enable this field to be blank and that it's an edge case. i have an issue floating around to apply missing validations to a handful of fields. will add this to it.

* https://github.com/cityofaustin/atd-data-tech/issues/20468

what you have here is good for now 🚢

editor/configs/peopleColumns.tsx
johnclary: We can do this (soon!) here: https://github.com/cityofaustin/atd-data-tech/issues/20468

johnclary (approved):
LVGTM! 🚢 🚢 🚢 🚢 🚢 🚢

PR #1733, opened 2025-03-11
Update Fatality Review Board (FRB) partners and schema

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/21450

## Testing

**URL to test:** 
Local

**Steps to test:**
1. Run the up migrations and metadata
2. Make sure you can add, delete, and edit coordination partners in the FRB section
3. Extra points if you want to compare to the values in the issue to make sure they align :trophy: 
4. Test the down migration

---

#### Ship list

- [ ] Code reviewed


database/migrations/default/1741719786807_run_sql_migration/up.sql
```diff
@@ -0,0 +1,27 @@
...
+
+insert into lookups.coordination_partners (label)
+values
+('Homeless Strategy Office'),
+('AUS (Airport)'),
+('Project Delivery'),
+('Sidewalks and Urban Trails Division'),
+('Office of the City Engineer');
+
+delete from lookups.coordination_partners where label = 'Public Works Department';
```
chiaberry, line 27: are there existing records in prod that have the Public Works Department coordination partner? Is updating the partner to TPW an option?

chiaberry (commented):
I was testing the up and down migrations, and then when i went back up I noticed that the ids for the inserted records incremented. One time I ran the up migration and I got ids 21-25 for the new records, the down migration made Public Works have id 26 and the next up migration went 27-31.

johnclary (commented):
Rose, this looks great! it is very nice to see this table get moved to it's proper home 🙌 

 i'm curious what you thinking about removing the auto-incrementing `id` column 👀

database/migrations/default/1741719786807_run_sql_migration/up.sql
```diff
@@ -0,0 +1,27 @@
...
+update lookups.coordination_partners set label = 'North Area Engineering' where label = 'North Area Engineering Office';
+update lookups.coordination_partners set label = 'South Area Engineering' where label = 'South Area Engineering Office';
+
+insert into lookups.coordination_partners (label)
+values
+('Homeless Strategy Office'),
+('AUS (Airport)'),
+('Project Delivery'),
+('Sidewalks and Urban Trails Division'),
+('Office of the City Engineer');
```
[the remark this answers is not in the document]
roseeichelmann: yeah nixing the sequence and manually assigning sounds good to me! idk what i think about the soft deletes for lookups though, could you give an example of how that could help with cris schema changes?

database/migrations/default/1741719786807_update_coord_partners_lookup/up.sql
```diff
@@ -0,0 +1,29 @@
+-- Updating table name to coordination_partners and moving to lookups schema
+alter table if exists atd__coordination_partners_lkp rename to coordination_partners;
+alter table if exists coordination_partners set schema lookups;
+alter table if exists lookups.coordination_partners rename column coord_partner_desc to label;
+alter table if exists lookups.coordination_partners add column source text default 'vz' not null;
+
+alter index lookups.atd__coordination_partners_lkp_pkey rename to coordination_partners_pkey;
+
+-- Cannot undo dropping the sequence in the down migration
+drop sequence lookups.atd__coordination_partners_lkp_id_seq cascade;
```
roseeichelmann, line 10: i cant really undo this without things getting weird, i would have to create the sequence again in the down migration but instead have it start at 8 to account for the IDs that are already assigned. is it okay to leave it like this and not account for it in the down?

roseeichelmann:
@johnclary this is ready for re-review!

database/migrations/default/1741719786807_update_coord_partners_lookup/up.sql
```diff
@@ -0,0 +1,29 @@
...
+update lookups.coordination_partners set label = 'Central Area Engineering' where label = 'Central Area Engineering Office';
+update lookups.coordination_partners set label = 'Capital Delivery Services (Corridor Program)' where label = 'Corridor Program Office';
+update lookups.coordination_partners set label = 'North Area Engineering' where label = 'North Area Engineering Office';
+update lookups.coordination_partners set label = 'South Area Engineering' where label = 'South Area Engineering Office';
+
+insert into lookups.coordination_partners (label, id)
+values
+('Homeless Strategy Office', 15),
+('AUS (Airport)', 16),
+('Project Delivery', 17),
```
mddilley, line 25: Might be worth adding "Division" at the end of these since it was added to a couple other partners. Probably not a detail to dwell on since this is what was requested. 🤷

mddilley (approved):
Works great and much more readable! Another addition to the lookup schema as well 🙌 🚀 🚢

database/migrations/default/1741719786807_update_coord_partners_lookup/up.sql
johnclary: ah, thanks, Mike. @roseeichelmann would you add `Division` to `Project Delivery` ? The other's are correct as-is.

johnclary (approved):
Looks great! Would you mind adding "Division" to that one lookup entry?

🚢 🚢 🚢 🚢

PR #1735, opened 2025-03-14
Show all contributing factors in Units table

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/21404

## Testing

**URL to test:** 
Local bc of metadata

**Steps to test:**
1. Start your local DB and apply the metadata
2. Go to a couple crash details pages, observe the Contributing factors column, all factors should be rendered now each other their own new line prefixed with **Primary:** or **Possible:**. This field is not editable now
3. Check out crash id http://localhost:3002/editor/crashes/13871901 for an example of multiple diff contrib factors
4. If the contributing factor id = 0 aka "None" it shouldnt be rendered

---

#### Ship list

- [ ] Code reviewed


editor/configs/unitColumns.tsx
```diff
@@ -76,15 +76,46 @@ export const ALL_UNIT_COLUMNS = {
...
+        record.contrib_factr_p1,
+        record.contrib_factr_p2,
+      ];
+      return (
+        <>
+          {primaryContribFactors.map(
+            (factor) =>
+              !!factor?.label &&
+              factor.id !== 0 && (
+                <tr key={factor.label}>
```
johnclary, line 97: i'm seeing a console warning about `span` being a child of `tr`. maybe this could be a plain `div` ?

johnclary (changes requested):
this code looks great and it's very cool to see these additional factors in the UI! so much data used to be hidden.

i think the design that was provided to you in the mockup is a bit meh, but it gets the job done. 
Screenshot 2025-03-17 at 5 04 09 PM
Screenshot 2025-03-17 at 5 03 24 PM

i requested changes because of a little DOM issue.

editor/components/RelatedRecordTableRow.tsx
```diff
@@ -138,9 +138,9 @@ export default function RelatedRecordTableRow<
               }}
             >
               {!isEditingThisColumn && (
-                <p style={{ whiteSpace: "pre-wrap" }}>
+                <div className="mb-3" style={{ whiteSpace: "pre-wrap" }}>
```
roseeichelmann, line 141: updated this bc i was also getting console errors about how `<div>` cannot be a child of `<p>` because the contributing factors field returns a `<div>`

johnclary: ah—much better 🙏 

as i'm looking at this, i think we can do away with that bottom padding in the div. i prefer the more compact spacing and kinda wonder if the padding introduced by the paragraph element was unintentional.

w/ padding
Screenshot 2025-03-18 at 1 22 08 PM

w/o padding
Screenshot 2025-03-18 at 1 22 14 PM

johnclary (approved):
Console error is gone 🙏 

Remove that padding if you agree w/ me?

editor/components/RelatedRecordTableRow.tsx
roseeichelmann: agreed!

mddilley (approved):
This is a really nice change to see all factors in the unit rows. I tested out manually changing a factor to id = 0 ("None"), and I see that it is left out of the render as expected. 😎 🚢 🚀

PR #1738, opened 2025-03-19
Create migration to drop null recommendations_partners

mateoclarke:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/21639

## Testing
Local testing required.

**Steps to test:**
1. Run VZE stack with a fresh-ish db replica
2. Run this SQL script to see the count of recommendation_partners that are null in your DB. For me it was 102:
  ```sql
SELECT * FROM public.recommendations_partners WHERE partner_id IS NULL;
```
You can also run this SQL query to see a list of crash IDs in your DB instance that would be effected by the bug:
```sql
SELECT 
    rp.id, 
    rp.recommendation_id, 
    r.crash_pk,
    c.cris_crash_id
FROM 
    public.recommendations_partners rp
LEFT JOIN 
    public.recommendations r ON rp.recommendation_id = r.id
LEFT JOIN 
    public.crashes c ON r.crash_pk = c.id
WHERE 
    rp.partner_id IS NULL;
```
4. Perform migrations and metadata updates
5. Run the same SQL script to see the count go to 0
6. Check the crash with the bug referenced in the issue: http://localhost:3002/crashes/20325372 or one of the records from the second SQL command in Step 2 (ex: http://localhost:3002/crashes/20256496) and confirm the mysterious commas are 👻.
7. Test the down migration.

Screenshot 2025-03-19 at 2 49 16 PM

Screenshot 2025-03-19 at 2 53 07 PM

---

#### Ship list

- [ ] Code reviewed


johnclary (approved):
Looks very great to me. Thanks for making quick work of this! 🚢 🚢 🚢 🚢 🚢

database/migrations/default/1742417173127_fix_null_recommendation_partners/up.sql
```diff
@@ -0,0 +1,8 @@
+-- delete records where partner_id is null
+delete from public.recommendations_partners
+where partner_id is null;
```
frankhereford, line 3: These deletions are not restored in the down migrations; is that a concern?

database/migrations/default/1742417173127_fix_null_recommendation_partners/up.sql
```diff
@@ -0,0 +1,8 @@
...
+
+-- alter table to enforce not null constraints
+alter table public.recommendations_partners
+alter column recommendation_id set not null,
+alter column partner_id set not null;
```
frankhereford, line 8: Can you add some linefeeds at the end of these files to bring them into posix compliance? It doesn't break `hasura`'s handling of the files, but there are plenty of unix tools that expect it to be there. Not a huge deal, but it's helpful.

database/migrations/default/1742417173127_fix_null_recommendation_partners/down.sql
```diff
@@ -0,0 +1,4 @@
+-- remove not null constraints
+alter table public.recommendations_partners
+alter column recommendation_id drop not null,
+alter column partner_id drop not null;
```
chiaberry, line 4: it needs a lil breathing room aka an extra line

frankhereford (commented):
Thanks for this Mateo - esp on short notice.

database/migrations/default/1742417173127_fix_null_recommendation_partners/up.sql
johnclary: Thanks to you and @chiaberry for flagging—i'll push a newline since Mateo is ooo 👍

database/migrations/default/1742417173127_fix_null_recommendation_partners/up.sql
johnclary: no concern—thanks for checking.

frankhereford (approved):
shipshipship!

johnclary:
thanks, all!

roseeichelmann (commented):
this looks great, thanks mateo!

wanted to point out a bug i found while testing that is unrelated to this PR. if you edit a recommendation in a certain way the save button disables.

[simplescreenrecorder-2025-03-20_10.30.40.webm](https://github.com/user-attachments/assets/2e5484ba-73a9-4db5-9daf-f7e4a0fbf2ce)

It seems like it has to do with the state or refetching or something, bc it only happens when your edited state is the same as when you first open the page. So if you open a crash details page and there are no partners, you add one, and then save. And then you edit it again and remove that partner and save disables.

johnclary:
thanks @roseeichelmann — i will test that and open an issue 🙏

PR #1742, opened 2025-03-24
Update crash details page <title>

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/21537

At first i tried implementing this by creating a `layout.tsx` page in the `app/crashes/[record_locator]` folder and exporting metadata title from there like [this comment](https://stackoverflow.com/a/76935731) explains 

But since i need to use data from the client component crash details page for the title it made most sense to just use the `<title>` component in the details page for this. 

## Testing

**URL to test:** 
[Netlify](https://deploy-preview-1742--atd-vze-staging.netlify.app/editor/crashes)

**Steps to test:**
1. Go to a crash details page, see that the title in your tab is a combo of the {cris crash id} - {primary address & secondary address}
2. Edit the address somehow on the page, the title in your tab should update as well

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/app/crashes/[record_locator]/page.tsx
+++ b/editor/app/crashes/[record_locator]/page.tsx
@@ -32,6 +32,7 @@ import {
   scrollToElementOnKeyPress,
   useKeyboardShortcut,
 } from "@/utils/shortcuts";
+import { formatAddresses } from "@/utils/formatters";
 
 const typename = "crashes";
 
@@ -72,9 +73,7 @@ export default function CrashDetailsPage({
   // When data is loaded or updated this sets the title of the page inside the HTML head element
   useEffect(() => {
     if (!!data) {
-      document.title = `${data[0].cris_crash_id} - ${data[0].address_primary ? data[0].address_primary : ""} ${
-        data[0].address_secondary ? "& " + data[0].address_secondary : ""
-      }`;
+      document.title = `${data[0].cris_crash_id} - ${formatAddresses(data[0])}`;
     }
   }, [data]);
 
--- a/editor/app/crashes/[record_locator]/page.tsx
+++ b/editor/app/crashes/[record_locator]/page.tsx
@@ -83,6 +83,9 @@ export default function CrashDetailsPage({
 
   return (
     <>
+      <title>{`${crash.cris_crash_id} - ${crash.address_primary ? crash.address_primary : ""} ${
+        crash.address_secondary ? "& " + crash.address_secondary : ""
+      }`}</title>
       <CrashHeader crash={crash} />
       {
         // show alert if crash on private drive or outside of Austin full purpose
--- a/editor/app/crashes/[record_locator]/page.tsx
+++ b/editor/app/crashes/[record_locator]/page.tsx
@@ -1,6 +1,6 @@
 "use client";
 import { notFound } from "next/navigation";
-import { useCallback } from "react";
+import { useCallback, useEffect } from "react";
 import Col from "react-bootstrap/Col";
 import Row from "react-bootstrap/Row";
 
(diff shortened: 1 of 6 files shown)


roseeichelmann:
First build failed bc our new exhaustive-deps are working so well, pushed a change to fix those errors

Screenshot from 2025-03-24 15-45-23

editor/components/CrashRecommendationCard.tsx
```diff
@@ -124,14 +124,7 @@ export default function CrashRecommendationCard({
...
         };
-  }, [recommendation]);
-
-  /**
-   * Reset the form values after the recommendation is saved
-   */
-  useEffect(() => {
-    reset(defaultValues);
-  }, [defaultValues]);
+  }, [recommendation, crash_pk, user]);
```
johnclary, line 127: 😮 🙏

editor/app/crashes/[record_locator]/page.tsx
```diff
@@ -83,6 +83,9 @@ export default function CrashDetailsPage({
 
   return (
     <>
+      <title>{`${crash.cris_crash_id} - ${crash.address_primary ? crash.address_primary : ""} ${
+        crash.address_secondary ? "& " + crash.address_secondary : ""
+      }`}</title>
```
Comment (johnclary) [inline editor/app/crashes/[record_locator]/page.tsx:88]: This title looks great, but it looks like it's duplicating the `title` element in the DOM. Maybe using the [`next/head`](https://nextjs.org/docs/pages/api-reference/components/head)
component would replace instead of duplicate the title?

Screenshot 2025-03-25 at 10 14 02 AM

Comment (roseeichelmann) [reply editor/app/crashes/[record_locator]/page.tsx:88]: The next/head component [doesnt work](https://github.com/vercel/next.js/discussions/53529#discussioncomment-6628991) in the app router :thinking:

Comment (chiaberry) [reply editor/app/crashes/[record_locator]/page.tsx:88]: I know having multiple `title` elements isn't valid, but I was wondering what happens if a page has more than one title element, and so far it seems like it affects SEO and browsers could potentially randomly choose which title to use

Comment (johnclary) [reply editor/app/crashes/[record_locator]/page.tsx:88]: I only did a quick bit of searching on this problem, but it seems like the only way to deal with this is with `useEffect`. Here's a little component the bot suggested

```javascript
export default function DynamicTitle({ title }) {
  useEffect(() => {
    // Update the document title using the browser API
    if (title) {
      document.title = title;
    }
  }, [title]);

  // This component doesn't render anything visible
  return null;
}
```

Comment (roseeichelmann) [reply editor/app/crashes/[record_locator]/page.tsx]: ha this is what i found too and i just pushed up that change

editor/app/crashes/[record_locator]/page.tsx
```diff
@@ -69,6 +69,15 @@ export default function CrashDetailsPage({
...
   }, [refetch]);
 
+  // When data is loaded or updated this sets the title of the page inside the HTML head element
+  useEffect(() => {
+    if (!!data) {
+      document.title = `${data[0].cris_crash_id} - ${data[0].address_primary ? data[0].address_primary : ""} ${
+        data[0].address_secondary ? "& " + data[0].address_secondary : ""
+      }`;
+    }
+  }, [data]);
```
Comment (johnclary) [inline editor/app/crashes/[record_locator]/page.tsx:79]: looks great! since we're doing this address formulation in multiple places, it might be nice to move to a helper.

Comment (johnclary) [reply editor/app/crashes/[record_locator]/page.tsx:79]: Really, I'd like to fix this at the DB level with a generated field that spits out the concatenated address - out of scope for sure though

johnclary (approved):
This is a really nice upgrade, thank you!

chiaberry (approved):
🚢

editor/app/crashes/[record_locator]/page.tsx
Comment (johnclary) [reply editor/app/crashes/[record_locator]/page.tsx:79]: oh, i was referring to the address concat, but yeah maybe a component for the title setter makes sense too, eh?

Comment (roseeichelmann) [reply editor/app/crashes/[record_locator]/page.tsx:78]: cool i put the address concat helper in the `formatters.ts`, i think that makes sense to keep it there?

johnclary (approved):
Thanks, Rose! 🚢

PR #1744, opened 2025-03-25
Create database trigger(s) to link EMS incidents to CR3 crashes

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/21713

## Testing

**URL to test:** Local

Here is the [slide deck](https://docs.google.com/presentation/d/1rtlMla6fU_phe6JVkkJypygXfPBKSzjT8Ut0Hnr8Pzw/edit?usp=sharing) i shared about this trigger.

1. If you haven't replicated a fresh copy of prod lately, now would be a good time.
2. Start your local database, hasura, and editor
3. Apply migrations and metadata
4. Open crash [20636421](http://localhost:3002/editor/crashes/20636421) in your editor, and edit its location by incrementing the last decimal place of the latitude (such that the location _barely_ moves).
5. Now use this query to view `ems__incidents` which have been linked to this crash. You should see two EMS records (`55776` and `52910`). Notice that the `crash_match_status` is `matched_by_automation` 👍 

```sql
SELECT
    e.id as ems_id,
    c.cris_crash_id,
    e.incident_location_address ems_address,
    c.address_primary || ' / ' || c.address_secondary as crash_address,
    e.incident_received_datetime ems_timestamp,
    c.crash_timestamp,
    e.apd_incident_numbers as ems_apd_case_ids,
    c.case_id as crash_case_id,
    geometry as ems_position,
    e.crash_match_status,
    matched_crash_pks
FROM
    ems__incidents e
    left join crashes c on e.crash_pk = c.id
WHERE
     crash_pk is not null
```

6. Edit the same crash again, this time moving it's location to be well more than 1k meters from it's current location. Use the same query to verify there are no EMS records linked to this crash.

7. Edit the crash once more, moving it back to it's original location (lat: `30.2531291`, lon: `-97.82332062`). Use the same sql query to verify there are once again two EMS incidents linked to this crash.

8. Now execute this SQL to prevent the trigger from updating these records' `crash_pk` column

```sql
update ems__incidents set crash_match_status = 'matched_by_manual_qa' where crash_pk = 1351900;
```

9.  Edit the crash yet again,  moving it's location to be well more than 1k meters from it's current location. Use the same query to verify that the two EMS records are still linked to this crash.

10.  Now we will execute a bulk update of crashes that will trigger the matching function. Run this sql, which may take a minute or so to complete.

```sql
update crashes set latitude = latitude + .0000001 where crash_timestamp > '2024-01-01';
```

8. Now execute the `select` ☝️ query again. You should see ~4800 results.

9. Nice! Now, run a slightly modified version of the query to view EMS records that were linked to multiple crashes. You should see ~260 results. Notice that the `matched_crash_pks` is populated for these results.

```sql
SELECT
    e.id as ems_id,
    c.cris_crash_id,
    e.incident_location_address ems_address,
    c.address_primary || ' / ' || c.address_secondary as crash_address,
    e.incident_received_datetime ems_timestamp,
    c.crash_timestamp,
    e.apd_incident_numbers as ems_apd_case_ids,
    c.case_id as crash_case_id,
    geometry as ems_position,
    e.crash_match_status,
    matched_crash_pks
FROM
    ems__incidents e
    left join crashes c on e.crash_pk = c.id
WHERE
     crash_match_status = 'multiple_matches_by_automation';
```

10. Lastly, test the afd/ems import ETL by following the instructions in that readme. There are a few chonky files in the `/inbox` that can be processed. Please use the `--skip-archive` directive to leave the files in the `/inbox`.

```shell
 docker run -it --rm -v $PWD:/app --env-file .env atddocker/vz-incident-import:latest python incident_import.py ems --skip-archive
```

---

#### Ship list

- [ ] Code reviewed


database/migrations/default/1742408937781_crash_ems_link/up.sql
```diff
@@ -0,0 +1,391 @@
...
+add column matched_crash_pks integer [],
+add constraint matched_crash_pks_non_null check (
+    (crash_match_status = 'multiple_matches_by_automation')
+    or (matched_crash_pks is NULL)
+);
+comment on column ems__incidents.crash_match_status is 'The status of the crash record match';
+comment on column ems__incidents.matched_crash_pks is 'The IDs of multiple crashes that were found to match this record. Can only be populated when crash_match_status is ''multiple_matches_by_automation''';
+
+--
+-- add column to assign injury severity
```
johnclary, line 44: i created this column based on an Excel formula Xavier provided

database/migrations/default/1742408937781_crash_ems_link/up.sql
```diff
@@ -0,0 +1,391 @@
...
+            or (lower(pcr_cause_of_injury) like '%auto%' and lower(pcr_cause_of_injury) like '%ped%')
+            or (lower(pcr_patient_complaints) like '%auto%' and lower(pcr_patient_complaints) like '%ped%') then 'Pedestrian'
+        when
+            lower(pcr_cause_of_injury) like '%motorcy%'
+            or lower(pcr_patient_complaints) like '%mc %'
+            then 'Motorcycle'
+        when lower(mvc_form_vehicle_type) = 'motorized scooter' then 'E-Scooter'
+        else 'Motor Vehicle'
+    end
+) stored;
```
johnclary, line 141: i created this column based on an Excel formula Xavier provided

database/migrations/default/1742408937781_crash_ems_link/up.sql
```diff
@@ -0,0 +1,391 @@
...
+create or replace trigger crash_update_ems_match_trigger
+after update on crashes
+for each row when (
+    (new.latitude is distinct from old.latitude)
+    or (new.longitude is distinct from old.longitude)
+    or (new.crash_timestamp is distinct from old.crash_timestamp)
+) execute function update_crash_ems_match();
+
+
+--
```
johnclary, line 288: this trigger used to parse out the first to incident numbers and assign them to their own columns. thanks to our updated EMS query, it's really common for there to be more than two incident numbers, and their order is completely random. these columns no longer serve any useful purpose.

database/migrations/default/1742408937781_crash_ems_link/up.sql
```diff
@@ -0,0 +1,391 @@
...
+    when (old.latitude is distinct from new.latitude or old.longitude is distinct from new.longitude)
+    execute function ems_incidents_trigger();
+
+--
+-- drop these individual incident number columns. a record may contain dozens of apd_incident_numbers
+-- 
+alter table ems__incidents drop column if exists apd_incident_number_1;
+alter table ems__incidents drop column if exists apd_incident_number_2;
+
+--
```
johnclary, line 334: we had an unused `ems_id` column on the `people` table, but i think it's best to keep this on the `ems__incidents` just to keep things organized.

database/migrations/default/1742408937781_crash_ems_link/up.sql
```diff
@@ -0,0 +1,391 @@
...
+create or replace function update_crash_ems_match()
+returns trigger as $$
+DECLARE
+    matching_ems RECORD;
+    ems_record RECORD;
+    match_count INTEGER;
+    matched_crash_ids INTEGER[];
+    meters_threshold INTEGER := 1000;
+    time_threshold INTERVAL := '60 minutes';
```
johnclary, line 187: I'm still working with Xavier to tune these to their optimal settings. We'll ultimately probably settle on tighter thresholds.

database/migrations/default/1742408937781_crash_ems_link/up.sql
```diff
@@ -0,0 +1,391 @@
...
+            then 1
+        when
+            upper(right(pcr_transport_priority, 1)) = '3'
+            and lower(pcr_patient_acuity_initial) = 'emergent (yellow)'
+            then 1
+        when
+            upper(right(pcr_transport_priority, 1)) = '3'
+            and lower(pcr_patient_acuity_level) = 'high acuity'
+            then 1
+        -- what is this pcr_transport_priority?
```
chiaberry, line 79: is this wondering what is CHARLIE?

johnclary: it was wondering about `upper(left(pcr_transport_priority, 1)) = 'D'` — i think i could clean these up or harden them a bit...

database/migrations/default/1742417174000_crash_ems_link/up.sql
```diff
@@ -0,0 +1,391 @@
...
+add constraint ems__incidents_crashes_crash_pk_fkey foreign key (crash_pk)
+    references public.crashes (id) on update cascade on delete set null;
+
+comment on column ems__incidents.crash_pk is 'Crash ID matched to this record';
+
+--
+-- add people fk column
+--
+alter table ems__incidents
+    add column person_id integer unique deferrable initially deferred,
```
roseeichelmann, line 15: what is this `deferrable initially deferred` do here?

database/migrations/default/1742417174000_crash_ems_link/up.sql
```diff
@@ -0,0 +1,391 @@
...
+        'matched_by_manual_qa'
+    )
+),
+add column matched_crash_pks integer [],
+add constraint matched_crash_pks_non_null check (
+    (crash_match_status = 'multiple_matches_by_automation')
+    or (matched_crash_pks is NULL)
+);
+comment on column ems__incidents.crash_match_status is 'The status of the crash record match';
+comment on column ems__incidents.matched_crash_pks is 'The IDs of multiple crashes that were found to match this record. Can only be populated when crash_match_status is ''multiple_matches_by_automation''';
```
roseeichelmann, line 41: super helpful comments you are leaving on these columns :ok_hand:

database/migrations/default/1742417174000_crash_ems_link/up.sql
```diff
@@ -0,0 +1,391 @@
...
+            upper(right(pcr_transport_priority, 1)) = '3'
+            and lower(pcr_patient_acuity_final) = 'emergent (yellow)'
+            then 1
+        when
+            upper(right(pcr_transport_priority, 1)) = '3'
+            and lower(pcr_patient_acuity_initial) = 'critical (red)'
+            then 1
+        when
+            upper(right(pcr_transport_priority, 1)) = '3'
+            and lower(pcr_patient_acuity_initial) = 'emergent (yellow)'
```
roseeichelmann, line 73: is this one a repeat of what you have a few lines up?

database/migrations/default/1742417174000_crash_ems_link/up.sql
```diff
@@ -0,0 +1,391 @@
...
+            then 3
+        else 5
+    end
+) stored;
+
+alter table ems__incidents
+add constraint ems__incidents_patient_injry_sev_id_fk foreign key (
+    patient_injry_sev_id
+)
+references lookups.injry_sev on update restrict on delete restrict;
```
roseeichelmann, line 108: so if this lookup tables ids were ever updated this constraint would prevent that update?

database/migrations/default/1742417174000_crash_ems_link/up.sql
```diff
@@ -0,0 +1,391 @@
...
+            then 'E-Scooter'
+        when
+            lower(pcr_cause_of_injury) like '%bike%'
+            or lower(pcr_cause_of_injury) like '%bicy%'
+            or lower(pcr_patient_complaints) like '%bike%' or lower(pcr_patient_complaints) like '%bicy%' then 'Bicycle'
+        when
+            lower(pcr_cause_of_injury) like '%pede%'
+            or lower(pcr_patient_complaints) like '%pede%'
+            or (lower(pcr_cause_of_injury) like '%auto%' and lower(pcr_cause_of_injury) like '%ped%')
+            or (lower(pcr_patient_complaints) like '%auto%' and lower(pcr_patient_complaints) like '%ped%') then 'Pedestrian'
```
roseeichelmann, line 133: i think some records are getting flagged as `travel_mode = ped` when really they were inside the car. see `ems__incident.pcr_key = 1290311`. the complaint says "auto vs ped" but the record also says the `vehicle_type` is "Motor Vehicle" and that airbags were deployed.

database/migrations/default/1742408937781_crash_ems_link/up.sql
roseeichelmann: yeah, fwiw the meter threshold feels far too liberal for me. when i was testing moving the crash so that it would unlink from the ems records, i felt like i had to move them so far away to finally unlink. feels too easy that the wrong crashes would get linked to ems records as is

database/migrations/default/1742417174000_crash_ems_link/up.sql
```diff
@@ -0,0 +1,391 @@
...
+            e.id,
+            e.incident_received_datetime,
+            e.geometry
+        FROM 
+            ems__incidents e
+        WHERE
+            -- ignore manual qa matches: these should not be modified
+            e.crash_match_status != 'matched_by_manual_qa'
+            AND e.incident_received_datetime  >= (NEW.crash_timestamp - time_threshold)
+            AND e.incident_received_datetime  <= (NEW.crash_timestamp + time_threshold)
```
roseeichelmann, line 201: i figured time threshold means we only want to link crashes and ems records that are within 60 min of each other, but this is implemented to allow a 2 hour difference. as is this would allow an ems record with `incident_received_datetime` of 2pm to be linked to a `crash_timestamp` of 4pm. just double checking this is whats intended

chiaberry: how does it allow a crash thats 2 hours later / before to be linked?
incident received 2pm, crash_timestamp is 4pm
2pm isnt  >= 3pm, so even tho 2pm <= 5pm this wouldn't be linked, would it?

database/migrations/default/1742417174000_crash_ems_link/up.sql
```diff
@@ -0,0 +1,391 @@
...
+                    AND matching_ems.incident_received_datetime  >= (c.crash_timestamp - time_threshold)
+                    AND matching_ems.incident_received_datetime  <= (c.crash_timestamp + time_threshold)
+                    AND c.position IS NOT NULL
+                    AND ST_DWithin(matching_ems.geometry::geography, c.position::geography, meters_threshold)
+            ) INTO matched_crash_ids;
+            
+        -- Get the count from the array length (handling when array_length is null as 0)
+        SELECT COALESCE(array_length(matched_crash_ids, 1), 0) INTO match_count;
+        
+        IF match_count = 0 THEN
```
roseeichelmann, line 224: im having trouble wrapping my head around this whole loop part - how would the match count be 0 if we already know the EMS records we are looping thru can at least be linked to the crash that triggered this?

roseeichelmann: ah - i guess if the crash is a temp record is one example. but then why not move that logic further up into what crashes can even trigger this trigger function at all

database/migrations/default/1742417174000_crash_ems_link/up.sql
```diff
@@ -0,0 +1,391 @@
...
+            -- multiple matching crashes found - cannot assign crash_pk
+            UPDATE ems__incidents 
+            SET crash_match_status = 'multiple_matches_by_automation',
+                crash_pk = NULL,
+                matched_crash_pks = matched_crash_ids
+            WHERE id = matching_ems.id;
+        END IF;
+    END LOOP;
+
+    IF TG_OP = 'UPDATE' THEN
```
roseeichelmann, line 247: could you add a comment here explaining what this next chunk is doing?

database/migrations/default/1742417174000_crash_ems_link/up.sql
```diff
@@ -0,0 +1,391 @@
...
+
+--
+-- create update trigger
+--
+create or replace trigger crash_update_ems_match_trigger
+after update on crashes
+for each row when (
+    (new.latitude is distinct from old.latitude)
+    or (new.longitude is distinct from old.longitude)
+    or (new.crash_timestamp is distinct from old.crash_timestamp)
```
roseeichelmann, line 284: and only when the crash record in not temporary maybe?

database/migrations/default/1742417174000_crash_ems_link/up.sql
roseeichelmann: in the example i was trying to do in my head i somehow decided that
4pm >= (2pm - 60min = 1pm) is true
4pm <= (2pm + 60min = 3pm) is also true

after looking at that for a very  long time ive decide 4pm is indeed not less than 3pm :melting_face: 

so nvm this is totally working as expected!

roseeichelmann (approved):
most of my comments are me just trying to understand all the code, but the test steps worked out and this seems to be working as expected! such an awesome feature you are implementing :clap: i think my only real concern is my comment about the `travel_mode`

database/migrations/default/1742417174000_crash_ems_link/up.sql
johnclary: deferred constraints are not checked until the end of a transaction.  This makes it possible to, for example, swap the `person_id` values of two records without bumping into the unique constraint. I believe we have a similar setup related to `prsn_nbr` and `unit_nbr`.

database/migrations/default/1742417174000_crash_ems_link/up.sql
```diff
@@ -0,0 +1,391 @@
...
+comment on column ems__incidents.matched_crash_pks is 'The IDs of multiple crashes that were found to match this record. Can only be populated when crash_match_status is ''multiple_matches_by_automation''';
+
+--
+-- add column to assign injury severity
+--
+alter table ems__incidents
+add column patient_injry_sev_id integer generated always as (
+    case
+        when lower(pcr_provider_impression_primary) = 'death on scene' then 4
+        when pcr_outcome = 'DECEASED ON SCENE' then 4
```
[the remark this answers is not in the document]
johnclary: agreed, we should be consistent. i will clean this up 🙏

database/migrations/default/1742417174000_crash_ems_link/up.sql
johnclary: yes. although it seems very unlikely we would ever do that, i got a little paranoid that we would forget that this non-CRIS table is also dependent on the same lookup table as `people` 🤷

database/migrations/default/1742417174000_crash_ems_link/up.sql
johnclary: you're right, the only time `match_count` would = 0 would be if a record is a temp record is deleted. 

 was a little hesitant to remove it in the interest of future-proofing. e.g. if in the future we decide to allow, for example, temp records to be changed to permanent records, or we enable soft deleting of non-temp records. i guess if we made such a drastic change we would hopefully search our db schema for references to those columns?

database/migrations/default/1742417174000_crash_ems_link/up.sql
johnclary: quoting my other comment, i'm leaning toward handling temp and deleted records in the trigger—curious what opinion folks have on this:

> in the interest of future-proofing. e.g. if in the future we decide to allow, for example, temp records to be changed to permanent records, or we enable soft deleting of non-temp records. i guess if we made such a drastic change we would hopefully search our db schema for references to those columns?

database/migrations/default/1742417174000_crash_ems_link/up.sql
johnclary: thanks for flagging this. i noticed some more oddities as i was digging into this. will discuss with xavier. it seems like the EMS techs may be filling in the vehicle type even when the victim was a ped. anyway, here's what i sent Xavier:

> for the ems travel_mode column, i'm sending you the case logic i created from your excel formula. the travel_mode is the right most column. i am seeing a handful of cases where an airbag is shown as deployed but the mode is pedestrian. and of course there are a decent number where the mvc_form_vehicle_type is Automobile but we're assigning the mode as pedestrian. i understand we're probably just assuming the injured party was the ped, and if so i'd like to document that detail in our code.
>  
> you'll notice a bunch of these where the mvc_form_vehicle_type is Motorized Scooter but we're flagging as a pedestrian mode because pcr_cause_of_injury  is pedestrian - vehicle accident. i think we may want to change those to scooter?

chiaberry:
I am currently testing this PR, on step 7 I was thankful the original lat/lon for the crash was included until I inputted the coordinates and its not actually the original lat/lon. 30.23555464, -97.82459722 is what the crash had originally.

database/migrations/default/1742417174000_crash_ems_link/up.sql
```diff
@@ -0,0 +1,391 @@
...
+    patient_injry_sev_id
+);
+create index ems__incidents_crash_match_status_index on ems__incidents (
+    crash_match_status
+);
+create index crashes_case_id_index on crashes (case_id);
+create index crashes_is_temp_record_index on crashes (
+    is_temp_record
+);
+-- these indexes optimize our ST_DWithin operations, which require geography types and are the fastes way to check point-to-point proximity
```
chiaberry, line 168: little typo in fastest

database/migrations/default/1742417174000_crash_ems_link/up.sql
```diff
@@ -0,0 +1,409 @@
...
+            or upper(left(pcr_transport_priority, 1)) = 'D'
+            then 1
+        when
+            right(pcr_transport_priority, 1) = '3'
+            and lower(pcr_patient_acuity_initial) = 'lower acuity (green)'
+            then 2
+        when
+            upper(left(pcr_transport_priority, 1)) = 'minor injury'
+            or upper(left(pcr_transport_priority, 1)) = 'possible injury'
+            then 2
```
mddilley, line 91: i wanted to check in on my understand, but will this ever result in TRUE? since it is comparing a lowercase string of words to an uppercase single letter.

database/migrations/default/1742417174000_crash_ems_link/up.sql
mddilley: idk, but as an latecomer to this PR and someone who is often confused about time, time window does help me think about the total time, 60 min, instead of in terms of +/- 30 min. ⏳ 🪟

etl/afd_ems_import/utils/columns.py
```diff
@@ -6,51 +6,8 @@
...
-            "mvc_form_weather",
-            "pcr_additional_agencies",
-            "pcr_transport_priority",
-            "pcr_patient_acuity_initial",
-            "pcr_patient_acuity_final",
-            "unparsed_apd_incident_numbers",
-            "apd_incident_numbers",
-        ],
+        # we assume EMS records do not change upstream, per Lynn C. at EMS
+        "update_columns": [],
```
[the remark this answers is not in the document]
mddilley: this is a tough call, but i've also lost context on when the `updated_at` does change. With this change, what would cause it to be different than the initial value? or would it always be the same as `created_at`?

mddilley (changes requested):
The request change is tiny and is in the down migration. The local steps worked perfectly, and I was able to see the expected changes in the DB. Thanks for creating the slides, too, which helped me spin back on this one quickly. 🙏 

I did leave a couple questions, and the important one is about the case statement in the generated column for patient_injury_sev_id.

database/migrations/default/1742408937781_crash_ems_link/up.sql
johnclary: the thresholds we ultimately settled on are 1200 meters and a 30-min threshold (60-min window), which is even greedier! 

our thinking is that we will very likely adjust this after we-go live with the [UI updates](https://github.com/cityofaustin/atd-data-tech/issues/21852), which will make it easier for Xavier to start seeing how many false-positives we're creating, and how many "multiple matches" we're creating when the nearest crash was the obvious correct match.

the process of linking an EMS record to an actual _person_ record will be a manual process, so ultimately any data we use to inform official Vision Zero reporting will have go through manual QA.

etl/afd_ems_import/utils/columns.py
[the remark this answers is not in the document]
johnclary: the plan is to add a bit of EMS editing capability, whereby users can manually set `crash_pk` and `person_id`, so those would be triggering the audit fields. which reminds me that we need to go ahead and add `created_by` and `updated_by` to this table. i'll put that into a separate issue 🙏

database/migrations/default/1742417174000_crash_ems_link/up.sql
johnclary: you're totally right, thanks. i'm giving all of these cases a finally copy edit now 😵‍💫

database/migrations/default/1742408937781_crash_ems_link/up.sql
johnclary: done. ditto for travel mode 🙏

database/migrations/default/1742417174000_crash_ems_link/up.sql
johnclary: I ended up just adding a comment in the function args, because the dividing by a postgres `INTERVAL` type is not so simple.

This thread did remind me of a TIL from this work, which is that using `INTERVAL` for these selections is much faster than doing match on the epoch timestamps, which is how i first approached it:

```sql
and extract(epoch from e.incident_received_datetime) >= extract(epoch from NEW.crash_timestamp) - 1800
and extract(epoch from e.incident_received_datetime) <= extract(epoch from NEW.crash_timestamp) + 1800
```

PR #1757, opened 2025-04-04
Indicate that EMS features are in beta

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/21982

## Testing

**URL to test:** Netlify

1. Visit the `/ems` page and look at the updated card header
2. Find an ems record with a matched crash, then visit the crash details page and look at the EMS related record table card header

Does this look ok?
---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/app/ems/page.tsx
+++ b/editor/app/ems/page.tsx
@@ -1,21 +1,40 @@
 "use client";
+import Badge from "react-bootstrap/Badge";
 import Card from "react-bootstrap/Card";
+import AlignedLabel from "@/components/AlignedLabel";
 import { emsListViewColumns } from "@/configs/emsColumns";
 import TableWrapper from "@/components/TableWrapper";
 import { emsListViewQueryConfig } from "@/configs/emsListViewTable";
+import { FaCircleInfo } from "react-icons/fa6";
+
 const localStorageKey = "emsListQueryConfig";
 
 export default function EMS() {
   return (
-      <Card className="mt-3">
-        <Card.Header className="fs-3 fw-bold">EMS Patient care</Card.Header>
-        <Card.Body>
-          <TableWrapper
-            columns={emsListViewColumns}
-            initialQueryConfig={emsListViewQueryConfig}
-            localStorageKey={localStorageKey}
-          />
-        </Card.Body>
-      </Card>
+    <Card className="mt-3">
+      <Card.Header>
+        <div className="d-flex mb-2">
+          <span className="fs-3 fw-bold me-2">EMS Patient care</span>
+          <div className="align-self-center">
+            <Badge bg="info">Beta</Badge>
+          </div>
+        </div>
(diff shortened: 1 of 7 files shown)


editor/components/RelatedRecordTable.tsx
johnclary, line 1: This turned into a bit of a refactor n order to support arbitrary content at the top of the related record table card. 

* Rename the `title` prop to `header`, and enable to be a full-blown `ReactNode`
* Remove the `headerActionComponent` prop, since that can be accomplished with the `header`
* Add dedicated `noRowsMessage` prop, since we can no longer reliably infer that from the card `title`

I'm happy to go in a different direction to get accomplish this. It seemed to me like the best approach to avoid bloating the component props 🤔

mddilley: Extending to support custom headers and no rows messages seems totally reasonable to me! The prop names and comments make it clear how to make future updates to me.

editor/components/RelatedRecordTable.tsx
```diff
@@ -101,7 +103,7 @@ export default function RelatedRecordTable<T extends Record<string, unknown>>({
                   colSpan={columns.length + (rowActionComponent ? 1 : 0)}
                   className="text-center text-secondary"
                 >
-                  No {title.toLowerCase()} found
+                  {noRowsMessage ? noRowsMessage : "No records found"}
```
mddilley, line 106: cool to have a fallback here in case the prop is left out. 😎

mddilley (approved):
Very cool to give users more insight into the status of this on-going work. I tested on the deploy preview, and I see the message and badge in both the EMS table and crash view table.

editor/app/ems/page.tsx
```diff
@@ -1,21 +1,40 @@
...
+          <span className="fs-3 fw-bold me-2">EMS Patient care</span>
+          <div className="align-self-center">
+            <Badge bg="info">Beta</Badge>
+          </div>
+        </div>
+        <Card.Subtitle className="fw-light text-secondary">
+          <AlignedLabel>
+            <FaCircleInfo className="me-2" />
+            <span>
+              EMS is currently in beta. Data may be inaccurate or change significantly as we continue to refine the system.
```
chiaberry, line 26: did Joel /VZ come up with this message? how about "EMS record linking is currently is beta" or "EMS analysis is currently in beta" ?

chiaberry (approved):
looks good! 

will the package.json version be bumped only in the patch to prod PR?

editor/app/ems/page.tsx
johnclary: Nope, it was. I like the second one — will do 🙏

johnclary:
Thanks, y'all! I'll open a companion PR to patch prod

PR #1761, opened 2025-04-08
Non-CR3 data quality enhancements

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/21948
* https://github.com/cityofaustin/atd-data-tech/issues/22042

This PR addresses some data quality issues which are blocking https://github.com/cityofaustin/atd-data-tech/issues/21863. 

* Replaces the timezone-naive `date` and `hour` columns with `case_timestamp`, a proper `timestamptz` column.
* Updates the Non-CR3 import tool to use `case_timestamp` and deal with naive local timestamps in the import payload
* Adds soft-deletes to the non-cr3 crashes table, and a trigger which automatically soft-deletes non-CR3s if there is a CR3 crash with the same `case_id`

## Testing

**URL to test:** Local

2. from `/editor`: `npm install`
3. Go to the non-cr3 uploads page (`/upload-non-cr3`) and download the template. Try importing it and observe errors on lines 41-47

```
Row # | Field | Error
-- | -- | --
41 | case_timestamp | Case timestamp must be in format YYYY-MM-DD HH:MM:SS or YYYY-MM-DD H:MM:SS
42 | address | Address is missing or invalid
43 | longitude | Longitude is required and must be a number
44 | longitude | Longitude is less than the minimum bounds (-99.7404)
45 | latitude | Latitude is required and must be a number
46 | latitude | Latitude is greater than the maximum bounds (32.2747)
47 | longitude | Longitude is required and must be a number
47 | latitude | Latitude is required and must be a number
```

4. Delete or correct the problematic rows from the sheet and complete the import/upload process.

5. Let's make sure the timestamps were handled properly. Run this query and verify the results align with the corresponding row in the imported CSV

```sql
select case_id, case_timestamp, case_timestamp at time zone 'America/Chicago' as case_timestamp_local from atd_apd_blueform where case_id = 140000001;
```

6. Now let's verify that soft-deletes are working. The last two rows in the import sheet have case IDs which matching case IDs of CR3 crashes. As such, they will have been soft deleted. You can inspect the database for this:

```sql
select case_id, is_deleted from atd_apd_blueform where case_id = 150181115 or case_id = 232520707;
```

---

#### Ship list

- [ ] Code reviewed


database/migrations/default/1743441451777_noncr3_soft_del_dedupe/up.sql
```diff
@@ -0,0 +1,56 @@
...
+);
+
+create index crashes_case_id_idx on public.crashes (
+    case_id
+);
+
+--
+-- Trigger which soft-deletes atd_apd_blueform records if there is a cr3 crash
+-- with the same case_id
+--
```
johnclary, line 40: This trigger handles situations where dupe/redundant non-CR3s are imported after a matching CR3 crash already exists, but we don't have a mechanism to detect this problem when a CR3 crash is created/updated.

I was hesitant to set up yet another trigger on the crashes table. One workaround Xavier has is to import the non-CR3 crashes at a higher frequency than he currently does—e.g., monthly instead of quarterly—and to include a full year of records in the import. This is going to be necessary of him anyway, I think, as we look to associate non-cr3s with EMS records.

database/migrations/default/1744126149551_noncr3_drop_date_hour_cols/up.sql
```diff
@@ -0,0 +1,244 @@
...
+non_cr3_crash_counts as (
+    select
+        atd_apd_blueform.location_id,
+        count(atd_apd_blueform.location_id) as crash_count,
+        count(atd_apd_blueform.location_id) * 10000 as noncr3_comp_costs_total
+    from atd_apd_blueform
+    where
+        atd_apd_blueform.location_id is not null
+        and atd_apd_blueform = false
+        and atd_apd_blueform.case_timestamp > (now() - '5 years'::interval)
```
johnclary, line 64: here's the change to this view ☝️

database/migrations/default/1744126149551_noncr3_drop_date_hour_cols/up.sql
```diff
@@ -0,0 +1,244 @@
...
+        aab.est_comp_cost_crash_based,
+        ''::text as collsn_desc,
+        ''::text as movement_desc,
+        ''::text as travel_direction,
+        ''::text as veh_body_styl_desc,
+        ''::text as veh_unit_desc
+    from atd_apd_blueform as aab
+    where
+        aab.is_deleted = false
+        and aab.case_timestamp >= (now() - '5 years'::interval)::date
```
johnclary, line 236: and here we're using `case_timestamp` and excluding soft-deletes

.gitignore
```diff
@@ -75,8 +75,8 @@ build-iPhoneSimulator/
 *.log
 *.xml
 
-# allow the VZE Non-CR3 upload upload template
-!app/public/files/non_cr3_template.csv
+# allow the VZE Non-CR3 upload template
+!editor/public/files/non_cr3_template.csv
```
johnclary, line 79: oops, this file wasn't available in staging or prod! i never noticed the issue because i had the file locally

editor/types/nonCr3.ts
```diff
@@ -30,23 +64,30 @@ export const NonCr3UploadSchema = z.object({
...
+    z.number({
+        invalid_type_error: "Latitude is required and must be a number",
+      })
+      .min(ATX_BBOX.latitude.min, {
+        message: `Latitude is less than the minimum bounds (${ATX_BBOX.latitude.min})`,
+      })
+      .max(ATX_BBOX.latitude.max, {
+        message: `Latitude is greater than the maximum bounds (${ATX_BBOX.latitude.max})`,
+      })
+  ),
```
johnclary, line 90: prior to this change, we were not properly validating if latitude or longitude were null. `z.coerce` is not our friend.

database/migrations/default/1743441451777_noncr3_soft_del_dedupe/up.sql
```diff
@@ -0,0 +1,56 @@
...
+alter table ems__incidents add column atd_apd_blueform_case_id integer;
+alter table ems__incidents
+add constraint ems__incidents_atd_apd_blueform_case_id_fkey foreign key (
+    atd_apd_blueform_case_id
+)
+references public.atd_apd_blueform (
+    case_id
+) on update cascade on delete set null;
+
+comment on column ems__incidents.atd_apd_blueform_case_id is 'The non-CR3 case ID matched to this record';
```
johnclary, line 13: ah yeah, i left this in here as prep work for https://github.com/cityofaustin/atd-data-tech/issues/22042

database/migrations/default/1743441451777_noncr3_soft_del_dedupe/up.sql
chiaberry: if im understanding this correctly, we dont have a mechanism to catch when we have non cr3s that are then duplicated when a cr3 crash is created?

editor/types/nonCr3.ts
```diff
@@ -19,7 +20,40 @@ const ATX_BBOX = {
  * Validation schema for a non-CR3 record uploaded through the UI
  */
 export const NonCr3UploadSchema = z.object({
-  date: z.string().regex(/^\d{4}-\d{2}-\d{2}$/, "Date is missing or invalid"),
+  case_timestamp: z
+    .string()
+    .refine(
+      (val) => {
+        // expected format: YYYY-MM-DD H:MM:SS or YYYY-MM-DD HH:MM:SS
```
chiaberry, line 27: thanks for including this regex decipher

chiaberry (approved):
Test steps work out, thank you for the test sql 

Could be nice to add `npm install` to the test instructions

database/migrations/default/1743441451777_noncr3_soft_del_dedupe/up.sql
johnclary: correct. it's a concern to discuss with Xavier and also a consideration when configuring the DB trigger that will associate non-cr3s to EMS incidents 😵‍💫

johnclary:
> Dunno if you wanna change the timestamp

Ewww—yes, fixed 🙏

database/migrations/default/1744126149550_noncr3_soft_del_dedupe/up.sql
```diff
@@ -0,0 +1,56 @@
...
+-- Trigger which soft-deletes atd_apd_blueform records if there is a cr3 crash
+-- with the same case_id
+--
+create or replace function remove_dupe_non_cr3s()
+returns trigger as $$
+DECLARE
+    text_case_id text;
+BEGIN
+    text_case_id := NEW.case_id::text;
+    if exists (select 1 from crashes where case_id = text_case_id) then
```
roseeichelmann, line 47: afaik `case_id` on the `crashes` table is not unique, so how can we trust them to have 1:1 relationships with the non cr3 crashes?

database/migrations/default/1744126149551_noncr3_drop_date_hour_cols/up.sql
```diff
@@ -0,0 +1,244 @@
+--
+-- add a timestamptz column
+--
+alter table atd_apd_blueform add column case_timestamp timestamptz;
+comment on column atd_apd_blueform.case_timestamp is 'The timestamp at which the APD case occurred';
```
roseeichelmann, line 5: could maybe be helpful to add `created_at` and `created_by` columns to the blueform table?

roseeichelmann (approved):
test steps worked! i just have a question about using `case_id` and the suggestion on adding audit fields, but this looks good!

database/migrations/default/1744126149551_noncr3_drop_date_hour_cols/up.sql
johnclary: thanks for that suggestion: https://github.com/cityofaustin/vision-zero/pull/1761/commits/dea6d4d41a6a11142c72e2410533d6d0aa9de23b

database/migrations/default/1744126149550_noncr3_soft_del_dedupe/up.sql
roseeichelmann: i guess its been some time since ive had context on `crash.case_id`, i thought i remembered that there are sometimes diff crashes with the same `case_id` so they cant be relied on to identify a crash, so it seemed weird to me that we are using them to identify and soft delete a non cr3. but maybe im lost on what the intent is here. 

thats an interesting idea to add `investigat_agency_id` to the mix, i just ran a query to see if we ever have duplicate `crash.case_id`'s where the agencies are the same, and my query returned around 200 instances of crashes w the same case id and investigat agencies. some of them seem to be complete duplicates, while others seem like they are two distinct crashes. :thinking:

johnclary:
👋 re-requesting review after reworking this to incorporate timestamps and investigating agency ID into the soft-delete trigger logic. Thanks, Rose, for checking my assumptions on that!

frankhereford (approved):
John, this looks really great. Tested out perfectly and your implementation is masterful. Thank you! 🚢 🚢🚢

editor/public/files/non_cr3_template.csv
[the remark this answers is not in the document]
johnclary: Good question. I mostly copied this file from the old VZE, which did include this crufty data in the template. 

I am leaning toward keeping the template as-is, since it is handy for testing validations. But obviously we wouldn't want a user to import these rows in prod. What do y'all think?

database/migrations/default/1744126149551_noncr3_drop_date_hour_cols/up.sql
```diff
@@ -0,0 +1,346 @@
...
+    is_deleted
+);
+
+create index crashes_case_id_investigat_agency_id_idx on public.crashes (
+    case_id, investigat_agency_id
+);
+
+--
+-- Trigger which soft-deletes atd_apd_blueform records if there is a cr3 crash
+-- with the same case_id and investigat_agency_id is 74 (Austin PD)
```
roseeichelmann, line 64: and if they occurred within a 12 hour range of each other? could add that detail to this comment

roseeichelmann (approved):
the instructions say to check that `case_id = 150181115` got soft deleted but it didnt bc of your recent update, which is actually the correct thing to do bc looking at them they are two distinct crashes with totally different times and locations. i instead added a row to the non cr3 import of a dupe case id with a timestamp in the 12 hour range and invest agency of 74 and it soft deleted as expected! so yay this seems to be working, im glad we came up with a solution i hope it works well in the long run

johnclary:
Thanks for retesting that bit, Rose!

chiaberry (commented):
I cant hit approve post merging, but just wanted to say I still approve this PR!

PR #1762, opened 2025-04-09
EMS details page: add summary and patient cards to EMS details page

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/22025
* https://github.com/cityofaustin/atd-data-tech/issues/22029

Screenshot 2025-04-09 at 2 21 47 PM

## Testing

**URL to test:** Local, because we've branched from a branch

1. Apply migrations and metadata and start your app. 
2. Run this in your SQL client so that you have some matched EMS records to work with

```sql
update crashes set latitude = latitude - .00000001 where crash_timestamp > '2025-01-01';
```

3. Use the EMS list page to navigate to an EMS record with a matched crash. Check that the data cards look good, including the hyperlinked **Crash ID** field on the summary card
4. Inspect some more EMS records, including an unmated one.
5. Test out the bread crumb navigation from the EMS details page and the crash details page and the location details page

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/components/AppBreadCrumb.tsx
+++ b/editor/components/AppBreadCrumb.tsx
@@ -5,6 +5,7 @@ import Col from "react-bootstrap/Col";
 import { routes } from "@/configs/routes";
 
 interface Crumb {
+  path: string;
   label: string;
   type: "page" | "id";
 }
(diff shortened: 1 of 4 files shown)


editor/configs/emsColumns.tsx
```diff
@@ -64,6 +64,11 @@ export const ALL_EMS_COLUMNS = {
     path: "id",
     label: "ID",
     sortable: true,
+  },
+  id_hyperlinked: {
+    path: "id",
+    label: "ID",
+    sortable: true,
```
johnclary, line 71: curious how y'all feel about repeating columns when they have different formatting/definition needs. this is the first time we're doing things this way. for crashes, we have a totally separate column config for the crashes list and crash details, with a bit of redundancy between them

editor/components/AppBreadCrumb.tsx
```diff
@@ -19,17 +21,19 @@ const useCrumbs = (path: string): Crumb[] =>
       // we are at root
       return crumbs;
     }
+    // find the formatted route label
+    const pathPart = parts[0];
+    const route = routes.find((route) => route.path === pathPart.toLowerCase());
     crumbs.push({
-      // if we don't remove the query string, nextjs can hit a server/client mismatch on login
-      // todo: this can't be the right way to fix this
-      // todo: test if still an issue with app router
```
johnclary, line 25: this was indeed resolved completely by the app router. apparently the `usePathName` strips out the query string automatically

chiaberry (approved):
🚢

roseeichelmann (approved):
tested and reviewed the code, this looks great! i will say the fact that the address in the `<title>` is all caps when on the crash details page but not on ems pages sticks out like a sore thumb to me

johnclary:
I'm going to close this PR in favor of https://github.com/cityofaustin/vision-zero/pull/1777, which did branch form here but has diverged quite a bit

PR #1777, opened 2025-04-17
EMS incident details page with record matching UI

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/22026
* https://github.com/cityofaustin/atd-data-tech/issues/22196
* https://github.com/cityofaustin/atd-data-tech/issues/22025

This PR provides a basic proof of concept for linking EMS records to person records. There are definitely additional enhancements we'll want to make before releasing this into the wild, some of which I hope to surface through your feedback on this PR. 

## Testing

**URL to test:** Local

1. If you haven't replicated since [v2.9](https://github.com/cityofaustin/vision-zero/releases/tag/v2.9.0), do that
2. Start the whole VZE stack and apply migrations and metadata
3. From the EMS list, search for **Incident number** `24203-0059`
4. This search should return two records. Click on the hyperlinked **Incident #** column on either row to open the incident details page.
5. On the incident details, observe that there are two EMS records and three person records across two different crashes. 
6. Use the **Select person** button for the first row (`id`= `49644`) and then click the **Select**  button for the matching person record (`id` = `836332`).
7. Confirm that the EMS record now has a **Crash match status** of **Matched by review/QA**, and the **Crash ID** and **Person ID** columns are populated.
8. Also observe that the other EMS record (`id` = `55108`) now has a crash ID populated but no person ID populated (this value was assigned via the new DB trigger in this PR)
9. Also observe that the **Associated people records** table now only displays records from crash ID `20308227`.
10. Complete the record matching by clicking into the **Person ID** column of the second EMS record (`55108`) and manually keying in the person ID value `848014`.
11. Now edit this EMS row again by entering an invalid person ID, such as `123456789`. Observe that this results in an unhandled graphql error. John will open an issue for that 👍 

---

#### Ship list

- [ ] Code reviewed


editor/utils/ems.ts
johnclary, line 1: I wrote out this proof of concept for scoring EMS-person matches based on a handful of fields, but didn't implement this in the UI: https://github.com/cityofaustin/atd-data-tech/issues/22191

database/metadata/databases/default/tables/public_ems__incidents.yaml
johnclary, line 1: Adds editor audit fields and permission to edit a few EMS columns

database/metadata/databases/default/tables/public_people_list_view.yaml
```diff
@@ -2,6 +2,15 @@ table:
   name: people_list_view
   schema: public
 object_relationships:
```
johnclary, line 4: adds a manual gql relationship between the people list view and crashes so that we can query related crash data through a person-level record

database/migrations/default/1744826895218_ems_remove_match_status_constraint/up.sql
```diff
@@ -0,0 +1,47 @@
+alter table ems__incidents
+drop constraint matched_crash_pks_non_null,
+add column created_by text not null default 'system',
+add column updated_by text not null default 'system';
+
+comment on column ems__incidents.matched_crash_pks is 'The IDs of multiple crashes that were found to match this record. Set via trigger.''';
```
johnclary, line 6: Removes the constraint check that only allowed `matched_crash_pks` when the match status was `multiple_matches_by_automation`.

By removing this constraint, we can link an EMS record to a person record, change its status from `multiple_matches_by_automation` to `matched_by_manual_qa`, and leave the `matched_crash_pks` column untouched.

These kinds of details have me thinking about https://github.com/cityofaustin/atd-data-tech/issues/21714

database/migrations/default/1744826895218_ems_remove_match_status_constraint/up.sql
```diff
@@ -0,0 +1,47 @@
...
+
+--
+-- Trigger function that sets the ems__incident `crash_pk` based on the
+-- provided person_id and updates the crash_pk of all EMS records that
+-- share the same incident number
+--
+create or replace function update_person_crash_ems_match()
```
johnclary, line 13: This trigger covers two use cases. 

First, it allows the editor to assign a person ID to an EMS record, and this trigger takes care of ensuring that the `crash_pk` column is correct. Otherwise, we have a situation where the `crash_pk` column is out of sync with the `person_id` column. 

Consider this example:
- The crash-EMS trigger automatically assigns a crash PK to an EMS record
- A user discovers that this match was incorrect, and uses the UI to assign a person ID from a *different crash*
- This database trigger detects the the EMS record `person_id` has changed, and updates the record's crash_pk to match that of the person ID

editor/app/ems/[incident_number]/page.tsx
```diff
@@ -0,0 +1,205 @@
...
+          <RelatedRecordTable<EMSPatientCareRecord, EMSLinkRecordButtonProps>
+            records={ems_pcrs}
+            isValidating={isValidating}
+            noRowsMessage="No crashes found"
+            header="EMS patient(s)"
+            columns={emsDataCards.patient}
+            mutation={UPDATE_EMS_INCIDENT}
+            onSaveCallback={onSaveCallback}
+            rowActionComponent={EMSLinkRecordButton}
+            rowActionComponentAdditionalProps={linkRecordButtonProps}
```
Comment (johnclary) [inline editor/app/ems/[incident_number]/page.tsx:182]: This PR extends the `rowActionComponent` so that it can accept additional props, which enables the table to share state with other components. Getting the typing right was pretty tricky, and I have some doubts about how well this scales. I'll add my thoughts on this in the `RelatedRecordTable` component

editor/components/AppBreadCrumb.tsx
johnclary, line 1: Fixing that casing issue in the breadcrumb

editor/components/EMSLinkRecordButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { FaLink } from "react-icons/fa6";
+import PermissionsRequired from "@/components/PermissionsRequired";
+import AlignedLabel from "@/components/AlignedLabel";
+
+const allowedLinkRecordRoles = ["vz-admin", "editor"];
+
+export interface EMSLinkRecordButtonProps extends Record<string, unknown> {
+  onClick: (emsId: EMSPatientCareRecord) => void;
+  selectedEmsPcr: EMSPatientCareRecord | null;
+}
```
johnclary, line 13: Here's how we can define the type signature of the abstract `RowActionComponent`. This typing is handed to the `RelatedRecordTable` as a generic that makes its way down to the tree to the `RelatedRecordTableRow`

editor/components/RelatedRecordTable.tsx
```diff
@@ -3,7 +3,10 @@ import Table from "react-bootstrap/Table";
 import RelatedRecordTableRow from "@/components/RelatedRecordTableRow";
 import { ColDataCardDef } from "@/types/types";
 
-interface RelatedRecordTableProps<T extends Record<string, unknown>> {
+interface RelatedRecordTableProps<
+  T extends Record<string, unknown>,
+  P extends Record<string, unknown> = Record<string, unknown>,
+> {
```
johnclary, line 9: Where as this component used to just take a generic for the record type, it now takes a second, optional generic for the the `RowActionComponent`'s additional props.

This pattern is working fine, and it's pretty cool to see the strong typing in action when you're developing on a custom row action component. My concern is that this abstracted row action component is not going to scale very well if we want to, say, have multiple action components or to move this the first column instead of the last column in the table.

Thinking about the MUI data grid API, a more resilient way to design this, I think, would be to have a more flexible column definitions in `ColDataCardDef`. Instead of limiting the column definitions to column names, we could enable a column to be an arbitrary component with additional props. We would then find ourselves constructing the column arrays in a [hook like this](https://github.com/cityofaustin/atd-moped/blob/c79b0ce8476e2720e985e9ecca521f753851e76f/moped-editor/src/views/projects/projectView/ProjectPhases.js#L26)

editor/configs/crashesColumns.tsx
chiaberry, line 1: was this file renamed because you anticipate it getting react components later?

editor/configs/emsMatchingPeopleColumns.ts
```diff
@@ -0,0 +1,16 @@
+"use client";
```
chiaberry, line 1: why do we need  "use client" at the top of this file? I am still fuzzy on what the "use client" directive does and when to use it

database/migrations/default/1744826895218_ems_remove_match_status_constraint/up.sql
```diff
@@ -0,0 +1,47 @@
...
+    WHERE
+        incident_number = NEW.incident_number
+        AND crash_pk is distinct from matching_person_record.crash_pk
+        AND person_id IS NULL;
+
+    RETURN NEW;
+END;
+$$ language plpgsql;
+
+comment on function update_person_crash_ems_match is 'Sets the ems__incident `crash_pk` based -- Trigger function that sets the ems__incident crash_pk based on the provided person_id and updates the crash_pk of all EMS records that share the same incident number';
```
chiaberry, line 41: i believe this comment has some typos

editor/components/EMSLinkRecordButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+const EMSLinkRecordButton: React.FC<
+  RowActionComponentProps<EMSPatientCareRecord, EMSLinkRecordButtonProps>
+> = ({ record, additionalProps }) => {
+  //   const { mutate, loading: isMutating } = useMutation(mutation);
```
chiaberry, line 18: ^^

chiaberry (approved):
thanks for demoing this last week. 

fyi for other testers, when you are searching for the incident number in the instructions, your date filter needs to be 1Y or greater, the incident happened in 2024

it isnt obvious to me that you can click into person ID column to edit, but I think frequent editors know that pattern by now. I hope. 

I dont entirely understand the proof of concept for scoring EMS person matches, the score would tick up by one for every variable that matches? The number (0, 1 or 2) would then be displayed in the table? or would it just be used for sorting?

database/migrations/default/1744826895218_ems_remove_match_status_constraint/up.sql
```diff
@@ -0,0 +1,47 @@
...
+-- provided person_id and updates the crash_pk of all EMS records that
+-- share the same incident number
+--
+create or replace function update_person_crash_ems_match()
+returns trigger as $$
+DECLARE
+    matching_person_record RECORD;
+BEGIN
+    IF NEW.person_id is null
+        THEN return null;
```
roseeichelmann, line 19: so by deleting / nulling out a `person_id`, the `crash_id` the ems record is linked to will remain instead of nulling out -- is that what we want to happen? should there be a way to null a crash link?

database/migrations/default/1744826895218_ems_remove_match_status_constraint/up.sql
roseeichelmann, line 41: this comment repeats itself "Sets the ems__incident `crash_pk` based --" at the beginning

database/migrations/default/1744826895218_ems_remove_match_status_constraint/up.sql
johnclary: yeah, thanks for pointing this out. i left the behavior as-is because i wasn't sure what the desired behavior would be in different scenarios. but i guess if you're nullifying the person ID you probably want to nullify the crash ID as well, right?

johnclary:
> it isnt obvious to me that you can click into person ID column to edit, but I think frequent editors know that pattern by now. I hope.

I agree, it's not obvious. Might be good to have a button or some callout that makes it more obvious that you can edit the person ID 🤔 

> I dont entirely understand the proof of concept for scoring EMS person matches, the score would tick up by one for every variable that matches? The number (0, 1 or 2) would then be displayed in the table? or would it just be used for sorting?

I wasn't sure how we'd use the scoring. One idea was that once you "activate" an EMS record for matching, the people row(s) with the hight score would be highlighted or brought into focus somehow.

database/migrations/default/1744826895218_ems_remove_match_status_constraint/up.sql
johnclary: Note to self—if we decide to change this block so that the trigger nullifies the `crash_pk` column as well, we need to check to make sure that `OLD.person_id` is not null.

editor/configs/emsMatchingPeopleColumns.ts
johnclary: this is just copy/pasta cruft. we don't need it here 🙏

editor/configs/crashesColumns.tsx
johnclary: ah—an earlier version of this PR included JSX, and then i forgot to change it back. tbh i think i'd rather just leave this and future configs as JSX if they can potentially hold JSX components?

database/migrations/default/1744826895218_ems_remove_match_status_constraint/up.sql
roseeichelmann: why do we need to check that `OLD.person_id` was not null?

but yeah i do think it makes sense to be able to null out the `crash_id`, but im not sure if vz team would ever want to be able to null out person id but keep the crash id around. could be something that we fine tune once the vz team starts using this or maybe we can ask xavier?

database/migrations/default/1744826895218_ems_remove_match_status_constraint/up.sql
```diff
@@ -0,0 +1,47 @@
...
+    END IF;
+    -- Find matching person record
+    SELECT id, crash_pk INTO matching_person_record 
+    FROM people_list_view 
+    WHERE id = NEW.person_id;
+    
+    -- Update crah_pk of related EMS if not already matched to a crash
```
roseeichelmann, line 26: little typo here crah_pk

also - so this will allow users to assign different ems records of the same incident to different crash_pks. which i think is okay right?

i was playing around with adding in random person ids to the ems records so that the records would be assigned different crash ids, and it does seem like a needed feature to be able to null out crash id bc now when i null out the person ids i added that pulled up random crashes, i am no longer getting the recommended list of people from the automated crash matches, i am getting lists of people from the random ids i put in. so esp if someone makes a typo typing something in and it changes the crash id and they cant undo that, that would be a problem

editor/app/ems/[incident_number]/page.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+  /**
+   * Hook which manages which related crash PKs we should
+   * use to query people records
+   */
+  const relatedCrashPks: number[] | null = useMemo(() => {
+    if (!ems_pcrs) {
+      return null;
+    }
+    const relatedCrashPks: number[] = [];
```
Comment (roseeichelmann) [inline editor/app/ems/[incident_number]/page.tsx:65]: at first it confused me that this var was being created twice and i was gonna suggest to just use a .map() but it makes sense how you did it with having to dedupe and all that :+1:

editor/app/ems/[incident_number]/page.tsx
```diff
@@ -0,0 +1,205 @@
...
+  });
+
+  const onSaveCallback = useCallback(async () => {
+    await refetch();
+  }, [refetch]);
+
+  /**
+   * Memoize the additional components props for related record tables
+   */
+  const linkRecordButtonProps: EMSLinkRecordButtonProps = useMemo(
```
Comment (roseeichelmann) [inline editor/app/ems/[incident_number]/page.tsx:97]: i dont get exactly whats going on here, are you just redefining the props?

roseeichelmann (approved):
this was a huge chunk of work!! it looks great, you are making huge strides with this feature. i do think crash id nulling is something that should be implemented but maybe in a follow-up PR bc this one has already gotten so long. also, i think we may want to disable the select person button when you are already editing the person id text box and vice versa? other than that this and the code looks awesome, test steps all checked out :ok_hand:

database/migrations/default/1744826895218_ems_remove_match_status_constraint/up.sql
johnclary: > why do we need to check that OLD.person_id was not null?

Many EMS records will have a crash ID and  null person ID as a result of the crash-EMS trigger. So we need to make sure that the event that triggers this update is actually intending to nullify the person ID, and not the result of some other update that would unintentionally null out the crash ID. 

I'm with you that we need a mechanism to null out crash ID and it makes sense to do it in this trigger. I think it will just be one or two lines so maybe I'll push that up in this PR 👍

database/migrations/default/1744826895218_ems_remove_match_status_constraint/up.sql
johnclary: > so this will allow users to assign different ems records of the same incident to different crash_pks. which i think is okay right?

Yes, I think that's ok. We could have a situation where multiple EMS records are assigned to the wrong crash/person, and the user would have to re-assign each one individually. So we'd have a state during the QA/QC process where EMS records of the same incident are temporarily associated with a different crash.

editor/app/ems/[incident_number]/page.tsx
Comment (johnclary) [reply editor/app/ems/[incident_number]/page.tsx:97]: it is a type declaration saying that `linkRecordButtonProps` should be of type `EMSLinkRecordButtonProps`. would by nice if GitHub did a better job with typescript syntax highlighting—it's really hard to follow in diffs.

editor/components/RelatedRecordTable.tsx
johnclary: Sounds good to me! 🙏

database/migrations/default/1744826895218_ems_remove_match_status_constraint/up.sql
```diff
@@ -0,0 +1,47 @@
...
+    UPDATE ems__incidents 
+    SET 
+        crash_pk = matching_person_record.crash_pk,
+        updated_by = NEW.updated_by,
+        crash_match_status = 'matched_by_manual_qa'
+    WHERE
+        incident_number = NEW.incident_number
+        AND crash_pk is distinct from matching_person_record.crash_pk
+        AND person_id IS NULL;
```
johnclary, line 35: just want to mention that this `WHERE` filter (along with the early `RETURN` when `person_id` is null) is preventing infinite recursion, because this trigger is fired after update and updates the record itself. 

I believe we do want to bind the trigger to `after update` instead of `before update`—I'm getting a bit lost thinking about cascading record updates (including to this record itself) inside a `before` trigger. I'll double check this assumption when I revisit this next week 👍

editor/components/RelatedRecordTable.tsx
mddilley: i like this idea, and it is cool to think about building up these types of components for a chance!

editor/queries/ems.ts
```diff
@@ -20,10 +24,113 @@ export const GET_EMS_RECORD = gql`
...
+        veh_body_styl {
+          id
+          label
+        }
+      }
+    }
+  }
+`;
+
+export const UPDATE_EMS_INCIDENT_CRASH_AND_PERSON = gql`
```
mddilley, line 111: You mentioned a separate unhandled graphql error so I wanted to mention one related to this mutation too. When matching an unmatched EMS patient, the rows in the associated people records table still shows an enabled select button next to people records that have already been matched in the UI. 

There might be plans to address this beyond this PoC but didn't want to let it slip by.

https://github.com/user-attachments/assets/c90d21ca-f735-4d7b-a0f9-f3f928094e0e

mddilley (approved):
Love where this is going! Very cool to read through the code too. I left some feedback, and I'm eager to see this continue to grow! 🙌

database/migrations/default/1744826895218_ems_remove_match_status_constraint/up.sql
```diff
@@ -23,25 +54,25 @@ BEGIN
...
+    RETURN null;
 END;
 $$ language plpgsql;
 
-comment on function update_person_crash_ems_match is 'Sets the ems__incident crash_pk based on the provided person_id and updates the crash_pk of all EMS records that share the same incident number';
+comment on function ems_update_incident_crash_pkis 'Sets the ems__incident crash_pk based on the provided person_id and updates the crash_pk of all EMS records that share the same incident number';
 
-create or replace trigger ems_update_person_match_trigger
+create or replace trigger ems_update_incident_crash_pk_trigger
 after update on ems__incidents
```
roseeichelmann, line 75: so this trigger `ems_update_incident_crash_pk_trigger` runs after the `ems_update_person_crash_id_trigger`, maybe include in the comments that that is important. "Must run after the `ems_update_person_crash_id` trigger"

johnclary: 👍 https://github.com/cityofaustin/vision-zero/pull/1777/commits/5d61a9f867a033f47f9c9e91795733f44705dc57

chiaberry (commented):
I am going to try again, but I got a graphql error when going through the test steps
```
graphql.ts:170 GraphQL Mutation Error: ClientError: expecting a value for non-nullable variable: "crash_pk": {"response":{"errors":[{"message":"expecting a value for non-nullable variable: \"crash_pk\"","extensions":{"path":"$","code":"validation-failed"}}],"status":200,"headers":{}},"request":{"query":"\n  mutation UpdateEMSIncident($id: Int!, $person_id: Int!, $crash_pk: Int!) {\n    update_ems__incidents(\n      where: { id: { _eq: $id } }\n      _set: {\n        person_id: $person_id\n        crash_pk: $crash_pk\n        crash_match_status: \"matched_by_manual_qa\"\n      }\n    ) {\n      affected_rows\n    }\n  }\n","variables":{"id":49644,"person_id":836332}}}
```

database/migrations/default/1744826895218_ems_remove_match_status_constraint/up.sql
```diff
@@ -23,25 +54,25 @@ BEGIN
...
-    -- Update crah_pk of related EMS if not already matched to a crash
+    -- Update crash_pk of related EMS if not already matched to a person
     UPDATE ems__incidents 
     SET 
         crash_pk = matching_person_record.crash_pk,
         updated_by = NEW.updated_by,
         crash_match_status = 'matched_by_manual_qa'
     WHERE
-        incident_number = NEW.incident_number
+        id != NEW.person_id
```
roseeichelmann, line 64: im very confused by this line, what does `ems__incidents.id` have to do with `NEW.person_id`?

johnclary: that's a bug! just changed to `NEW.id` 🙏

chiaberry (approved):
github seems to have reverted to my previous approval, but just wanted to say that ive tested this through its iterations today and got it working.

roseeichelmann (approved):
tested and your updates are working as expected!

mddilley (approved):
I retested as well, and all the steps still work out for me! Thanks for the extra details in the slack thread too 🙏

johnclary:
thanks @mateoclarke for flagging that. I added a link to your comment in https://github.com/cityofaustin/atd-data-tech/issues/22299, which i _think_ should handle this.

johnclary:
Appreciate all the reviews and bug catching 🙏

PR #1779, opened 2025-04-23
EMS details: show people from crashes that occurred on the same date as the incident

roseeichelmann:
## Associated issues
Closes https://github.com/issues/assigned?issue=cityofaustin%7Catd-data-tech%7C22200

## Testing

**URL to test:** 
local

**Steps to test:**
1. Go to the EMS page, click on some EMS records that have a crash status of unmatched.
2. See that the people list is populated with  people from crashes that occurred within 12 hours of the ems incident
3. Test linking an ems incident to a person record and see the status turn from unmatched to matched by manual Q/A and the person list should update with only people from that crash id
4. Go to a EMS page that has at least two unmatched records
5. Link one, see that both records will turn to matched with Q/A and the people list will shorten to only people from that crash id. 
6. Then in your hasura console or db manager test deleting the crash pk and update crash_status to "unmatched" again on one of the records in the `ems__incidents` tables, so that one record is unmatched and the other record is matched. See that the people list populates with records both from the crash id you manually Q/Aed and from crashes that occurred within 12 hours of the record that is now unmatched. This is to test the behavior that in the future users will be able to null out a crash pk they manually added and revert to the non-fiddled with crash match status for one of the ems records while keeping the other ems record manual QA

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/app/ems/[incident_number]/page.tsx
+++ b/editor/app/ems/[incident_number]/page.tsx
@@ -84,9 +84,9 @@ export default function EMSDetailsPage({
     return Array.from(new Set(relatedCrashPks));
   }, [ems_pcrs]);
 
-  const areAnyUnmatchedRecords = ems_pcrs
-    ? ems_pcrs.some((ems) => ems.crash_match_status === "unmatched")
-    : null;
+  const areAnyUnmatchedRecords: boolean | undefined = ems_pcrs?.some(
+    (ems) => ems.crash_match_status === "unmatched"
+  );
 
   const incidentTimestamp = incident?.incident_received_datetime
     ? parseISO(incident?.incident_received_datetime)
@@ -98,15 +98,12 @@ export default function EMSDetailsPage({
     ? addHours(incidentTimestamp, 12)
     : null;
 
-  console.log(
-    time12HoursAfter,
-    time12HoursBefore,
-    incidentTimestamp,
-    areAnyUnmatchedRecords
-  );
-
-  const { data: unmatchedCrashes, error: errorUnmatched } = useQuery<Crash>({
-    query: ems_pcrs ? GET_UNMATCHED_EMS_CRASHES : null,
+  /**
+   * Get all crash records that occurred within 12 hours of the incidents
+   * if the incidents have a crash match status of "unmatched"
+   */
+  const { data: unmatchedCrashes } = useQuery<Crash>({
+    query: areAnyUnmatchedRecords ? GET_UNMATCHED_EMS_CRASHES : null,
     variables: {
       timestamp12HoursBefore: time12HoursBefore,
       timestamp12HoursAfter: time12HoursAfter,
@@ -116,23 +113,18 @@ export default function EMSDetailsPage({
 
   const unmatchedCrashPks = unmatchedCrashes?.map((crash) => crash.id);
 
-  console.log(relatedCrashPks, "related crash pks");
-
   /**
-   * Get all people records linked to crashes associated with these incidents
+   * Get all people records that are either linked to crashes associated
+   * with these incidents or that occurred within 12 hours of the incidents
+   * if they have a crash status of unmatched
    */
-  const { data: matchingPeople, error: errormatchingpeople } =
-    useQuery<PeopleListRow>({
-      query: relatedCrashPks || unmatchedCrashPks ? GET_MATCHING_PEOPLE : null,
-      variables: {
-        crash_pks: relatedCrashPks?.[0] ? relatedCrashPks : unmatchedCrashPks,
-      },
-      typename: "people_list_view",
-    });
-
-  console.log(unmatchedCrashPks, "unmatched crashes", errorUnmatched);
-
-  console.log(matchingPeople, "matching people", errormatchingpeople);
+  const { data: matchingPeople } = useQuery<PeopleListRow>({
+    query: relatedCrashPks || unmatchedCrashPks ? GET_MATCHING_PEOPLE : null,
+    variables: {
+      crash_pks: relatedCrashPks?.[0] ? relatedCrashPks : unmatchedCrashPks,
+    },
+    typename: "people_list_view",
+  });
 
   const onSaveCallback = useCallback(async () => {
     await refetch();
--- a/editor/app/ems/[incident_number]/page.tsx
+++ b/editor/app/ems/[incident_number]/page.tsx
@@ -80,30 +80,36 @@ export default function EMSDetailsPage({
   }, [ems_pcrs]);
 
   /**
-   * Hook which manages getting the 12 hour timestamp interval
+   * Function that gets the 12 hour timestamp interval
    * to be used for fetching people list for unmatched EMS records
    */
-  const unmatchedTimeInterval: Date[] = useMemo(() => {
-    // if any of the ems records have a crash match status of unmatched
-    if (ems_pcrs?.some((ems) => ems.crash_match_status === "unmatched")) {
-      if (ems_pcrs[0].incident_received_datetime) {
-        const incidentTimestamp = parseISO(
-          ems_pcrs[0].incident_received_datetime
-        );
-        const time12HoursBefore = subHours(incidentTimestamp, 12);
-        const time12HoursAfter = addHours(incidentTimestamp, 12);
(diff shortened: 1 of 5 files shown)


editor/app/ems/[incident_number]/page.tsx
```diff
@@ -58,9 +61,9 @@ export default function EMSDetailsPage({
    * Hook which manages which related crash PKs we should
    * use to query people records
    */
-  const relatedCrashPks: number[] | null = useMemo(() => {
+  const relatedCrashPks: number[] = useMemo(() => {
     if (!ems_pcrs) {
-      return null;
+      return [];
```
Comment (roseeichelmann) [inline editor/app/ems/[incident_number]/page.tsx:66]: this possibly being null was causing me typescript headaches later

roseeichelmann:
At first the way i wrote this code it expected that if one record was unmatched then they would all be unmatched. So it would only check the first record of the incident for the match status and it would only either query people with the unmatched crash pks OR with the related crash pks. But, assuming that ems records associated with the same incident should be able to be manually edited to be assigned to different crashes, and users should be able to revert their updates, i updated this code to check if any of the records are unmatched, and it will query people using both unmatched and related crash pks if the records have varying crash match statuses.

chiaberry (commented):
I am doing the test steps and I'm a little confused on step 4. Go to a EMS page that has two unmatched records -- should it be an ems incident page that *only* has two records? Trying to understand the difference between what im testing in step 1 vs step 4

roseeichelmann:
@chiaberry its just incase in the first step the incident you clicked on only had one record, for the steps in 4 and after it needs to be at least two ems records on the page

editor/queries/ems.ts
```diff
@@ -34,6 +34,23 @@ export const GET_EMS_RECORDS = gql`
...
 
+export const GET_UNMATCHED_EMS_CRASHES = gql`
+  query EMSUnmatchedCrashes(
+    $time12HoursBefore: timestamptz!
+    $time12HoursAfter: timestamptz!
+  ) {
+    crashes(
+      where: {
+        crash_timestamp: { _gte: $time12HoursBefore, _lte: $time12HoursAfter }
+      }
```
chiaberry, line 45: I think it would be easier to visually scan the results if these were returned in date order

johnclary: ah, i see—the people list is sorting by `{ cris_crash_id: asc, unit_nbr: asc, prsn_nbr: asc }`. Rose, what if we replaced `cris_crash_id` with `crash_timestamp` as the first sort column? That should keep things sorted chronologically and by crash ID.

johnclary (commented):
Rose, this looks really great! It's cool to be able to do some work with these unmatched EMS records.

I'm happy to open a follow-up for this, but I think it'd be preferable to always show the **Associated people records** table, even if there are no results. I got a little confused when I was looking at an EMS incident that was very recent and had no crashes in the 12-hour window. I think rendering the table with the message **No people found** would be fine?

Screenshot 2025-04-28 at 11 35 37 AM

editor/app/ems/[incident_number]/page.tsx
```diff
@@ -77,12 +80,58 @@ export default function EMSDetailsPage({
...
+  };
+
+  // Check if any of the ems incidents have an unmatched status
+  const areAnyUnmatchedEMSRecords = ems_pcrs?.some(
+    (ems) => ems.crash_match_status === "unmatched"
+  );
+
+  const unmatchedTimeInterval = areAnyUnmatchedEMSRecords
+    ? getUnmatchedTimeInterval()
+    : null;
```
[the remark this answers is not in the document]
Comment (roseeichelmann) [reply editor/app/ems/[incident_number]/page.tsx:105]: gah i [originally](https://github.com/cityofaustin/vision-zero/pull/1779/commits/498e13bacc9864cce4cefe9b04539d2816a304e0) had it in a `useMemo` but convinced myself that was overkill. i will memoize this!

editor/app/ems/[incident_number]/page.tsx
```diff
@@ -77,12 +80,58 @@ export default function EMSDetailsPage({
...
+   * if the incidents have a crash match status of unmatched
+   */
+  const { data: unmatchedCrashes } = useQuery<Crash>({
+    query: unmatchedTimeInterval ? GET_UNMATCHED_EMS_CRASHES : null,
+    variables: {
+      time12HoursBefore: unmatchedTimeInterval?.[0],
+      time12HoursAfter: unmatchedTimeInterval?.[1],
+    },
+    typename: "crashes",
+  });
```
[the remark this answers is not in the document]
Comment (roseeichelmann) [reply editor/app/ems/[incident_number]/page.tsx:118]: i was thinking about that approach but landed on this bc like you said then i can just query all the combined crash pks in the next person query. also figured its better to query a the crashes table than the person list view to minimize view querying but idk if that rly makes a performance diff

editor/app/ems/[incident_number]/page.tsx
```diff
@@ -183,23 +227,21 @@ export default function EMSDetailsPage({
...
-            />
-          </Col>
-        </Row>
-      )}
+      <Row>
+        <Col sm={12} className="mb-3">
+          <RelatedRecordTable<PeopleListRow, EMSLinkToPersonButtonProps>
+            records={matchingPeople ? matchingPeople : []}
+            isValidating={isValidating}
+            noRowsMessage="No people found"
```
Comment (roseeichelmann) [inline editor/app/ems/[incident_number]/page.tsx:235]: we already had a helpful no rows message in place but i needed to delete the `matchingPeople` check above for the table to still render with no people, also updated this message to say No people found instead of No crashes found

roseeichelmann:
@johnclary updated with all your feedback!

johnclary (approved):
This is ready to gooooo! 🚢 🚢 🚢 🚢 

I did have a question about date string formatting.

editor/app/ems/[incident_number]/page.tsx
```diff
@@ -80,39 +85,31 @@ export default function EMSDetailsPage({
...
+   * if any of the ems pcrs have a crash match status of unmatched
    */
   const { data: unmatchedCrashes } = useQuery<Crash>({
-    query: unmatchedTimeInterval ? GET_UNMATCHED_EMS_CRASHES : null,
+    query: unmatchedTimeInterval[0] ? GET_UNMATCHED_EMS_CRASHES : null,
     variables: {
-      time12HoursBefore: unmatchedTimeInterval?.[0],
-      time12HoursAfter: unmatchedTimeInterval?.[1],
+      time12HoursBefore: unmatchedTimeInterval[0],
+      time12HoursAfter: unmatchedTimeInterval[1],
```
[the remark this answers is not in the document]
Comment (roseeichelmann) [reply editor/app/ems/[incident_number]/page.tsx:112]: it is interesting that it just handles it under the hood, im guessing somehow the gql query/postgres can just infer a timestamptz from a date object? but i went ahead and converted them to iso in the `unmatchedInterval` function bc it is confusing!

chiaberry (approved):
👍🏼 👍🏼 
finally followed the test steps correctly and it all checks out

editor/app/ems/[incident_number]/page.tsx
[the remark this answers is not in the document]
Comment (mddilley) [reply editor/app/ems/[incident_number]/page.tsx:112]: thanks for sharing about this in planning today, y'all! 🧠 makes me wonder how many times i could have avoided formatting a date obj!

editor/app/ems/[incident_number]/page.tsx
```diff
@@ -77,12 +85,53 @@ export default function EMSDetailsPage({
...
+   * Get all people records linked to crashes that were either automatically
+   * matched with the ems record or that occurred within 12 hours of the incident
+   * if it has a crash status of unmatched
    */
   const { data: matchingPeople } = useQuery<PeopleListRow>({
-    query: relatedCrashPks ? GET_MATCHING_PEOPLE : null,
+    query: allCrashPks[0] ? GET_MATCHING_PEOPLE : null,
     variables: {
-      crash_pks: relatedCrashPks,
+      crash_pks: unmatchedTimeInterval[0] ? allCrashPks : relatedCrashPks,
```
Comment (mddilley) [inline editor/app/ems/[incident_number]/page.tsx:134]: Ok cool now I see how the results are being controlled in the different steps of the matching 👍

mddilley (approved):
Tested and all the steps worked as described. Cool to test the un-fiddle functionality in here too. 🚢

PR #1781, opened 2025-04-28
EMS details: hide select button for people records that are already linked

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/22207

Best to test this after https://github.com/cityofaustin/vision-zero/pull/1779 for your own context

## Testing

**URL to test:** 
Local

**Steps to test:**
1. Make sure you apply metadata locally
2. Go to an incident with unmatched PCRs, start linking them to people and see that after a person gets linked the select button no longer shows up on that row when you try to link other pcrs
3. Go to a different incident with unmatched PCRs with a timestamp that is very close to the previous incident, you should see that when you go to select people some of them are already unavailable to be selected if you linked them with the previous incident in step 1

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/database/metadata/databases/default/tables/public_people_list_view.yaml
+++ b/database/metadata/databases/default/tables/public_people_list_view.yaml
@@ -20,6 +20,15 @@ object_relationships:
         remote_table:
           name: drvr_ethncty
           schema: lookups
+  - name: ems_pcr
+    using:
+      manual_configuration:
+        column_mapping:
+          id: person_id
+        insertion_order: null
+        remote_table:
+          name: ems__incidents
+          schema: public
   - name: gndr
     using:
       manual_configuration:
--- a/editor/components/EMSLinkToPersonButton.tsx
+++ b/editor/components/EMSLinkToPersonButton.tsx
@@ -18,11 +18,8 @@ const EMSLinkToPersonButton: React.FC<
   RowActionComponentProps<PeopleListRow, EMSLinkToPersonButtonProps>
 > = ({ record, additionalProps }) => {
   const isLinkingInProgress = !!additionalProps?.selectedEmsPcr;
-  const isPersonAlreadyLinked = additionalProps?.matchedPersonIds
-    ? additionalProps?.matchedPersonIds.some(
-        (personId) => personId === record.id
-      )
-    : null;
+  // Does the person already have a non-null ems_pcr relationship
+  const isPersonAlreadyLinked = !!record.ems_pcr;
 
   if (!isLinkingInProgress || isPersonAlreadyLinked) {
     return null;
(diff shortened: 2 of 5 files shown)


database/metadata/databases/default/tables/public_people_list_view.yaml
```diff
@@ -20,6 +20,15 @@ object_relationships:
         remote_table:
           name: drvr_ethncty
           schema: lookups
+  - name: ems_pcr
```
roseeichelmann, line 23: new object relationship between `people_list_view` and `ems__incidents` so that when we query people on the incident details page we know if they have already been linked so users wont be able to try to link them again

editor/types/peopleList.ts
```diff
@@ -26,4 +26,5 @@ export type PeopleListRow = {
   unit_id: number;
   unit?: Unit;
   crash?: Crash;
+  ems_pcr: LookupTableOption;
```
roseeichelmann, line 29: technically not a lookup table but.. is this fine?

johnclary: eeek, yeah, the contract we promise to uphold between our type definitions and our graphql queries 😬 

i think we probably want to replace `LookupTableOption` with `EMSPatientCareRecord` and make it optional. 

i have been really sloppy with this, but we should probably be making all properties of our graphql types optional, since we can't guarantee what will be included in the graphql query. That would force us to write type guards that handle that case in our code. 

That feels probably out of scope for this PR, but fwiw most of the properties of the `EMSPatientCareRecord` are already nullable, so i would think we could make them optional as well without too much pain.

So I'll write up a follow-up issue to review and tighten up our graphql typings, and i think you should change this line to be an optional `EMSPatientCareRecord`.

johnclary (approved):
@roseeichelmann this looks great!  🙌 🚢

editor/types/peopleList.ts
roseeichelmann: thanks for pointing me to the pcr type, i should have found it myself and not been lazy lol. pushed an update!

chiaberry (approved):
no more postgres tx error! 🚢

editor/components/EMSLinkToPersonButton.tsx
```diff
@@ -17,8 +17,10 @@ const EMSLinkToPersonButton: React.FC<
   RowActionComponentProps<PeopleListRow, EMSLinkToPersonButtonProps>
 > = ({ record, additionalProps }) => {
   const isLinkingInProgress = !!additionalProps?.selectedEmsPcr;
+  // Does the person already have a non-null ems_pcr relationship
+  const isPersonAlreadyLinked = !!record.ems_pcr;
 
-  if (!isLinkingInProgress) {
+  if (!isLinkingInProgress || isPersonAlreadyLinked) {
```
mddilley, line 23: Looks great!

Screenshot 2025-04-30 at 5 00 16 PM

mddilley (approved):
Thanks for updating - I can no longer try to match a previously matched person. And, one less Texas error!  🚢 😎

PR #1786, opened 2025-05-09
Rework the EMS `patient_injry_sev_id` generated column

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/22033

## Testing

**URL to test:** Local

2. Query the EMS incidents table to inspect results. Here's a query that will be helpful:

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/database/migrations/default/1746807631787_patient_inj_sev_update/down.sql
+++ b/database/migrations/default/1746807631787_patient_inj_sev_update/down.sql
@@ -1,0 +1,71 @@
+alter table ems__incidents drop column patient_injry_sev_id;
+
+alter table ems__incidents
+add column patient_injry_sev_id integer generated always as (
+    case
+        when
+            lower(pcr_provider_impression_primary) = 'death on scene'
+            or lower(pcr_outcome) = 'deceased on scene'
+            or lower(pcr_patient_acuity_final)
+            = 'dead without resuscitation efforts (black)'
+            then 4
+        when
+            right(pcr_transport_priority, 1) = '3'
+            and lower(pcr_patient_acuity_final) = 'lower acuity (green)'
+            then 2
+        when
+            right(pcr_transport_priority, 1) = '3'
+            and lower(pcr_patient_acuity_final) = 'critical (red)'
+            then 1
+        when
+            right(pcr_transport_priority, 1) = '3'
+            and lower(pcr_patient_acuity_final) = 'emergent (yellow)'
+            then 1
+        when
+            right(pcr_transport_priority, 1) = '3'
+            and lower(pcr_patient_acuity_initial) = 'critical (red)'
+            then 1
+        when
+            right(pcr_transport_priority, 1) = '3'
+            and lower(pcr_patient_acuity_initial) = 'emergent (yellow)'
+            then 1
+        when
+            right(pcr_transport_priority, 1) = '3'
+            and lower(pcr_patient_acuity_level) = 'high acuity'
+            then 1
+        -- what is this pcr_transport_priority?
+        when
+            lower(left(pcr_transport_priority, 1)) = 'charlie'
+            or lower(left(pcr_transport_priority, 1)) = 'd'
+            then 1
+        when
+            right(pcr_transport_priority, 1) = '3'
+            and lower(pcr_patient_acuity_initial) = 'lower acuity (green)'
+            then 2
+        when
+            lower(left(pcr_transport_priority, 1)) = 'minor injury'
+            or lower(left(pcr_transport_priority, 1)) = 'possible injury'
+            then 2
+        when
+            right(pcr_transport_priority, 1) != '3'
+            and lower(pcr_patient_acuity_initial) = 'lower acuity (green)'
+            then 3
+        when
+            lower(right(pcr_outcome, 7)) = 'refused'
+            and left(lower(pcr_patient_acuity_level), 10) = 'low acuity'
+            then 3
+        else 5
+    end
+) stored;
+
+alter table ems__incidents
+add constraint ems__incidents_patient_injry_sev_id_fk foreign key (
+    patient_injry_sev_id
+)
+references lookups.injry_sev on update restrict on delete restrict;
+
+comment on column ems__incidents.patient_injry_sev_id is 'The patient injury severity as mapped to the CRIS injury severity lookup. Based on methodology provided by Xavier A.';
+
+create index ems__incidents_patient_injry_sev_id_index on ems__incidents (
+    patient_injry_sev_id
+);
--- a/database/migrations/default/1746807631787_patient_inj_sev_update/up.sql
+++ b/database/migrations/default/1746807631787_patient_inj_sev_update/up.sql
@@ -1,0 +1,261 @@
+alter table ems__incidents drop column patient_injry_sev_id;
+
+alter table ems__incidents
+add column patient_injry_sev_id integer generated always as (
+    case
+        --
+        -- fatal injuries
+        --
+        when
+            (lower(pcr_provider_impression_primary) = 'death on scene')
+            or
+            (lower(pcr_provider_impression_secondary) = 'death on scene')
+            or
+            (lower(pcr_outcome) = 'deceased on scene')
+            or
+            (
+                lower(pcr_patient_acuity_final)
+                = 'dead without resuscitation efforts (black)'
+            )
+            or (
+                (
+                    lower(pcr_patient_acuity_initial)
+                    = 'dead without resuscitation efforts (black)'
+                )
+                and (
+                    pcr_patient_acuity_final is NULL
+                    or
+                    lower(pcr_patient_acuity_final)
+                    = 'dead without resuscitation efforts (black)'
+                )
+            )
+            then 4
+        --
+        -- serious injuries
+        --
+        when
+            (lower(pcr_patient_acuity_final) = 'critical (red)')
+            or
+            (
+                lower(pcr_patient_acuity_initial) = 'critical (red)'
+                and pcr_patient_acuity_final is NULL
+            )
+            or
+            (
+                lower(pcr_patient_acuity_final) = 'emergent (yellow)'
+                and lower(pcr_outcome) in ('transported', 'care transferred')
+            )
+            or
+            (right(pcr_transport_priority, 1) = '3')
+            or
+            (lower(pcr_transport_priority) like 'charlie%')
+            or
+            (lower(pcr_transport_priority) like 'delta%')
+            or
+            (lower(pcr_transport_priority) like 'echo%')
+            or
+            (
+                lower(pcr_patient_acuity_initial) = 'emergent (yellow)'
+                and lower(pcr_patient_acuity_final) is NULL
+                and lower(pcr_outcome) in ('transported', 'care transferred')
+            )
+            or
+            lower(pcr_provider_impression_primary) like 'fracture%'
+            or
+            lower(pcr_provider_impression_secondary) like 'fracture%'
+            or
+            (
+                lower(pcr_patient_acuity_level_reason)
+                = 'evidence of major trauma/hemorrhage'
+            )
+            or
+            (
+                lower(pcr_provider_impression_primary) in
+                (
+                    'burn',
+                    'cardiac - cardiac arrest',
+                    'cardiac - cardiac arrest - traumatic',
+                    'syncope',
+                    'unconsciousness'
+                )
+            )
+            or
+            (
+                lower(pcr_provider_impression_secondary) in
+                (
+                    'burn',
+                    'cardiac - cardiac arrest',
+                    'cardiac - cardiac arrest - traumatic',
+                    'syncope',
+                    'unconsciousness'
+                )
+            )
+            then 1
+        --
+        -- minor injuries
+        --
+        when
+            (lower(pcr_outcome) in ('transported', 'care transferred'))
+            or
+            (mvc_form_patient_injured_flag = 1)
+            or
+            (lower(pcr_provider_impression_primary) like 'injury%')
+            or
+            (lower(pcr_provider_impression_secondary) like 'injury%')
+            or
+            (
+                lower(pcr_provider_impression_primary) in
+                (
+                    'Altered Mental Status',
+                    'Assault - Physical',
+                    'Burn - Chemical',
+                    'Cardiac - Arrhythmia/Dysrythmia',
+                    'Cardiac - Myocardial Infarction',
+                    'Cardiac - Palpitations',
+                    'Confusion',
+                    'Diarrhea',
+                    'Dizziness',
+                    'Epistaxis',
+                    'Fatigue',
+                    'Fever',
+                    'Foreign Body - Alimentary Tract',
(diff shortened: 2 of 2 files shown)


database/migrations/default/1746807631787_patient_inj_sev_update/up.sql
```diff
@@ -0,0 +1,258 @@
...
+            or (
+                (
+                    lower(pcr_patient_acuity_initial)
+                    = 'dead without resuscitation efforts (black)'
+                )
+                and (
+                    pcr_patient_acuity_final is NULL
+                )
+            )
+            then 4
```
johnclary, line 29: this block now checks two fields we weren't previously looking at:
* `pcr_patient_acuity_initial` (when `pcr_patient_acuity_final` is null)
* `pcr_provider_impression_secondary`

database/migrations/default/1746807631787_patient_inj_sev_update/up.sql
```diff
@@ -0,0 +1,258 @@
...
+        --
+        -- serious injuries
+        --
```
johnclary, line 32: Serious vs minor injuries is a real guessing game. Here's the TxDOT definition:

> Severe injury that prevents continuation of normal activities
> • Severe laceration, broken or distorted extremity, crush injuries
> • Skull, chest, or abdominal injury other than bruises or minor lacerations
> • Significant burns
> • Unconsciousness when taken from the crash scene
> • Paralysis

database/migrations/default/1746807631787_patient_inj_sev_update/up.sql
```diff
@@ -0,0 +1,258 @@
...
+                and lower(pcr_outcome) in ('transported', 'care transferred')
+            )
+            or
+            (right(pcr_transport_priority, 1) = '3')
+            or
+            (lower(pcr_transport_priority) like 'charlie%')
+            or
+            (lower(pcr_transport_priority) like 'delta%')
+            or
+            (lower(pcr_transport_priority) like 'echo%')
```
johnclary, line 58: The `pcr_transport_priority` column uses a mix of letter codenames plus generic `Code 1` and `Code 3. Our current best understanding:

- Code 1: transport without lights + sirens
- Code 3: transport with lights + sirens

And for the latter codes:

- A (Alpha Response) refers to a low-priority situation.
- B (Bravo Response) indicates a mid-priority condition.
- C (Charlie Response) represents a potentially life-threatening condition.
- D (Delta Response) indicates a life-threatening emergency.
- E (Echo Response) refers to full cardiac arrest or imminent death.

database/migrations/default/1746807631787_patient_inj_sev_update/up.sql
```diff
@@ -0,0 +1,258 @@
...
+            (lower(pcr_outcome) in ('transported', 'care transferred'))
+            or
+            (mvc_form_patient_injured_flag = 1)
+            or
+            (lower(pcr_provider_impression_primary) like 'injury%')
+            or
+            (lower(pcr_provider_impression_secondary) like 'injury%')
+            or
+            (
+                lower(pcr_provider_impression_primary) in
```
johnclary, line 104: I came up with this list based on my own review of the values in this column 🤷

johnclary:
Folks, I'm going to go ahead and request review, although it's quite like that this will evolve as Xavier (cc'd) has a chance to review.

database/migrations/default/1746807631787_patient_inj_sev_update/up.sql
chiaberry: I wonder if the `ob` impressions should be moved to the excluded list/possible injury, or at least the generic `ob` because it appears that is what is listed if the person is pregnant at all. So this would label a pregnant person in a fender bender as "minor injury"

chiaberry (approved):
looks good! I pulled back from getting sucked into splitting hairs about provider impression categorizing, I assume this will be refined later if it turns out some crashes are routinely being miscategorized.

database/migrations/default/1746807631787_patient_inj_sev_update/up.sql
johnclary: I wondered this as well. And, yes, currently this would label a pregnant person as a minor injury—EMS ID `51602` being an example of someone listed as `OB` with no other injuries that was not transported.

I'll move the generic `OB` into the "possible injury" list. 

Looking at this more closely, I'm realizing that I generated the unique lists of provider impressions based on the values remaining in this column _after_ all major injuries had been assigned. Which is to say, I'm missing some values. I will update "minor" and "possible" injury lists and report back.

database/migrations/default/1746807631787_patient_inj_sev_update/up.sql
roseeichelmann: My first thought was to give users the ability to manually edit the injury status, but i think lets avoid that until xavier makes it known that is a need if weird cases like that keep stacking up? for now using the higher injury severity when a fatality is involved seems okay

database/migrations/default/1746807631787_patient_inj_sev_update/up.sql
```diff
@@ -0,0 +1,258 @@
...
+                and lower(pcr_provider_impression_secondary) not in
+                (
+                    'exam - adult - no finding',
+                    'exam - child - no finding',
+                    'exam - general',
+                    'no complaints'
+                )
+            )
+            then 3
+        else 5
```
roseeichelmann, line 241: does 5 mean no injury? could you comment that

database/migrations/default/1746807631787_patient_inj_sev_update/down.sql
```diff
@@ -0,0 +1,71 @@
+alter table ems__incidents drop column patient_injry_sev_id;
```
roseeichelmann, line 1: maybe a brief comment at this top of this file summarizing what the change is

database/migrations/default/1746807631787_patient_inj_sev_update/up.sql
johnclary: New plan after discussing with Xavier: Going to keep things simple and if the `provider_impression_` columns contain anything other than `injury%`, will assign as "possible injury".

We will re-assess once we have this in place. Our expectation is there are a few borderline cases that will be misclassified but the error rate should be <1%

johnclary:
Thanks for your previous reviews. I've requested re-review after some significant changes. The test steps are still the same.

Summary of changes:
* simplified logic for minor injuriess—no longer combing through the `provider_impression_` columns
* Add the `patient_injry_sev_reason` column, which contains the snipped of sql used to assign the injury severity. This adds transparency to our classification process
* Use a trigger instead of generated columns to set these values ☝️

database/migrations/default/1746807631787_patient_inj_sev_update/up.sql
```diff
@@ -0,0 +1,143 @@
...
+    RETURN NEW;
+END;
+$$ language plpgsql;
+
+--
+-- Trigger is only deployed on inserts, because we do not allow updates to the dependent columns
+--
+create or replace trigger ems_incidents_trigger_insert_set_patient_injry_sev
+before insert on ems__incidents
+for each row execute function update_ems_patient_injry_sev();
```
johnclary, line 132: i suppose we could set up the update trigger just for good measure?

frankhereford (commented):
I'd love to maybe learn some background on your work with Xavier that lead to this business logic and to maybe have a discussion about resilience to upstream changes. I think I need to know more of the bigger picture to see how this work fits into it.

database/migrations/default/1746807631787_patient_inj_sev_update/up.sql
johnclary: I agree—I will put together the update trigger. Thanks for that nudge!

database/migrations/default/1746807631787_patient_inj_sev_update/up.sql
```diff
@@ -0,0 +1,143 @@
...
+        NEW.patient_injry_sev_reason = 'pcr_provider_impression_secondary is not NULL and lower(pcr_provider_impression_secondary) not in (''exam - adult - no finding'', ''exam - child - no finding'', ''exam - general'', ''no complaints'')';
+    --
+    -- not injured
+    --
+    else 
+        NEW.patient_injry_sev_id = 5;
+        NEW.patient_injry_sev_reason = NULL;
+    end case;
+    RETURN NEW;
+END;
```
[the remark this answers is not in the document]
johnclary: > concerned about the number of repeating strings going into the DB

Yeah, copying the literal SQL string into the column is a bit hair-brained. The thinking is that having this detail will help Xavier/Joel/VZ team speak confidently about why we have assigned an injury severity to a crash that conflicts with what was reported by the investigating officer. 

Alternatively, we could construct this audit trail on the fly with a view, but that would require us to duplicate this trigger logic in the view's query and remember to keep it in sync with changes to this trigger 😕

> When a FK value gets changed upstream...

Glad you mentioned this. I've pondered this case as well. The implications of having a non-CRIS table dependent on a CRIS lookup table is a bit icky. I guess I could set up a separate lookup table specifically for EMS records that holds the same set of values 🤔  At the end of the day, we are going to have to deal with fallout if CRIS adds or changes values in the `injry_sev` table. That change would be so impactful to VZ ops that it obviously wouldn't fly under the radar—so I'm inclined to live with this dependency. I did [add a note to include this](https://github.com/cityofaustin/atd-data-tech/issues/22006) in forthcoming docs about our enhanced EMS data.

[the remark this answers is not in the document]
johnclary: I'm definitely open to alternative ideas on both these points 🙏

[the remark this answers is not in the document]
frankhereford: Thank you for that extra information, I think you and I share a lot of the same views and concerns. I also think that the proposed solution you have here is both

1) Here and working ✅ 
2) The current best proposal that's been put forward ✅ 

I don't have any ideas jumping out at me here. Your observation about moving the business logic into a view is compelling. The business logic has to live somewhere, and here is a fine place. I'll keep this in the back of my mind in, but as of right now, I'm on board with your solution. It's a hard problem, and this solves it.

---

You ponder the idea of cloning that FK table under the assumption that CRIS and EMS will drift in time as they update values. I think this is very wise. 

There is not a "real world" constraint operating out there that is going to keep EMS and CRIS in sync, and therefore, I think the best way to represent this ground truth in the database is to have two foreign key tables. Having them be a duplicate of each other **now** feels very "un-DRY", but I'd argue that it's actually a more accurate representation of what is happening in reality. It won't feel repetitious at all once they drift -- it will feel like you saw into the future. (Or that's my bet at least 😉)

frankhereford (approved):
Thanks for that extra background and insight into your reasoning about how you built this. You got my vote 🚢

database/migrations/default/1746807631787_patient_inj_sev_update/up.sql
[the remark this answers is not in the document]
johnclary: Thanks, Frank! I'll set up a separate lookup table. You're right—we've already seen that kind of separation entities pay off in the long run. If there's anything i regret about or current `lookups` schema, it's that we didn't establish a naming convention to make it clear when a lookup table is CRIS-owned versus some other custom set. Oh well—I'll try to do that in this PR change. 🙏

johnclary:
👋 I am re-requesting review based on Frank's feedback. Hoping this is good to go now!

chiaberry (approved):
✅  

fwiw, when I ran through the test steps, the patient_injry_sev_id was 2, but the reason was `lower(pcr_provider_impression_primary) like 'injury%')`, which is what I think we expect based on the trigger code. Just commenting because the test steps describe something different.

database/migrations/default/1746807631787_patient_inj_sev_update/up.sql
```diff
@@ -0,0 +1,187 @@
...
+        NEW.patient_injry_sev_reason = 'lower(pcr_provider_impression_primary) in (''burn'', ''cardiac - cardiac arrest'', ''cardiac - cardiac arrest - traumatic'', ''syncope'', ''unconsciousness'')';
+    when (lower(NEW.pcr_provider_impression_secondary) in ('burn', 'cardiac - cardiac arrest', 'cardiac - cardiac arrest - traumatic', 'syncope', 'unconsciousness')) then
+        NEW.patient_injry_sev_id = 1;
+        NEW.patient_injry_sev_reason = 'lower(pcr_provider_impression_secondary) in (''burn'', ''cardiac - cardiac arrest'', ''cardiac - cardiac arrest - traumatic'', ''syncope'', ''unconsciousness'')';
+    --
+    -- minor injuries
+    --
+    when (NEW.mvc_form_patient_injured_flag = 1) then
+        NEW.patient_injry_sev_id = 2;
+        NEW.patient_injry_sev_reason = 'mvc_form_patient_injured_flag = 1)';
```
mddilley, line 118: Not a huge deal but saw this reason and the next two have trailing parentheses when I was scoping out the reasons in my local DB

database/migrations/default/1746807631787_patient_inj_sev_update/up.sql
```diff
@@ -0,0 +1,187 @@
...
+for each row execute function update_ems_patient_injry_sev();
+
+--
+-- Backfill our new columns by temporarily creating an update trigger to call the new function
+--
+create or replace trigger ems_incidents_trigger_update_set_patient_injry_sev
+before update on ems__incidents
+for each row execute function update_ems_patient_injry_sev();
+
+update ems__incidents set id = id, updated_by = 'dts_automation';
```
mddilley, line 162: Is this ok that this will overwrite Xavier as the last updater on some of the production records? It makes sense to me since this is showing the last updater which is this trigger but wanted to check.

mddilley (changes requested):
Really enjoyed the discussion in this one, and I think that storing the logic behind the reasons provides a nice way to trace the severity categorization. 

I tested the UI and noticed that the new `ems_patient_injry_sev` lookup table is missing permissions that VZE needs to render the EMS table. After that update, this is 🚢 🙌

database/migrations/default/1746807631787_patient_inj_sev_update/up.sql
johnclary: There were a bunch of cruft lines—fixed. Thank you!

database/migrations/default/1746807631787_patient_inj_sev_update/up.sql
johnclary: Yeah, not ideal—we need https://github.com/cityofaustin/atd-data-tech/issues/21714!

I'll grab a snapshot of the prod data and let him know in case we want to restore his edit attributions.

mddilley: ahhhh gotcha! makes sense to me and cool to know that issue is out there 😎

johnclary:
I just ran the EMS incident import ETL locally as one last test of these triggers ✅

PR #1788, opened 2025-05-14
Misc updates to EMS and people tables

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/22558

## Testing

**Steps to test:** Local

2. Visit the EMS incident details page for incident # `24138-0534`.
3. Use the matching UI to link the EMS patient to the matching person record. 
4. Observe that both tables now have a **Transported to** column
5. Observe that the **Possible matching people** table shows `PEDALCYCLIST` in the **Travel mode** column
6. Once linked, navigate to the crash details page for the person record (crash ID `20184776`)
7. On the crash details page, observe the new **Person** number column on the **People table** and the **Unit** and **Person** number columns on **EMS Patient Care** table
8. Still on the crash details page, use the hyperlinked **Incident** number column in the **EMS Patient Care** table to navigate back to the incident details page.

---

#### Ship list

- [ ] Code reviewed


editor/types/ems.ts
```diff
@@ -18,7 +19,10 @@ export type EMSPatientCareRecord = {
   pcr_patient_age: number | null;
   pcr_patient_gender: string | null;
   pcr_patient_race: string | null;
+  pcr_transport_destination?: string | null;
   person_id: number | null;
+  // we are using Omit to avoid a circular reference
+  person?: Omit<PeopleListRow, "ems_pcr"> | null;
```
mddilley, line 25: thanks for noting this! TIL about using Omit for this way. I replicated the circular reference in my code editor and immediately see the TS error that this solves. 🙏

mddilley (approved):
Tested locally, and I see the updates and was able to map them to the diff too. Looks amazing! 🚢

editor/app/ems/[incident_number]/page.tsx
```diff
@@ -246,7 +246,7 @@ export default function EMSDetailsPage({
             records={matchingPeople ? matchingPeople : []}
             isValidating={isValidating}
             noRowsMessage="No people found"
-            header="Associated people records"
+            header="Possible matching people"
```
Comment (chiaberry) [inline editor/app/ems/[incident_number]/page.tsx:249]: does "Possible people matches" sound better? I genuinely dont know

editor/types/ems.ts
chiaberry, line 25: I understand what Omit does, but I dont understand why you had to add it here

chiaberry: ahh i should try removing Omit and seeing what the error says...

chiaberry (approved):
Nice refinements! I think a potential future enhancement could be to highlight the linked row in the "Possible matching people" table, or to shade it in some way.

editor/types/ems.ts
johnclary: Thanks for checking that! I think that _maybe_ circular references can be ok, but i believe the problem was stemming from our fancy `Path` type that itself is recursive ♾️

editor/app/ems/[incident_number]/page.tsx
Comment (roseeichelmann) [reply editor/app/ems/[incident_number]/page.tsx:249]: i think maybe i like chias idea?

roseeichelmann (approved):
changes look good, these updates will be super helpful in matching!!

editor/app/ems/[incident_number]/page.tsx
Comment (johnclary) [reply editor/app/ems/[incident_number]/page.tsx]: Same—thank you!

editor/configs/peopleColumns.tsx
```diff
@@ -70,6 +78,10 @@ export const ALL_PEOPLE_COLUMNS = {
       min: { value: 0, message: "Age cannot be negative" },
     },
   },
+  prsn_taken_to: {
+    path: "prsn_taken_to",
+    label: "Transported to",
```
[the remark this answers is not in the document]
johnclary: Correct—since it's only useful for record matching we don't need to show it on the crash details page. As an aside, I'd love to get a column visibility picker implemented for the related record tables.

johnclary:
Thanks, y'all!

editor/types/ems.ts
johnclary: Just to wrap this up—[here's](https://github.com/cityofaustin/vision-zero/blob/3af0e8e607fcf86ff01e78fc862e5ca6f60ce1ae/editor/types/utils.ts#L18-L19) the part of our `Path` type that's recursive. So having a circular reference of interfaces causes typescript to abort after some recursion limit is reached.

We technically don't need the `Omit` on both the `PeopleListRow` and `EMSPatientCareRecord`, because Omitting the self reference on either blocks the recursion, but it seemed like a good idea to future proof it.

PR #1789, opened 2025-05-14
EMS details: show an error message when saving an invalid person ID

roseeichelmann:
## Associated issues
Closes https://github.com/issues/assigned?issue=cityofaustin%7Catd-data-tech%7C22208

## Testing

**URL to test:** 
Netlify or local

**Steps to test:**
1. Go to an EMS incident. Click on the person id column of an EMS patient record to edit the input. The "Select person" button should get disabled so you cant link while also editing the person id col, the falafel menu should also disappear. Type in and save a bogus person id like `111111111`. You should get an error message indicating that the id is either invalid or in use. When you cancel and click back in to edit the input, the error message should be gone.
2. Now in your dev tools throttle your network to be offline. Type in something to the person id column and save it, you should get the "Something went wrong" error.
3. Turn network back online and test matching this record to a person in the Possible people matches table. Copy the person id, go to a completely different EMS incident, and paste that person id into one of the EMS patient record person ID cols, you should now get the same error this time because you already matched that person.
4. Test resetting the record, the person ID should nullify and the match status should return to what is was before.
5. Go to a crash details page and make sure the other editable field input validation is still working as expected -- ie. editing the speed limit to be `70a` should get an error, as should editing the unit Year column to something before 1900

---

#### Ship list

- [ ] Code reviewed


editor/components/EMSLinkRecordButton.tsx
```diff
@@ -57,21 +63,21 @@ const EMSLinkRecordButton: React.FC<
       <div className="d-flex">
         <Button
           size="sm"
-          variant={isLinkingThisRecord ? "secondary" : "primary"}
-          disabled={isLinkingInProgress && !isLinkingThisRecord}
+          variant={isInLinkMode ? "secondary" : "primary"}
+          disabled={(isLinkingInProgress && !isInLinkMode) || !!isEditingColumn}
```
roseeichelmann, line 67: i decided to disable the Select person button when you are editing the input of the person id column otherwise things can get confusing. And then lower down the falafel menu to reset/set unmatched just wont render if u are editing the input

editor/components/EditableField.tsx
```diff
@@ -108,9 +120,9 @@ export default function EditableField({
             <option value="false">No</option>
           </Form.Select>
         )}
-        {errors.value && (
+        {(errors.value || mutationError) && (
```
roseeichelmann, line 123: if there is a validation error ie. you are typing in a string in an integer input, OR if you tried to save and got a mutation error, you will get an error message

editor/configs/emsColumns.tsx
```diff
@@ -106,6 +106,12 @@ export const ALL_EMS_COLUMNS = {
     label: "Person ID",
     editable: true,
     inputType: "number",
+    getMutationErrorMessage: (error) => {
+      if (!!error) {
+        return "Person ID is invalid or in use";
```
roseeichelmann, line 111: this could be updated to have more specific messages based on what the mutation error is

editor/components/EditableField.tsx
```diff
@@ -55,8 +60,15 @@ export default function EditableField({
...
   });
 
+  // Sets the message to be shown if there was an error in the graphql mutation
+  const [mutationError, setMutationError] = useState<string | null>(null);
+
   const onSubmit = async (data: FormValues) => {
-    await onSave(data.value);
+    await onSave(data.value).catch((error) => {
+      setMutationError(
+        getMutationErrorMessage ? getMutationErrorMessage(error) : null
```
roseeichelmann, line 69: I found out that this `onSave` only runs if there are no validation errors from RHF, so this wont run for example if you type in `70a` to the speed limit col in the crash details page, RHF validates the input and catches the error before handling the `onSave`. This will only run if there are no validation errors, and then it will catch any graphql mutation errors to show as error feedback in the form.

editor/configs/emsColumns.tsx
johnclary: Love it—i think this pattern works really well!

And I love your use of `unknown` typing for `error` in this function signature. Something that would be a nice touch here would be to add more specificity  when checking the `error`, just in case we're actually suppressing some bug or exception that was not at all related to the graphql request.

something like:

```javascript
import { ClientError } from "graphql-request";

if (error instanceof ClientError) {
        // assume the problem is related to the person ID
       // ...although we could also inspect error.response.errors
        return "Person ID is invalid or in use";
} else if (error instanceof Error) {
    // Handle other errors
    return "Something went wrong"
} 
```

editor/types/types.ts
```diff
@@ -82,6 +82,9 @@ export interface ColDataCardDef<T extends Record<string, unknown>> {
    * @see https://react-hook-form.com/docs/useform/register
    */
   inputOptions?: RegisterOptions<FormValues, "value">;
+
+  /** Function that gets the message to be displayed in the case of a graphql mutation error */
+  getMutationErrorMessage?: (error: unknown) => string | null;
```
johnclary, line 87: it might be more accurate to name this `onSaveErrorHandler` — since technically the `onSave` callback doesn't have to make a mutation

editor/components/EMSLinkRecordButton.tsx
```diff
@@ -17,9 +17,15 @@ export interface EMSLinkRecordButtonProps extends Record<string, unknown> {
...
+> = ({
+  record,
+  additionalProps,
+  mutation,
+  onSaveCallback,
+  isEditingColumn,
+}) => {
   const isLinkingInProgress = !!additionalProps?.selectedEmsPcr;
-  const isLinkingThisRecord =
+  const isInLinkMode =
```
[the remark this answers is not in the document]
johnclary: I agree it was confusing. is this better?
- `isLinkingAnyRecord`
- `isLinkingThisRecord`

editor/components/EditableField.tsx
johnclary: Niiiice!

editor/components/EMSLinkRecordButton.tsx
johnclary: I like this. The disabled/visible states still feel a little uneven. What do you think about this?

After clicking **Select person** — hide the falafel menus for other rows as well?

Screenshot 2025-05-16 at 2 38 42 PM

When inputting a person ID — disable the other **Select person ** buttons and hide the falafel menus?

Screenshot 2025-05-16 at 2 38 47 PM

johnclary (commented):
Woohoo—I love how you solved this! It looks fantastic. I think you should had more targeted error handling but otherwise this is ready to gooooooo!

editor/components/EMSLinkRecordButton.tsx
roseeichelmann: it may be more trouble than its worth to try to disable/hide the buttons on all the rows when one of the rows is being edited bc edit state is determined in `RelatedRecordTableRow` so id have to somehow extract that edit status up into the parent component of `RelatedRecordTable`? im gonna push up a change where i just hide the select person button also, lmk what you think

editor/components/EMSLinkRecordButton.tsx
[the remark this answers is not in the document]
roseeichelmann: ahh okay see i still didnt understand what the diff in the two vars was, but those var names made it make sense for me! pushing up that change now tyyy

editor/configs/emsColumns.tsx
roseeichelmann: ooo good idea! i just pushed up that change, if you throttle your network to be offline you will get the something went wrong error now :pray:

editor/components/EditableField.tsx
johnclary: What do you think about adding a fallback/default error message when `getMutationErrorMessage` is null? Like instead of `null` we use the generic `Something went wrong message`? That _should_ give us error feedback across the whole app ✨ 

Edit: I did a quick test of this and we'll need a few minor tweaks to make that work universally. I'll open a follow-up issue.

editor/configs/emsColumns.tsx
johnclary: Looks great!

johnclary (approved):
Looks fantastic—thank you!

🚢

chiaberry (approved):
This is a great way to handle the different error messages, test steps worked perfectly. 🚢

PR #1791, opened 2025-05-16
v2.12.1 - Update crash-EMS trigger

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/22612

I will PR this against prod as well once we have approvals.

## Testing

**URL to test:** Local

2. Visit this crash's details: `http://localhost:3002/editor/crashes/17357278`
3. Scroll down to **EMS Patient care** and observe that there is a single EMS record with status of **Matched automatically**
4. Edit the crash's **Longitude** to be `-98.12642144`. Scroll down to the EMS table and notice that there are no EMS records listed.
5. Edit crash's **Longitude** back to the original value of `-97.676720393005`. Notice that the matched EMS record is listed in the EMS table. Use it's hyperlinked incident # to visit the incident details page.
6. Use the match UI to match the EMS record to any person record. Notice that it's status changes to **Matched by review/QA**
7. Navigate back to the crash details page, and again edit the longitude to  `-98.12642144`. Notice that the EMS record remains listed in the table.
8. Navigate back to the EMS incident details page, and use the falafel menu to "Reset" the record's status.
9. Notice that the EMS record now has a status of **Unmatched**
10. Finally, go back to the crash details and once again restore the crash to it's original **Longitude**: `-97.676720393005`. Notice that the EMS record is again listed in the table with a status of **Matched automatically**

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/database/migrations/default/1747404534187_patch_crash_ems_trigger/down.sql
+++ b/database/migrations/default/1747404534187_patch_crash_ems_trigger/down.sql
@@ -1,0 +1,111 @@
+-- Updating this trigger to always keep the matched_crash_pks up to date with automatic matching results
+-- regardless of crash match status
+CREATE OR REPLACE FUNCTION public.update_crash_ems_match()
+ RETURNS trigger
+ LANGUAGE plpgsql
+AS $function$
+DECLARE
+    matching_ems RECORD;
+    ems_record RECORD;
+    match_count INTEGER;
+    matched_crash_ids INTEGER[];
+    meters_threshold INTEGER := 1200;
+    -- 30 min threshold equates to a 60 minute window (+/- 30 minutes of crash timestamp)
+    time_threshold INTERVAL := '30 minutes';
+BEGIN
+    -- Find all EMS records near the crash location + time
+    FOR matching_ems IN (
+        SELECT 
+            e.id,
+            e.incident_received_datetime,
+            e.geometry,
+            e.crash_match_status
+        FROM 
+            ems__incidents e
+        WHERE
+            e.incident_received_datetime  >= (NEW.crash_timestamp - time_threshold)
+            AND e.incident_received_datetime  <= (NEW.crash_timestamp + time_threshold)
+            AND e.geometry IS NOT NULL
+            AND NEW.position IS NOT NULL
+            AND ST_DWithin(e.geometry::geography, NEW.position::geography, meters_threshold)
+    ) LOOP
+        -- Find all crashes which match this EMS record location + time
+        SELECT 
+            ARRAY(
+                SELECT c.id
+                FROM crashes c
+                WHERE 
+                    c.is_deleted = false
+                    and not c.is_temp_record
+                    AND matching_ems.incident_received_datetime  >= (c.crash_timestamp - time_threshold)
+                    AND matching_ems.incident_received_datetime  <= (c.crash_timestamp + time_threshold)
+                    AND c.position IS NOT NULL
+                    AND ST_DWithin(matching_ems.geometry::geography, c.position::geography, meters_threshold)
+            ) INTO matched_crash_ids;
+            
+        -- Get the count from the array length (handling when array_length is null as 0)
+        SELECT COALESCE(array_length(matched_crash_ids, 1), 0) INTO match_count;
+
+        -- For records that have been manually matched we want to only update the automated match column
+        IF matching_ems.crash_match_status === 'matched_by_manual_qa' THEN
+          IF match_count = 0 THEN
+            UPDATE ems__incidents 
+            SET matched_crash_pks = NULL
+            WHERE id = matching_ems.id;
+          ELSE 
+            -- if the match count is more than one we will update the matched_crash_pks
+            UPDATE ems__incidents 
+            SET matched_crash_pks = matched_crash_ids
+            WHERE id = matching_ems.id;
+          END IF;
+        -- If the record has not been manually matched
+        ELSE
+          IF match_count = 0 THEN
+              UPDATE ems__incidents 
+              SET crash_pk = NULL,
+                  crash_match_status = 'unmatched',
+                  matched_crash_pks = NULL
+              WHERE id = matching_ems.id;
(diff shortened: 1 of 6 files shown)


database/migrations/default/1747404534187_patch_crash_ems_trigger/down.sql
```diff
@@ -0,0 +1,111 @@
...
+                    AND c.position IS NOT NULL
+                    AND ST_DWithin(matching_ems.geometry::geography, c.position::geography, meters_threshold)
+            ) INTO matched_crash_ids;
+            
+        -- Get the count from the array length (handling when array_length is null as 0)
+        SELECT COALESCE(array_length(matched_crash_ids, 1), 0) INTO match_count;
+
+        -- For records that have been manually matched we want to only update the automated match column
+        IF matching_ems.crash_match_status === 'matched_by_manual_qa' THEN
+          IF match_count = 0 THEN
```
johnclary, line 51: This is another bug I uncovered while testing. This block never ran because match count is never `0` here. I moved this handling down to the `IF TG_OP = 'UPDATE' THEN` section.

johnclary:
👋 I didn't add this to the test instructions, but I'm going to run the latest CRIS import locally to make sure that's working as expected

database/migrations/default/1747404534187_patch_crash_ems_trigger/up.sql
```diff
@@ -0,0 +1,118 @@
...
+              or NEW.is_deleted is TRUE
+              OR (NOT ST_DWithin(geometry::geography, NEW.position::geography, meters_threshold))
+          )
+          AND 
+            -- ignore manual qa matches
+            crash_match_status != 'matched_by_manual_qa';
+       
+        -- update the matched_crash_pks column only for manually qa matches
+        UPDATE ems__incidents
+        SET matched_crash_pks = NULL
```
mddilley, line 102: @johnclary I follow this whole function, but I have one question about setting `matched_crash_pks` to `null` here but not in the block on lines 50-53. I think that I'm getting lost on when we want to use `matched_crash_pks` as a holding place for recovering previously matches or when we don't want to.

frankhereford (approved):
Great instructions, and everything worked as expected. This is ready to ship -- thank you for such a quick fix.

Maybe it's something I have setup funny, but I get the crash URL in my local stack at http://localhost:3002/crashes/17357278 without the `editor` portion of the path.

The match UI is great, and this is a nothing, but FWIW, I was unable to figure out how to scroll the matching UI table to the right without having to open up the mac and use the trackpad. I wonder if anyone else has any issue when using just a traditional mouse -- I'm going to try to remember to ask in the dev sync.

Also, it'd be neat to get the moped view bot in effect in VZ to help see the diffs of these more complex views. Maybe that will help in the future for this sort of thing.

database/migrations/default/1747404534187_patch_crash_ems_trigger/up.sql
roseeichelmann: i believe the logic here is that `matched_crash_pks` should always stay up to date with what our automated matching finds, and in this case this incident doesnt have any matches, so `matched_crash_pks` would update to null. but this wouldnt affect the manual value in `crash_pk` that the VZ team manually matched. in lines 50-53 the record did have matching crash ids, so thats why `matched_crash_pks` was updated there.

mddilley: ohhh now i see it. the WHERE conditions in this update statement are how we know there are no automated matches - that is what I was missing. 🙏 Thanks for helping me through this because I was looking earlier in the code and not seeing that kind of check.

mddilley (approved):
All the test steps check out, and the updates to the function make sense to me! 🚢

roseeichelmann (approved):
this looks great, super thorough test instructions, thank u for fixing my oopsie :heavy_equals_sign:

database/migrations/default/1747404534187_patch_crash_ems_trigger/up.sql
johnclary: > when we want to use `matched_crash_pks as a holding` place for recovering previous matches and when we don't want to.

@mddilley Our intention is that `matched_crash_pks` _always_ holds the automatically matched crash PKs regardless of any user edits. This way we can reset to the original/automatic state of the EMS record. So the trigger is keeping this column in sync with any edits that are happening to the crash.

johnclary: I added this thread to https://github.com/cityofaustin/atd-data-tech/issues/22006

chiaberry (approved):
:ship: !

chiaberry:
That's a good idea Frank, to also see if we could get that dev experience not only on views like in moped but also on triggers
https://github.com/cityofaustin/atd-data-tech/issues/18791

frankhereford:
@johnclary wrote:

> 1. Hold down the Shift key while scrolling the mouse wheel - this will convert vertical scrolling to horizontal scrolling

😲 **I've sat in front of these computers for 30 years, and today I learn that shift+scroll → horizontal-scroll. What other deep magic do have?!?!**

My mouse is a [Corsair Nightswoard RGB Gaming Mouse](https://www.amazon.com/dp/B07QX9C9WH) and 1) I wouldn't buy it again and it lacks that Logitech wheel-tilt feature. I've used mice with that before though, I totally know what you mean. It'd do the trick.

I don't think it's a big deal since it appears that I may be the only user who doesn't always have a track-pad handy, but I did have to puzzle for a second where the UI's button when matching an EMS record because my window was just wide enough to have the edge of the table land between columns. The button was well hidden in my particular, isolated case.

Thank you also for the `/editor` tip; I've made that configuration here now too. 🙏

johnclary:
Yeah, I'm not really sure how to get away from side-scrolling in this app. It would be nice if the buttons were stuck the right and always visible. We could also move them to the left 🤔

PR #1795, opened 2025-05-19
Adds a non-CR3-EMS match trigger

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/21863

## Testing

**URL to test:** 

**Steps to test:**

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/database/metadata/databases/default/tables/public_ems__incidents.yaml
+++ b/database/metadata/databases/default/tables/public_ems__incidents.yaml
@@ -2,6 +2,9 @@ table:
   name: ems__incidents
   schema: public
 object_relationships:
+  - name: atd_apd_blueform
+    using:
+      foreign_key_constraint_on: atd_apd_blueform_case_id
   - name: crash
     using:
       foreign_key_constraint_on: crash_pk
@@ -22,6 +25,7 @@ select_permissions:
     permission:
       columns:
         - apd_incident_numbers
+        - atd_apd_blueform_case_id
         - crash_match_status
         - crash_pk
         - created_at
@@ -32,7 +36,9 @@ select_permissions:
         - incident_problem
         - incident_received_datetime
         - matched_crash_pks
+        - matched_non_cr3_case_ids
         - mvc_form_position_in_vehicle
+        - non_cr3_match_status
         - patient_injry_sev_id
         - pcr_patient_age
         - pcr_patient_gender
@@ -50,6 +56,7 @@ select_permissions:
     permission:
       columns:
         - apd_incident_numbers
+        - atd_apd_blueform_case_id
         - crash_match_status
         - crash_pk
         - created_at
@@ -60,7 +67,9 @@ select_permissions:
         - incident_problem
         - incident_received_datetime
         - matched_crash_pks
+        - matched_non_cr3_case_ids
         - mvc_form_position_in_vehicle
+        - non_cr3_match_status
         - patient_injry_sev_id
         - pcr_patient_age
         - pcr_patient_gender
@@ -78,6 +87,7 @@ select_permissions:
     permission:
       columns:
         - apd_incident_numbers
+        - atd_apd_blueform_case_id
         - crash_match_status
         - crash_pk
         - created_at
@@ -88,7 +98,9 @@ select_permissions:
         - incident_problem
         - incident_received_datetime
         - matched_crash_pks
+        - matched_non_cr3_case_ids
         - mvc_form_position_in_vehicle
+        - non_cr3_match_status
         - patient_injry_sev_id
         - pcr_patient_age
         - pcr_patient_gender
--- a/database/migrations/default/1747681652276_non_cr3_ems_trigger/down.sql
+++ b/database/migrations/default/1747681652276_non_cr3_ems_trigger/down.sql
@@ -1,0 +1,9 @@
+drop trigger if exists non_cr3_insert_ems_match_trigger on atd_apd_blueform;
+
+drop trigger if exists non_cr3_update_ems_match_trigger on atd_apd_blueform;
+
+drop function public.update_noncr3_ems_match;
+
+alter table ems__incidents
+    drop column non_cr3_match_status,
+    drop column matched_non_cr3_case_ids;
--- a/database/migrations/default/1747681652276_non_cr3_ems_trigger/up.sql
+++ b/database/migrations/default/1747681652276_non_cr3_ems_trigger/up.sql
@@ -1,0 +1,153 @@
+alter table ems__incidents
+add column non_cr3_match_status text not null default 'unmatched',
+add column matched_non_cr3_case_ids integer[],
+add constraint ems__incidents_non_cr3_match_status_check
+check (
+    non_cr3_match_status in (
+        'unmatched',
+        'matched_by_automation',
+        'multiple_matches_by_automation',
+        'matched_by_manual_qa',
+        'unmatched_by_manual_qa'
+    )
+);
+
+comment on column ems__incidents.non_cr3_match_status is 'The status of the non-CR3 crash record match';
+comment on column ems__incidents.matched_non_cr3_case_ids is 'The IDs of non-cr3 crases that were found to match this record. Set via trigger, always kept up to date regardless of EMS match status.';
+
+-- todo: any other indexes needed?
+CREATE INDEX ems__incidents_non_cr3_match_status_index on public.ems__incidents (non_cr3_match_status);
+
+create or replace function public.update_noncr3_ems_match()
+returns trigger
+language plpgsql
+as $function$
+DECLARE
+    matching_ems RECORD;
+    ems_record RECORD;
+    match_count INTEGER;
+    matched_case_ids INTEGER[];
+    meters_threshold INTEGER := 1200;
+    -- 30 min threshold equates to a 60 minute window (+/- 30 minutes of crash timestamp)
+    time_threshold INTERVAL := '30 minutes';
+BEGIN
+    -- Find all EMS records near the crash location + time
+    FOR matching_ems IN (
+        SELECT 
+            e.id,
+            e.incident_received_datetime,
+            e.geometry,
+            e.non_cr3_match_status
+        FROM 
+            ems__incidents e
+        WHERE
+            e.incident_received_datetime  >= (NEW.case_timestamp - time_threshold)
+            AND e.incident_received_datetime  <= (NEW.case_timestamp + time_threshold)
+            AND e.geometry IS NOT NULL
+            AND NEW.position IS NOT NULL
+            AND ST_DWithin(e.geometry::geography, NEW.position::geography, meters_threshold)
+    ) LOOP
+        -- Find all non-cr3 crashes which match this EMS record location + time
+        SELECT 
+            ARRAY(
+                SELECT c.case_id
+                FROM atd_apd_blueform c
+                WHERE 
+                    c.is_deleted = false
+                    AND matching_ems.incident_received_datetime  >= (c.case_timestamp - time_threshold)
+                    AND matching_ems.incident_received_datetime  <= (c.case_timestamp + time_threshold)
+                    AND c.position IS NOT NULL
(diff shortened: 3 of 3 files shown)


database/migrations/default/1748878240290_non_cr3_ems_trigger/up.sql
```diff
@@ -0,0 +1,232 @@
...
+        'unmatched',
+        'matched_by_automation',
+        'multiple_matches_by_automation',
+        'matched_by_manual_qa',
+        'unmatched_by_manual_qa'
+    )
+);
+
+comment on column ems__incidents.non_cr3_match_status is 'The status of the non-CR3 crash record match';
+comment on column ems__incidents.matched_non_cr3_case_ids is 'The IDs of non-CR3 crases that were found to match this record. Set via trigger, always kept up to date regardless of EMS match status.';
```
roseeichelmann, line 19: typo "crases" instead of "crashes"

database/migrations/default/1748878240290_non_cr3_ems_trigger/up.sql
```diff
@@ -0,0 +1,232 @@
...
+                    UPDATE ems__incidents 
+                    SET atd_apd_blueform_case_id = NULL,
+                        non_cr3_match_status = 'unmatched',
+                        matched_non_cr3_case_ids = NULL
+                    WHERE id = ems_record.id;
+                ELSIF remaining_case_ids_count = 1 THEN
+                    -- One non-cr3 remaining: match it to the EMS record
+                    raise debug 'Matching EMS incident # %, ID % to non-cr3 case ID %', ems_record.incident_number, ems_record.id, updated_matched_case_ids[1];
+                    UPDATE ems__incidents 
+                    SET atd_apd_blueform_case_id = updated_matched_case_ids[1],
```
roseeichelmann, line 193: if `remaining_case_ids_count = 1` then doesnt that mean there is only one `case_id` for this record, so this should be `updated_matched_case_ids [0]`?

database/migrations/default/1748878240290_non_cr3_ems_trigger/up.sql
```diff
@@ -0,0 +1,232 @@
...
+    --
+    IF TG_OP = 'UPDATE' THEN
+        --
+        -- Check if there are any EMS records that reference this non-cr3
+        --
+        SELECT COUNT(*) INTO ems_cleanup_count
+        FROM ems__incidents 
+        WHERE NEW.case_id = ANY(matched_non_cr3_case_ids);
+        
+        raise debug '% EMS records need to be checked for possible update', ems_cleanup_count;
```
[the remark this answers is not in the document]
roseeichelmann: ohh darn so in the [CR3 matching trigger](https://github.com/cityofaustin/vision-zero/blob/ba63d802a8d59ea9c48f075b6765a8a5ea05a5e4/database/migrations/default/1747404534187_patch_crash_ems_trigger/up.sql#L81) we are not handling the case where the extant EMS record still has 1 or more valid CR3 crash matches,  we are just assuming that the EMS record was only linked to the one crash record which has been updated, and thus updating everything to null.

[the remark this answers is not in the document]
johnclary: Correct ☹️

roseeichelmann (commented):
thanks for the super thorough test instructions. the test steps all checked out, and i reviewed the code thoroughly and it looks great to me. bummer about the other edge case you realized we forgot to handle, but im glad you did!! this stuff is hairy..

just have one comment about where you have `updated_matched_case_ids [1]`, other than that im ready to approve!

database/migrations/default/1748878240290_non_cr3_ems_trigger/up.sql
johnclary: 'twas a TIL for me that Postgres uses one-based array indexing. `updated_matched_case_ids[1]` grabs the first element!

roseeichelmann: :open_mouth: i never would have guessed that. now im wondering if ive ever indexed incorrectly w Postgres :thinking:

database/migrations/default/1748878240290_non_cr3_ems_trigger/up.sql
```diff
@@ -0,0 +1,232 @@
...
+            e.geometry,
+            e.non_cr3_match_status
+        FROM 
+            ems__incidents e
+        WHERE
+            e.incident_received_datetime  >= (NEW.case_timestamp - time_threshold)
+            AND e.incident_received_datetime  <= (NEW.case_timestamp + time_threshold)
+            AND e.geometry IS NOT NULL
+            AND NEW.position IS NOT NULL
+            AND ST_DWithin(e.geometry::geography, NEW.position::geography, meters_threshold)
```
frankhereford, line 59: I guess i'm from a pre-`geography` time in postGIS when things were only ever `geometry`s, and I always think about distances as being defined by the projection. That said, it's cool here to see some of that projection complexity hidden by postGIS via the `geography` data type with its default distance of meters. Thank you postGIS because I know the math of calculating distances based on WGS84 unprojected latitude, longitude pairs is nothing to sneeze at. I am a poor candidate to become a flat-earther, even though the math behind their geography is deeply appealing.

When I ponder a crash 100 meters from the north pole being in match range of every single line of longitude, I really appreciate what `geography` does.

database/migrations/default/1748878240290_non_cr3_ems_trigger/up.sql
```diff
@@ -0,0 +1,232 @@
...
+                --
+                IF ems_record.non_cr3_match_status = 'matched_by_manual_qa' THEN
+                    -- For manually matched records, only update the case_id array
+                    raise debug 'Updating matched_non_cr3_case_ids for manually matched EMS incident # %, ID % to %', ems_record.incident_number, ems_record.id, NULLIF(updated_matched_case_ids, '{}');
+                    UPDATE ems__incidents
+                    SET matched_non_cr3_case_ids = NULLIF(updated_matched_case_ids, '{}')
+                    WHERE id = ems_record.id;
+                ELSIF remaining_case_ids_count = 0 THEN
+                    -- No other non-cr3s matched to this EMS record
+                    raise debug 'Updating EMS incident # %, ID % as `umatched`', ems_record.incident_number, ems_record.id;
```
frankhereford, line 183: Typo here on `unmatched`

frankhereford (approved):
It's stellar work John, it's really great. Tested out and reads a dream.

⚠️ I bumped into a problem during testing this because I believe we are using two different DB clients with different date handling behavior. Until I figured out what I think is up, nothing would match for me. 

The testing instructions include dates in the form of `2025-03-31 06:56:49` without a timezone declaration, and I needed to use the time `2025-03-31 01:56:49.000 -0500` borrowed from the `ems__incidents`' records as shown in the very first query. 

I say that for others who may follow behind me with DBeaver or if you may be concerned that my testing was potentially invalidated by the change I had to make to your testing queries. 

Thank you for this work, it's a ton and it really is ✨ .

database/migrations/default/1748878240290_non_cr3_ems_trigger/up.sql
mddilley: +1 - i feel like this is going to be a TIL in the future many times for me 😅 Feels like i should know this already

mddilley (approved):
All of the test steps worked as expected, and I read through the code too. Nothing jumped out at me that didn't have comments already. 

Cool to see some of the existing matching patterns come into play here, and it looks tricky overall so I'm glad to it come together! 🚢 🙌

mddilley:
Unrelated to this PR, but I saw this error returned from the crashes query (CR3 and non-CR3) on the locations page. I see it on the main branch too but it is obviously working in staging and prod. I'll debug more tomorrow - probably something going on with my local stack.

Screenshot 2025-06-11 at 7 12 46 PM

database/migrations/default/1748878240290_non_cr3_ems_trigger/up.sql
frankhereford: ## TLDR

Kinda, sorta -- in a sentence, it's because meters is the "default" length unit of the world and we're letting postGIS handle a ton of decisions in the background when using `geography`. In our part of the world and the size of our city -- this is totally fine.

But .. the full answer is actually a lot of stuff, and but I find it really interesting and if you're curious -- 

## Let's talk maps

We don't actually use meters normally here in Austin, we use feet .. so what gives. It's a chain of things: 

We (vz DB) store our data as unprojected latitude and longitude, (-90° through 90°, -180° - 180°). These coordinates represent a location on a **sphere**, which is a non-euclidian surface. It's a globe, not a flat map.

This globe-ness has implications on geometry functions. An example of how this coordinate system is not like coordinate systems we're used to working with: if you draw a triangle along the surface of a sphere with straight lines, the sum of the angles will be > 180°. Or, a more earth-y example of how latitude and longitude don't act like regular coordinates: if you stand on the north pole, every direction you can travel is in the negative-y direction, which is a property you can not find for any point on a 2D plane. 

So - the solution that geographers use is to take the 3D latitude and longitude grid and "project" it down onto a 2D plane. There are innumerable ways to do this, but for example, imagine a globe that is made of glass with black lines for borders. If you were to put a point light source in the smack middle inside the sphere at the "center of the earth" and cast shadows of the black lines out onto a cylinder wrapped around the globe and trace the shadows of those lines onto the cylinder -- now unroll that cylinder so it lays flat on a table like a map -- you've got a distorted, **projected** map of the earth, but now on a normal, 2D, euclidian surface with X,Y coordinates that have the properties of geometry we learned in grade-school.  

All projections incur a distortion of at least one of the following three properties: area (size), shape (angles), or scale (distances). 

So, depending on what job you want to accomplish with your map, you pick a projection that doesn't distort the property of the map you want to use. 

For example, Mercator projection (a la web maps) are great for measuring distance and angles, but not size (Greenland is not the same size as Africa problem). Sailors (and web developers) use this one.

Another one you see is conical projections -- like the UN logo, where all the shapes have the right relative area to one another. 

A final example is `NAD83/Texas Central (ftUs)`, where we live. This projection uses US feet as its distance unit as its x,y units. A particular origin for the coordinate plane is defined by the "datum", and this origin and set of transformations are optimized to most accurately represent the terrain and geographic features found in [central texas, particularly central texas at this latitude] (https://spatialreference.org/ref/epsg/2277/). 

Anyway - there are tons of projections, and the weird ones are /really/ weird.  

In GIS, a projection is defined as a set of mathematical transformations that you can use to go back and forth between unprojected and projected points. The idea is you pick a projection tailored to the region you want to measure with the units you want to measure in, project into it, and then you can use normal Pythagorean theorem math to calculate distance.  And it's really only after you did all that crazy math that you can really do that distance calculation -- it was not valid on a non-euclidain plane, which back to the beginning, is how we store our location data.

So - back to postgresql - `geometry` is datatype that it's always had and the fundamental data type that you build geometry out of is a x,y pairs. All of the [various methods](https://postgis.net/docs/manual-3.5/reference.html) are assuming that the x,y pairs in the geometries are on a Euclidean, flat plane, like distance is Pythagorean theorem, to borrow from above.  

As a developer, one has to think about what projection your data was stored in and project it to and from different projections based on what you were using it for. For example, to make a circle of radius 500 feet from a given point, you would project your lat,long data into a projection so you could do that distance calculation / circle drawing and only then you could project that answer back out into whatever you wanted, maybe lat, long again so you could draw it with mapbox. It's a lot to keep track of. 

And maybe a decade ago? something like that -- postgis added a second fundamental data type called `geography` which is to hold unprojected latitude and longitude points. A subset of the postGIS methods were given a second version to deal with `geography` and some remain `geometry` based on restrictions imposed by math and dimensionality and other considerations. 

One of the methods that has a `geography` variant is `ST_DWithin()` and it's doing all that gnarly math under the hood behind the scenes to tell you if one point is within so many linear meters of another. It's a lot of stuff being hidden behind the scenes, and it's making a lot of choices for you (meters as the unit, for example), but it sure is easy.

PR #1810, opened 2025-06-20
Add Non-CR3 crash match fields to EMS list

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/23068

## Testing

**URL to test:** 
Local

**Steps to test:**
1. Go to the EMS page, see the two new non-cr3 columns at the end of the table
2. Test sorting by these columns
3. Click on the Filters button to get the dropdown with filters. You may need to click the Reset button to see the new filter card. Test filtering by Non-CR3 match status 
4. There arent any multiple matches by automation yet, so run the following commands to test that filter which sets these two non cr3 records to match the same EMS incident:

``` sql
UPDATE atd_apd_blueform
SET
    case_timestamp = '2025-03-31 06:56:49',
    latitude = 30.38802,
    longitude = -97.759782,
    is_deleted = FALSE
WHERE
    case_id = 250050598;

UPDATE atd_apd_blueform
SET
    case_timestamp = '2025-03-31 06:56:49',
    latitude = 30.38802,
    longitude = -97.759782,
    is_deleted = false
WHERE
    case_id = 250050225;
```
5. Now after running the above command, when you turn on the Non CR3 multiple matches filter you will get results

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/package.json
+++ b/editor/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vision-zero-editor",
-  "version": "2.14.0",
+  "version": "2.15.0",
   "description": "ATD Vision Zero Editor",
   "author": "TPW Data & Technology Services",
   "license": "MIT",
--- a/editor/testing.md
+++ b/editor/testing.md
@@ -61,6 +61,7 @@ The below features should be tested with each role. Features with role-based acc
 - Crash map: crash map displays crash location with nearmap aerials
 - Crash map: edit crash location by dragging map
 - Crash map: edit crash location by keying in lat/lon
+- Observe in change log that council district, jurisdiction, APD sector, engineer area update when crash location is edited to a distant position
 - Crash map: in edit mode, use the address search to find a location within Austin metro area
 - Crash map: verify **Location ID** updates when crash is moved to another intersection
 - Crash map: validation restricts keying in lat/lon with alpha characters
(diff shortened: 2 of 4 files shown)


editor/configs/emsColumns.tsx
```diff
@@ -38,6 +38,17 @@ export const ALL_EMS_COLUMNS = {
...
   },
+  non_cr3_match_status: {
+    path: "non_cr3_match_status",
+    label: "Non-CR3 match status",
+    valueFormatter: formatCrashMatchStatus,
+    sortable: true,
+  },
+  atd_apd_blueform_case_id: {
+    path: "atd_apd_blueform_case_id",
+    label: "Non-CR3 Case ID",
```
roseeichelmann, line 49: Issue said to use this column, but just to double check, we dont want to use the `matched_non_cr3_case_ids` column incase more than one were matched? @johnclary

johnclary: I suppose we could add `matched_non_cr3_case_ids` and `matched_crash_pks` to this table and make them hidden by default. That might be nice to have here?

mddilley (approved):
Tested locally, and I see the new columns and the new filters all working as described. 🚢 💯

PR #1812, opened 2025-06-26
Create UI for non-CR3 -> EMS matching

roseeichelmann:
## Testing

**URL to test:** 

**Steps to test:**

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/app/ems/[incident_number]/page.tsx
+++ b/editor/app/ems/[incident_number]/page.tsx
@@ -10,6 +10,7 @@ import {
   GET_EMS_RECORDS,
   GET_MATCHING_PEOPLE,
   GET_UNMATCHED_EMS_CRASHES,
+  GET_NON_CR3_CRASHES,
   UPDATE_EMS_INCIDENT,
   UPDATE_EMS_INCIDENT_CRASH_AND_PERSON,
 } from "@/queries/ems";
@@ -22,10 +23,13 @@ import EMSLinkToPersonButton, {
   EMSLinkToPersonButtonProps,
 } from "@/components/EMSLinkToPersonButton";
 import { emsMatchingPeopleColumns } from "@/configs/emsMatchingPeopleColumns";
+import { emsNonCR3Columns } from "@/configs/nonCR3Columns";
 import { PeopleListRow } from "@/types/peopleList";
 import { FaTruckMedical } from "react-icons/fa6";
 import { parseISO, subHours, addHours } from "date-fns";
 import { Crash } from "@/types/crashes";
+
+console.log(emsNonCR3Columns);
 
 export default function EMSDetailsPage({
   params,
@@ -61,7 +65,6 @@ export default function EMSDetailsPage({
    * Use the first EMS record as the "incident"
    */
   const incident = ems_pcrs?.[0];
-
   /**
    * Hook which manages which related crash PKs we should
    * use to query people records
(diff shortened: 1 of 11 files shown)


editor/app/ems/[incident_number]/page.tsx
```diff
@@ -255,6 +307,21 @@ export default function EMSDetailsPage({
...
+      <Row>
+        <Col sm={12} className="mb-3">
+          <RelatedRecordTable
+            records={nonCR3Crashes ? nonCR3Crashes : []}
+            isValidating={isValidatingNonCR3}
+            noRowsMessage="No crashes found"
+            header="Possible Non-CR3 matches"
+            columns={emsNonCR3Columns}
+            mutation=""
+            onSaveCallback={onSaveCallback}
```
Comment (roseeichelmann) [inline editor/app/ems/[incident_number]/page.tsx:319]: as of right now this table doesnt actually have any editable fields so this `onSaveCallback` isnt doing anything... but neither is the `onSaveCallback` in the DataCard above for the "Summary"? also `isValidating` isnt doing anything here or in the DataCard component either :thinking: 

could update `RelatedRecordTable` and `DataCard` to not require these props?

editor/configs/nonCR3Columns.ts
```diff
@@ -0,0 +1,26 @@
+import { ColDataCardDef } from "@/types/types";
+import { NonCR3Record } from "@/types/nonCr3";
+
+export const nonCR3Columns = {
+  address: {
+    path: "address",
+    label: "Address",
+    sortable: true,
+  },
+  case_timestamp: {
```
johnclary, line 10: wold you add a formatter to this so that it renders in the same format as the crash date and incident date?

editor/types/nonCr3.ts
```diff
@@ -125,3 +127,9 @@ export type NonCr3ValidationError = {
   fieldName: string;
   message: string;
 };
+
+export type NonCR3Record = {
```
johnclary, line 131: I was going to suggest that you construct this type using `<Partial<NonCr3Upload>>`, but it does feel right to have a separate type for this post-import record.

editor/app/ems/[incident_number]/page.tsx
```diff
@@ -255,6 +307,21 @@ export default function EMSDetailsPage({
           />
         </Col>
       </Row>
+      <Row>
+        <Col sm={12} className="mb-3">
+          <RelatedRecordTable
+            records={nonCR3Crashes ? nonCR3Crashes : []}
+            isValidating={isValidatingNonCR3}
+            noRowsMessage="No crashes found"
+            header="Possible Non-CR3 matches"
```
Comment (johnclary) [inline editor/app/ems/[incident_number]/page.tsx:316]: Would you change the `noRowsMessage` to **No non-CR3 crashes found** and change the header casing so that  **Non-CR3** is **non-CR3**?

editor/app/ems/[incident_number]/page.tsx
Comment (johnclary) [reply editor/app/ems/[incident_number]/page.tsx:319]: thanks for flagging this. definitely some code smell here. i wonder if we've not fixed this because it will cause some typing headaches, but i'm all for making those props optional if you want to do that 🙏

johnclary (commented):
Rose, this looks really good! Thanks for putting these pieces in place. I left some minor feedback, and have two other requests:

1. What do you think about squeezing [this snackoo](https://github.com/cityofaustin/atd-data-tech/issues/23370) into this PR
2. Would you you update this section of [the test docs](https://github.com/cityofaustin/vision-zero/blob/main/editor/testing.md#ems-incident-details---emsincident-number) with some test steps for Non-CR3s? You'll want to pull in `main` because I've updated that testing doc since you branched.

editor/app/ems/[incident_number]/page.tsx
```diff
@@ -248,13 +296,26 @@ export default function EMSDetailsPage({
             noRowsMessage="No people found"
             header="Possible people matches"
             columns={emsMatchingPeopleColumns}
-            mutation=""
+            mutation={UPDATE_PERSON}
```
Comment (roseeichelmann) [inline editor/app/ems/[incident_number]/page.tsx:299]: we didnt have a mutation for this even though some of the columns are editable, you can test on staging and see that when you try to edit one of the editable columns you get an error. so i updated this, unless we dont want to allow editing people records from ems incidents pages? i think its useful though

johnclary:
I've pushed up a few minor edits to resolve conflicts. In the process of doing that I think I erased some work Rose did to make `onSaveCallback` optional. And so I fixed that as well.

PR #1820, opened 2025-07-25
Ability to edit multiple address fields at the same time

roseeichelmann:
## Associated issues
Closes https://github.com/issues/assigned?issue=cityofaustin%7Catd-data-tech%7C17308

## Testing

**URL to test:** 

**Steps to test:**

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/components/EditCrashAddressModal.tsx
+++ b/editor/components/EditCrashAddressModal.tsx
@@ -1,5 +1,15 @@
 import Modal from "react-bootstrap/Modal";
-import { Form } from "react-bootstrap";
+import { Form, Col, Row } from "react-bootstrap";
+import Button from "react-bootstrap/Button";
+import { useForm, SubmitHandler } from "react-hook-form";
+import { LookupTableOption } from "@/types/relationships";
+import { UPDATE_CRASH } from "@/queries/crash";
+import { useMutation } from "@/utils/graphql";
+import { crashDataCards } from "@/configs/crashDataCard";
+import { useLookupQuery, useQuery } from "@/utils/graphql";
+import { Crash } from "@/types/crashes";
+import { useMemo, Dispatch, SetStateAction } from "react";
+
 // import Button from "react-bootstrap/Button";
 
 interface EditCrashAddressModalProps {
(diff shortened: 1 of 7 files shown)


editor/components/CrashSwapAddressButton.tsx
```diff
@@ -1,58 +1,55 @@
...
+      rpt_sec_street_desc: currentValues.rpt_street_desc,
+      rpt_sec_road_part_id: currentValues.rpt_road_part_id,
+      rpt_sec_rdwy_sys_id: currentValues.rpt_rdwy_sys_id,
+      rpt_sec_hwy_num: currentValues.rpt_hwy_num,
+    };
+
+    // Set the new swapped values in the form state
+    Object.entries(swappedValues).forEach(([key, value]) => {
+      // Setting shouldDirty on setValue allows us to enable the save button when addresses are swapped
+      setValue(key as keyof AddressFormInputs, value, { shouldDirty: true });
```
roseeichelmann, line 46: this was a cool [TIL](https://github.com/react-hook-form/react-hook-form/issues/72#issuecomment-656500986), previously if you made and edit and saved and then opened modal and reverted the edit, the form thought it wasnt dirty so the save button was disabled

editor/components/EditCrashAddressModal.tsx
```diff
@@ -0,0 +1,249 @@
...
+    });
+
+    onSaveCallback();
+  };
+
+  /**
+   * Component that renders the form field based on the input type
+   * of the column, such as text or select
+   */
+  const FormField = ({ col }: FormColumnProps) => {
```
roseeichelmann, line 152: i didnt really want to pull this out into its own file but what do yall think?

editor/components/EditCrashAddressModal.tsx
```diff
@@ -0,0 +1,249 @@
...
+            size="sm"
+            disabled={!selectOptions}
+          >
+            <option value="">Select...</option>
+            {selectOptions?.map((option) => (
+              <option key={option.id} value={String(option.id)}>
+                {option.label}
+              </option>
+            ))}
+          </Form.Select>
```
roseeichelmann, line 187: i didnt handle any form validation or error handling in this PR, but i think thats out of scope

editor/components/EditCrashAddressModal.tsx
```diff
@@ -0,0 +1,255 @@
...
+      show={show}
+      onHide={() => {
+        reset();
+        setShowEditAddressModal(false);
+      }}
+      size="lg"
+    >
+      <Modal.Header>
+        <Modal.Title className="me-2">Edit crash address</Modal.Title>
+        <CrashSwapAddressButton getValues={getValues} setValue={setValue} />
```
johnclary, line 204: Would you make this button aligned to the right side of the modal header? It now looks off to me on the left.

johnclary (changes requested):
This is looking so great, Rose! I really like the way you approached the react-hook-form components + typings. I ran into a few bugs that hopefully aren't too hard to sort out 😄

editor/components/EditCrashAddressModal.tsx
```diff
@@ -0,0 +1,255 @@
...
+        query,
+        // we don't need to refetch lookup table options
+        options: { revalidateIfStale: false },
+        typename,
+      });
+    const inputType = col.inputType;
+    return (
+      <div className="mb-2">
+        {(inputType === "text" || inputType === "number") && (
+          <Form.Control
```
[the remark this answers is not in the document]
roseeichelmann: i went ahead and added the data-1p-ignore attribute which didnt fix it, but moving the `FormField` component into its own component DID fix this, i think it had something to do with it rerendering?? and now that doesnt happen :sparkles:

editor/components/EditCrashAddressModal.tsx
roseeichelmann: i went ahead and moved it into its own component called `AddressFormField` bc it fixed the error you described in another comment about the field losing focus. its true we have repeated this code a few times now :grimacing: i think kicking the can and addressing it app-wide is a good idea

also i see your point about the monolithic components folder, its getting hugeeee. could be another good improvement for the codebase in the future, to reorganize our folder structure

editor/styles/global.scss
```diff
@@ -83,6 +84,29 @@ $dark: $night;
   }
 }
 
+[data-bs-theme="light"] .edit-address-button {
```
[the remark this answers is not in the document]
roseeichelmann: simple fix, i just needed to add !important to that part of the css. this is why its so great we all use different web browsers!

editor/components/EditCrashAddressModal.tsx
roseeichelmann: done, you can test this by changing something in the `crashesColumns` config to make the lookup table search not work, like changing the table name to something bogus

editor/utils/formHelpers.ts
```diff
@@ -185,7 +185,6 @@ export const commonValidations = {
     // Allow empty string for nullable fields
     if (!value) return true;
     return /^\d+$/.test(value) || "This field must be a number";
-    return true;
```
roseeichelmann, line 188: unreachable return here

editor/configs/peopleColumns.tsx
```diff
@@ -33,7 +33,7 @@ export const ALL_PEOPLE_COLUMNS = {
       labelColumnName: "label",
     },
     valueRenderer: (record) => {
-      const value = record.injry_sev.label;
+      const value = record.injry_sev?.label;
```
roseeichelmann, line 36: this is random but fixed a bug where if you edit the injury severity to be null be saving the Select... option, the crash details page for that crash breaks. u can see this in staging

editor/components/AddressFormField.tsx
```diff
@@ -0,0 +1,90 @@
...
+  });
+  const inputType = col.inputType;
+
+  return (
+    <div className="mb-2">
+      {(inputType === "text" || inputType === "number") && (
+        <Form.Control
+          {...register(col.path, {
+            setValueAs: (value) => value?.trim() || null,
+            onChange: (e) => (e.target.value = e.target.value.toUpperCase()), // uppercase input as user is typing
```
roseeichelmann, line 56: i made the executive decision to just uppercase user input as they are typing bc its gonna get save like that anyways so i think it makes sense for users to see whats going to get saved :shrug:

roseeichelmann:
@johnclary this is ready for re-review, thanks for your thorough comments!

PR #1821, opened 2025-07-31
Support shorter token lifespans

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/23893

This PR refactors token retrieval to ensure that we always grab a fresh user token before making an API call. It resolves the issue Xavier encountered in [#23893](https://github.com/cityofaustin/atd-data-tech/issues/23893).

## Testing

ℹ️ Our staging Auth0 is configured to use JWTs with a 1-minute TTL. So you can test locally or on netlify with the same results.

**URL to test:** Netlify

1. Log out and log back in to the VZE as an admin user.
2. Observe that the crashes list loads normally. 
3. Visit a crash details page such as CRIS crash ID #20713044. Make sure the crash has a crash diagram visible. If the crash doesn't have a digram, find a different crash that does.
4. Open the **Network** inspector in your browser's debug console. then sit quietly for 60+ seconds.
5. Now edit the crash's location. Observe that when you save your changes, a network request is sent to `https://atd-datatech.auth0.com/oauth/token` to fetch a fresh token. 
6. Sit quietly again for 60 seconds.
7. You may start to notice that a token request is being sent to auth0 every time you interact with the app. That's because `getTokenSilently` fetches a new token whenever your token is within one minute of expiring. When we set the token TTL to 5-minutes we will see fewer token refresh requests.
9. Once you have sat quietly for 60 seconds, click the **Download CR3** button, located at the top of the crash narrative card. One of two things should happen: you are prompted to save the CR3 pdf, or you encounter a `NoSuchKey` error.  Both of these results are ok! The staging app has far fewer CR3 PDFs available (the diagram image source is pointed at the production s3 bucket).
10. Navigate to the Users list and verify that it loads properly.
11. Use the **Add user** button to add a read-only user. For the email address, use `<your.email>+pr-1821@austintexas.gov`
12. Take a 60 second break 🧘
13. Edit the name of the user you just created. Then delete the user.

You have completed the test steps for this PR 👍 

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/utils/auth.ts
+++ b/editor/utils/auth.ts
@@ -1,8 +1,5 @@
-import { useState, useEffect, useCallback } from "react";
-import { usePathname } from "next/navigation";
+import { useCallback } from "react";
 import { useAuth0, User } from "@auth0/auth0-react";
-
-const BASE_PATH = process.env.NEXT_PUBLIC_BASE_PATH || "";
 
 /**
  * Add our claims to the Auth0 ID token—these are
@@ -83,8 +80,9 @@ export const useGetToken = (): (() => Promise<string | undefined>) => {
     } catch (err) {
       console.error("Error getting access token:", err);
     }
+    // we can ignore getAccessTokenSilently in our dep array - 
+    // Auth0 didn't bother to memoize it for us and `isAuthenticated`
+    // has us covered.
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    // we can ignore getAccessTokenSilently - Auth0 didn't bother
-    // to memoize it for us and `isAuthenticated` has us covered
   }, [isAuthenticated]);
 };
--- a/editor/utils/auth.ts
+++ b/editor/utils/auth.ts
@@ -1,4 +1,4 @@
-import { useState, useEffect } from "react";
+import { useState, useEffect, useCallback } from "react";
 import { usePathname } from "next/navigation";
 import { useAuth0, User } from "@auth0/auth0-react";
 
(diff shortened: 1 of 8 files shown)


editor/utils/auth.ts
```diff
@@ -59,33 +59,32 @@ export const formatRoleName = (role: string): string => {
...
-        console.error("Error getting access token:", err);
-        await loginWithRedirect({
-          appState: {
-            returnTo: BASE_PATH + pathName,
-          },
-        });
-      }
-    };
-    getToken();
-  }, [getAccessTokenSilently, isAuthenticated, loginWithRedirect, pathName]);
```
johnclary, line 88: as it turns out, neither `getAccessTokenSilently()` nor `loginWithRedirect()` are memoized and this was executing on every render 😖

editor/utils/graphql.ts
```diff
@@ -100,20 +100,20 @@ export const useQuery = <T extends Record<string, unknown>>({
...
   const { user } = useAuth0();
-  const token = useToken();
+  const getToken = useGetToken();
 
   const fetchWithAuth = async ([query, variables]: [
     RequestDocument,
     Variables,
   ]): Promise<HasuraGraphQLResponse<T, typeof typename>> => {
     const hasuraRoleName = getHasuraRoleName(getRolesArray(user));
+    const token = await getToken();
```
johnclary, line 110: again, here you can see the `getToken()` called on demand, right before we execute the request.

editor/utils/graphql.ts
```diff
@@ -100,20 +108,20 @@ export const useQuery = <T extends Record<string, unknown>>({
...
     return fetcher([query, variables, token || "", hasuraRoleName]);
   };
 
-  // todo: document falsey query handling
```
johnclary, line 113: ✅

roseeichelmann (approved):
this looks great, bug is fixed and awesome that you fixed the memoization problem and cleaned up those hooks. i cant tell if there is a lag now or if its in my head.. see my comment.

mddilley (approved):
Tested on the deploy preview, had some nice 60 second rests, and all works as expected. Cool to see how this ties into the requests in the code. 🚀 😎

PR #1811, opened 2025-06-20
Create EMS -> person match trigger

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/23198

Like I mentioned yesterday, I am looking for your feedback about this approach in general in addition to your feedback on the all the details. If you look through the commit history (e.g. [here](https://github.com/cityofaustin/vision-zero/pull/1811/commits/15e8f4f887bff3751057e5e554441c21f92b992f)) you can get a sense of how lost I was before switching to the event-based approach.

I've thought a little about what this might look like as an external API living outside the DB, but I think we'd need a fairly complex solution (probably with a redis queue or what have you) to try and orchestrate all of the different updates going on with the `ems__incidents table` 🤔 

A few out-of-scope items to consider:

* Backfill the existing EMS records to see what can be automatically matched to a person record
* Figure out what (if any) fields we want to show in the EMS UI wrt to person matching
* Think through if the non-CR3 matching trigger should be reworked to tie into the event handler

## Testing

Start your local stack with a fresh replica and apply migrations and metadata.

## Part 1 - VZE

In the VZE, navigate to EMS incident `25145-0421`: http://localhost:3002/editor/ems/25145-0421

Use the **Select person** button to match the 34yo/female to person ID `3525212`.

Observe that the other EMS patient is matched automatically to person ID `3525336` and that both EMS patients have a **Crash match status** of **Matched by review/QA** with the **Crash ID** populated as well.

Now edit one of the patient records by using the falafel dropdown menu to **Reset** the match status. Obsever that the record reverts to a **Crash match status** of **Multiple** and the **Crash ID** and **Person ID** columns are blank.

Now use the falafel menu and select the **Match not found** option. Observe that the record has a **Crash match status** of **Unmatched by review/QA** and the crash and person ID columns are blank.

Now edit the same patient record by clicking into the empty **Person ID** field and inputting its matching person ID. Click the **Save** button and observe that the record now has a **Crash match status** of **Matched by review/QA** with the **Crash ID** populated as well.

Lastly, use the falafel menu to **Reset** the match status of both records. We're going to continue testing with these records using your DB client.

## Part 2 - DB Client

Enable debug messaging

```sql
set client_min_messages to 'DEBUG';
```

Bring up the two EMS records we've been working with by querying the `incident_number`:

```sql
select
    id,
    incident_number,
    crash_pk,
    pcr_patient_age,
    pcr_patient_gender,
    pcr_patient_race,
    crash_match_status,
[...]

PR #1823, opened 2025-07-31
Dispatch EMS match events from people record inserts + updates 

johnclary:
## Associated issues

This PR builds on https://github.com/cityofaustin/vision-zero/pull/1811 by taking people record inserts and edits into consideration.

## Testing

**URL to test:** Local

1. Start your local stack with a recent copy of prod
3. Navigate to crash ID [17804147](http://localhost:3002/editor/crashes/17804147) and edit the crash location by moving it just a few feet. We do this to "backfill" the initial record matching state.
4. Scroll down to the **EMS Patient Care** card and notice that we have one EMS-person match and three unmatched records.
5. Navigate to the matched EMS incident by clicking on the link in the **EMS Patient Care** table or here: [#20215-0374](http://localhost:3002/editor/ems/20215-0374)
6. In the **Possible people matches** table, edit the **Race/Ethnicity** of person ID `391216` by changing the person's race from **White** to **Hispanic**. Upon saving your changes, notice that this person ID is now assigned to an EMS patient record. Although this kind of edit should be rare via the UI, demographic updates will be more common from CRIS. So we've now tested that case.
7. Notice that we have an EMS patient, age 17, that is unacconted for in the **Possible people matches** table. Use your DB client to create a person record that will match to this EMS record:

```sql
-- enable debug logging
set client_min_messages to 'DEBUG';

-- create person record
INSERT INTO
    people (id, prsn_nbr, prsn_age, prsn_ethnicity_id, prsn_gndr_id, unit_id, prsn_injry_sev_id)
VALUES
    (99999999, 9, 17, 2, 1, 285142, 2);
```

8. Refresh the incident details page and confirm that person ID `99999999` is now matched to an EMS patient record.
9. In the **Possible people matches** table, edit person ID `359980` by setting their age to `17`. Confirm that there is now only one EMS record matched to a person. Because we have two people records with same demographic identifiers, the trigger removed the automated match.
10. Use the inline **Select person** button to match the age 17 EMS patient to person ID `359980`
11. In the **Possible people matches** table, again edit person ID `359980` by setting their age back to `49`. Confirm that that this person is still matched to the age `17` EMS record. The match has persisted because you manually assigned it ✅
12. I am running additional tests using the CRIS import ETL, but you don't have to do that unless you'd really like to 😅 

That covers the testing for this PR. The parent PR has further testing that will need to be re-checked once this PR is merged 👍

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/database/migrations/default/1750435376415_ems_person_match_auto/up.sql
+++ b/database/migrations/default/1750435376415_ems_person_match_auto/up.sql
@@ -311,7 +311,12 @@ BEGIN
             END IF;
         END IF;
     END IF;
-    -- todo: do we need this line?
+    -- 
+    -- Finally, we set NEW._match_event_name = null as a failsafe to make sure
+    -- that this function never returns before nullifying _match_event_name.
+    -- If _match_event_name is not nullified before this function returns we 
+    -- will fail a check constraint on this column
+    --
     NEW._match_event_name = null;
     return NEW;
 END;
--- a/database/migrations/default/1750435376415_ems_person_match_auto/up.sql
+++ b/database/migrations/default/1750435376415_ems_person_match_auto/up.sql
@@ -491,7 +491,7 @@ $function$;
 
 DROP TRIGGER IF EXISTS people_dispatch_ems_match_trigger ON people;
 CREATE TRIGGER people_dispatch_ems_match_trigger AFTER
-UPDATE ON people FOR EACH ROW
+INSERT OR UPDATE ON people FOR EACH ROW
 EXECUTE FUNCTION people_dispatch_ems_match();
 
 -- delete this trigger/function which is now redundant
--- a/database/migrations/default/1750435376415_ems_person_match_auto/up.sql
+++ b/database/migrations/default/1750435376415_ems_person_match_auto/up.sql
@@ -282,10 +282,26 @@ BEGIN
                 return NEW;
             ELSIF array_length(matching_person_ids, 1) = 1 THEN
                 raise debug 'One person match found for EMS ID %', NEW.id;
-                NEW.matched_person_ids = matching_person_ids;
-                NEW.person_id = matching_person_ids[1];
-                NEW.person_match_status = 'matched_by_automation';
-                return NEW;
+                --
+                -- Before we can assign the person_id, we must make sure that no other
+                -- EMS records are matched to this person ID. This is an edge case that 
+                -- can happen when one or more EMS records with different
+                -- **incident numbers** (`incident_number`) are matched to the same
+                -- crash. This is an edge case that occurs in < .5% of records
+                --
+                IF NOT EXISTS (
+                    SELECT 1 FROM ems__incidents 
+                    WHERE person_id = matching_person_ids[1] 
+                    AND id != NEW.id
+                ) THEN
+                    NEW.matched_person_ids = matching_person_ids;
+                    NEW.person_id = matching_person_ids[1];
+                    NEW.person_match_status = 'matched_by_automation';
+                    return NEW;
+                ELSE
+                    raise debug 'Skipping person_id assignment because person_id is matched to another EMS record';
+                    return NEW;
+                END IF;
             ELSE
                 raise debug 'Multiple person matches found for EMS ID %', NEW.id;
                 NEW.matched_person_ids = matching_person_ids;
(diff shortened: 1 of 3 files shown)


database/migrations/default/1750435376415_ems_person_match_auto/up.sql
```diff
@@ -459,7 +465,36 @@ BEGIN
...
+        _match_event_name = 'handle_person_record_created_or_updated'
+    WHERE
+       ems__incidents.crash_pk = person_crash_pk;
+    RETURN null;
+END;
+$function$;
+
+DROP TRIGGER IF EXISTS people_dispatch_ems_match_trigger ON people;
+CREATE TRIGGER people_dispatch_ems_match_trigger AFTER
+INSERT OR UPDATE ON people FOR EACH ROW
```
johnclary, line 494: Something obvious that I completely missed in the parent PR: if we try to match an EMS record to a person record when a crash is inserted, it will never match because people records are created by the ETL after the crash is created. 

So prior to this change, an EMS-person level match was only ever possible if a crash's time or location were edited. A match would never happen when the crash was inserted 🙃

editor/app/ems/[incident_number]/page.tsx
```diff
@@ -250,7 +251,7 @@ export default function EMSDetailsPage({
             noRowsMessage="No people found"
             header="Possible people matches"
             columns={emsMatchingPeopleColumns}
-            mutation=""
+            mutation={UPDATE_PERSON}
```
Comment (johnclary) [inline editor/app/ems/[incident_number]/page.tsx:254]: Enables editing of people records from the **Possible matching people** table

chiaberry (approved):
I'm glad that by working on this you uncovered some flaws in the previous version of the migration. I think I could spend the rest of the day trying to go through this step by step, there is so much good stuff in this EMS match event work.

database/migrations/default/1750435376415_ems_person_match_auto/up.sql
mddilley: Nice catch! I'm sad that I didn't catch this either 😅 Lots of moving parts!

database/migrations/default/1750435376415_ems_person_match_auto/up.sql
```diff
@@ -151,18 +152,22 @@ BEGIN
...
+        WHERE people.id = NEW.person_id;
 
-        IF matching_person_record.crash_pk IS DISTINCT FROM NEW.crash_pk then
-            raise debug 'updating EMS record ID % crash_pk to % to match updated person_id', NEW.id, matching_person_record.crash_pk;
-            NEW.crash_pk = matching_person_record.crash_pk;
+        IF matching_person_record_crash_pk IS DISTINCT FROM NEW.crash_pk then
+            raise debug 'updating EMS record ID % crash_pk to % to match updated person_id', NEW.id, matching_person_record_crash_pk;
+            NEW.crash_pk = matching_person_record_crash_pk;
             NEW.crash_match_status = 'matched_by_manual_qa';
-            NEW.person_match_status = 'matched_by_manual_qa';
```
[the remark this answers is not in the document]
mddilley: @johnclary After step 11, I'm seeing every row in the EMS patients table show "Matched automatically". Is that what is expected even after manually selecting a match with the "Select Person" button? Maybe I am confusing manual with entering a person ID (or maybe not?)

Just wanted to check since I've lost some context on the UI so sorry if this has been covered already!

database/migrations/default/1750435376415_ems_person_match_auto/up.sql
```diff
@@ -459,7 +487,36 @@ BEGIN
...
+BEGIN
+    raise debug '**function: people_dispatch_ems_match **';
+    
+    SELECT units.crash_pk INTO person_crash_pk FROM units where units.id = NEW.unit_id;
+    
+    raise debug 'Dispatching update event to EMS records matched to crash_pk %', person_crash_pk;
+
+    UPDATE ems__incidents 
+    SET 
+        _match_event_name = 'handle_person_record_created_or_updated'
```
mddilley, line 507: awesome so this is where the event is queued up to be handled by `ems_update_handle_record_match_event` which has the new event handling ready to go.

database/migrations/default/1750435376415_ems_person_match_auto/up.sql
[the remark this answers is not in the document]
mddilley: Thanks for helping me get there - makes total sense to me but wanted to make sure. I'll slap an approval on here ✅

mddilley (approved):
🚢 😎 🚀

PR #1824, opened 2025-08-04
[Discovery day] Add crashes map view

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/23975

## Testing

**URL to test:** 

**Steps to test:**

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/components/TableMap.tsx
+++ b/editor/components/TableMap.tsx
@@ -1,0 +1,83 @@
+import { useCallback, MutableRefObject, useMemo } from "react";
+import MapGL, {
+  FullscreenControl,
+  NavigationControl,
+  ViewStateChangeEvent,
+  MapRef,
+  Source,
+  Layer,
+} from "react-map-gl";
+import MapGeocoderControl from "@/components/MapGeocoderControl";
+import {
+  DEFAULT_MAP_PAN_ZOOM,
+  DEFAULT_MAP_PARAMS,
+  MAP_MAX_BOUNDS,
+} from "@/configs/map";
+import { MapAerialSourceAndLayer } from "./MapAerialSourceAndLayer";
+import "mapbox-gl/dist/mapbox-gl.css";
+import { FeatureCollection } from "geojson";
+import { TableMapConfig } from "@/types/tableMapConfig";
+
+export interface LatLon {
+  latitude: number;
+  longitude: number;
+}
+
+export interface LatLonString {
+  latitude: string;
+  longitude: string;
+}
+
+interface TableMapProps<T extends Record<string, unknown>> {
+  /**
+   * Ref object which will hold the mapbox instance
+   */
+  mapRef: MutableRefObject<MapRef | null>;
+  /**
+   * The data that will be rendered as map features
+   */
+  geojson: FeatureCollection;
+  mapConfig: TableMapConfig;
+}
+
+/**
+ * Map component which renders an editable point marker
+ */
+export const TableMap = <T extends Record<string, unknown>>({
+  mapRef,
+  geojson,
+  mapConfig,
+}: TableMapProps<T>) => {
+  const onDragEnd = useCallback((e: ViewStateChangeEvent) => {
+    // todo: set a value to refetch data
+    console.log("drag end");
+  }, []);
+
+  return (
+    <MapGL
+      ref={mapRef}
+      initialViewState={{
+        latitude: DEFAULT_MAP_PAN_ZOOM.latitude,
+        longitude: DEFAULT_MAP_PAN_ZOOM.longitude,
+        zoom: DEFAULT_MAP_PAN_ZOOM.zoom - 5,
+      }}
+      {...DEFAULT_MAP_PARAMS}
+      cooperativeGestures={true}
+      // Resize the map canvas when parent row expands to fit crash
+      onLoad={(e) => e.target.resize()}
+      onDragEnd={onDragEnd}
+      maxZoom={21}
+    >
+      <FullscreenControl position="bottom-right" />
+      <NavigationControl position="top-right" showCompass={false} />
+
+      {/* custom point source and layer from data */}
+      {/* Point source and layer from data */}
+      <Source id="custom-source" type="geojson" data={geojson}>
+        <Layer id="custom-layer" type="circle" {...mapConfig?.layerProps} />
+      </Source>
+
+      <MapGeocoderControl position="top-left" />
+    </MapGL>
+  );
+};
--- a/editor/components/TableMapToggle.tsx
+++ b/editor/components/TableMapToggle.tsx
@@ -1,0 +1,81 @@
+import { Dispatch, SetStateAction } from "react";
+import Button from "react-bootstrap/Button";
+import {
+  FaSliders,
+  FaMapLocationDot,
+  FaMapLocation,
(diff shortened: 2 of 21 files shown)


editor/types/tableMapConfig.ts
```diff
@@ -0,0 +1,57 @@
+import { FeatureCollection } from "geojson";
+import { LayerProps } from "react-map-gl";
+
+/**
+ * An index of functions that transform input data into a geojson feature collection
+ */
+export const geoJsonTransformers = {
```
johnclary, line 7: This can theoretically be expanded to handle other geometry types, like the location polygons. 

We can of course make a DB view that returns geojson directly, but by doing the transform client-side we're able to re-use the same query/query-builder that powers the table view.

editor/components/MapFitBoundsControl.tsx
```diff
@@ -0,0 +1,39 @@
...
+        cursor: "pointer",
+      }}
+      onClick={() => {
+        if (bounds) {
+          mapRef?.current?.fitBounds(bounds, { padding: 10 });
+        }
+      }}
+    >
+      <AlignedLabel>
+        <FaHome className="fs-4" />
```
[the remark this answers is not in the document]
roseeichelmann: I almost feel like a reset button could be more clear?

image

Also -- i accidentally clicked on the EMS tab when I went to click on this Home button bc the sidebar expanded on me. Maybe this button could go on the right side along with the zoom buttons, would also just make it easier to use the zoom buttons and then quickly  reset back to the bounding box

editor/components/TableMap.tsx
```diff
@@ -0,0 +1,96 @@
...
+   */
+  const [initialViewState] = useState(() => {
+    if (geojsonBounds) {
+      return { bounds: geojsonBounds };
+    }
+    // Fallback when bounds are not available
+    return {
+      latitude: DEFAULT_MAP_PAN_ZOOM.latitude,
+      longitude: DEFAULT_MAP_PAN_ZOOM.longitude,
+      zoom: 9,
```
roseeichelmann, line 60: :sparkles:

editor/components/TableMap.tsx
```diff
@@ -0,0 +1,96 @@
...
+  /**
+   * After initialization, this hook updates the map extent
+   * when data changes
+   */
+  useEffect(() => {
+    if (geojsonBounds) {
+      mapRef?.current?.fitBounds(geojsonBounds, { padding: 10 });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [geojsonBounds]); // mapRef is stable (ref object)
```
roseeichelmann, line 73: :+1: good reminder of when to use that eslint disable line

editor/components/TableMap.tsx
```diff
@@ -0,0 +1,96 @@
...
+      cooperativeGestures={true}
+      // Resize the map canvas when parent row expands to fit crash
+      onLoad={(e) => e.target.resize()}
+      maxZoom={21}
+    >
+      <FullscreenControl position="bottom-right" />
+      <NavigationControl position="top-right" showCompass={false} />
+
+      {/* custom point source and layer from data */}
+      {/* Point source and layer from data */}
```
roseeichelmann, line 89: are these dupe comments?

editor/components/TableMapWrapper.tsx
```diff
@@ -0,0 +1,41 @@
...
+ * Card component that renders the crash map and edit controls
+ */
+export default function TableMapWrapper<T extends Record<string, unknown>>({
+  mapConfig,
+  data,
+}: TableMapWrapperProps<T>) {
+  const mapRef = useRef<MapRef | null>(null);
+  /**
+   * Trigger resize() when the map container size changes - this ensures that
+   * the map repaints when the sidebar is collased/expanded.
```
roseeichelmann, line 22: typo "collased"

editor/configs/crashesListViewTable.ts
```diff
@@ -267,21 +268,21 @@ export const crashesListViewQueryConfig: QueryConfig = {
...
-  limit: DEFAULT_QUERY_LIMIT,
+  limit: 1000,
   offset: 0,
   sortColName: "crash_timestamp",
   sortAsc: false,
   searchFilter: {
     id: "search",
     value: "",
-    column: "record_locator",
+    column: "address_primary",
```
roseeichelmann, line 278: :+1: i was already finding it annoying that this wasnt the default anymore

editor/configs/crashesListViewTable.ts
```diff
@@ -267,21 +268,21 @@ export const crashesListViewQueryConfig: QueryConfig = {
   exportable: true,
   exportFilename: "crashes",
   tableName: "crashes_list_view",
-  limit: DEFAULT_QUERY_LIMIT,
+  limit: 1000,
```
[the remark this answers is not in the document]
frankhereford: It's pretty amazing how many points these mapping libraries can support, and I think it'd awesome to give the user control over how many to render on their machine if it's capped. I'd push the limit and render tons of them and wait for someone to complain it's slow though.

Maybe this could be another justification to making a geojson or shapefile export -- if you have a /huge number/ of points, it's time to look to some GIS packages.

I saw in the query definition that it has a default sorting behavior - which is awesome - and I think this map makes that even more important. If the postGIS types in the table had any influence what order the database served back a limited, unordered query, it could exhibit on the map as large-scale patterns that don't exist.

I think the way you have it here is great, and I can't wait to see where you go from here related to the dataset of crashes on the map.

editor/components/TableMap.tsx
johnclary: yeah! and it still makes me nervous 😬 

This was a TIL—eslint normally ignores refs as hook dependencies, but it was barking because this ref was passed to the component as a prop 🤷 

i think the right move here might actually be to stick the ref in the dep array. i'm going to read about this some more.

johnclary:
> I was kind of on the fence if the "download results" tooltip and settings button should be updated to reflect its for the list view only, but I think its obvious to frequent users.

I went ahead and disabled the col visibility menu when the map is active. I think it's fine to leave the download button enabled since it works normally. I do think we might combine the download with frank's idea of limiting results to current map extent by offering an option to "include visible features only" in the download menu :)

johnclary:
> Maybe a future idea to think about is letting the map's extent become an additional search filter for the tabular interface. The existing criteria + even a bbox based geospatial filter can be really powerful i think.

Yes—it'd be really cool to have a search mode that lets a user drill into a geographic area and know they're seeing all crashes from the DB. And we have the VZV to copy from for drawing arbitrary shapes for query/analysis 🚀

editor/components/TableMapToggle.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+export interface TableMapToggleProps {
+  queryConfig: QueryConfig;
+  setQueryConfig: Dispatch<SetStateAction<QueryConfig>>;
+}
+
+/**
+ * Table component that controls advanced search filters
+ */
+export default function TableMapToggle({
```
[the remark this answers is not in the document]
roseeichelmann: that sounds like a good plan to me!

PR #1828, opened 2025-08-08
Remove non-cr3 matching UI

johnclary:
This PR updates Rose's original work (https://github.com/cityofaustin/vision-zero/pull/1812) by removing the non-CR3 matching UI but leaving the Non-CR3 record table in place.

## Testing

**URL to test:** Local

**Steps to test:**

1. Go to the EMS list page, and, if visible, click the **Reset** button to reset the filter state you have in local storage. 

2. Filter the EMS list for "Matched" Non CR3 crash match status. Click on one of those EMS records, and verify there is a record listed in the **Possible non-CR3 matches** table at the bottom of the incident details page.

3. Repeat these steps by filtering for EMS records with a non-cr3 match status of **Multiple matches**

4. Repeat these steps for EMS records which have a non-cr3 match status of  **Unmatched**. On the incident details page, observe that the  **Possible non-CR3 matches** table displays the message **No non-CR3 crashes found**.

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/configs/emsListViewTable.ts
+++ b/editor/configs/emsListViewTable.ts
@@ -264,7 +264,7 @@ const emsListViewFilterCards: FilterGroup[] = [
         enabled: false,
         filters: [
           {
-            id: "matched_by_automation",
+            id: "Matched",
             column: "non_cr3_match_status",
             operator: "_eq",
             value: "matched_by_automation",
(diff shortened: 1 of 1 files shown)


johnclary:
> i wondered if it would make sense to have the the filters dropdown state be something that is saved locally

Thanks for bringing up the clunkiness with filtering. For better or worse, I've replaced the accordion menu with a dropdown menu in https://github.com/cityofaustin/vision-zero/pull/1830. I agree that the filter experience needs some tuning. Maybe we should implement some UI event logging like Moped as way to start collecting more stats about feature usage 🤔 

We could also look at adding Moped-style filter chips so that users can quickly disable active ones without opening the menu. This also makes me want to hear about how saved views have been received in Moped, although our filters are lot less complex than Moped's at the moment.

I'm also having doubts about the filter **Reset** button. It completely purges your current table config, but users might want to simply reset the advanced search filters. That could be done with a toggle we add to the top of the filter menu 🤔

chiaberry (approved):
Test steps are all a-ok

I cant remember who is pro accordion, but I found i wanted either a way to collapse the possible people matches card or keyboard shortcut to the possible non cr3 matches because the incident i chose had a long list of possible people matches

PR #1835, opened 2025-08-14
Ability to select table page size

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/24160

This adds a basic page-size selector to our fancy table component. I think it might be cool to have different page sizes/defaults between the map and list view. Or maybe a checkbox on the map view that says **Load all records** —something like that. Anyway, I left that out of scope.

Screenshot 2025-08-14 at 2 08 12 PM

## Testing

**URL to test:** Local because we're branched from a branch's branch

Visit the crashes, Locations, and fatalities list views and test out the different page sizes. Refresh your page and confirm that the page size persists.

Confirm that when you hover over the page size dropdown a tooltip appears that says **Page size**

Also, notice how the loading spinner now replaces the "1-100 of 62,934 results" text until the page has loaded.

Lastly, how do you feel about these page size options: 50, 250, 1,000, 5,000?

---

#### Ship list

- [x] Check migrations for any conflicts with latest migrations in `main` branch
- [ ] Code reviewed

### DIFF
--- a/editor/components/TablePageSizeSelector.tsx
+++ b/editor/components/TablePageSizeSelector.tsx
@@ -40,7 +40,7 @@ export default function TablePageSizeSelector({
               const draftQueryConfig = produce(
                 queryConfig,
                 (draftQueryConfig) => {
-                  // type gaurd which enables us to assert that the selected page limit is allowed
+                  // type guard which enables us to assert that the selected page limit is allowed
                   if (
                     !ALLOWED_QUERY_PAGE_SIZES.includes(
                       newPageLimit as AllowedPageSize
@@ -49,7 +49,7 @@ export default function TablePageSizeSelector({
                     return;
                   }
                   draftQueryConfig.limit = newPageLimit as AllowedPageSize;
-                  // reset the current offset to return to the 1st page os results
+                  // reset the current offset to return to the 1st page of results
                   draftQueryConfig.offset = 0;
                   return draftQueryConfig;
                 }
--- a/editor/components/TablePageSizeSelector.tsx
+++ b/editor/components/TablePageSizeSelector.tsx
@@ -18,7 +18,7 @@ interface TablePageSizeSelectorProps {
 }
 
 /**
- * Table component that controls column visibility
+ * Table component that controls the the query limit, aka the page size
  */
 export default function TablePageSizeSelector({
   queryConfig,
--- a/editor/components/TablePageSizeSelector.tsx
+++ b/editor/components/TablePageSizeSelector.tsx
@@ -1,0 +1,68 @@
+import { Dispatch, SetStateAction } from "react";
+import { QueryConfig } from "@/types/queryBuilder";
+import Form from "react-bootstrap/form";
+import OverlayTrigger from "react-bootstrap/OverlayTrigger";
+import Tooltip from "react-bootstrap/Tooltip";
+import { produce } from "immer";
+import { ALLOWED_QUERY_PAGE_SIZES, AllowedPageSize } from "@/utils/constants";
+
+interface TablePageSizeSelectorProps {
+  /**
+   * The graphql query configuration
+   */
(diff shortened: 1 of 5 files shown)


editor/components/TablePageSizeSelector.tsx
```diff
@@ -0,0 +1,68 @@
...
+                    )
+                  ) {
+                    return;
+                  }
+                  draftQueryConfig.limit = newPageLimit as AllowedPageSize;
+                  // reset the current offset to return to the 1st page os results
+                  draftQueryConfig.offset = 0;
+                  return draftQueryConfig;
+                }
+              );
```
chiaberry, line 56: I meant to mention this in a previous PR but I had forgotten about using `immer` and this has been a good reminder that its a library we can reach for.

also a few typos in the comments, should be `type guard` and `1st page of results`

editor/utils/constants.ts
```diff
@@ -1,8 +1,12 @@
...
 
 export const COLORS = {
   primary: "#1176d0",
   danger: "#dd0426",
-  warning: "#ffd22f"
+  warning: "#ffd22f",
 };
+
+export const ALLOWED_QUERY_PAGE_SIZES = [50, 250, 1000, 5000] as const;
+export type AllowedPageSize = (typeof ALLOWED_QUERY_PAGE_SIZES)[number];
```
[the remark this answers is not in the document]
chiaberry: so much to learn!!

I think the options are fine for the crashes map, but I cant really see why anyone would want 5000 results for a table

editor/types/queryBuilder.ts
```diff
@@ -1,4 +1,9 @@
 import { TableMapConfig } from "@/types/tableMapConfig";
+import {
+  AllowedPageSize,
+  ExportPageSize,
+  MAX_RECORD_EXPORT_LIMIT,
```
chiaberry, line 5: `MAX_RECORD_EXPORT_LIMIT` isnt being used in this file

editor/utils/constants.ts
```diff
@@ -1,8 +1,12 @@
 export const DEFAULT_QUERY_LIMIT = 50;
-export const MAX_RECORD_EXPORT_LIMIT = 1_000_000;
+export const MAX_RECORD_EXPORT_LIMIT = 1_000_000 as const;
+export type ExportPageSize = typeof MAX_RECORD_EXPORT_LIMIT;
```
chiaberry, line 3: Why define the type like this, is it so that the type has to be 1000000 as a const? I guess I don't know another way to define a type that is 1,000,000 const...so this is how to define a constant variable?

chiaberry (approved):
Found an unused import, but otherwise its good to go. 

I left some questions about types, and a typo or two.

editor/utils/constants.ts
johnclary: Good question—turns out the `as const` is not necessary for the number declaration. 

Typescript knows that you just defined a `const` number, so it ain't changing. The inferred type is the same:

Screenshot 2025-08-20 at 1 06 52 PM

So I will remove `as const` here 🙏

editor/components/TablePageSizeSelector.tsx
```diff
@@ -0,0 +1,68 @@
...
+   */
+  queryConfig: QueryConfig;
+  /**
+   * Sets the queryConfig
+   */
+  setQueryConfig: Dispatch<SetStateAction<QueryConfig>>;
+}
+
+/**
+ * Table component that controls column visibility
```
roseeichelmann, line 21: copy pasta remnants in this comment :p

roseeichelmann (approved):
love this update, tested and it works like a charm! i like the way you updated the spinner behavior. i also think it could be a good idea to update page size options for the map view. good to go besides updating the comment for the `TablePageSizeSelector` component :ship: :page_facing_up:

PR #1837, opened 2025-08-20
Crashes map: ability to navigate to crash details from feature pop-up

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/24153

## Testing

Needs to be local because of branching

**Steps to test:**

In the table map view, click on crashes and see a popup appear. 
Click on the link to navigate to crash details page

Try toggling off the address and date column in the table view, confirm the popup still shows the information. 

---

#### Ship list

- [ ] Code reviewed


editor/components/PopupWrapper.tsx
```diff
@@ -0,0 +1,31 @@
...
+  longitude,
+  featureProperties,
+  PopupContent,
+  onClose,
+}: PopupWrapperProps) {
+  return (
+    <Popup
+      latitude={latitude}
+      longitude={longitude}
+      // The popup won't render after multiple map feature click without this prop being false :/
```
chiaberry, line 24: I copied this prop (and the comment) from the map configuration in the data and perfomance hub because I found the same issue

mddilley: yep so weird! i was able replicate without this prop so thanks for covering, Chia! 🙏

editor/components/PopupWrapper.tsx
mddilley, line 1: I noticed that the width of the popup seems to be tied to the address length, and the content and X icon are a bit crowded on shorter ones. Maybe a minimum width for the popup would help?

Image

editor/components/TableMap.tsx
johnclary, line 1: @chiaberry one other small thing—what do you think about increasing the minimum circle radius. Maybe from `2` to `6`?

editor/components/TableMap.tsx
```diff
@@ -71,22 +74,57 @@ export const TableMap = ({ mapRef, geojson, mapConfig }: TableMapProps) => {
     }
   }, [geojsonBounds, mapRef]);
 
+  // todo, check if this is the right type to use
```
mddilley, line 77: does this TODO need to stick around? it looks like the right type when i inspected in VSCode 👍

editor/components/TableMap.tsx
```diff
@@ -71,22 +74,57 @@ export const TableMap = ({ mapRef, geojson, mapConfig }: TableMapProps) => {
...
-        <Layer id="custom-layer" type="circle" {...mapConfig?.layerProps} />
+        <Layer type="circle" {...mapConfig?.layerProps} />
       </Source>
       <MapFitBoundsControl mapRef={mapRef} bounds={geojsonBounds} />
+      {selectedFeature && (
+        <PopupWrapper
+          longitude={selectedFeature?.properties?.longitude}
+          latitude={selectedFeature?.properties?.latitude}
+          featureProperties={selectedFeature.properties}
+          PopupContent={TableMapPopupContent}
```
mddilley, line 124: probably me being old school but i was expecting the popup wrapper to use `children` to render the content. it doesn't change the functionality at all!

mddilley (approved):
Looks amazing! I reviewed with the latest commits from John's feedback. I offered one suggestion about the popup width but this is ready for 🚢 🚢  🙌

editor/components/PopupWrapper.tsx
mddilley: this comment was before you added extra padding so maybe not as big of a concern now!

editor/configs/crashesListViewTable.ts
```diff
@@ -305,7 +305,7 @@ export const crashesListViewQueryConfig: QueryConfig = {
           ["zoom"],
           // zoom is 5 (or less)
           5,
-          2,
```
chiaberry, line 308: increased per https://github.com/cityofaustin/vision-zero/pull/1837/files#r2289354470

PR #1838, opened 2025-08-21
Crashes map: Navigate to crash details from feature pop-up

chiaberry:
Things got very confusing with branches going every which way and my [previous PR](https://github.com/cityofaustin/vision-zero/pull/1837) is now in some "files changed" nightmare state, so here is my code again, only my changes. 

## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/24153

## Testing
Needs to be local because of branching

Steps to test:

In the table map view, click on crashes and see a popup appear.
Click on the link to navigate to crash details page

Try toggling off the address and date column in the table view, confirm the popup still shows the information.

### DIFF
--- a/editor/components/TableMap.tsx
+++ b/editor/components/TableMap.tsx
@@ -77,7 +77,6 @@ export const TableMap = ({ mapRef, geojson, mapConfig }: TableMapProps) => {
   const [selectedFeature, setSelectedFeature] = useState<GeoJSONFeature | null>(
     null
   );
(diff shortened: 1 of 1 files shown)


johnclary (approved):
🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 

thanks Chia!!

chiaberry:
thanks for the re-review!

PR #1845, opened 2025-08-27
Handle errors when fetching a new access token

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/24371

## Testing

**URL to test:** [Netlify](https://deploy-preview-1845--atd-vze-staging.netlify.app/)

**Steps to test:**

1. Login to the VZE deploy preview on Netlify
2. Visit the `/locations` list view
3. We can simulate a refresh token expiration by manually editing the Auth0 data saved in local storage. Using your browser's dev console find the key/value pair that contains the `access_token` and `refresh_token` properties. It will be named `@@auth0spajs@@::random-string`. Take notice—there is a similar storage key that holds the `id_token` object which is not of concern here.
4. Invalidate your refresh token by making two changes to the local storage value:
- change the `expiresAt` property to be `0` (or any date in the past): this ensures a token refresh will be attempted.
- corrupt the `refresh_token` property by adding one or more characters to its string: this ensures the token refresh will fail.

5. Save your changes to the token object, now click on one of the list view table's column headers to sort the table. Once you click the column header, this will initiate a token refetch, and the app should redirect you to the sign in page. This may happen very quickly before you are immediately returned to the locations page—this is because your Auth0 session is still valid. We have successfully tested this bug fix!
6. Now sign out of the app. Without signing back in, try to visit a crash details URL such as https://deploy-preview-1845--atd-vze-staging.netlify.app/editor/crashes/20712641. Complete the sign in flow and confirm that you are redirected to the correct page after signing in. This is the new `onRedirectCallback` handler in action 🎉 

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/components/AppNavBar.tsx
+++ b/editor/components/AppNavBar.tsx
@@ -46,8 +46,11 @@ export default function AppNavBar({ user, logout }: NavBarProps) {
           </Navbar.Brand>
           <EnvironmentBadge />
         </Col>
-        <Col className="d-none d-md-flex justify-content-end align-items-center">
-          <NavBarSearch />
+        <Col className="d-flex justify-content-end align-items-center">
+          {/* Hide navbar  search on xs and sm */}
+          <div className="d-none d-md-flex">
+            <NavBarSearch />
+          </div>
           <Dropdown align="end">
             <Dropdown.Toggle id="avatar-toggle" as={DropdownAnchorToggle}>
               <Image
--- a/editor/components/SidebarLayout.tsx
+++ b/editor/components/SidebarLayout.tsx
@@ -12,8 +12,6 @@ import AppBreadCrumb from "@/components/AppBreadCrumb";
 import AppFooter from "@/components/AppFooter";
 import { darkModeLocalStorageKey } from "@/components/DarkModeToggle";
 
-const BASE_PATH = process.env.NEXT_PUBLIC_BASE_PATH || "";
-
 /**
  * The app sidebar component
  */
--- a/editor/components/SidebarLayout.tsx
+++ b/editor/components/SidebarLayout.tsx
@@ -86,7 +86,7 @@ export default function SidebarLayout({ children }: { children: ReactNode }) {
           onLogin={() =>
             loginWithRedirect({
               appState: {
-                returnTo: BASE_PATH + pathName,
+                returnTo: pathName,
               },
             })
           }
--- a/editor/contexts/Auth.tsx
+++ b/editor/contexts/Auth.tsx
@@ -1,5 +1,7 @@
 "use client";
-import { Auth0Provider } from "@auth0/auth0-react";
+import { Auth0Provider, AppState } from "@auth0/auth0-react";
+import { useRouter } from "next/navigation";
+import { useCallback } from "react";
 
 const DOMAIN = process.env.NEXT_PUBLIC_AUTH0_DOMAIN!;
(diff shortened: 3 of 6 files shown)


editor/components/AppNavBar.tsx
```diff
@@ -46,8 +46,11 @@ export default function AppNavBar({ user, logout }: NavBarProps) {
           </Navbar.Brand>
           <EnvironmentBadge />
         </Col>
-        <Col className="d-none d-md-flex justify-content-end align-items-center">
-          <NavBarSearch />
+        <Col className="d-flex justify-content-end align-items-center">
+          {/* Hide navbar  search on xs and sm */}
+          <div className="d-none d-md-flex">
+            <NavBarSearch />
+          </div>
```
johnclary, line 53: Fixes a bug I introduced in the last release in which the navbar dropdown menu is hidden on `xs` and `sm`. 

We have https://github.com/cityofaustin/atd-data-tech/issues/24283 to deal with the mobile nav

editor/contexts/Auth.tsx
```diff
@@ -18,6 +20,20 @@ export default function AuthProvider({
...
+   * Callback function that routes the user to their original destination
+   * after login
+   */
+  const onRedirectCallback = useCallback(
+    (appState: AppState | undefined) => {
+      const returnTo = appState?.returnTo || window.location.pathname;
+      router.push(returnTo);
+    },
+    [router]
+  );
```
johnclary, line 35: As you can see below, we were already setting the `{ returnTo: "some-url"}` data when logging in, but we did not have a handler configured to do anything about it. Now the app properly routes to the `returnTo` destination 👍

editor/utils/auth.ts
```diff
@@ -78,9 +79,15 @@ export const useGetToken = (): (() => Promise<string | undefined>) => {
...
       return accessToken;
     } catch (err) {
-      console.error("Error getting access token:", err);
+      console.warn(
+        "Redirecting to login page due to error getting access token:",
+        err
+      );
+      loginWithRedirect({
+        appState: { returnTo: window.location.pathname },
+      });
```
johnclary, line 88: Here's the missing error handler. If we can't get a new token (most likely because the user's refresh token has expired) we route the user to the login page and save their current window location in state. 

The `appState` payload gets sent to the Auth0 sign in page, and Auth0 sends it back to the client after sign-in by encoding it in the app's URL. The new `onRedirectCallback` in `Auth.tsx` handles this payload.

editor/components/SidebarLayout.tsx
```diff
@@ -86,7 +84,7 @@ export default function SidebarLayout({ children }: { children: ReactNode }) {
           onLogin={() =>
             loginWithRedirect({
               appState: {
-                returnTo: BASE_PATH + pathName,
+                returnTo: pathName,
```
johnclary, line 87: ☝️ The previous string formatting (`BASE_PATH + pathName`) was wrong. We had never tested it because there was no `onRedirectCallback` function

editor/utils/auth.ts
mddilley: very nice! cool to see that there is an clean way to pass the previous route. 🙌 ✨

mddilley (approved):
I tested on the deploy preview, and I wasn't able to break it. Very cool enhancement ↩️ and fun to see some parallel work happening across these projects. 🙌 🚢

PR #1846, opened 2025-08-28
Crashes map: ability to select basemap 

roseeichelmann:
## Associated issues
Closes https://github.com/issues/assigned?issue=cityofaustin%7Catd-data-tech%7C24155

## Testing

**URL to test:** 

**Steps to test:**

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/components/DarkModeToggle.tsx
+++ b/editor/components/DarkModeToggle.tsx
@@ -17,6 +17,12 @@ export default function DarkModeToggle() {
     const htmlElement = document.documentElement;
     htmlElement.setAttribute("data-bs-theme", appColorMode);
     localStorage.setItem(darkModeLocalStorageKey, appColorMode);
+    // Dispatch custom event
+    window.dispatchEvent(
+      new CustomEvent("themeChange", {
+        detail: { mode: appColorMode },
+      })
+    );
   }, [appColorMode]);
 
   return (
--- a/editor/components/MapSelectBasemap.tsx
+++ b/editor/components/MapSelectBasemap.tsx
@@ -1,0 +1,34 @@
+import { Form } from "react-bootstrap";
+import "mapbox-gl/dist/mapbox-gl.css";
+import { Card } from "react-bootstrap";
+
+/**
+ * Custom map control that fits the map to current bounds
+ */
+export default function MapSelectBasemap() {
+  return (
+    <div className="map-select-basemap-bottom-left">
+      <Card>
+        <Card.Header className="pb-0 ">Basemap</Card.Header>
+        <Card.Body className="py-1">
+          <Form.Check
+            type="radio"
+            label="Streets"
+            // value="readonly"
+            // inline
+            // id="readonly"
+            // checked={field.value === "readonly"}
+          />
+          <Form.Check
+            type="radio"
+            label="Aerial"
+            // value="editor"
+            // inline
+            // id="editor"
+            // checked={field.value === "editor"}
+          />
+        </Card.Body>
+      </Card>
+    </div>
+  );
+}
--- a/editor/components/MapSelectBasemap.tsx
+++ b/editor/components/MapSelectBasemap.tsx
@@ -1,11 +1,12 @@
 import { Form } from "react-bootstrap";
 import "mapbox-gl/dist/mapbox-gl.css";
 import { Card } from "react-bootstrap";
+import { mapStyleOptions } from "@/configs/map";
 
 /**
  * Custom map control that fits the map to current bounds
  */
-export default function MapSelectBasemap() {
+export default function MapSelectBasemap({ basemap, setBasemap }) {
   return (
     <div className="map-select-basemap-bottom-left">
       <Card>
@@ -14,18 +15,21 @@ export default function MapSelectBasemap() {
           <Form.Check
             type="radio"
             label="Streets"
-            // value="readonly"
-            // inline
-            // id="readonly"
-            // checked={field.value === "readonly"}
+            checked={
+              basemap === mapStyleOptions.darkStreets ||
+              basemap === mapStyleOptions.lightStreets
+            }
+            onChange={() => {
+              setBasemap(mapStyleOptions.lightStreets);
+            }}
           />
           <Form.Check
             type="radio"
             label="Aerial"
-            // value="editor"
-            // inline
-            // id="editor"
-            // checked={field.value === "editor"}
+            checked={basemap === mapStyleOptions.aerial}
+            onChange={() => {
+              setBasemap(mapStyleOptions.aerial);
+            }}
           />
         </Card.Body>
       </Card>
--- a/editor/components/MapSelectBasemap.tsx
+++ b/editor/components/MapSelectBasemap.tsx
@@ -2,11 +2,20 @@ import { Form } from "react-bootstrap";
 import "mapbox-gl/dist/mapbox-gl.css";
 import { Card } from "react-bootstrap";
 import { mapStyleOptions } from "@/configs/map";
+import { Dispatch, SetStateAction } from "react";
+
+interface MapSelectBasemapProps {
+  basemap: string;
+  setBasemap: Dispatch<SetStateAction<string | undefined>>;
+}
 
 /**
  * Custom map control that fits the map to current bounds
  */
-export default function MapSelectBasemap({ basemap, setBasemap }) {
+export default function MapSelectBasemap({
+  basemap,
+  setBasemap,
+}: MapSelectBasemapProps) {
   return (
     <div className="map-select-basemap-bottom-left">
       <Card>
--- a/editor/components/TableMap.tsx
+++ b/editor/components/TableMap.tsx
@@ -48,17 +48,18 @@ interface TableMapProps {
  * Map which can be configured to render in the Table component
  */
 export const TableMap = ({ mapRef, geojson, mapConfig }: TableMapProps) => {
+  // Use custom hook to get the app theme
   const isDarkMode = useGetTheme();
-  const [basemap, setBasemap] = useState(
-    isDarkMode ? mapStyleOptions.darkStreets : mapStyleOptions.lightStreets
-  );
+  const [basemap, setBasemap] = useState();
 
-  /** Check dark mode */
+  /** Changes the basemap to match app theme unless in aerial mode */
   useEffect(() => {
-    setBasemap(
-      isDarkMode ? mapStyleOptions.darkStreets : mapStyleOptions.lightStreets
-    );
-  }, [isDarkMode]);
+    if (basemap !== mapStyleOptions.aerial) {
+      setBasemap(
+        isDarkMode ? mapStyleOptions.darkStreets : mapStyleOptions.lightStreets
+      );
+    }
+  }, [isDarkMode, basemap]);
 
   const geojsonBounds = useCurrentBounds(geojson);
   /**
@@ -132,7 +133,7 @@ export const TableMap = ({ mapRef, geojson, mapConfig }: TableMapProps) => {
           <Layer type="circle" {...mapConfig?.layerProps} />
         </Source>
         <MapFitBoundsControl mapRef={mapRef} bounds={geojsonBounds} />
-        <MapSelectBasemap />
+        <MapSelectBasemap basemap={basemap} setBasemap={setBasemap} />
         {selectedFeature && (
           <PopupWrapper
             longitude={selectedFeature?.properties?.longitude}
--- a/editor/components/TableMap.tsx
+++ b/editor/components/TableMap.tsx
@@ -50,7 +50,7 @@ interface TableMapProps {
 export const TableMap = ({ mapRef, geojson, mapConfig }: TableMapProps) => {
   // Use custom hook to get the app theme
   const isDarkMode = useGetTheme();
-  const [basemap, setBasemap] = useState();
+  const [basemap, setBasemap] = useState<string | undefined>();
 
   /** Changes the basemap to match app theme unless in aerial mode */
   useEffect(() => {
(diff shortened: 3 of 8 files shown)


editor/configs/map.ts
```diff
@@ -18,15 +18,20 @@ export const MAP_MAX_BOUNDS: [[number, number], [number, number]] = [
   [-96, 32],
 ];
 
+/** Different basemap style options */
+export const mapStyleOptions = {
+  darkStreets: "mapbox://styles/mapbox/dark-v11",
+  lightStreets: "mapbox://styles/mapbox/streets-v12",
```
roseeichelmann, line 24: this is another option for light mode but it was looking very white, what do yall think?

Screenshot from 2025-08-28 15-06-57
 Screenshot from 2025-08-28 15-07-14

editor/utils/darkMode.ts
```diff
@@ -0,0 +1,27 @@
+import { useEffect, useState } from "react";
+import { darkModeLocalStorageKey } from "@/components/DarkModeToggle";
+
+/**
+ * Custom hook that returns whether the app is in dark mode
+ * and uses event listener to listen for updates
+ */
+export const useCheckDarkMode = () => {
```
roseeichelmann, line 8: When i first tried to implement this feature, the map would load the correct basemap color on first render, but if i toggled Dark mode while having the map open, the basemap wouldnt update. This useEffect paired with the event listener allows the map component to keep the app theme state up to date with when the user toggles dark mode.

It can also be used elsewhere in the app as needed -- i thought about using it [here in SidebarLayout.tsx](https://github.com/cityofaustin/vision-zero/blob/63adb99f91fe99eb78bf4d256f9b6c4d941f8dd1/editor/components/SidebarLayout.tsx#L49) but the useEffect there also sets the document attribute, so idk :thinking:

editor/configs/map.ts
roseeichelmann: VS what i currently have in this PR:

Screenshot from 2025-08-28 15-12-07

editor/components/MapSelectBasemap.tsx
```diff
@@ -0,0 +1,45 @@
...
+  /** The mapbox basemap type to be used in the map */
+  basemapType: "streets" | "aerial";
+  /** Sets the state for the basemap type */
+  setBasemapType: Dispatch<SetStateAction<"streets" | "aerial">>;
+}
+
+/**
+ * Custom map component for selecting the basemap option
+ */
+export default function MapSelectBasemap({
```
johnclary, line 16: would you add `Control` to the end of this name to be consistent with the other control components we have?

editor/components/TableMap.tsx
```diff
@@ -88,42 +97,49 @@ export const TableMap = ({ mapRef, geojson, mapConfig }: TableMapProps) => {
...
-      </Source>
-      <MapFitBoundsControl mapRef={mapRef} bounds={geojsonBounds} />
-      {selectedFeature && (
-        <PopupWrapper
-          longitude={selectedFeature?.properties?.longitude}
-          latitude={selectedFeature?.properties?.latitude}
-          featureProperties={selectedFeature.properties}
-          PopupContent={TableMapPopupContent}
-          onClose={() => setSelectedFeature(null)}
+    !!basemapType && (
```
johnclary, line 100: this value is not nullable, so maybe we don't need this conditional render?

editor/configs/map.ts
johnclary: I like the full color street map in light mode! I think the `light` map could be useful to draw more contrast between the map and the crash points, esp for seeing clusters, but I think `streets` is the best default and has tons of utility 🚢

editor/utils/map.ts
```diff
@@ -67,3 +68,18 @@ export const useCurrentBounds = (
       [bounds[2], bounds[3]],
     ];
   }, [geojson]);
+
+/**
+ * Returns the URL for the basemap type depending on app theme
+ */
+export const getBasemapURL = (
```
johnclary, line 75: probably not much of a concern because this is a super simple computation, but maybe a nice touch to convert this to a `useBasemap` hook?

editor/components/DarkModeToggle.tsx
```diff
@@ -17,6 +17,12 @@ export default function DarkModeToggle() {
     const htmlElement = document.documentElement;
     htmlElement.setAttribute("data-bs-theme", appColorMode);
     localStorage.setItem(darkModeLocalStorageKey, appColorMode);
+    // Dispatch custom event which is used in useGetTheme hook
+    window.dispatchEvent(
+      new CustomEvent("themeChange", {
+        detail: { mode: appColorMode },
```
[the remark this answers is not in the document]
johnclary: TIL—neat!

editor/utils/darkMode.ts
johnclary: This is really cool! Having proper dark mode state is going to be super handy for other places where we want to get fancy with our light/dark mode UI. 

One thing that jumps out at me is that it probably makes sense to move this hook into a global context (e.g. `ThemeProvider`) so that we don't duplicate this event handling if we add the hook to multiple components (like the `SidebarLayout`). Would you mind giving that a try?

editor/components/TableMap.tsx
roseeichelmann: this is correct, remnants from a previous iteration where it was nullable :smile: ty!

editor/contexts/ThemeProvider.tsx
```diff
@@ -0,0 +1,40 @@
+"use client";
+import { useEffect, useState, createContext, useContext } from "react";
+import { darkModeLocalStorageKey } from "@/components/DarkModeToggle";
+
+export const ThemeContext = createContext<string | null>(null);
+
+/**
+ * Global context provider that returns whether the app is in dark or light mode
+ */
+export const ThemeProvider: React.FC<{ children: React.ReactNode }> = ({
```
roseeichelmann, line 10: @johnclary this is the new `ThemeProvider`, this is my first time working with react global contexts and providers so lots of TILs for me today, does this all look right to you?

i do find it a little confusing that there is a React import called ThemeProvider as well, so could name our custom one something slightly diff to avoid confusion but i didnt have better ideas :shrug:

editor/utils/map.ts
```diff
@@ -67,3 +69,19 @@ export const useCurrentBounds = (
       [bounds[2], bounds[3]],
     ];
   }, [geojson]);
+
+/**
+ * Returns the URL for the basemap type depending on app theme
+ */
+export const useBasemap = (basemapType: "streets" | "aerial") => {
+  const themeMode = useTheme();
```
roseeichelmann, line 77: @johnclary converting this into a custom hook and using the global theme context here I think actually made the basemap transition when toggling between dark/light look sooooo much better. i feel like before you would toggle to dark and the basemap would stay in light mode for longer than the rest of the app before reloading in dark but now everything is just more seamless, so yay! :sparkles:

editor/components/TableMap.tsx
```diff
@@ -45,6 +47,12 @@ interface TableMapProps {
  * Map which can be configured to render in the Table component
  */
 export const TableMap = ({ mapRef, geojson, mapConfig }: TableMapProps) => {
+  const [basemapType, setBasemapType] = useState<"streets" | "aerial">(
+    "streets"
+  );
+
+  const basemapURL = useBasemap(basemapType);
```
mddilley, line 54: Doesn't change how this works at all, but it could be nice to package up the `basemap` state in the `useBasemap` custom hook since it will likely be needed for any map with dynamic basemaps.

editor/contexts/ThemeProvider.tsx
mddilley: in case it is helpful, this is a similar pattern that we use in Moped to make user/session details available throughout the app ([code](https://github.com/cityofaustin/atd-moped/blob/main/moped-editor/src/auth/user.js)). We also instantiated state from local storage, and it performed solidly.

we memoize there, but I don't see a need here since the state in this context contains a boolean.

anyways, this is really cool code to observe the theme changing!! 🚀

editor/styles/global.scss
```diff
@@ -316,6 +316,13 @@ to make sure it renders on top is a common fix */
   background-color: rgba(255, 255, 255, 0.95) !important;
 }
 
+.map-select-basemap-bottom-left {
+  position: absolute;
+  bottom: 10px;
+  left: 10px;
```
mddilley, line 322: nice detail to use the same pixel values as the OOTB Mapbox controls ✨

mddilley (approved):
Looks and works great! Very cool to see how the light/dark mode is now available outside Bootstrap elements 🚢 🚀

editor/contexts/ThemeProvider.tsx
```diff
@@ -0,0 +1,40 @@
+"use client";
+import { useEffect, useState, createContext, useContext } from "react";
+import { darkModeLocalStorageKey } from "@/components/DarkModeToggle";
+
+export const ThemeContext = createContext<string | null>(null);
```
johnclary, line 5: It looks like you can remove this `export`

editor/contexts/ThemeProvider.tsx
johnclary: This looks great!

It looks like it's `react-bootstrap` that exports a `ThemeProvider` as well. That's annoying, but I think what you have is fine. Alternatively: `AppThemeProvider`?

editor/contexts/ThemeProvider.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Global context provider that returns whether the app is in dark or light mode
+ */
+export const ThemeProvider: React.FC<{ children: React.ReactNode }> = ({
+  children,
+}) => {
+  const [themeMode, setThemeMode] = useState(
+    localStorage.getItem(darkModeLocalStorageKey)
+  );
```
johnclary, line 15: It might be good to copy the typing (and var name?) we use in the `DarkModeToggle`, since it is tighter:

```typescript
  const [appColorMode, setAppColorMode] = useState<"dark" | "light">(
    localStorage.getItem(darkModeLocalStorageKey) === "dark" ? "dark" : "light"
  );
```

Or instead of `themeMode`, short and sweet: `theme` ?

johnclary: Hmm, now I'm having doubts about including this `localStorage.getItem` call in the `DarkModeToggle`. It feels a bit off that we are pulling from `localStorage` to check for dark mode in two different places.

Maybe `DarkModeToggle` should use the value from the new context instead? I suppose that is complicated by the toggle itself using this state.

Going a step further, maybe the useEffect code in `DarkModeToggle` could be centralized in the context, and `useTheme` provides that as well, so you can do this

```typescript
const [themeMode, setThemeMode] = useTheme()
```

Am I overthinking this?

editor/utils/map.ts
johnclary: 🚀 Very awesome—and I love that we have this hook available for whatever other dark mode needs we have.

editor/components/MapBasemapControl.tsx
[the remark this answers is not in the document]
johnclary: Thanks, Chia! I think I like the card header style. And yes, I did fix the missing outline/shadow in the branch I have going to refine this control. Will get that PR'd v soon 🙏

editor/configs/map.ts
johnclary: I keep going back and forth on streets versus `light`. I like the full-color `streets` because I find it helpful for wayfinding.

I had initially conceived of three choices:
- **Analysis** - `light` or `dark` (default)
- **Streets** - `streets`
- **Aerial**

But Rose pointed out that **Analysis** is confusing and, echoing Tilly, it's better to keep the options simple. 

I'm open to going with `light` instead of `streets` though. In any case, I will open an issue to fix the poor contrast on streets and especially with aerials. We might want to ship that as a patch, even.

johnclary (commented):
Rose, this is looking fantastic and pretty much ready to ship. I am wondering about moving the theme setter into the context.

editor/contexts/ThemeProvider.tsx
johnclary: Agh, I just noticed I get a build error locally that the app seems to recover from. 

```
⨯ contexts/ThemeProvider.tsx (14:5) @ localStorage
 ⨯ ReferenceError: localStorage is not defined
    at ThemeProvider (./contexts/ThemeProvider.tsx:19:87)
digest: "3899942164"
  12 | }) => {
  13 |   const [themeMode, setThemeMode] = useState(
> 14 |     localStorage.getItem(darkModeLocalStorageKey)
     |     ^
  15 |   );
```

It might be that the initial `localStorage` interaction needs to be inside `useEffect`? 😐

roseeichelmann: Okay so I was able to move all of the state setting and getting from `DarkModeToggle` along with the contents of its `useEffect` that sets the local storage and DOM into the `AppThemeProvider` and it worked beautifully.  This allowed me to actually cut down on some logic and remove the first `useEffect` that added the event listener bc all of the theme state context is contained here and is global so we dont even need that custom event listener stuff anymore :smile:

roseeichelmann: Oh i am also now getting that build error... lemme see whats up

roseeichelmann: Okay yeah, it was a server side rendering problem, we were trying to access localStorage during SSR. as you said, had to employ a `useEffect` to update the state w the local `localStorage` val

editor/utils/map.ts
```diff
@@ -67,3 +69,29 @@ export const useCurrentBounds = (
...
     ];
   }, [geojson]);
+
+/**
+ * Custom hook that manages basemap state and returns the appropriate basemap URL
+ * depending on the selected basemap type and app theme
+ */
+export const useBasemap = (initialBasemapType: "streets" | "aerial") => {
+  const [basemapType, setBasemapType] = useState<"streets" | "aerial">(
+    initialBasemapType
```
roseeichelmann, line 79: thinking ahead, this `initialBasemapType` will allow us to pick what is the default basemap for various maps in the app, like in the crash details page it will be the Aerial map

PR #1855, opened 2025-09-08
Add basemap selector to the crash details map

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/24467

## Testing

**URL to test:** 
Netlify, but local to see the Nearmap imagery

**Steps to test:**
1. Go to a crash details page, the "Aerial" basemap should be toggled be default
2. Toggle between Streets and Aerial, see the basemap change.
3. Toggle on/off dark mode to see the Streets basemap change color
4. On page refresh the map will default back to Aerial

---

#### Ship list

- [ ] Code reviewed


editor/components/PointMap.tsx
```diff
@@ -140,14 +144,16 @@ export const PointMap = ({
           color={isEditing ? COLORS.danger : undefined}
         />
       )}
-      {/* add nearmap raster source and style */}
-      <MapAerialSourceAndLayer />
```
roseeichelmann, line 144: this component was responsible for rendering nearmap aerial imagery as the top layer, in a follow up PR we are going to make it so that the "Aerial" option in all the basemap pickers uses this nearmap imagery but for now i removed this

frankhereford (approved):
# 🚀 

Rose you were right -- this is super satisfying to put in the streets overlay and flip from dark to light mode. Such a nice touch. Thank you!

# 🤔 

Day dreaming here, tangential to this PR -- what do you think? 

Would it would be possible to write a component that itself contained the crash diagram, but the component could also listen for the light / dark mode event, right? And conditionally apply CSS to modify the colors of the crash diagram in HSV color space and invert them? 

We might be able to do it strictly in CSS too even. i'm thinking like something like this when it goes to dark mode. If we ever get color images, we'd want to add a 180 degree rotation on the hue as well.

```
@media (prefers-color-scheme: dark) {
  img.diagram { invert(0.8) brightness(1.2) }
}
```

Image


Also, if we reprocessed our diagrams to snip the little bands of black (white in this image, since it's inverted) at the top and bottom, it'd look ever better. Or maybe we can snip those off in CSS - i bet we could.

chiaberry (approved):
test steps check out! 

The selector feels like it takes up a lot of real estate on the map if I'm looking at the crash details on my laptop, and not external monitor. Maybe it could be a little smaller, but since the the crash details map only shows one point, maybe its just fine as is.

roseeichelmann:
@johnclary added the basemap control to the locations page and the satellite map as default!

johnclary (approved):
Thanks, Rose! 🚢

PR #1856, opened 2025-09-09
Add toolbox script to set CR3 PDF ContentType

johnclary:
## Associated issues
* https://github.com/cityofaustin/atd-data-tech/issues/24384

## Testing

**URL to test:** Local

1. `docker compose build`
2. `docker compose run backfill`

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/toolbox/backfill_content_type/.gitignore
+++ b/toolbox/backfill_content_type/.gitignore
@@ -1,0 +1,4 @@
+extracts/*
+!extracts/*.md
+*.zip
+*.jpeg
--- a/toolbox/backfill_content_type/backfill.py
+++ b/toolbox/backfill_content_type/backfill.py
@@ -1,0 +1,40 @@
+# docker compose build
+# docker compose run backfill
+import os
+import boto3
+
+ENV = os.getenv("BUCKET_ENV")
+BUCKET_NAME = os.getenv("BUCKET_NAME")
+folder_prefix = f"{ENV}/cr3s/pdfs/"
+
+# set up our S3 bucket paginator
+s3 = boto3.client("s3")
+paginator = s3.get_paginator("list_objects_v2")
+pages = paginator.paginate(Bucket=BUCKET_NAME, Prefix=folder_prefix)
+
+count = 0
+for page in pages:
+    if "Contents" in page:
+        # for each item in the buck
+        for obj in page["Contents"]:
+            count += 1
+            key = obj["Key"]
+            # if the obj is a .pdf file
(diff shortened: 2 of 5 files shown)


toolbox/backfill_content_type/README.md
```diff
@@ -0,0 +1,11 @@
+# Backfill CR3 PDF content type
+
+This script set's the `ContentType` header of all CR3 PDFs in our S3 bucket.
```
chiaberry, line 3: nitpick but I don't think the `'` is necessary here in sets.

toolbox/backfill_content_type/backfill.py
```diff
@@ -0,0 +1,40 @@
...
+folder_prefix = f"{ENV}/cr3s/pdfs/"
+
+# set up our S3 bucket paginator
+s3 = boto3.client("s3")
+paginator = s3.get_paginator("list_objects_v2")
+pages = paginator.paginate(Bucket=BUCKET_NAME, Prefix=folder_prefix)
+
+count = 0
+for page in pages:
+    if "Contents" in page:
```
chiaberry, line 17: just curious, this is to handle empty pages from the paginator?

toolbox/backfill_content_type/.gitignore
```diff
@@ -0,0 +1,4 @@
+extracts/*
+!extracts/*.md
```
frankhereford, line 2: TIL there's a negation operator you can use here - this is a good pattern for "nothing in here except for this particular thing."

frankhereford (approved):
Thank you for this John, it looks great. Thank you especially for the compose stack around it, beyond the convienece to the developer, I think they do a good job of defining / documenting the usage, environment, and other little details like that. This is great 🚢

johnclary:
> Thank you especially for the compose stack around it

yeah! and TIL that you can do `docker compose run -d --rm backfill` to force a container to be pruned when it's done

frankhereford:
> > Thank you especially for the compose stack around it
> 
> yeah! and TIL that you can do `docker compose run -d --rm backfill` to force a container to be pruned when it's done

Yea totally - to me `run` is a really useful way to think of one-off, short-lived tasks in a compose stack, where I'm often tempted to fall into the mindset where everything is an `up`, long-running service.

AWS reflects this too, but in a way we don't use and it's immediately obvious to me from the web console -- in ECS, you can define short lived tasks in task definitions and then kick them off, read `run`, with `aws ecs run-task`

Thanks again!!

PR #1861, opened 2025-09-17
Ability to filter EMS records by Austin full purpose jurisdiction

roseeichelmann:
## Associated issues
Closes https://github.com/issues/assigned?issue=cityofaustin%7Catd-data-tech%7C24580

Currently in Prod the EMS list page includes results outside of Austin full purpose. This PR updates the query so that by default the results do not include outside AFP, but you can toggle the filter button to include them

## Testing

**URL to test:** 
Local bc migrations

**Steps to test:**
1. Apply metadata
2. Go to the EMS page and toggle the Austin full purpose filter, when you toggle it on you will see more records added to the total
3. I cross checked w another branch and confirmed that when you toggle on "Include all" in this branch the total is the same as other branches that dont have this update yet

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/configs/emsListViewTable.ts
+++ b/editor/configs/emsListViewTable.ts
@@ -365,6 +365,28 @@ const emsListViewFilterCards: FilterGroup[] = [
       },
     ],
   },
+  {
+    id: "geography_filter_card",
+    label: "Jurisdiction",
+    groupOperator: "_and",
+    filterGroups: [
+      {
+        id: "austin_full_purpose",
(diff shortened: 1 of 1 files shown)


chiaberry (approved):
Filtered and could see more records and fewer records. 

At first I couldn't find the filter and I swore I had scrolled down to the bottom of the menu, but it must have been user error because I cant reproduce it. I think the upcoming work to refine the filter menu will be helpful to the non-power users

PR #1862, opened 2025-09-22
Process lookup table updates for CRIS v29.1.1

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/24756

## Testing

**URL to test:** Local

1. Start the DB from fresh replica. Apply migrations and metadata.
2. Try the down migration: `hasura migrate apply --down 1`

---

#### Ship list

- [ ] Code reviewed


database/migrations/default/1758552073255_cris_29_1_1_lookup_tables/up.sql
```diff
@@ -0,0 +1,85 @@
+-- Add crash notes about removal of agency ID 2429
```
johnclary, line 1: CRIS has removed  agency ID `2429 - WILLACY COUNTY SHERIFF'S OFFICE`. We have a single crash in our DB  (CRIS ID `11350054`) referencing this value in  `investigat_agency_id`. It is from from 2010 and does not have a latitude or longitude. 

This migration removing the reference to the ID and adds a note about it.

Other options I considered:
 
* Write a handler in our toolbox script that ignores agency ID so that we can persist it in our database
* Delete the entire crash from the DB

CRIS states that they do not warranty crash data that is +10 years old, and this is an example of how that can create headaches for us.

johnclary: I checked the staging DB to make sure we don't have any other conflicts there. Fortunately, staging has just the one identical conflict that production does. Ideally, I suppose we would have the toolbox generate these handlers for all deleted agency IDs to ensure the migrations run successfully on staging and prod.

PR #1863, opened 2025-09-24
Ability to soft-delete duplicate EMS PCRs

roseeichelmann:
## Associated issues
Closes https://github.com/issues/assigned?issue=cityofaustin%7Catd-data-tech%7C24416

## Testing

**URL to test:** 
Local

**Steps to test:**
1. Apply migrations and metadata locally
2. Go to an EMS incident that has a crash match status of "Matched automatically"
3. Open the crash details page the EMS record has been matched to (just keep this open in a diff tab)
4. Now back on the EMS detail page, copy the EMS record ID and use that to find it in `ems__incidents` in your hasura console or whatever db manager you are using (you will refer back to this in a later step)
5. Now back on the VZE use the falafel menu on the EMS record and Delete the record, it should open a confirmation modal. Click Ok to soft delete. If the record you deleted was the only record for the incident, you should get a 404 page.

7. If you refresh the record in your db, you should see the `person_id` and `crash_pk` have been updates to null, and all the match statuses updated to "unmatched"
8. Now go back to the crash details page for that crash and edit and move the coordinate just the tiniest bit and save
9. Now refresh the record in your db and make sure everything in step 7 is still the case -- we are wanting to make sure some triggers that would usually have automatically rematched the crash are being disabled on soft deleted ems records
10. Soft delete a record on an incident that has more than one EMS PCR records, make sure the records you didnt soft delete are still there

Now we are going to test the update to the `find_matching_person_ids` trigger function

1. Go to the [following EMS incident](http://localhost:3002/editor/ems/24283-0323) and note the two matching EMS PCR records -- 64 yr old hispanic males
2. Soft delete one of the 64 yr old male records, then go to the crashes details page matched w the record and bump the crash coordinate a tiny bit
3. Refresh the EMS incident page and you should now see that the remaining 64 yr old male has been automatically matched. This proves that my update to `find_matching_person_ids` works and soft deleted records are ignored when it comes to duplicate ems person records not being matched automatically -- in the case of a soft delete you want the remaining duplicate to be able to be matched

Lastly test the down migras

---

#### Ship list

- [ ] Code reviewed


database/migrations/default/1759336215844_update_crash_ems_match/up.sql
```diff
@@ -0,0 +1,129 @@
...
+        SELECT 
+            e.id,
+            e.incident_received_datetime,
+            e.geometry,
+            e.crash_match_status
+        FROM 
+            ems__incidents e
+        WHERE
+            e.incident_received_datetime  >= (NEW.crash_timestamp - time_threshold)
+            AND e.is_deleted IS FALSE
```
roseeichelmann, line 30: update here

editor/components/EMSLinkRecordButton.tsx
roseeichelmann, line 1: Critique on this modal is welcome

editor/components/EMSLinkRecordButton.tsx
```diff
@@ -89,7 +93,58 @@ const EMSLinkRecordButton: React.FC<
...
+                        is_deleted: true,
+                        crash_match_status: "unmatched",
+                        person_match_status: "unmatched",
+                        non_cr3_match_status: "unmatched",
+                        crash_pk: null,
+                        person_id: null,
+                      },
+                    });
+                    if (onSaveCallback) await onSaveCallback();
+                    setShowDeleteModal(false);
```
johnclary, line 141: If we're deleting the only EMS record with this incident number (which is not a case that should happen IRL), can we re-route back to the EMS list? Looks like that might require some trickery.

editor/app/ems/[incident_number]/page.tsx
```diff
@@ -63,7 +63,15 @@ export default function EMSDetailsPage({
...
    */
-  const incident = ems_pcrs?.[0];
+  const incident = all_ems_pcrs?.[0];
+
+  /**
+   * Filter out any soft deleted EMS records
+   */
+  const ems_pcrs = all_ems_pcrs?.filter(
+    (record) => record.is_deleted === false
+  );
```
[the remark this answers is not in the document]
Comment (roseeichelmann) [reply editor/app/ems/[incident_number]/page.tsx:73]: yeah originally i was doing that and letting the 404 page do its thing and was gonna ask about rerouting, but then i thought maybe thats not the behavior we want. but i can undo this change

editor/components/EMSLinkRecordButton.tsx
roseeichelmann: i updated let the 404 page do its thing, im tempted to leave re-routing for a future issue though? just cuz it does seem like it will take some noodling around

editor/queries/ems.ts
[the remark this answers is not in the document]
roseeichelmann: fixed this using contextFilters, thank you!

chiaberry (approved):
Test steps all work out. I was wondering if we should prevent people from being able to delete the only EMS record with the incident number, but John's comment said it shouldnt happen IRL, so maybe its fine. (Or maybe thats more of a reason to prevent it?)

mddilley (commented):
Still looking over the code, but I noticed that I'm getting the 404 page after deleting in step 5. I'm seeing this on http://localhost:3002/ems/25270-0048, http://localhost:3002/ems/25268-0433, and http://localhost:3002/ems/25134-0409.

editor/components/EMSLinkRecordButton.tsx
```diff
@@ -89,7 +93,58 @@ const EMSLinkRecordButton: React.FC<
...
+              show={showDeleteModal}
+              onHide={() => {
+                setShowDeleteModal(false);
+              }}
+            >
+              <Modal.Header closeButton>
+                <Modal.Title>Delete record</Modal.Title>
+              </Modal.Header>
+              <Modal.Body>
+                Are you sure you want to delete this EMS patient record?
```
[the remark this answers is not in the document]
johnclary: I agree it'd be nice to have more context to make you're deleting the correct record, but because we don't currently show the record IDs to the user anywhere I don't think this would be much help. So I'm inclined to leave it as is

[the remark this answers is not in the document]
johnclary: derp—we DO show the ID to users. it is not the first column in the table, which is where i was looking for it 🤔

editor/components/EMSLinkRecordButton.tsx
```diff
@@ -89,7 +93,58 @@ const EMSLinkRecordButton: React.FC<
...
               </Dropdown.Item>
+              <Dropdown.Item
+                onClick={async () => {
+                  setShowDeleteModal(true);
+                }}
+              >
+                Delete
+              </Dropdown.Item>
             </Dropdown.Menu>
+            <Modal
```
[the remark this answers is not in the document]
johnclary: yuck! i did not realize Firefox did this (and prop for good reason). iirc iOS does the same thing with dialogs. well, i was going to point out Github has an example that uses plain old dialogs, but they have recently changed this!

Screenshot 2025-10-03 at 11 56 48 AM
Screenshot 2025-10-03 at 11 57 51 AM

editor/components/EMSLinkRecordButton.tsx
johnclary: Fair enough—considering that this is an extreme edge case I am ok pushing it into a follow-up issue 🙏

database/migrations/default/1758732024749_ems__incidents_add_is_deleted/down.sql
```diff
@@ -0,0 +1,3 @@
+alter table "public"."ems__incidents" drop column "is_deleted";
+
+drop index ems__incidents_is_deleted_index;
```
mddilley: ~I'm seeing an error when running this down migration, and I think this index is dropped with the column.~ you fixed it! 🙏🙏

johnclary (approved):
Heavily tested—everything looks good. Thanks, Rose!!!! 🚢 🚢 🚢 🚢 🚢

mddilley (approved):
Awesome! So cool to be able to handle dupes now. I tested all the steps, and it all worked as laid out in the test steps. 🚢 🚀

roseeichelmann:
@mddilley im curious also about potential need to handle weird edge-cases like this!

johnclary:
> I'm thinking about how we show a modal on deleted Moped projects that says to contact us to restore, etc. 

i think mirroring Moped's behavior would be a nice upgrade @mddilley. will open an issue

PR #1866, opened 2025-09-29
Add map view to location crashes table

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/24457

## Testing

**URL to test:** 

https://deploy-preview-1866--atd-vze-staging.netlify.app/editor/locations/36CD245FAC

**Steps to test:**

Navigate to a location, scroll down to the list of crashes associated with that location. 
Scroll down the page to the crashes list and toggle the map button

The crashes from the table are rendered on the map. Some of the crashes may be non-cr3 crashes, the map popup says the crash type and if it is not a cr3 crash, it will show the case id instead of the crash id. 

Screenshot 2025-10-01 at 5 03 05 PM

Test the other table maps (crashes map and fatality maps) and confirm those popups still behave as before. 

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/schema/queryBuilder.ts
+++ b/editor/schema/queryBuilder.ts
@@ -88,5 +88,5 @@ export const QueryConfigSchema = z.object({
   filterCards: z.array(FilterGroup),
   exportable: z.boolean().optional(),
   exportFilename: z.string().optional(),
-  mapConfig: TableMapConfigSchema.optional(),
+  mapConfig: TableMapConfigSchema.strict().optional(),
 });
--- a/editor/schema/tableMapConfig.ts
+++ b/editor/schema/tableMapConfig.ts
@@ -6,5 +6,5 @@ export const TableMapConfigSchema = z.object({
   layerProps: z.record(z.unknown()).optional(),
   featureLimit: z.number().optional(),
   popupComponentName: z.literal("locationTableMap").optional(),
-  defaultBaseMap: z.enum(["streets", "aerial"]),
+  defaultBasemap: z.enum(["streets", "aerial"]),
 });


roseeichelmann (approved):
i like the hook you made and how you handled the different popup content. steps all worked as expected!

one thing i noticed and im curious if others can replicate, are you able to change the basemap by clicking the labels "Aerial" or "Streets"? For me its not updating unless i click on the radio button

editor/styles/global.scss
```diff
@@ -328,7 +328,7 @@ to make sure it renders on top is a common fix */
 }
 
 .table-map-container {
-  height: 100%;
+  height: 85vh;
```
johnclary, line 331: Sorry for not providing a better spec for this layout challenge. In the interest of preserving the `100%` height behavior on the crashes and fatalities lists, I think this styling should do the trick:

* add a fixed height to the crashes `Card` component. `75vh` looks good to me?
* add `d-flex flex-column` to the the <Card.Body> component

Image

editor/components/TableMap.tsx
```diff
@@ -92,6 +91,8 @@ export const TableMap = ({ mapRef, geojson, mapConfig }: TableMapProps) => {
     setCursor("grab");
   }, []);
 
+  const popupComponent = getPopupComponent(mapConfig.popupComponentName);
```
johnclary, line 94: I believe this should be capitalized

editor/types/tableMapConfig.ts
```diff
@@ -56,4 +73,8 @@ export interface TableMapConfig {
    * todo: implement this ;)
    */
   featureLimit?: number;
+  /**
+   * Optional name of popup component, if not provided then Map will use TableMapPopupContent component
+   */
+  popupComponentName?: "locationTableMap";
```
johnclary, line 79: looks good! 🙌

editor/configs/locationCrashesTable.ts
```diff
@@ -73,4 +73,39 @@ export const locationCrashesQueryConfig: QueryConfig = {
     }),
   },
   filterCards: locationCrashesFiltercards,
+  mapConfig: {
```
johnclary, line 76: I was going to ask if we could default this map to the aerial imagery basemap instead of streets, but it looks like that will require some refactor of the TableMap component that feels beyond scope. Right?

johnclary (changes requested):
Chia—this map looks great! I am requesting changes to the card/map container styles, otherwise this is ready to gooo!

johnclary:
> one thing i noticed and im curious if others can replicate, are you able to change the basemap by clicking the labels "Aerial" or "Streets"? For me its not updating unless i click on the radio button

Ah yeah—I forgot to mention that I noticed that little bug as well. On the table basemap picker, if you click the radio button **label** that controls the basemap picer on the location map above it!

editor/components/LocationsTableMapPopupContent.tsx
```diff
@@ -0,0 +1,41 @@
...
+      </div>
+      {properties?.type === "CR3" ? (
+        <div className="d-flex justify-content-between">
+          <span className="fw-bold">Crash ID</span>
+          <Link
+            href={`/crashes/${properties?.record_locator}`}
+            prefetch={false}
+          >
+            {properties?.record_locator}
+          </Link>
```
mddilley, line 25: Yay! This is so nice to be able to navigate to the crash page from the map. 🙌

mddilley (commented):
I tested with the latest commits, and I'm noticing the same bug that Rose and John saw with the basemap radios controls crossing over between the location and crash-location map below. I quickly looked at the code, and I think it might have to do with the explicit ids on the radios maybe?

Love the map!! 😎

chiaberry:
Thank you @mddilley! I pushed up a change to include the map type in the id and that seems to have fixed it

mddilley (approved):
Just retested with John's branch merged in here. I was able to see the before and after so I can confirm that I had to press reset before and didn't have to after with the config updating automatically. Great stuff, y'all! And I love the bug fix for the duplicate ids, Chia! 🙌 🚢 😎

mddilley:
@chiaberry you probably saw but I missed a dependency warning in the code that made the build fail in the deploy preview.
> 12:16:50 PM: 225:6  Error: React Hook useEffect has a missing dependency: 'initialQueryConfig'. Either include it or remove the dependency array.  react-hooks/exhaustive-deps

johnclary:
that's my bad about the missing useEffect dep. would you mind adding that, chia?

johnclary (approved):
🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

PR #1869, opened 2025-10-02
Improve handling of when table configs are updated

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/24932

This PR makes it so that you don't have to click that **Reset** filters button on table views in order to see updates to the table. This was previously necessary whenever we added new filters, or when we added a map to a view that previously didn't have one (as is done int he parent PR).

## Testing

We must  test on local because we're off of chia's branch.

1. Use your db client to run `refresh materialized view location_crashes_view`. This will take a minute or two.
2. Navigate to any location details page, such as `http://localhost:3002/editor/locations/6DE627CF2B`
3. Scroll down to the **Crashes** table, and notice that the list/map toggle is visible and you can click the **Map** toggle to switch to the map view.
4. Open your browser's dev console and locate the local storage item called `locationCrashesQueryConfig`.  Edit the value of this key by pasting the below config. Notice that it has no `_version` property, it has no `mapConfig` property, and it has a key called `oranges`. 

```json
{
  "oranges": 3,
  "exportable": true,
  "exportFilename": "location-crashes",
  "tableName": "location_crashes_view",
  "limit": 50,
  "offset": 0,
  "sortColName": "crash_timestamp",
  "sortAsc": false,
  "searchFilter": {
    "id": "search",
    "value": "",
    "column": "case_id",
    "operator": "_ilike",
    "wildcard": true
  },
  "searchFields": [
    { "label": "Case ID", "value": "case_id" },
    { "label": "Crash ID", "value": "record_locator" },
    { "label": "Primary address", "value": "address_primary" },
    { "label": "Secondary address", "value": "address_secondary" },
    { "label": "Collision type", "value": "collsn_desc" }
  ],
  "dateFilter": {
    "column": "crash_timestamp",
    "mode": "5y",
    "filters": [
      {
        "id": "date_start",
        "operator": "_gte",
        "value": "2020-10-02T04:00:00.000Z",
        "column": "crash_timestamp"
      }
    ]
  },
  "filterCards": [
    {
      "id": "primary_filter_card",
      "label": "Crash type",
      "groupOperator": "_or",
      "filterGroups": [
        {
          "id": "cr3_crashes",
          "label": "CR3 crashes",
          "groupOperator": "_and",
          "enabled": false,
          "inverted": false,
          "filters": [
            {
              "id": "cr3_crashes",
              "column": "type",
              "operator": "_eq",
              "value": "CR3"
            }
          ]
        },
        {
          "id": "non_cr3_crashes",
          "label": "Non-CR3 crashes",
          "groupOperator": "_and",
          "enabled": false,
          "inverted": false,
          "filters": [
            {
              "id": "cr3_crashes",
              "column": "type",
              "operator": "_eq",
              "value": "NON-CR3"
            }
          ]
        }
      ]
    }
  ]
}
```
5. After saving this updated config, refresh your page. Notices that the crashes config loads normally, and the **list/map** toggle is visible (with the **List** view currently active)
6. Nice! To see the old behavior, visit the same page in [Chia's PR's deploy preview](https://deploy-preview-1866--atd-vze-staging.netlify.app/editor/locations/36CD245FAC), and save this same config into the local storage value. Notice the **Reset** button is visible and you must click it to enable the map view.
7. Back in your local VZE, change the sort column on the location crashes table. Refresh the page, and confirm that the current sort column is persisted, and that clicking the **Reset** filters button restores the default sort.
8. Visit the Crashes, Fatalities, and EMS pages to confirm they load normally

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/components/TableWrapper.tsx
+++ b/editor/components/TableWrapper.tsx
@@ -155,7 +155,6 @@ export default function TableWrapper<T extends Record<string, unknown>>({
       queryConfigFromStorage &&
       queryConfigFromStorage?._version !== initialQueryConfig._version
     ) {
-      // deal with this
       const initialQueryConfigKeys = Object.keys(initialQueryConfig);
       const queryConfigFromStorageKeys = Object.keys(queryConfigFromStorage);
 
--- a/editor/components/TableWrapper.tsx
+++ b/editor/components/TableWrapper.tsx
@@ -150,6 +150,8 @@ export default function TableWrapper<T extends Record<string, unknown>>({
      * Check if stored config version is current and add/remove unknown
      * properties if needed - this has the effect of bringing the old
      * config in sync with the latest settings
+     *
+     * Todo: nested properties (e.g., mapConfig) are not handled
      */
     if (
       queryConfigFromStorage &&
@@ -179,6 +181,8 @@ export default function TableWrapper<T extends Record<string, unknown>>({
           delete queryConfigFromStorage[key as keyof QueryConfig];
         });
       }
+      // we're now on the latest config version ✨
+      queryConfigFromStorage._version = initialQueryConfig._version;
     }
 
     /**
--- a/editor/components/TableWrapper.tsx
+++ b/editor/components/TableWrapper.tsx
@@ -147,10 +147,46 @@ export default function TableWrapper<T extends Record<string, unknown>>({
       return;
     }
     /**
+     * Check if stored config version is current and add/remove unknown
+     * properties if needed - this has the effect of bringing the old
+     * config in sync with the latest settings
+     */
+    if (
+      queryConfigFromStorage &&
+      queryConfigFromStorage?._version !== initialQueryConfig._version
+    ) {
+      // deal with this
+      const initialQueryConfigKeys = Object.keys(initialQueryConfig);
+      const queryConfigFromStorageKeys = Object.keys(queryConfigFromStorage);
+
+      const missingKeys = initialQueryConfigKeys.filter(
+        (key) => !(key in queryConfigFromStorage)
+      ) as Array<keyof QueryConfig>;
+
+      const unknownKeys: string[] = queryConfigFromStorageKeys.filter(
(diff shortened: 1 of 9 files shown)


editor/components/TableWrapper.tsx
```diff
@@ -146,11 +146,46 @@ export default function TableWrapper<T extends Record<string, unknown>>({
...
+        });
+      }
+    }
+
     /**
      * Validate the query config we found in local storage
      */
     try {
-      QueryConfigSchema.parse(queryConfigFromStorage);
+      QueryConfigSchema.strict().parse(queryConfigFromStorage);
```
johnclary, line 188: Oops—I completely forgot that we are using Zod to do schema validation on the config that's loaded from local storage. The default behavior ignores unknown properties, which is why this hasn't been throwing an error (in which the default query config would be used instead).

Adding `.strict()` now means that we **must** update the QueryConfigSchema whenever we modify the `QueryConfig` type. If we don't, the user's config will never be read from local storage and default will always be used.

chiaberry (approved):
Such a good improvement. 🚢

editor/types/queryBuilder.ts
```diff
@@ -113,6 +113,11 @@ export type DateFilterMode = "ytd" | "all" | "5y" | "1y" | "custom";
  * Configuration object for the graphql query builder
  */
 export interface QueryConfig {
+  /**
+   * The current version of the config—is arbitrary and should be incremented when
+   * changes are made to the config so that earlier version can be migrated when possible
```
chiaberry, line 118: Does this mean opening the door to migrate back to an earlier version, or does "earlier version can be migrated" mean updated to the newer versions?

johnclary: "so that earlier versions can be migrated to the latest version"—I'll update the it 🙏

roseeichelmann (approved):
all the test steps work as expected, thanks for fixing this! :rocket:

PR #1871, opened 2025-10-03
Restores fit-bounds control on table maps

johnclary:
## Testing

**URL to test:** 

**Steps to test:**

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/components/TableMapWrapper.tsx
+++ b/editor/components/TableMapWrapper.tsx
@@ -34,7 +34,7 @@ export default function TableMapWrapper<T extends Record<string, unknown>>({
   }, [data, mapConfig.geojsonTransformerName]);
 
   return (
-    <div className="d-flex flex-grow-1" ref={mapContainerRef}>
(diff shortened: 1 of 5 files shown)


frankhereford (approved):
This looks great -- walked through the steps and everything checks out. Thank you for fixing this. Code looks 💯

This observation isn't related to this PR at all, but I did notice that the labeling on the EMS details page in street view mode is hard to read -- Maybe a second label layer could be added for use in street view mode that doesn't have a white font with black stroke and just had a black front instead?

Image

roseeichelmann (approved):

johnclary:
I've seen this recently, but can't reproduce it on this branch. I do see it on the v2.20 deploy preview (crash details pictured here):

Screenshot 2025-10-06 at 11 04 21 AM

My hunch is that what's happening is that we're actually rendering two label layers: the automatic one from the basemap plus our custom `streetLabels` layer that is meant to add labels to the nearmap imagery. I'll open an issue if we're still seeing this after the release 🙏

PR #1874, opened 2025-10-07
Crash details map: add fit-bounds controls

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/24743

## Testing

**URL to test:** 

**Steps to test:**

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/components/PointMap.tsx
+++ b/editor/components/PointMap.tsx
@@ -88,7 +88,6 @@ export const PointMap = ({
   const { basemapURL, basemapType, setBasemapType } = useBasemap("aerial");
 
   const geojsonBounds = useCurrentBoundsFromPoint(mapLatLon);
-  console.log(geojsonBounds)
 
   const onDrag = useCallback(
     (e: ViewStateChangeEvent) => {
--- a/editor/components/PointMap.tsx
+++ b/editor/components/PointMap.tsx
@@ -19,8 +19,9 @@ import {
   MAP_COORDINATE_PRECISION,
   MAP_MAX_BOUNDS,
 } from "@/configs/map";
-import { useBasemap } from "@/utils/map";
+import { useBasemap, useCurrentBoundsFromPoint } from "@/utils/map";
 import MapBasemapControl from "@/components/MapBasemapControl";
+import MapFitBoundsControl from "./MapFitBoundsControl";
 import { COLORS } from "@/utils/constants";
 import { z, ZodFormattedError } from "zod";
 import "mapbox-gl/dist/mapbox-gl.css";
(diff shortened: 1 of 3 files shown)


frankhereford (approved):
I think this looks really good, and I agree with you @chiaberry, I think the zoom is a couple of zoom levels too close when you pan and zoom to the bbox of a point. 

I tried zooming in and out trying to find the sweet spot, and I think it's about where the map is "whole intersection scale," maybe something like this. 

Image

Thank you so much for this -- it's another really nice touch.

editor/utils/map.ts
```diff
@@ -56,18 +56,35 @@ export function useResizeObserver<T extends HTMLElement>(
  * Returns undefined if the geojson has no features
  */
 export const useCurrentBounds = (
-  geojson: FeatureCollection
+  geojson: AllGeoJSON
```
mddilley: this is a cool TIL that turf has this type. idk how strict we are being about types in here but would it makes sense to limit the type to just Feature and FeatureCollection?

editor/components/PointMap.tsx
```diff
@@ -132,6 +138,7 @@ export const PointMap = ({
       {basemapType === "aerial" && <MapAerialSourceAndLayer />}
       <FullscreenControl position="bottom-right" />
       <NavigationControl position="top-right" showCompass={false} />
+      <MapFitBoundsControl mapRef={mapRef} bounds={geojsonBounds} />
```
mddilley: agreed about the zoom level for this crash one. Looks like adding a prop for max zoom or other parameter of the `fitBounds` call could open the door to customizing per view. These components are slick! 🚀

roseeichelmann:
I think when you click the fit to bounds button it should zoom to the same or similar extent that is loaded when you first open the page :thinking:

johnclary:
Hmm, I am seeing some bugginess when I:

1. navigate to an address using the search
2. click the fitbounds button -> the map zooms in on the searched address instead of taking me back to the crash location 

Looks like we might need some extra state to manage the address lat/lon?

editor/utils/map.ts
chiaberry: updated to remove the manual bbox construction mentioned in this comment https://github.com/cityofaustin/vision-zero/pull/1874#discussion_r2411701363 
and including this check for Infinity

editor/components/PointMap.tsx
chiaberry: added the maxZoom limit!

mateoclarke (approved):
smoooth ✨

PR #1875, opened 2025-10-07
Create basic fatal crash details page with header

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/24771

## Testing

**URL to test:** 

**Steps to test:**

---

#### Ship list

- [ ] Code reviewed


editor/app/fatalities/[record_locator]/page.tsx
```diff
@@ -0,0 +1,52 @@
...
+      : "Vision Zero Editor",
+    true // exclude the suffix
+  );
+
+  return (
+    <>
+      {data && (
+        <div className="d-flex justify-content-between mb-3">
+          <span className="fs-3 fw-bold text-uppercase">
+            {formatAddresses(data[0])}
```
Comment (mddilley) [inline editor/app/fatalities/[record_locator]/page.tsx]: Probably out of the scope of this issue, but I noticed that the code errors in the `formatAddress` call if I entered a non-existent crash record locator like http://localhost:3002/fatalities/123

editor/app/fatalities/[record_locator]/page.tsx
```diff
@@ -0,0 +1,52 @@
...
+          </span>
+          <span className="fs-4">
+            {formatYear(data[0].crash_timestamp)} Fatal Crash #
+            {data[0].law_enforcement_ytd_fatality_num}
```
Comment (frankhereford) [inline editor/app/fatalities/[record_locator]/page.tsx]: > So if you go to the fatality details page of a crash without one manually filled in yet, it will be empty AKA will say "2025 Fatal Crash # ". I talked to Mary about this and she is okay with this behavior, and if anything its a reminder that they still need to fill it in.

I'm fully 👍 on this staying as is because you've talked to Mary, but still - I want to advocate for putting in a condition that emits a message that the YTD fatality number has not been completed. If it's undefined, maybe show the string "Undefined YTD Fatality Number." I'm thinking that it makes more sense than the incomplete concatenation and it is very explicit about what is wrong, so a new user doesn't need to figure out something in the future when they are learning the app, and hopefully they instantly know exactly how to fix it to get it to show a YTD number.

editor/app/fatalities/[record_locator]/page.tsx
Comment (johnclary) [reply editor/app/fatalities/[record_locator]/page.tsx]: Right, so we can turn on the [`noUncheckedIndexedAccess`](https://www.typescriptlang.org/tsconfig/#noUncheckedIndexedAccess) compiler option if we want to check for this kind of error. 

Switching it on will require a bit of refactor across the app, because we will have to write some new type guards to deal with it. I'm going to open an issue for this!

editor/app/fatalities/[record_locator]/page.tsx
Comment (johnclary) [reply editor/app/fatalities/[record_locator]/page.tsx]: Thanks, Frank—I totally agree. We will definitely polish this up.

editor/utils/formatters.ts
```diff
@@ -51,6 +51,16 @@ export const formatTime = (value: unknown): string => {
...
 
+/**
+ * Format year as: 2025
+ */
+export const formatYear = (value: unknown): string => {
+  if (!value || typeof value !== "string") {
+    return "";
+  }
+  return format(parseISO(value), "yyyy") || "";
+};
```
johnclary: ✨

editor/app/fatalities/[record_locator]/page.tsx
Comment (mddilley) [reply editor/app/fatalities/[record_locator]/page.tsx]: thanks for sharing and all of this is TIL! 🙏

PR #1968, opened 2026-02-19
Adds ability to transfer data from temporary crashes when they are deleted via modal

### LINKED ISSUE #26663: Ability to transfer data from temporary crashes when they are deleted
User's need the ability to transfer data from temporary crash records so that important notes can be preserved.

On the **Crash details** page, add a modal that appears when the user clicks the **Delete** button. The modal should present a form where the user can use a combo box to select a crash to which data will be transferred.

* Modal is opened by clicking the **Delete** button

Image

* Modal form presents a combox box. "Delete" is disabled unless the user toggles the **I don't want to transfer data** toggle

Image

* Delete enabled

Image

* If a crash is selected, display text that lists the information that will be transferred

Image

* On submit, if a crash was selected, navigate to that crash details page

mateoclarke:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/26663

##
Adds the ability to transfer data from a temporary crash record to another crash before deleting it. When a user clicks "Delete" on a temporary crash, a modal opens that lets them:

- Search for a target crash by Crash ID or primary address
- Transfer any combination of: notes, FRB recommendation, ~edited card fields (Summary/Flags/Other)~, lat/lon and victim photo
- Skip transfer and delete the temp crash without transferring anything

The victim photo transfer is handled via a new Flask API endpoint that copies the image in S3 and updates the target person's metadata — this avoids permission issues with accessing image metadata through GraphQL.
Also refactors the repeated `@cross_origin` header lists in `server.py` into a reusable `CORS_HEADERS` constant.

One limitation of the victim photo transfer is that it only works if the source and target crash each only have one fatality (which is the majority of cases). Otherwise, extra logic would be required to try to match persons between the source and target crashes, which could be considered as a future enhancement. 

Screenshot 2026-02-25 at 12 05 40 PM

## Testing

**URL to test:** Local

**Steps to test:**

1. Start up your local stack, including the user/image API by running: `./vision-zero local-stack-up`
2. Test the delete with no transferable data found:
    - Create a new Temporary record by clicking the button in the top right of the [Crash list](http://localhost:3002/editor/crashes) page.
    - Open the Crash details page of the new temp record you just made. Make no changes and click the Delete button in the yellow banner.
    - Expect: Modal shows with message "No transferrable data was found on this temporary crash record." and Delete button is enabled. 
    - Click Delete and you should be routed back to the Crash list and the record should be archived. 
3. Test the change detection code & transfer flow:
    - Navigate to a temporary crash record (one with a T prefix in the record locator). You can use the table filter to only show temp records (down at the bottom of toggles under "Internal"). Or you can create one like in the step above.
    - Check the "Record history" at the bottom of the page. If there isn't much, edit some fields like: notes, FRB     recommendation, latitude, longitude, fields on the summary, flags, and other cards.
    - Click the Delete button in the yellow banner at the top of the page.
    - Expect: Delete button should be disabled before any target crash has been selected.
    - Go ahead and search for a target crash by typing a Crash ID or address
    - Select a result from the dropdown
    - Expect: In the modal, the "The following will be transferred" summary lists the correct edit items
    - Click Delete and confirm you are redirected to the target crash
    - On the target crash, verify the transferred data is present (notes, recommendations, and victim photo if both crashes had exactly one fatality)
6. Test skip-transfer flow:
    - Open the delete modal again on a different temp crash
    - Toggle "I don't want to transfer data"
    - Click Delete and confirm the crash is deleted and you are redirected to `/crashes`
7. Test photo transfer specifically:
    - Find a temp crash with a fatality. Using that crash ID, search the fatality list table and navigate to the temp crash's related fatality page. (ex: http://localhost:3002/editor/fatalities/T1798087)
    - Upload a photo to the victim record
    - Search for a target crash that also has exactly one fatality (ex: http://localhost:3002/editor/crashes/21214030)
    - Verify "Victim photo" appears in the transfer summary
    - Complete the transfer and verify the photo appears on the target crash's fatality
    - Bonus: Try it again but select a target record without any fatalities or with more than 1 fatality and confirm the photo doesn't transfer.
8. Run new API integration tests:
    - Run: `pytest api/tests/test_person_images.py -vv`
    - All 15 tests should pass. If there's an error, you may need replicate-db data since the test person and target person IDs are hardcoded.

---

#### Ship list

- [ ] Code reviewed


api/server.py
```diff
@@ -55,6 +56,7 @@
 EDITOR_ROLE_NAME = "editor"
 MAX_IMAGE_SIZE_MEGABYTES = 5
 CORS_URL = "*"
+CORS_HEADERS = ["Content-Type", "Authorization", "Access-Control-Allow-Origin", CORS_URL]
```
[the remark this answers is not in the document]
frankhereford: I'm glad you did this -- you cleaned up a lot of repetition, I think this is good.

frankhereford (approved):
Wow, thanks for this big chunk of work Mateo. There's a lot going on here -- I spent an hour reading over the code, and I think it's really well done. In particular, I really like the type-ahead component. 

It tested out perfectly as I worked through the steps you provided.

Of my comments, I would consider the request about the `updated_by` audit field to be the most important.

I don't see anything that should hold this back from shipping; thanks again. 🚢 🚢 🚢

api/utils/images.py
```diff
@@ -284,6 +285,56 @@ def _handle_image_upload(person_id, file, s3):
...
+            404,
+            description=f"No image found for source person ID: {source_person_id}",
+        )
+
+    source_image_source = make_hasura_request(
+        query=GET_PERSON_IMAGE_METADATA_FULL,
+        variables={"person_id": source_person_id},
+    )["people_by_pk"]["image_source"]
+
+    ext = source_obj_key.rsplit(".", 1)[-1] if "." in source_obj_key else "jpg"
```
johnclary: it feels kinda odd to fallback to `jpg` here. we might want to just throw an error since we should never encounter `source_obj_key` without a file extension?


this is way out of scope, but this awkward file ext handling makes me realize we probably should have included it has a separate piece of image metadata in the first place. NBD though 😅

editor/components/DeleteTemporaryCrashModal.tsx
```diff
@@ -0,0 +1,394 @@
...
+    id: number;
+    people_list_view: { id: number; prsn_injry_sev_id: number | null }[];
+  }>({
+    query: show && selectedTarget ? GET_TARGET_CRASH_FATALITY : null,
+    variables: { id: selectedTarget?.id ?? 0 },
+    typename: "crashes",
+  });
+  const targetFatalities = targetFatalityData?.[0]?.people_list_view ?? [];
+  // Photo transfer requires exactly one fatality on the target so the
+  // recipient is unambiguous. Skip if zero or multiple.
```
johnclary: 👍

api/utils/images.py
```diff
@@ -284,6 +285,56 @@ def _handle_image_upload(person_id, file, s3):
...
+    if not source_obj_key:
+        abort(
+            404,
+            description=f"No image found for source person ID: {source_person_id}",
+        )
+
+    source_image_source = make_hasura_request(
+        query=GET_PERSON_IMAGE_METADATA_FULL,
+        variables={"person_id": source_person_id},
+    )["people_by_pk"]["image_source"]
```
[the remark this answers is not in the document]
chiaberry: Just want to cosign Frank's comments here, I don't know enough to be able to assess how likely it could be for this to fail, but just wondering if we should protect it anyway.

api/utils/images.py
```diff
@@ -284,6 +285,56 @@ def _handle_image_upload(person_id, file, s3):
...
+        variables={
+            "person_id": target_person_id,
+            "object": {
+                "image_s3_object_key": target_obj_key,
+                "image_source": source_image_source,
+                "image_original_filename": source_original_filename,
+                "updated_by": get_user_email(),
+            },
+        },
+    )
```
chiaberry: along the lines of what Frank said above, but do we need to check the response at all from this request?

chiaberry (commented):
If there is a temp crash created, but I only input info into the initial create modal (case ID, address, unit) and I try to transfer the data to another crash, should I see anything listed under what information is going to be moved?

editor/components/DeleteTemporaryCrashModal.tsx
```diff
@@ -0,0 +1,394 @@
...
+      crash.people_list_view?.filter((p) => p.prsn_injry_sev_id === 4) ?? [],
+    [crash.people_list_view]
+  );
+  const tempFatalityId =
+    tempFatalities.length > 0 ? tempFatalities[0].id : null;
+
+  // Detect photo existence via the API
+  const [hasPhotoToTransfer, setHasPhotoToTransfer] = useState(false);
+  useEffect(() => {
+    if (!show || !tempFatalityId) {
```
roseeichelmann: this `show` was ambiguous to me.. could you make this var name more descriptive (showModal)? i had to root around to figure out what this was referring to

editor/components/DeleteTemporaryCrashModal.tsx
```diff
@@ -0,0 +1,394 @@
...
+    async (e: React.FormEvent) => {
+      e.preventDefault();
+      if (!canDelete) return;
+      await runTransferThenDelete();
+    },
+    [canDelete, runTransferThenDelete]
+  );
+
+  useEffect(() => {
+    if (!show) return;
```
roseeichelmann: Why do you need this `useEffect` + resetting these values here if you already reset them in the `handleClose` function? ( i could be overlooking something obvious)

roseeichelmann (approved):
followed all the test steps and this works great! wahoo thanks mateo, this is such a cool feature ✨ 

i will say that i found searching by address kind of unhelpful -- a lot of addresses have more than the limit of 20 crashes in the query, so i couldnt even find the crash i was looking for when doing that. but maybe in some select cases it could be helpful. i just wonder if it would be more straightforward to only search by Crash ID?

editor/utils/transferTempCrash.ts
```diff
@@ -0,0 +1,146 @@
...
+    updateRecommendation: MutateFn;
+    deleteRecommendation: MutateFn;
+  };
+}
+
+/**
+ * Execute all data transfers from a temporary crash to a target crash.
+ * Each step is independently conditional — only runs if there is data to transfer.
+ */
+export async function executeTransfer(config: TransferConfig): Promise<void> {
```
johnclary: I really like the way you've bundled all the mutations together. Obviously, if for some reason one of these mutations fails after an earlier one completes we will have a minor headache on our hands, but I can't think of a way to avoid this without a special SQL function or API route that handles this transactionally.

editor/components/CrashSearchTypeahead.tsx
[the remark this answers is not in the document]
johnclary: Agreed!

johnclary (commented):
@mateoclarke this feature is looking so good. It looks great and works great. And thank you for filling in the blanks on details that were missed in scoping. I'm happy to discuss any of my feedback—let me know 🙏

johnclary (commented):
Mateo, thanks for pushing all the updates. This looks so good. I've asked for a few minor changes:

- Add conditional style and label to the delete button
- Update the `bg-light` classes to `bg-light-use-them`
- Sort out an issue with the person ID test fixtures

johnclary (approved):
🚢 🚢 🚢 🚢 🚢 🚢   🙌

Charlie-Henry (commented):
I am getting these errors trying to upload an image, but the other test steps are working for me

Image
Image
Image

PR #1973, opened 2026-02-23
Adds more fields and filters to the locations list

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/26710
* https://github.com/cityofaustin/atd-data-tech/issues/27211

## Testing

**URL to test:** Local

On the `/locations` list, use the settings menu to enable visibility of all table columns. Test out the new filter cards on the filter drop down menu. Test out sorting on the **Signalized** and **High injury network columns**. Sorting is disabled on the array type columns.

On the location details page, observe the additional fields on the **Details** card

- Area engineer
- Signal engineer
- Signalized
- Council district(s)
- High injury network
- APD sector

---

#### Ship list

- [ ] Code reviewed


johnclary (commented):
I left comments on my own PR

chiaberry (commented):
I tested filtering on the new fields and ✔️ 

I expected the signalized filter to be a toggle yes or no and not able to filter on "yes signalized" and "not signalized" but I see thats how the fields are defined `is_signalized_yes` and `is_signalized_no`

The down migration is giving me an error " "cannot drop columns from view". I'll change this comment to approval after its addressed!

chiaberry (approved):
up and down migrations work perfectly, ✅

mddilley (changes requested):
Looks great, John! Requesting changes because I think there is a bug in the council district and street level filters when I was looking at the # of records returned. Otherwise, this is good to go! 🙏

mddilley (approved):
Tested again and this all looks great! 🚢 🙌 ✨

Charlie-Henry (approved):
I was able to sort by these fields and I see them on the locations details page!

editor/configs/locationsListViewTable.ts
```diff
@@ -2,6 +2,271 @@ import { QueryConfig, FilterGroup } from "@/types/queryBuilder";
...
+          },
+        ],
+      },
+    ],
+  },
+  // is_signalized
+  {
+    id: "is_signalized",
+    label: "Signalized intersection",
+    groupOperator: "_or",
```
mateoclarke: At first I expected the "Signalized intersection" toggle to behave more like the "High injury network locations only" toggle since they both are using the same data type, booleans. It felt weird to me that Signalized intersection filtering results would be the same when 1) neither Yes or No was toggled or 2) when both are Yes and No are toggled on. 

Maybe the use case for those toggles is different, but why not have "Signalized intersections only" that shows all (yes and no) when untoggled and only the Yeses when toggled on?

mateoclarke (approved):
:shipit:

roseeichelmann (approved):
These changes look great 👌

PR #1980, opened 2026-03-05
Fixes Viewer accessibility issues

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/13454

I looked at addressing the AA level issues, that is the target goal we need to meet by late April. 
The AAA issues do not seem too difficult to address, changing some color contrast and making buttons bigger.

I attempted a fix for "Interactive element does not meet minimum size nor spacing" and left comments about it in the code. The gist is we have nested interactive elements and I disabled the inner button, but that means we lost the mouseover hover effect. I can try another solution if we want that effect back. 

There are a few issues on the map page ("Empty headings" and "Text not included in an ARIA landmark" plus one that I see using the chrome extension, but not on the siteimprove site) that are related to the links on the mapbox map in the lower right. I checked the map on our traffic cameras site, which is running a later version of the mapbox package, and those issues have been fixed. I did not try updating the mapbox package, but I can go down that road if we want. 

## Testing

**URL to test:** 

https://deploy-preview-1980--atd-vzv-staging.netlify.app/viewer/

**Steps to test:**

Visit the deploy preview and compare the summary cards to whats in production. I changed them from `h3`s to `spans`

I installed the site improve chrome extension and ran it on the deploy preview:

Screenshot 2026-03-06 at 10 50 07 AM

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/viewer/src/Components/Widgets/SummaryWidget.js
+++ b/viewer/src/Components/Widgets/SummaryWidget.js
@@ -109,10 +109,10 @@ const SummaryWidget = ({
             />
           </span>
           <div className="d-flex flex-column widget-header-text">
-            <span className="mb-0">
+            <span className="mb-0 h5">
               {text} {infoPopover}
             </span>
-            <span className="text-muted">{`in ${currentYear}`}</span>
+            <span className="text-muted h5">{`in ${currentYear}`}</span>
           </div>
         </div>
       </CardBody>
--- a/viewer/src/Components/Widgets/SummaryWidget.js
+++ b/viewer/src/Components/Widgets/SummaryWidget.js
@@ -56,22 +56,10 @@ const SummaryWidget = ({
       bottom: 1px;
     }
 
-    .widget-header-text > h3 {
+    .widget-header-text > span {
       font-size: 1.2em;
     }
   `;
-
-  const renderIcon = () => (
-    <span className="fa-layers fa-3x fa-fw widget-icon">
-      <FontAwesomeIcon icon={faCircle} color={backgroundColor} />
-      <FontAwesomeIcon
-        icon={icon}
-        inverse
-        transform="shrink-6"
-        color={colors.white}
-      />
-    </span>
-  );
 
   const renderFooterBasedOnChange = (currentYearTotal, lastYearTotal) => {
     const icon =
@@ -111,12 +99,20 @@ const SummaryWidget = ({
           </Col>
         </Row>
         <div className="text-left d-flex flex-row">
-          {renderIcon()}
+          <span className="fa-layers fa-3x fa-fw widget-icon">
+            <FontAwesomeIcon icon={faCircle} color={backgroundColor} />
+            <FontAwesomeIcon
+              icon={icon}
+              inverse
+              transform="shrink-6"
+              color={colors.white}
+            />
+          </span>
           <div className="d-flex flex-column widget-header-text">
-            <h3 className="h5 mb-0">
+            <span className="mb-0">
               {text} {infoPopover}
-            </h3>
-            <h3 className="h5 text-muted">{`in ${currentYear}`}</h3>
+            </span>
+            <span className="text-muted">{`in ${currentYear}`}</span>
           </div>
         </div>
       </CardBody>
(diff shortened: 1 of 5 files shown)


viewer/src/views/nav/SideDrawerContent.js
```diff
@@ -35,6 +35,9 @@ const SideDrawerContent = ({ type }) => {
   return (
     <div className="side-menu">
       <StyledDrawerHeader>
+        <h1 className="sr-only">
+          Vision Zero Map -- Help Austin reach zero traffic deaths
+        </h1>
```
chiaberry: this isnt one of the AA requirements, but was easy to address. We have it on the summary page, just not on the map page

"Page does not start with a level 1 heading"

viewer/src/Components/Widgets/SummaryWidget.js
```diff
@@ -111,12 +99,20 @@ const SummaryWidget = ({
...
+            />
+          </span>
           <div className="d-flex flex-column widget-header-text">
-            <h3 className="h5 mb-0">
+            <span className="mb-0">
               {text} {infoPopover}
-            </h3>
-            <h3 className="h5 text-muted">{`in ${currentYear}`}</h3>
+            </span>
+            <span className="text-muted">{`in ${currentYear}`}</span>
```
chiaberry: This is another best practices recommendation "Content missing after heading" 

I was confused because they were `h3` with the class `h5` and the sizes weren't lining up -- but thats because the font size was actually being overridden by the style on line 59

viewer/src/views/nav/Header.js
```diff
@@ -138,16 +143,15 @@ const Header = () => {
...
-                        className="pr-0 pl-2 mr-0 ml-2"
+                        className="pr-0 pl-2 mr-0 ml-2 "
+                        style={{ color: "#fff" }}
                       >
-                        <Button
-                          className={`nav-button inactive-nav-button mx-xs-0 mx-lg-2`}
-                          onClick={() => trackPageEvent(config.eventKey)}
-                          active={currentPath === config.url}
+                        <span
+                          className={`btn nav-button inactive-nav-button mx-xs-0 mx-lg-2`}
```
chiaberry: this fixes "Issue: Interactive element does not meet minimum size nor spacing"

This Button is nested inside of a NavLink, the NavLink is what navigates to the map (or summary, depending on what page you are on). Siteimprove says the element doesnt meet the size or spacing, and the spacing it is references is both of the elements being on top of each other.

johnclary (approved):
Looks incredible, in that it looks exactly like it did before. Thanks, Chia! 🚢  🚢  🚢  🚢

mddilley (approved):
Thank you for updating to make this compliant and bug fixing too! 🙏 Looking at this code was quite the step back in time 😅

frankhereford (approved):
To my eyes, this retains the look and behavior of the existing VZV. I think this is great work, thank you so much Chia. You changes make sense to me. 🚢 🚢 🚢 

---

@johnclary, I want to propose a new / spin-off issue. In the [PeopleByDemographics.js](https://github.com/cityofaustin/vision-zero/blob/main/viewer/src/views/summary/PeopleByDemographics.js#L39-L46) file, on lines 39-45, we define different `labelCategories` to describe ethnicities. I would prefer to see `Other or unknown` as the last option, not `American Indian or Alaska Native`. I know that there is not an explicit meaning to the order, but having a more specific category after a more broad category violates the expected order of specificity and makes the last entry look too much like a late-add or an afterthought.

roseeichelmann (approved):
was trying to find the differences and i couldnt, then i realized that was the whole point! :p

mateoclarke (approved):
This all looks good. Just noting that I'm seeing a few `A` level issues on `/map` and that's higher severity than `AA`. Maybe those are the issues that aren't showing on the Siteimprove website that you mentioned?

Image

Also to @mddilley's note about `axe-core/react` not being maintained past React 18, here's an interesting alternative to investigate: https://accented.dev/blog/2025-08-20-accented-a-modern-alternative-to-axe-core-react

I do like the potential to have an a11y tool integrated into CI that could block or, at the very least, warn if a PR triggers new a11y issues.

Finally, @johnclary, this seems bug report worthy but unrelated to this a11y work. The calendar icon pops out of the date input field until you select it the field. 

Image

Image

PR #1983, opened 2026-03-10
Update VZV packages

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/27283

## Testing

**URL to test:** Local until we merge https://github.com/cityofaustin/vision-zero/pull/1980.

1. from `/viewer`: `npm install`
2. `npm run start`
3. Click around the summary and map tabs of the viewer 🕵️ 

---

#### Ship list

- [ ] Code reviewed


Charlie-Henry (commented):
I'm seeing a blank sidebar when only when it's wide enough

Image

Image

Also the "Go to Summary" button is not visible in the prod version?

I was noticing that I wasn't getting the same values as the prod viewer but it looks like it's referencing the test [dataset](https://data.austintexas.gov/Transportation-and-Mobility/-TEST-Austin-Crash-Report-Data-Crash-Level-Records/3aut-fhzp/about_data), we might want to update that dataset just so we can compare the values side by side?

Map looked good 👍

chiaberry (approved):
I forgot to npm install when I was first clicking around, and then when I did I could see no difference, so thats a big check mark. 

@Charlie-Henry theres an ancient issue to address the black sidebar, if you add a slash to the url (`http://localhost:3000/viewer/` vs `http://localhost:3000/viewer`) the sidebar goes away.

mddilley (approved):
Clicked around the map and dashboard locally. Everything looks and works like it did before. 🚢 🚀

mateoclarke (approved):
I'm seeing some errors loading the map layer from Mapbox but I'm going to assume its related to the local environment or maybe localhost:3000 isn't whitelisted. I also noticed we hardcode the mapbox key inline. Small bit of tech debt to move this into a .env
https://github.com/cityofaustin/vision-zero/blob/john/update-vzv-packages/viewer/src/views/map/Map.js#L33-L34

Would be good to see this running in staging before we completely merge but I'll approve assuming this isn't a real problem. 🤞 

Image

roseeichelmann (approved):
Looks good!

PR #2015, opened 2026-04-10
Paginate overlapping map pop-ups on table maps

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/26697

mapbox sends multiple features on click, i did not dig into how big of a radius the click encompasses. We choose the first feature in production now, this PR uses the entire array of features in the popups. 

## Testing

**URL to test:** 
https://deploy-preview-2015--atd-vze-staging.netlify.app/editor/crashes

**Steps to test:**

Toggle map view, and select past 5 years as a filter (this helps get overlapping points). Zooming out also helps. 

Click an area with overlapping markers. Use the left and right arrow buttons to cycle through the popups. 
Find a marker without others nearby (you may need to zoom in). Check that the popup for one crash works as before. 

Repeat the tests on the ems listing map and the fatalities map. 

Screenshot 2026-04-13 at 10 41 58 AM

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/components/FatalitiesMapPopupContent.tsx
+++ b/editor/components/FatalitiesMapPopupContent.tsx
@@ -10,13 +10,16 @@ export default function FatalitiesMapPopupContent({
   properties,
 }: FatalitiesMapPopupContentProps) {
   return (
-    <div className="h-100 m-1 px-1" style={{ minWidth: "125px" }}>
+    <div className="h-100 m-1 px-1">
       <div className="fw-bold fs-6 pb-2 border-bottom">
         {properties?.address_display}
       </div>
       <div className="d-flex justify-content-between pt-2">
         <span className="fw-bold">Crash ID</span>
-        <Link href={`/fatalities/${properties?.record_locator}`} prefetch={false}>
+        <Link
+          href={`/fatalities/${properties?.record_locator}`}
+          prefetch={false}
+        >
           {properties?.record_locator}
         </Link>
       </div>
@@ -28,9 +31,7 @@ export default function FatalitiesMapPopupContent({
       </div>
       <div className="d-flex justify-content-between">
         <span className="fw-bold">YTD Fatal Crash</span>
-        <span className="text-muted">
-          {properties?.ytd_fatal_crash}
-        </span>
+        <span className="text-muted">{properties?.ytd_fatal_crash}</span>
       </div>
     </div>
   );
--- a/editor/components/PopupWrapper.tsx
+++ b/editor/components/PopupWrapper.tsx
@@ -1,7 +1,9 @@
 import { useState } from "react";
 import { Popup } from "react-map-gl";
+import { Button } from "react-bootstrap";
 import { TableMapPopupContentProps } from "./TableMapPopupContent";
 import { GeoJsonProperties } from "geojson";
+import { LuSquareArrowLeft, LuSquareArrowRight } from "react-icons/lu";
 
 interface PopupWrapperProps {
   latitude: number;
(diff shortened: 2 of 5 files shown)


editor/components/PopupWrapper.tsx
```diff
@@ -1,22 +1,76 @@
...
+        latitude={latitude}
+        longitude={longitude}
+        // The popup won't render after multiple map feature click without this prop being false :/
+        closeOnClick={false}
+        onClose={onClose}
+      >
+        <PopupContent
+          properties={selectedFeatures[activeFeatureIndex]?.properties}
+        />
+        <div className="d-flex align-items-center justify-content-between border-top m-2">
```
johnclary: Looks just like the design! I'd like to make some light modifications:

- remove the bottom margin from this container
- use proper outline buttons so that we get hover effects
- make the feature count text primary and use "of" instead of "/"

I copied the button/icon styles from table navigation controls. How does this look?
 
Image
Image

```jsx
      <div className="d-flex align-items-center justify-content-between border-top pt-2">
          <Button
            size="sm"
            className="px-0 me-2 px-1"
            variant="outline-primary"
            onClick={handlePrevious}
          >
            <AlignedLabel>
              <LuArrowLeft className="fs-6" />
            </AlignedLabel>
          </Button>
          <span className="text-primary">
            {activeFeatureIndex + 1} of {selectedFeaturesLength}
          </span>
          <Button
            size="sm"
            className=" px-0 px-1"
            variant="outline-primary"
            onClick={handleNext}
          >
            <AlignedLabel>
              <LuArrowRight className="fs-6" />
            </AlignedLabel>
          </Button>
        </div>
```

editor/components/PopupWrapper.tsx
```diff
@@ -25,7 +79,7 @@ export default function PopupWrapper({
       closeOnClick={false}
       onClose={onClose}
     >
-      <PopupContent properties={featureProperties} />
+      <PopupContent properties={selectedFeatures[0]?.properties} />
```
johnclary: Looks like this could be dried up slightly by removing this duplicated `PopupContent` and conditionally rendering the button controls when there is more than one feature.

johnclary (commented):
@chiaberry this looks so good, wow! I have been testing on the EMS map where it is very common for incidents to overlap. 

I am requesting some slight design changes but otherwise this looks ready to go to me!

I have ideas for follow-up issues:

- Update the popup location/anchor when paginating through points so that it stays focused on the currently selected feature
- Show a label/indicator on overlapping points. E.g. like on this apartment finder map 👇 

Image

Charlie-Henry (approved):
One small thing I noticed that the width of the popup changes based on the length of the header, which then moves the backwards and forwards arrows. This makes it harder to quickly flip through the list of crashes because you might have to move your mouse back and forth to find the button.


Image

Image

editor/components/PopupWrapper.tsx
```diff
@@ -25,7 +47,36 @@ export default function PopupWrapper({
       closeOnClick={false}
       onClose={onClose}
     >
-      <PopupContent properties={featureProperties} />
+      <PopupContent
+        properties={selectedFeatures[activeFeatureIndex]?.properties}
+      />
+      {selectedFeaturesLength > 1 && (
+        <div className="d-flex align-items-center justify-content-between border-top pt-2 mx-2">
```
chiaberry, line 55: Thank you for including the styling in your feedback @johnclary, I am using the different icons and your style as you wrote it except I added `mx-2` so that the top border and buttons match the width of the popup content above it. Is that okay?

editor/components/PopupWrapper.tsx
chiaberry: done! thank you

editor/components/PopupWrapper.tsx
johnclary: Beautiful—thank you! ✨

editor/components/PopupWrapper.tsx
johnclary: @chiaberry one more minor change in consideration of the variable pop-up width that Charlie mentioned. 

I have been testing with these width changes and I think it's pretty good:
- remove the `minWidth: "125px"` styles from all of the component popup contents (`EMSMapPopupContent`, etc)
- Add an inline style `{width: "235px"}` to the `PopUp` component inside the `PopupWrapper`. 

So we're setting a fixed pop-up width that lets the header content wrap if needed.

editor/components/FatalitiesMapPopupContent.tsx
```diff
@@ -10,13 +10,16 @@ export default function FatalitiesMapPopupContent({
...
   return (
-    <div className="h-100 m-1 px-1" style={{ minWidth: "125px" }}>
+    <div className="h-100 m-1 px-1">
       <div className="fw-bold fs-6 pb-2 border-bottom">
         {properties?.address_display}
       </div>
       <div className="d-flex justify-content-between pt-2">
         <span className="fw-bold">Crash ID</span>
-        <Link href={`/fatalities/${properties?.record_locator}`} prefetch={false}>
+        <Link
```
chiaberry, line 19: just linting changes below this

editor/components/PopupWrapper.tsx
chiaberry: done! thank you

johnclary (approved):
🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

roseeichelmann (approved):
this looks so so great!! ✨ 

another idea for a future issue is when you zoom in a lot after selecting a point with overlapping popups, maybe the paginated popup updates based on how many points are still overlapping after zoom? that may be more complicate than its worth, but i think that would pair well with johns idea to update popup to be focused on the selected feature

mddilley (approved):
So cool! I love being able to page from the end back to the start. I tested on all the maps, and I see this nice handling in each one of them. 🚀 🚢

mateoclarke (approved):
Looks great and works just as described. 🚀

PR #2017, opened 2026-04-13
Ability to edit contributing factors for temporary crashes

### LINKED ISSUE #27638: Ability to edit contributing factors for temporary crashes
On the fatality details page, add the ability to edit unit contributing factors. 

There is currently no way to add/edit contributing factors for temp crashes. Adding a UI for editing contributing factors for temp records will allow VZ staff to present information consistently for all crashes, regardless of if the record is temporary. 

To do this, we should:

* On the fatality details page, always show the **Charges** and **Contributing** factors headings for temp records (they are currently hidden if empty).
* When no contributing factors are present, show the text `"None"` with a plus icon button to right
* Clicking the plus icon button will open a modal where the five contributing factor fields can be edited
* When contributing factors are present for temp records, show an edit icon button in the **Contributing factors** section header. Clicking this button opens the same modal where the contributing factors can be edited.
* The UI buttons should not be visible for readonly users.

### Additional notes

* Included in this work are any graphql API metadata updates need to enable this change.
* Contributing factors do not need to be transferred when transferring data from temp records to CRIS records
* Note that editing Charges is out of scope.

#### Initial state with no contributing factors. **Contributing factors** section has a "None" item displayed with plus icon 

Image

#### State with contributing factors present - edit icon displayed to right of **Contributing factors** heading

Image

#### Modal form for editing contributing factors

Image

#### Reference markup

* Inline plus button

```jsx
<div>
  <div className="fw-bold">Charges</div>
  <div className="d-flex justify-content-start align-items-center">
    <span className="text-secondary">None</span>
    <span>
      <Button size="sm" variant="outline-primary border-white">
        <AlignedLabel>
          <LuCirclePlus />
        </AlignedLabel>
      </Button>
    </span>
  </div>
</div>
```

* Edit button inline with section heading

```jsx
  <div className="d-flex justify-content-start align-items-center">
    <div className="fw-bold">Contributing factors</div>
    <Button size="sm" variant="outline-primary border-white">
      <AlignedLabel>
        <LuSquarePen />
      </AlignedLabel>
    </Button>
  </div>
```

roseeichelmann:
## Associated 
Closes https://github.com/cityofaustin/atd-data-tech/issues/27638

## Testing

**URL to test:** 
Netlify

**Steps to test:**
1. Go to a Fatality details page for a temp record
2. See that you can now either Edit or Add contributing factors
3. Test adding, editing, and nullifying (using the Select... option) contributing factors
4. After editing, go to the Crash details page for that crash ID and see the contrib factors have been added there
5. Make sure that contrib factors are only editable for temp records
6. Test that read only users cant edit, only view

---

#### Ship list

- [ ] Code reviewed


editor/utils/formHelpers.ts
```diff
@@ -39,6 +39,12 @@ export const valueToString = <T extends Record<string, unknown>>(
   return String(value || "");
 };
 
+export const stringToNumberNullable = (v: string | null | undefined) => {
```
roseeichelmann, line 42: converted this to an exportable util bc i found we were using it in 3 diff places

editor/components/CreateCrashRecordModal.tsx
```diff
@@ -278,7 +279,7 @@ export default function CreateCrashRecordModal({
                         {...register(`units_cris.${index}.unit_desc_id`, {
                           // coerce to number or null
                           required: true,
-                          setValueAs: (v) => Number(v) || null,
```
roseeichelmann, line 281: this was previously problematic because 0's were being coerced as null

only a problem if we have any IDs with a value of 0, which we do for contributing factors so thats how i realized this

editor/components/ContributingFactorsModal.tsx
```diff
@@ -0,0 +1,163 @@
...
+    path: "contrib_factr_3_id",
+    label: "Primary #3",
+  },
+  {
+    path: "contrib_factr_p1_id",
+    label: "Possible #1",
+  },
+  {
+    path: "contrib_factr_p2_id",
+    label: "Possible #2",
```
roseeichelmann, line 48: i suppose i could have followed that pattern we have of putting this in the config folder with `relationship:` and `inputType:` etc like we do with editable lookup table options for datacrd or related records rows but it seemed easier to just handle all of this here with its own explicit pattern

editor/components/ContributingFactorsModal.tsx
```diff
@@ -0,0 +1,163 @@
...
+  onSaveCallback: () => Promise<void>;
+}
+
+type ContributingFactorsInputs = {
+  contrib_factr_1_id: number | null;
+  contrib_factr_2_id: number | null;
+  contrib_factr_3_id: number | null;
+  contrib_factr_p1_id: number | null;
+  contrib_factr_p2_id: number | null;
+};
```
johnclary: Rather than re-declaring these properties to create a new type, it would be good to use the typescript `Pick` function to grab them from the `Unit`. That will help ensure that we keep this code in sync with any future changes to the Unit schema.

editor/components/ContributingFactorsModal.tsx
johnclary: I'm on the fence about this. I do think it would be nice to keep track of these fields in the `ALL_UNIT_COLUMNS` object just so that we have all the fields in one place. That would be consistent with how you handled the address edit modal. 

That would give you the benefit of being able to use the `relationship` object with the fancy `useLookupQuery` generator, but that wouldn't add any extra type safety as far as I can tell. And the rest of the `ColDataCardDef` structure is designed to abstract our edit components that only work when editing a single field (`<EditableField/>`). 

Not feeling particularly strongly so I am inclined to say keeping it simple like you have it is fine. One day it would be very cool to have a fully abstracted multi-field edit component that could replace all of our edit modals 😈

editor/queries/contribFactors.ts
johnclary, line 1: I think it'd be good to move this to the existing `queries/unit.ts` file.

editor/components/ContributingFactorsModal.tsx
```diff
@@ -0,0 +1,163 @@
...
+          onClick={() => {
+            onClose();
+            reset();
+          }}
+          disabled={isSubmitting}
+        >
+          Cancel
+        </Button>
+      </Modal.Footer>
+    </Modal>
```
johnclary, line 162: This form looks great!

editor/components/ContributingFactorsModal.tsx
```diff
@@ -0,0 +1,163 @@
...
+          {contribFactorLabels.map((factor) => {
+            return (
+              <Form.Group key={factor.path} className="mb-2">
+                <Form.Label>{factor.label}</Form.Label>
+                {!isLoading && factorOptions && (
+                  <Form.Select
+                    {...register(factor.path, {
+                      setValueAs: (v) => stringToNumberNullable(v),
+                    })}
+                  >
```
johnclary, line 126: We are overdue for having a typeahead component for this. I opened https://github.com/cityofaustin/atd-data-tech/issues/27975 🚀

editor/components/CreateCrashRecordModal.tsx
johnclary: thank you for catching this!

editor/components/FatalityUnitCardFooter.tsx
```diff
@@ -19,17 +29,33 @@ export default function FatalityUnitCardFooter({
...
+
+  const { user } = useAuth0();
+
+  const isReadOnlyUser = user && hasRole(["readonly"], user);
+
   return (
     <Card.Footer
       // If theres no card body remove extra padding
       className={
         "fatality-units-card-header-footer " + (!hasVictim ? "pt-0" : "")
```
johnclary: Would you remove this "pt-0"? The contrib factors are looking cramped.

Image

johnclary (commented):
@roseeichelmann this is looking and working great! my feedback is pretty minor 🚀

editor/utils/formHelpers.ts
```diff
@@ -39,6 +39,12 @@ export const valueToString = <T extends Record<string, unknown>>(
...
+  if (v === "" || v === null || v === undefined) return null;
+  const num = Number(v);
+  return isNaN(num) ? null : num;
+};
```
johnclary, line 46: looks good!

editor/components/ContributingFactorsModal.tsx
roseeichelmann: Im feeling inclined to keep it this simple way because i did start implementing the other way and it seemed like more hassle than it was worth :thinking:

editor/components/FatalityUnitCardFooter.tsx
```diff
@@ -42,21 +64,55 @@ export default function FatalityUnitCardFooter({
...
         <div>
-          <div className="fw-bold">Contributing factors</div>
-          {primaryContribFactors.map((factor) => (
-            <div className="ms-2" key={factor?.id}>
+          <span className="fw-bold">Contributing factors</span>
+          {isTempRecord && !isReadOnlyUser && (
+            <span className="ms-1">
+              <Button
+                size="sm"
+                variant="outline-primary border-white"
```
johnclary: @roseeichelmann two minor things I just noticed

1. The white outline doesn't look right on dark-mode. maybe `border-none` is solution?
2. The edit button/icon appears to be slightly off of vertical center with the **Contributing factors** heading. You might be able to fix this by adding another `AlignedLabel` or adding flex classes directly.

**Here's a picture of the problematic border**

Image


**And here i've tried to highlight the vertical alignment problem**

Image

chiaberry (approved):
Tested on the netlify deploy after running into the local dev issues that will be fixed by https://github.com/cityofaustin/atd-data-tech/issues/28075

Everything looks good, besides the border issue John mentioned in dark mode. 

I was wondering how the NONE contributing factor works, Is that another option to clear out the factor, besides choosing Select....? Because when I select NONE as one of the factors, it does not show up in the UI which makes sense unless None is intended to explicitly be its own factor i.e. "There are NO contributing factors, it just happened". I'll experiment with the code locally once i can get it back up and running, but wanted to check anyway.

editor/components/FatalityUnitCardFooter.tsx
roseeichelmann: Fixed both these issues in the follow up PR :)

roseeichelmann:
@chiaberry NONE is just another value in the lookup table option, so it doesn't mean NULL. i see how its confusing to have that as an option as well as Select as... which would mean nulling out the value, but they both render the same way as None. So maybe i remove the Select....?

@johnclary

editor/components/FatalityUnitCardFooter.tsx
johnclary: looks good! 🙏

johnclary:
> So maybe i remove the Select....?

Agree that having a `NONE` option is confusing, but I think we should leave this as-is. Across the app, the "Select..." placeholder text is meant to indicate no value/null.

chiaberry:
I guess my question could be rephrased then "why not show 'None' in the UI if its an option that is different from null?

mddilley (approved):
Tested with different roles and temp/non-temp, and everything works as described. Code looks awesome too! 🚢 😎

editor/components/ContributingFactorsModal.tsx
```diff
@@ -0,0 +1,164 @@
...
+  const onSave: SubmitHandler<ContributingFactorsInputs> = async (data) => {
+    setIsSubmitting(true);
+    await mutate({
+      id: unit.id,
+      updates: data,
+    });
+    await onSaveCallback();
+    onClose();
+    setIsSubmitting(false);
+  };
```
mateoclarke, line 98: This is a really clean async/await pattern. I really like the way the `isSubmitting` bookends the whole transaction. 👍

mateoclarke (approved):
Awesome work Rose! 🚢

johnclary (approved):
Let's goooo 🚢 🚢 🚢

PR #2020, opened 2026-04-14
Ensure the EMS popup shows all fields regardless of column visibility settings

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/27881

## Testing

**URL to test:** [Netlify](https://deploy-preview-2020--atd-vze-staging.netlify.app/)

1. Go to the EMS list. Confirm that the table is set to display 1,000 rows per page by default.
2. Use the column visibility settings menu (cog icon to the right of the table pagination controls) to hide the **Date**, **Incident address**, and **Incident number** columns.
3. Switch the map view and click on a circle marker — confirm that the incident address, date, and crash ID are all rendered correctly. Repeat steps 2 and 3 in prod and you will see the bug 🐛 
4. Repeat these steps (confirm 1k row limit and hide visible columns) on these pages as well:

* [Fatalities list](https://deploy-preview-2020--atd-vze-staging.netlify.app/editor/fatalities) (columns: Crash ID, YTD Fatal Crash, Crash Date)
* [Location details crashes list](https://deploy-preview-2020--atd-vze-staging.netlify.app/editor/locations/153D618A97) (columns: Type, Crash ID, Case ID, Date, Address)
* Crashes (columns: **Crash ID**)

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/configs/emsColumns.tsx
+++ b/editor/configs/emsColumns.tsx
@@ -154,6 +154,7 @@ export const ALL_EMS_COLUMNS = {
   cris_crash_id: {
     path: "crash.cris_crash_id",
     label: "Crash ID",
+    fetchAlways: true,
     relationship: {
       tableSchema: "public",
       tableName: "crashes",
@@ -181,6 +182,7 @@ export const ALL_EMS_COLUMNS = {
     path: "incident_location_address",
     label: "Incident address",
     sortable: true,
+    fetchAlways: true,
   },
   incident_number: {
     path: "incident_number",
@@ -204,6 +206,7 @@ export const ALL_EMS_COLUMNS = {
     style: { whiteSpace: "nowrap" },
     valueFormatter: formatIsoDateTime,
     sortable: true,
+    fetchAlways: true,
   },
   incident_received_datetime: {
     path: "incident_received_datetime",
@@ -238,11 +241,15 @@ export const ALL_EMS_COLUMNS = {
   },
   latitude: {
     path: "latitude",
-    label: "Latitude"
+    label: "Latitude",
+    fetchAlways: true,
+    exportOnly: true,
   },
   longitude: {
     path: "longitude",
-    label: "Longitude"
+    label: "Longitude",
+    fetchAlways: true,
+    exportOnly: true,
   },
   mvc_form_position_in_vehicle: {
     path: "mvc_form_position_in_vehicle",
--- a/editor/configs/emsColumns.tsx
+++ b/editor/configs/emsColumns.tsx
@@ -154,7 +154,6 @@ export const ALL_EMS_COLUMNS = {
   cris_crash_id: {
     path: "crash.cris_crash_id",
     label: "Crash ID",
-    fetchAlways: true,
     relationship: {
       tableSchema: "public",
       tableName: "crashes",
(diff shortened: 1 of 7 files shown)


editor/configs/emsListViewTable.ts
```diff
@@ -440,7 +440,7 @@ export const emsListViewQueryConfig: QueryConfig = {
   exportable: true,
   exportFilename: "ems_patient_care_records",
   tableName: "ems__incidents",
-  limit: DEFAULT_QUERY_LIMIT,
+  limit: 1000,
```
johnclary, line 442: This change matches our crashes table config. The benefit of showing 1,000 records by default is that it makes the map nice and full of features.

editor/configs/emsColumns.tsx
```diff
@@ -238,11 +241,15 @@ export const ALL_EMS_COLUMNS = {
   },
   latitude: {
     path: "latitude",
-    label: "Latitude"
+    label: "Latitude",
+    fetchAlways: true,
+    exportOnly: true,
```
johnclary, line 246: `exportOnly` causes these fields to be hidden from the list view entirely.

Charlie-Henry (commented):
I am not seeing date populated? 

Image

editor/configs/emsColumns.tsx
```diff
@@ -211,6 +213,7 @@ export const ALL_EMS_COLUMNS = {
     style: { whiteSpace: "nowrap" },
     valueFormatter: formatDate,
     sortable: true,
+    fetchAlways: true,
```
johnclary, line 216: @Charlie-Henry thanks for flagging that the date visibility was busted. I had attached this prop to the wrong date field. Should be good now!

chiaberry (approved):
looks good! 🚢

Charlie-Henry (approved):
worked!

editor/configs/emsColumns.tsx
```diff
@@ -154,6 +154,7 @@ export const ALL_EMS_COLUMNS = {
   cris_crash_id: {
     path: "crash.cris_crash_id",
     label: "Crash ID",
+    fetchAlways: true,
```
roseeichelmann: i think you need to make this update to the crashColumns, bc Crash ID is not populating on crashes map popups when that col viz is turned off

editor/configs/emsColumns.tsx
roseeichelmann, line 189: this incident field needs to be updated with the `fetchAlways` instead of the crash ID for EM bc this is the id field thats used on the popup

so could add to your test instructions to test turning off Incident from col viz and make sure its still showing on the popup

roseeichelmann:
We should go ahead and fix this for Crash ID and YTD fatal crash in Fatalities map view as well?

Also Type, Case ID, Address, and Date in the Non Cr3 map

johnclary:
@roseeichelmann thanks for uncovering the other places these settings were missing. I retested all the maps, and updated the settings across the board. The test instructions reflect that there are changes to be tested on the fatalities list, location crashes list, and crashes list as well 🙏

mddilley (approved):
Tricky bug! And very cool to see an existing setting waiting to help fix. I tested with the latest changes and instructions, and I see all the expected info in the popups. 🙌 🚢

mateoclarke (approved):
Image

PR #2028, opened 2026-04-21
Ability to edit charges for temp crashes

roseeichelmann:
## Associated issues
Closes https://github.com/issues/assigned?issue=cityofaustin%7Catd-data-tech%7C27645

## Testing

**URL to test:** 
Local

**Steps to test:**
2. Create a temp crash record with multiple units with fatalities and go to its fatality details page
3. Test adding charges to the units, newlines should be preserved and shown so that users can format/separate each charge
4. Now test editing and saving the same Charges text box
5. Test that contrib factors still works as before
6. Make sure you cant edit non non-temp crash records
7. Test read only user cant edit
8. Test down migration

---

#### Ship list

- [ ] Code reviewed


editor/components/ChargesModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+          <Form.Group>
+            <Form.Control
+              as="textarea"
+              aria-label="Charges text"
+              rows={6}
+              placeholder="Enter charges..."
+              autoFocus={true}
+              {...register("charge", {
+                // coerce empty fields to null
+                setValueAs: (v) => v?.trim() || null,
```
roseeichelmann, line 97: should we auto CAPS the users input as they type?

editor/components/ContributingFactorsModal.tsx
```diff
@@ -66,23 +66,19 @@ export default function ContributingFactorsModal({
     typename: "lookups_contrib_factr",
   });
 
-  const defaultValues = useMemo(() => {
```
roseeichelmann, line 69: i realized this `useMemo` was unnecessary

database/migrations/default/1777314214182_updates_to_charges_cris_table/up.sql
```diff
@@ -0,0 +1,23 @@
...
+-- via the Fatality Details page
+
+alter table "public"."charges_cris" add column "is_temp" boolean
+ not null default 'false';
+
+alter table "public"."charges_cris" add constraint "check_cris_crash_id_is_temp_dependency" check (cris_crash_id is not null or is_temp is true);
+
+alter table "public"."charges_cris" alter column "cris_crash_id" drop not null;
+
+alter table "public"."charges_cris" alter column "person_id" drop not null;
```
roseeichelmann: is this okay?

johnclary: I'd rather see us populate this for temp charges since we do have a value we can use. Can we pull this through the `units.people` relationship?

database/migrations/default/1777314214182_updates_to_charges_cris_table/up.sql
```diff
@@ -0,0 +1,23 @@
...
+alter table "public"."charges_cris" add column "updated_at" timestamptz
+ not null default now();
+
+alter table "public"."charges_cris" add column "updated_by" text
+ null default 'system';
+
+drop trigger if exists set_charges_cris_person_id_and_crash_pk on charges_cris;
+
+CREATE TRIGGER set_charges_cris_person_id_and_crash_pk BEFORE INSERT ON public.charges_cris FOR EACH ROW 
+when (NEW.is_temp = false)
```
johnclary, line 25: nice! much cleaner than rebuilding the whole function like i suggested ✨

database/migrations/default/1777314214182_updates_to_charges_cris_table/up.sql
```diff
@@ -0,0 +1,23 @@
...
+ not null default 'false';
+
+alter table "public"."charges_cris" add constraint "check_cris_crash_id_is_temp_dependency" check (cris_crash_id is not null or is_temp is true);
+
+alter table "public"."charges_cris" alter column "cris_crash_id" drop not null;
+
+alter table "public"."charges_cris" alter column "person_id" drop not null;
+
+alter table "public"."charges_cris" add column "updated_at" timestamptz
+ not null default now();
```
johnclary, line 16: thank you for adding these! to keep this field in sync you'll need to add the trigger that calls `set_current_timestamp_updated_at()`

editor/components/ChargesModal.tsx
johnclary: As an all-caps hater I am inclined to say no. In the examples we have from Mary she does not write charges in all caps (even though CRIS charges are all-caps).

editor/components/FatalityUnitCardFooter.tsx
```diff
@@ -93,16 +96,67 @@ export default function FatalityUnitCardFooter({
...
+                  <AlignedLabel>
+                    <LuCirclePlus />
+                  </AlignedLabel>
+                </Button>
+              </span>
+            )}
+          </div>
+        </div>
+      )}
+      {hasCharges && (
```
johnclary: Since we don't have the ability to delete charges, what do you think about handling the case of when the temp charge is blank, so that the plus button shows instead of the edit button?

Here's what I see when I add a charge, save, and then clear the charge text and save:

Image

Kinda gross, but something like this?

```
{hasCharges && isTempRecord && !!unitCharges[0].charge &&  <>}
```

johnclary (commented):
This is looking really good, Rose!

editor/components/FatalityUnitCardFooter.tsx
roseeichelmann: should we also show the None?

chiaberry: i like seeing the plus icon if its empty, +1

chiaberry (commented):
Test steps work out and I logged in as a read only user and was not able to edit anything. 

The down migration doesnt work because of the null constraint being added back to the charges table, but if the units.people relationship gets added, that situation will be handled. 

I'll swing back and do a quick re-test after the migrations are reworked and then approve the PR.

database/migrations/default/1777314214182_updates_to_charges_cris_table/up.sql
roseeichelmann: im glad you had me update this :ok_hand:

database/migrations/default/1777314214182_updates_to_charges_cris_table/up.sql
```diff
@@ -0,0 +1,26 @@
...
+
+-- This command cannot be undone in the down migration because if we delete this column then the following command with the 
+-- check_cris_crash_id_is_temp_dependency constraint will be violated
+alter table "public"."charges_cris" add column if not exists "is_temp" boolean
+ not null default 'false';
+
+alter table "public"."charges_cris" add constraint "check_cris_crash_id_is_temp_dependency" check (cris_crash_id is not null or is_temp is true);
+
+-- This command can not be undone in the down migration because the temp records added will violate
+-- the not null constraint
```
roseeichelmann, line 12: i dont know how else to handle this? 😬

database/migrations/default/1777314214182_updates_to_charges_cris_table/up.sql
```diff
@@ -0,0 +1,26 @@
+-- These updates enable users to insert and update charge records for temporary records
+-- via the Fatality Details page
+
+-- This command cannot be undone in the down migration because if we delete this column then the following command with the 
+-- check_cris_crash_id_is_temp_dependency constraint will be violated
+alter table "public"."charges_cris" add column if not exists "is_temp" boolean
```
mddilley, line 6: i was going to ask what happens to temp charges when the temp crash is deleted, but then i remembered that crash records soft-delete so it makes sense for the charge record and connection to remain. Both rows being marked as temp makes it easy to trace when I was exploring the DB. 👍

editor/components/ChargesModal.tsx
```diff
@@ -0,0 +1,122 @@
...
+      : {
+          updates: {
+            charge: data.charge,
+            person_id: unit.people.filter((person) => (person.prsn_nbr = 1))[0]
+              .id,
+            unit_nbr: unit.unit_nbr,
+            prsn_nbr: 1,
+            crash_pk: crashPk,
+            is_temp: true,
+            cris_schema_version: "2026",
```
mddilley: i'm not close enough to this work to know, but could this schema version be set app-wide? this is me making a huge assumption that the same version applies to every record type.

database/migrations/default/1777314214182_updates_to_charges_cris_table/up.sql
mddilley: this makes sense to me for it to be up-only and your comment captures the reason well too (that the data collected by the new functionality will block reversing it)

mddilley (approved):
I followed the test steps, caught up on the discussion, and I was able to add, edit, and null out temp charges as an editor. Looks like there might be some lingering questions about the migrations, but I am going to approve since this worked as described. 🚢 😎

chiaberry (approved):
I have retested and its all ✔️ ✔️

database/migrations/default/1777314214182_updates_to_charges_cris_table/up.sql
johnclary: Agreed 🙌

editor/components/ChargesModal.tsx
```diff
@@ -52,6 +52,8 @@ export default function ChargesModal({
       : {
           updates: {
             charge: data.charge,
+            person_id: unit.people.filter((person) => (person.prsn_nbr = 1))[0]
+              .id,
```
johnclary, line 57: This is highlighting a shortcoming in our `Unit` type, which is that `people` should be nullable as a safeguard against malformed queries. Would you mind updating that type and adding conditional chaining here?

Edit: i noticed another issue with this relationship that I mention further on.

editor/components/ChargesModal.tsx
johnclary: That's a pretty good idea. Looks like we're currently setting this as `2023` in the `CreateCrashRecordModal`. 

It's an awkward thing in general, because this field is pretty meaningless for user-created records. 

@roseeichelmann what do you think about adding schema year to our `constants.ts` file and updating it wherever its used?

editor/components/CrashNarrativeEditableCard.tsx
```diff
@@ -95,7 +95,9 @@ export default function CrashNarrativeEditableCard({
         <Card.Body className="crash-header-card-body">
           <Tab.Content>
             <Tab.Pane eventKey="narrative">
-              {crash.investigator_narrative || ""}
+              <div style={{ whiteSpace: "pre-wrap" }}>
+                {crash.investigator_narrative || ""}
+              </div>
```
johnclary, line 100: thank you!

editor/components/FatalityUnitCardFooter.tsx
```diff
@@ -117,7 +117,7 @@ export default function FatalityUnitCardFooter({
           </div>
         </div>
       )}
-      {hasCharges && (
+      {hasCharges && !!unitCharges?.[0].charge && (
```
johnclary, line 120: looks great—thanks!

editor/types/unit.ts
```diff
@@ -31,4 +32,5 @@ export type Unit = {
   autonomous_unit_id: number | null;
   autonomous_level_engaged: LookupTableOption | null;
   rpt_autonomous_level_engaged_id: number | null;
+  people: PeopleListRow[];
```
johnclary: Looking in Hasura, this relationship is actually tied to the `people` table and not  the `people_list_view`. 

Two possible routes here:
* Add a `People` or `Person` type to app. I guess this could be a super-slimmed down type that includes just the fields we're using at the moment.
* Add a `people_list_view` relationship to units

🤔

johnclary (commented):
Your updates look great, Rose. Just one issue I encountered with our rather confusing people vs people_list_view relationships.

editor/components/ChargesModal.tsx
roseeichelmann: done! i hit all the places its used

editor/components/ChargesModal.tsx
roseeichelmann: does this still apply for when using the people table? just updated to use that instead and i made the relationship nullable as you said here

editor/types/unit.ts
roseeichelmann: thanks for catching this!! update is in [this commit](https://github.com/cityofaustin/vision-zero/pull/2028/changes/ce9e353e3ca7c3ad5adaea3ca4e4a5989ba4f976)

johnclary (approved):
🚢 🚢 🚢 🚢 🚢 🚢

mateoclarke (approved):
Tested locally as admin and read-only and works as described. Very nice. I especially like the detail of the add icon vs edit icon changing if the content is null. 🚢 

_Couple of thoughts not related to Rose's enhancements, just the general flow of creating temp records as you had us do in your testing instructions. cc/ @johnclary_ 

1. When we create a temp record, should we auto-navigate to the details page of the new record on success? Seems like in most cases, you'd either want to see the details as confirmation or just go straight into adding/editing attributes. Maybe this is a bad assumption in the case where you're creating a batch of temp records? Not sure how common that is.
2. Do we have a way to navigate from the crash details page to the related fatality details page? If not, should we? I found some friction in navigating to the new temp crash I created by having to go from the Crashes index > the click in the side to navigate to Fatalities index >  Then eyeball scan for my temp crash which wasn't at the top of the list > Fatality detail
3. In the Fatalities index, when there are multiple fatality records for a crash, they both show in the table. At first, I thought maybe I created the temp twice as a dupe. I know technically there were 2 individual fatalities, but each row's link takes you to the same Fatality details page. Should we dedupe the table by Case or Crash ID?

johnclary:
@mateoclarke 

1. should we auto-navigate to the details page of the new [temp] record on success?

Xavier requested that we do not do this, because he enters multiple temp records at a time. I have also been second-guessing this because it can be pretty confusing in cases where you add a temp record and, due to either your list filters or the date of the temp crash, it does not appear on the screen after you create it.


2. Do we have a way to navigate from the crash details page to the related fatality details page? 

Agreed this is needed! Was thinking we might add the link to the FTT details to the **People** table but haven't gotten around to scoping that.

3. When there are multiple fatality records for a crash, they both show in the [fatalities list] table

My understanding is that this is preferred, since the table functions as a running list of VZ fatalities. The **YTD Fatality** column is meant to disambiguate these rows.

PR #2044, opened 2026-05-21
Adds ETL to group CAD records as a single incident

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/27761

This PR adds a new script that groups CAD incidents together based on spatial and temporal proximity. It introduces a bare bones `vz_incidents` table to track these relationships. 

## Testing

**URL to test:** Local

**Steps to test:**

1. Replicate your DB to get latest set of CAD incident data locally. Apply migrations and metadata.

2. Follow the `./etl/cad_incidents` readme to get set up for local dev. If you've already tested the other CAD ETLs you are good to go.

3. Run the ETL, making sure to use the `--limit` param or your script will run for a very long time.

```
docker compose -f docker-compose.yml -f docker-compose.local.yml run import incident_linker.py --limit 2000
```

4. You can use this query to inspect the way records are getting grouped:

```sql
SELECT
  vz_incident_id,
  count(*) AS member_count,
  string_agg(DISTINCT agency_type, ', ') AS agency_types,
  string_agg(DISTINCT address, ', ') AS addresses,
  min(response_date) AS earliest_response,
  max(response_date) AS latest_response,
  max(response_date) - min(response_date) AS time_span,
  ST_Length(
    ST_LongestLine(ST_Collect(geom), ST_Collect(geom))::geography
  ) AS max_spread_meters
FROM cad_incidents
WHERE geom IS NOT NULL and vz_incident_id is not null
GROUP BY vz_incident_id
ORDER BY vz_incident_id;
```

5. The output of this query can be used to visualize the incident groups on a map. Copy the result into [geojson.io](http://geojson.io/) or [placemark](https://www.placemark.io/):

```sql
SELECT
    jsonb_build_object(
        'type',
        'FeatureCollection',
        'features',
        jsonb_agg(
            jsonb_build_object(
                'type',
                'Feature',
                'geometry',
                ST_AsGeoJSON (geom)::jsonb,
                'properties',
                jsonb_build_object(
                    'vz_incident_id',
                    vz_incident_id,
                    'master_incident_id',
                    master_incident_id,
                    'agency_type',
                    agency_type,
                    'response_date',
                    response_date,
                    'address',
                    address,
                    'marker-color',
                    '#' || lpad(to_hex(('x' || substr(md5(vz_incident_id::text), 1, 6))::bit(24)::int), 6, '0'),
                    'marker-size',
                    'large'
                )
            )
        )
    ) AS geojson
FROM
    cad_incidents
WHERE
    vz_incident_id IS NOT NULL
LIMIT
    2000;
```

6. Run the down migration

---

#### Ship list

- [ ] Code reviewed


etl/cad_incidents_import/explore_incident_match.py
johnclary, line 1: This is LLM code that I am going to remove before merge. It might be of interest to folks interested in seeing the effects of different time + distance thresholds.

etl/cad_incidents_import/incident_linker.py
```diff
@@ -0,0 +1,161 @@
...
+            "geom": point_geojson(incident["longitude"], incident["latitude"]),
+            "start": start.isoformat(),
+            "end": end.isoformat(),
+            "distance": meters_to_degrees(DISTANCE_THRESHOLD_M),
+        },
+    )
+    return data["cad_incidents"]
+
+
+def is_group_closed(primary: dict, matches: list[dict]) -> bool:
```
johnclary: This is the function that identifies a close group/set of incidents. Incidents are only grouped together if every incident in group exclusively matches the other incidents.

Here are some examples:

This is a group of three incidents where all three occurred within the time and distance thresholds.

Image


Here we have a group of two incidents (yellow) that matched based on their time and location. A third incident occurred a few minutes outside of our time cutoff, so it was not included in the group. We may want to adjust our time/distance threshold to try to catch this case. In the future there will likely be a manual QA/QC process to review such cases.

Image

And here we have a situation where no incidents were grouped together. Their timestamps fall within our threshold, but westernmost and easternmost points are not within the required distance (illustrated by the fuchsia circles). So no grouping has been made at all. Further investigation needed if we want to adjust our rules to capture this incident.

Image

etl/cad_incidents_import/README.md
```diff
@@ -70,6 +72,16 @@ docker compose -f docker-compose.yml -f docker-compose.local.yml run import inci
...
 
+### `incident_linker.py`
+
+![diagram](docs/cad_incidents_matched.png)
+
+It is common for multiple CAD records to be created in response to a single incident that happens in the real-world. This script groups CAD incidents together based on their time and location.
+
+```shell
+docker compose -f docker-compose.yml -f docker-compose.local.yml run import incident_linker.py --limit 500
+```
```
johnclary: Before merging, I intend to expand on these docs by adding in some of the matching examples I provided in comments here.

etl/cad_incidents_import/incident_linker.py
```diff
@@ -0,0 +1,161 @@
...
+
+from utils.graphql import make_hasura_request
+from utils.queries import (
+    SET_VZ_INCIDENT_IDS,
+    GET_UNPROCESSED_INCIDENTS,
+    INSERT_VZ_INCIDENT,
+    GET_POTENTIAL_MATCHES,
+)
+
+MIN_RECORD_AGE_HOURS = 24
```
johnclary, line 15: We ignore the newest records in our database to ensure that aren't processing any live incidents or incidents at the edge of our daily export cutoff

etl/cad_incidents_import/incident_linker.py
```diff
@@ -0,0 +1,161 @@
...
+        counts["cad_incidents_processed"] += len(group_ids)
+        counts["ambiguous"] += 1 if not closed else 0
+        counts["skipped"] += len(group_ids) - 1
+
+    logging.info(f"\n=== Done ===")
+    logging.info(
+        f"  CAD recorsds processed (including skipped): {counts['cad_incidents_processed']:,}"
+    )
+    logging.info(f"  VZ incidents created: {counts['vz_incidents_created']:,}")
+    logging.info(f"  Ambiguous incidents seen: {counts['ambiguous']:,}")
```
johnclary: A group is flagged  "ambiguous" if it has partial membership but can't be closed (like this slaughter ln example from above). It has been helpful to see this logging while developing the script but we're not storing this flag anywhere.

etl/cad_incidents_import/incident_linker.py
johnclary: The updated "greedy" algo now groups all three of these CAD records (brown dots).

Screenshot 2026-05-29 at 3 25 50 PM

database/migrations/default/1778695662683_cad_incident_links/up.sql
```diff
@@ -0,0 +1,34 @@
...
+        OR lower(call_disposition) like '%training%'
+    false)
+) STORED,
+ADD COLUMN vz_incident_id bigint REFERENCES vz_incidents(id);
+
+COMMENT ON COLUMN public.cad_incidents.is_cancelled_call is 'If the call disposition indicated that the call was canceled, a duplicate, false alarm, etc.';
+COMMENT ON COLUMN public.cad_incidents.vz_incident_id is 'The vz_incidents foreign key. When null, indicates that this incident has not been processed by the VZ incident grouping ETL.';
+
+CREATE INDEX cad_incidents_geom_geography_idx 
+    ON cad_incidents USING GIST ((geom::geography));
```
johnclary, line 32: I found that this massively speeds up queries that cast to geography—see the sample queries for example. We have a similar index on the `crashes` table.

database/migrations/default/1778695662683_cad_incident_links/up.sql
```diff
@@ -0,0 +1,34 @@
...
+  coalesce(
+        lower(call_disposition) LIKE '%cancelled%'
+        or lower(call_disposition) LIKE '%canceled%'
+        OR lower(call_disposition) LIKE '%false alarm%'
+        OR lower(call_disposition) LIKE '%duplicate%'
+        OR lower(call_disposition) LIKE '%reassigned call%'
+        OR lower(call_disposition) LIKE '%test call%'
+        OR lower(call_disposition) like '%training%'
+    false)
+) STORED,
```
johnclary, line 25: Still TBD on how we want to handle these canceled/dupe calls, but they are very common and so I thought I'd go ahead and codify this.

etl/cad_incidents_import/incident_linker.py
```diff
@@ -0,0 +1,146 @@
...
+from utils.queries import (
+    SET_VZ_INCIDENT_IDS,
+    GET_UNPROCESSED_INCIDENTS,
+    INSERT_VZ_INCIDENT,
+    GET_POTENTIAL_MATCHES,
+)
+
+MIN_RECORD_AGE_HOURS = 24
+DISTANCE_THRESHOLD_M = 500
+TIME_THRESHOLD_MINUTES = 60
```
johnclary, line 17: i generated these based a few iterations of eyeballing the results. we'll almost definitely revisit once Xavier and Joel have a chance to review.

etl/cad_incidents_import/incident_linker.py
```diff
@@ -0,0 +1,146 @@
...
+    )
+    return data["cad_incidents"]
+
+
+def flood_fill_group(anchor: dict) -> set[str]:
+    """
+    Creates groups of incidents by recursively expanding a group starting
+    from anchor by bread-first search. Each newly discovered member's own
+    neighbors are explored in turn, until the frontier is exhausted.
+    """
```
johnclary, line 56: See `cad_incidents_chained.png` for a visual of what's going on here.

I initially approached this differently, using a check to make sure groups of incidents were "closed"—each incident in the group had to be within the time and date cutoff of all others. I ultimately found this to be too conservative—it was leaving some obvious incident groups unlinked. See the resolved comments and earlier commits in this PR if you want to see what that looked like.

etl/cad_incidents_import/incident_linker.py
```diff
@@ -0,0 +1,146 @@
...
+            nid = neighbor["master_incident_id"]
+            if nid not in group_ids:
+                group_ids.add(nid)
+                frontier.append(neighbor)
+
+    return group_ids
+
+
+def main(args):
+    record_limit = args.limit
```
johnclary, line 73: this limit function is handy for dev and for the backfill we'll need to do but won't be needed in prod, where we'll process all new incidents daily.

etl/cad_incidents_import/docs/cad_incidents_chained.png
johnclary: i don't love adding 1.7mb to the codebase. what's the best practice here?

etl/cad_incidents_import/incident_linker.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def time_window(response_date: datetime, minutes: float) -> tuple[datetime, datetime]:
+    delta = timedelta(minutes=minutes)
+    return response_date - delta, response_date + delta
+
+
+def meters_to_degrees(meters):
+    """Rough conversion for _st_d_within on a 4326 geometry column."""
+    return meters / 111_000
```
frankhereford: One degree along the x-axis is ~96K meters -- what do you think about taking the 111K and making it 104K to split the error approximately equally over both axes? I thought about asking about projections, but I think this is a fine approximation and super fast.

Image

etl/cad_incidents_import/docs/cad_incidents_chained.png
frankhereford, line 1: I think if you compress it with a more aggressive format like JPG, you can get better results. This is the same size and comes in at 150K. If you want to keep it loseless, dropping the alpha channel out of your PNG should let it pick up 1/4 of the data needed to recreate the image maybe. 

Image

etl/cad_incidents_import/incident_linker.py
```diff
@@ -0,0 +1,146 @@
...
+
+    while frontier:
+        current = frontier.pop()
+        for neighbor in fetch_potential_matches(current):
+            nid = neighbor["master_incident_id"]
+            if nid not in group_ids:
+                group_ids.add(nid)
+                frontier.append(neighbor)
+
+    return group_ids
```
frankhereford, line 69: This is written really well, it reads so clearly. It makes sense to me how it will explore and fill the linked chains of locations. ✨

frankhereford (approved):
This is a really great script, I enjoyed reading your work here. It ran just like expected and is fast. Thanks for the geojson-building query. It was interesting to see all the points distributed on a map.

database/README.md
```diff
@@ -268,12 +270,99 @@ The view `crashes_change_log_view` provides a unioned view of the unified table
...
+    JOIN cad_incidents c ON c.vz_incident_id = v.id
+WHERE
+    v.is_deleted = FALSE
+GROUP BY
+    v.id
+ORDER BY
+    v.id
+```
+
+* Generates a geojson of VZ incidents with simple styles that can be visualized in mapping tools such as https://geojson.io.
```
chiaberry, line 323: ✨

etl/cad_incidents_import/incident_linker.py
chiaberry: even though i love looking for 🍞 , its still a typo

chiaberry (approved):
thank you for including the test queries in the readme ✨

etl/cad_incidents_import/docs/cad_incidents_chained.png
johnclary: brilliant—thank you!!

I am going to use [`git-filter-repo`](https://github.com/newren/git-filter-repo) to purge this png from history and use the jpeg instead 😅

etl/cad_incidents_import/incident_linker.py
johnclary: Love it—thank you for taking a closer look at my lazy conversion!

database/migrations/default/1778695662683_cad_incident_links/up.sql
```diff
@@ -0,0 +1,34 @@
+CREATE TABLE vz_incidents (
+    id bigserial PRIMARY KEY,
```
mddilley, line 2: this is a good reminder that i need to create an issue to update our ecapris status id column to bigserial where the upserts burn up ids 🔥

etl/cad_incidents_import/utils/queries.py
```diff
@@ -25,3 +25,66 @@
...
+    $geom: geometry!
+    $start: timestamptz!
+    $end: timestamptz!
+    $distance: Float!
+) {
+    cad_incidents(
+        where: {
+            master_incident_id: { _neq: $incident_id }
+            response_date: { _gte: $start, _lte: $end }
+            geom: { _st_d_within: { distance: $distance, from: $geom } }
```
mddilley, line 61: this is either a TIL or a TIR (today i remembered) that these postgis functions are available in the gql queries 🚀

etl/cad_incidents_import/incident_linker.py
```diff
@@ -0,0 +1,146 @@
...
+        },
+    )
+    return data["cad_incidents"]
+
+
+def flood_fill_group(anchor: dict) -> set[str]:
+    """
+    Creates groups of incidents by recursively expanding a group starting
+    from anchor by breadth-first search. Each newly discovered member's own
+    neighbors are explored in turn, until the frontier is exhausted.
```
mddilley, line 55: this is super cool, and it makes sense to me. is that right and it is because the incidents in the data are sparse in time and proximity that the while loop won't take a really long time? 

i don't have any concern about the code but it isn't that often that we get to think through problems like this so i'm curious. 🙏

mddilley (approved):
I tested locally and read through the code, and everything looked as expected. Thanks for the docs that made it straightforward for me to spin up and get a handle on what this is doing. 🙏 🚢

etl/cad_incidents_import/README.md
```diff
@@ -70,6 +72,22 @@ docker compose -f docker-compose.yml -f docker-compose.local.yml run import inci
...
 
+### `incident_linker.py`
+
+It is common for multiple CAD records to be created in response to a single incident that happens in the real-world. The `incident_linker.py` script groups CAD incidents together based on their time and location, creating a new `vz_incidents` record for each group it finds.
+
+The script works as follows:
+
+* Only CAD incidents older than 24 hours are considered, to ensure records have had time to be fully populated before processing.
+
+* Starting from an unprocessed anchor incident, the script flood-fills outward, pulling in any neighboring incidents within 500 meters and 60 minutes of each group member. This process repeats for each newly added member until no new neighbors are found.
```
roseeichelmann, line 83: its a really cool pattern you have implemented w the "neighbors"!

roseeichelmann (approved):
tested the ETL and down migration and reviewed the code, this is great and its cool to see the overarching VZ "incidents" idea we've been throwing around getting put into action! im excited to see how we expand this out

sidenote but funny to randomly see 2 points on a diff continent when putting the output of the query into the geojson tool, must have been some typos on their coordinates, afaik Austin PD isnt stationed in Morocco 😄

etl/cad_incidents_import/incident_linker.py
johnclary: that's right. the guardrails keeping this from running recursively through the whole dataset are the narrow time + distance constraints. in the sample of 10k cad records i processed, the average group size was 1.8 and the largest group had 10 incidents.

johnclary:
> sidenote but funny to randomly see 2 points on a diff continent when putting the output of the query into the geojson tool

Thanks for mentioning that. It's bugging me to have that crufty geospatial data in there. I think we might want to set up a trigger that nulls out any point location that is obviously corrupted.

PR #2045, opened 2026-05-22
Ability to autocomplete/typeahead long select inputs

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/27975

The issue first said to extend the existing [CrashSearchTypeahead](https://github.com/cityofaustin/vision-zero/blob/main/editor/components/CrashSearchTypeahead.tsx) component to be able to be used throughout the app. I initially began doing that and then found that that component leans heavily on the query to to search and filter the options, users do not get a full list of options due to the number of crashes. So I reverted changes to that component and created this new one. 

## Testing

**URL to test:** 

https://deploy-preview-2045--vze-staging.netlify.app/editor/fatalities

**Steps to test:**

The component currently exists on the contributing factors inputs on the temporary fatalities pages. 

https://deploy-preview-2045--vze-staging.netlify.app/editor/fatalities/T1371165

Add or edit new contributing factors. Click into one of the inputs, the dropdown should open with all of the options. 
Begin typing in the input, try the word "over", as you type the options narrow down to fit your input. 
Selections are not saved until the form is submitted. 

Edit the contributing factors. Make sure the saved option is populated in the input. 

---

#### Ship list

- [ ] Code reviewed


editor/components/FormControlAutocomplete.tsx
johnclary, line 1: What do you think about calling this `FormControlTypeahead` or `FormControlAutocomplete`? That would be consistent with our generic `FormControlDatePicker`.

editor/components/InputLookupTypeahead.tsx
```diff
@@ -0,0 +1,130 @@
...
+          disabled={disabled}
+          value={searchInput}
+          onChange={(e) => {
+            setSearchInput(e.target.value);
+            setHighlightedIndex(0);
+            setShowDropdown(true);
+          }}
+          onKeyDown={handleKeyDown}
+          onFocus={() => setShowDropdown(true)}
+          onBlur={() => setTimeout(() => setShowDropdown(false), 200)}
```
johnclary: i didn't asked about this when I reviewed the CrashSearchTypeahead component—any idea why the setTimeout is needed here?

I think we also want to call `field.onBlur()` inside this function, because that triggers RHF validation. Not needed at the moment, but would be good to not have that muted.

This could also be a place to clear the searchInput if we wanted to take the approach of treating it as a "draft".

editor/components/InputLookupTypeahead.tsx
```diff
@@ -0,0 +1,130 @@
...
+          autoComplete="off"
+        />
+        {showDropdown && (
+          <ul
+            className="list-group w-100 mt-1 shadow-sm"
+            style={{ maxHeight: "240px", overflowY: "auto" }}
```
johnclary: Would you add absolute positioning to the dropdown, so that it does't push other elements around when it expands?

editor/components/InputLookupTypeahead.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const { field } = useController({ name, control });
+
+  useEffect(() => {
+    if (field.value) {
+      const selectedOption = options.find(
+        (option) => option.id === field.value
+      );
+      setSearchInput(selectedOption?.label ?? "");
+    }
+  }, [field.value, options]);
```
johnclary: There is some weirdness when clearing the input or if you type in a value that doesn't have a matching option. I think the desired behavior is:

1. If you clear a selected option (searchInput is `""`), this should cause the field to be saved as `null`. (Future enhancement: support required/non-nullable fields)
2. If you input text but click away (blur) without selecting an option, the input should revert to the previous value (null if the field was empty or the previously selected id)

The way this is currently wired, there's a minor bug where you can't select the already selected option. You can reproduce:

1. Save a contributing factor, e.g. `ANIMAL ON ROAD - DOMESTIC`.
2. Edit the contributing factor by clicking into the input and backspacing a few of the characters (`ANIMAL ON ROAD - DOME`).
3. Try to select the option `ANIMAL ON ROAD - DOMESTIC`

johnclary: This is looking to me like a case where "You might not need a useEffect".  Instead of trying to keep the `searchInput` in sync with either the selected value or the input text, perhaps the searchInput could be treated as a "draft" state that's only non-null while the user is actively typing. 

You could derive the displayed label from field.value: `const selectedOption = options.find(o => o.id === field.value)?.label ?? ""`.  Then the input `value` can be either the draft state or the selected label (``searchInput ?? selectedLabel``). 

I'm rambling here but would be happy to pair to see what we can do to nix that useEffect.

johnclary (commented):
@chiaberry thanks for brining this feature to the contributing factors modal—this is a major improvement versus picking from those mega dropdowns.

I left some feedback about how I think the input should work when we don't have a matching/selected value that I'm sure would have been helpful to have earlier. Happy to discuss/clarify.

editor/components/FormControlAutocomplete.tsx
chiaberry: sounds good to me!

editor/components/InputLookupTypeahead.tsx
chiaberry: I will put it back. Including it was preventing the dropdown from showing on Safari and Firefox, I will figure out what combination of positioning I need to fix that and have it absolutely positioned

editor/components/InputLookupTypeahead.tsx
chiaberry: I don't actually know why the timeout is needed here, I copied it from the CrashSearchTypeahead component thinking that maybe without it the dropdown closing was too jarring. But it feels fine with out it. I am removing it.

editor/components/InputLookupTypeahead.tsx
chiaberry: I removed the useEffect and followed your suggestions

editor/components/FormControlAutocomplete.tsx
```diff
@@ -0,0 +1,132 @@
...
+        onBlur={() => {
+          setShowDropdown(false);
+          field.onBlur();
+        }}
+        autoComplete="off"
+      />
+      {showDropdown && (
+        <ul
+          className="list-group w-100 mt-1 shadow-sm position-absolute"
+          style={{ zIndex: 100, maxHeight: "240px", overflowY: "auto" }}
```
johnclary, line 131: was z-index the issue in safari/firefox? (thank you for testing those)

editor/components/FormControlAutocomplete.tsx
```diff
@@ -0,0 +1,133 @@
...
+        onFocus={() => setShowDropdown(true)}
+        onBlur={() => {
+          setShowDropdown(false);
+          field.onBlur();
+          setSearchInput(null);
+        }}
+        autoComplete="off"
+      />
+      {showDropdown && (
+        <ul
```
johnclary: ah—just realizing why this looks off to me: would you replace the `<ul>` and `<il>`  elements with `<ListGroup>` and `<ListGroup.Item>` react-bootstrap components? that will also allow you to clean up some redundant classing, and ideally make use of the `action` prop (iirc). This keeps things idiomatic to our codebase and also gives us the ARIA freebies baked into react-bootstrap.

johnclary (commented):
Nice—this is working really cleanly now 😍

editor/components/FormControlAutocomplete.tsx
chiaberry: I will update here, fwiw `<ul>` and `<li>` are in the CrashSearchTypeahead

editor/components/FormControlAutocomplete.tsx
chiaberry: the issue ended up being a span that had `position: relative`. getting rid of the span and setting that style on the parent form div ended up working. I am confused because a) it works in Mateo's CrashSearchTypeahead and b) its what my searching said to do, wrap the `position: absolute` component in a `position: relative` span, but removing it ended up working

and it was Rose who alerted me to it not working on Firefox, so thank you Rose

editor/components/FormControlAutocomplete.tsx
```diff
@@ -0,0 +1,136 @@
...
+  }, [searchInput, options]);
+
+  const handleKeyDown = useCallback(
+    (e: KeyboardEvent<HTMLInputElement>) => {
+      if (!showDropdown || !results.length) return;
+
+      if (e.key === "ArrowDown") {
+        e.preventDefault();
+        setHighlightedIndex((prev) =>
+          prev + 1 < results.length ? prev + 1 : 0
```
roseeichelmann, line 83: 💡

roseeichelmann (approved):
chia, this looks and works great! im glad you figured out the safari/firefox bug, what a mystery that its not an issue in mateos component.

this looks good to ship to me!! ✨ 🚢

editor/components/FormControlAutocomplete.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const results = useMemo(() => {
+    if (searchInput && searchInput.length > 0) {
+      return options.filter((item: LookupTableOption) =>
+        item.label.toLowerCase().includes(searchInput.toLowerCase())
+      );
+    }
+    return options;
+  }, [searchInput, options]);
+
+  const handleKeyDown = useCallback(
```
frankhereford, line 76: I think it's something that one wouldn't run into during normal use, but I did find myself having "lost track of" the active row by scrolling and then starting to use the arrow keys. I expected the UI to scroll the list of possible options so that the active row was restored to being visible as feedback for having hit the arrow keys.

That said, again, I think if people know what they are looking for, they will just type into the search box and the component does a great job of keeping an active response visible and at the top of the list. If this component is going to be used widely across the app, it might be worth looking into this very niche request, but if not, than this is probably not worth the work.

frankhereford (approved):
I think this works great Chia. I have a suggestion below that is probably best ignored, but I thought I'd include it since it did cross my mind. 

Otherwise though, this is 🚢 🚢 🚢 - thank you for this great component and integration into the VZE. Works awesome!!

editor/components/FormControlAutocomplete.tsx
chiaberry: > I expected the UI to scroll the list of possible options so that the active row was restored to being visible as feedback for having hit the arrow keys.

I am not sure I can picture what you are describing, is it that you use the mouse to scroll down the options, but the keyboard focus is still at the top?

johnclary: yes. here's a screencap of what I think Frank is referring to. ideally the options list would scroll with the focused option.

https://github.com/user-attachments/assets/7d2e56dc-7f65-4ea1-8176-5233c7559a59

i agree, but have no idea what implementation would look like. i can kick to another issue unless you want to look into it @chiaberry

chiaberry: I can look into it after I finish moped stuff!

chiaberry:
@frankhereford I pushed up some code to scroll the options into view, does this address the behavior you were describing?

frankhereford:
@chiaberry wrote:

> I pushed up some code to scroll the options into view, does this address the behavior you were describing?

It sure does - I think that works really well. I like how it keeps the current selection centered when there are enough entries before and after in the list. ✨ 🚀

editor/components/FormControlAutocomplete.tsx
```diff
@@ -0,0 +1,161 @@
...
+            <ListGroup.Item
+              key={result.id}
+              role="button"
+              ref={(element) => {
+                if (element) {
+                  itemRefs.current.set(index, element);
+                } else {
+                  itemRefs.current.delete(index);
+                }
+              }}
```
johnclary, line 148: This "callback ref" pattern is a really interesting way to manage the dynamic list item refs 💡

editor/components/FormControlAutocomplete.tsx
```diff
@@ -0,0 +1,161 @@
...
+  control,
+}: FormControlAutocomplete<TFieldValues>) {
+  const [searchInput, setSearchInput] = useState<null | string>(null);
+  const [showDropdown, setShowDropdown] = useState(false);
+  const [highlightedIndex, setHighlightedIndex] = useState(0);
+
+  // Map to store refs to scroll options into view when using keyboard navigation
+  const itemRefs = useRef(new Map());
+
+  const scrollToItem = (index: number) => {
```
johnclary: Because we're dealing with lots of re-rendering on input, would you mind moving this function outside the component?

editor/components/FormControlAutocomplete.tsx
```diff
@@ -0,0 +1,161 @@
...
+    const node = itemRefs.current.get(index);
+    if (node) {
+      node.scrollIntoView({
+        behavior: "smooth",
```
johnclary: I tested this with `"instant"` instead of `"smooth"` scroll and I think that's preferable. It makes the scroll feel very snappy. Would you make that change?

editor/components/FormControlAutocomplete.tsx
```diff
@@ -0,0 +1,161 @@
...
+        block: "center", // Centers the item vertically within the scrollable box
+      });
```
johnclary: The block centering is really nice. TIL!

johnclary (commented):
Wow, this is so fancy now with the autoscrolling!

johnclary (approved):
Thanks, Chia! 🚢 🚢 🚢 🚢

mddilley (commented):
Glad that I came back to review this one! Awesome to see all of the event handling and how it maps to the UI interactions. 🙌 I tested with mouse and keyboard and was always able to arrive at my choice. Really cool work! 🚀

PR #2048, opened 2026-05-28
Migrate from Create React App to Vite

roseeichelmann:
## Associated issues
Closes https://github.com/issues/assigned?issue=cityofaustin%7Catd-data-tech%7C27940

This PR migrates the build environment from CRA to the latest Vite v8, updates all files from.js to .jsx if they use JSX, includes ESLint and formatting w Prettier, and sets up a config for Typescript that still allows .js and .jsx as we undergo that transition.

## Testing

**URL to test:** 
Local

Can also test the [deploy preview](https://deploy-preview-2048--vzv-staging.netlify.app/) for the new Netlify project i created for vzv staging

**Steps to test:**

1. Run the following commands:
```shell
rm -rf node_modules package-lock.json
nvm use
npm install
```
2. Update your local env file with the new var name in `env_template`, make sure env file is named `.env.local`
3. Then start the local dev environment with:
```shell
npm run dev
```
5. Make sure the app is working as expected, you can compare to the [production app](https://visionzero.austin.gov/viewer/map) if you want
6. Copy pastaing next steps from the Moped vite PR to test the Typescript integration:
7. Create test.ts in /src/ folder
8. Add `const x: number = "hello";` to the file, and you should see a TypeScript error highlighted
9. Now, update `test.ts` to be named `test.tsx` and paste the code below:
```js
import App from "src/App";

import { useEffect } from "react";

export const MyCoolApp = () => {
  const x: number = "hello";

  if (true) {
    useEffect(() => {
      console.log("do something");
    }, []);
  }

  return <App />;
};
```
10. You should see that:
   - Relative `src/` imports are working
   - the same TS error is working in a TSX file
   - the conditional useEffect is showing an error that it violates the React rules of hooks

**To do after this PR:**
- Address linting errors
- Run `npm run format` to have prettier formatting run on all files
- Spin down the old netlify deploys

---

#### Ship list

- [ ] Code reviewed


viewer/src/views/nav/Header.jsx
```diff
@@ -126,7 +126,7 @@ const Header = () => {
               <img
                 className="vz-logo"
                 // Need to adjust location of public folder to account for /viewer/ basepath
-                src={process.env.PUBLIC_URL + "/vz_logo.png"}
+                src={import.meta.env.BASE_URL + "vz_logo.png"}
```
roseeichelmann, line 129: `PUBLIC_URL` is now defined as `BASE_URL` in Vite, and the `BASE_URL` already is composed of "/" instead of empty string, so the "/" before `vz_logo` needs to be removed

and `process.env` has been updated to `import.meta.env`

viewer/.prettierrc
```diff
@@ -1,6 +1,10 @@
 {
-  "trailingComma": "es5",
-  "tabWidth": 2,
   "semi": true,
-  "singleQuote": false
+  "singleQuote": false,
+  "tabWidth": 2,
+  "trailingComma": "es5",
+  "printWidth": 100,
```
roseeichelmann: do yall have preferences on printWidth? a lot of the formatting changes that happened in this PR are from increasing the printWidth here, but i could lower it back down to, say, 75?

though i would prefer to mess with this more in a follow up PR that both addresses linting errors and formatting differences in all files, so that this PR doesnt get bogged up by changes

viewer/src/views/summary/Summary.jsx
```diff
@@ -49,25 +49,6 @@ const Summary = () => {
 
   const lastUpdated = format(dataEndDate, "MMMM dd, yyyy");
 
-  // Display disclaimer when viewing preview instance
-  const previewDisclaimer = process.env.REACT_APP_VZV_ENVIRONMENT ===
```
roseeichelmann, line 53: i deleted this and didnt create a new vite version of this env var because i cant think of a time we've last used this

viewer/eslint.config.mjs
```diff
@@ -0,0 +1,76 @@
...
+import globals from "globals";
+import reactPlugin from "eslint-plugin-react";
+import reactHooks from "eslint-plugin-react-hooks";
+import reactRefresh from "eslint-plugin-react-refresh";
+import tseslint from "typescript-eslint";
+import prettierConfig from "eslint-config-prettier";
+
+/**
+ * This file's extension is updated to .mjs to workaround ESLint's MODULE_TYPELESS_PACKAGE_JSON warning.
+ * We could use a package.json with "type": "module" (https://nodejs.org/api/packages.html#type) but this
```
roseeichelmann, line 12: had to do the same workaround as moped here, so i copy pasta'd this comment 😄

viewer/tsconfig.json
roseeichelmann, line 1: copied this from mike works in moped, its a good base to allow us to start migrating to TS without requiring it

viewer/vite.config.js
```diff
@@ -0,0 +1,29 @@
...
+      svgr({ svgrOptions: { icon: true } }),
+    ],
+    define: {
+      // Ensures the global variable is defined for libraries that expect it
+      global: "globalThis",
+    },
+    resolve: {
+      alias: {
+        // Provides the EventEmitter polyfill for mapbox-gl-geocoder and similar libraries
+        events: "events",
```
roseeichelmann, line 27: this doesnt work out of the box in Vite so have to manually add this for our mapbox stuffs

viewer/src/constants/time.js
```diff
@@ -5,9 +5,7 @@ import moment from "moment";
 // Number of past years data to fetch
 export const ROLLING_YEARS_OF_DATA = 4;
 // Number of months window slides in the past (to display most accurate data)
-// Accommodate for a preview instance that provide and extra month of data
-export const MONTHS_AGO =
-  process.env.REACT_APP_VZV_ENVIRONMENT === "PREVIEW" ? 1 : 2;
```
roseeichelmann, line 10: i deleted this because i dont think we ever use it but i can bring it back if theres a use case?

roseeichelmann:
I also created two new Netlify projects called vzv-staging and vzv-prod with the new build commands and env vars, etc. Are there any steps to follow up with in AWS/cloudfront now? @frankhereford

viewer/src/views/summary/Summary.jsx
chiaberry: this makes sense to me, but I think we'd want to bring back some kind of preview banner on netlify preview deploys just to make it really clear its not using production data

viewer/.prettierrc
chiaberry: i dont have a strong preference but i appreciate you holding off on linting everything in a follow up PR

chiaberry (approved):
I spun this up locally and clicked around comparing it to production and it all works this same. 

Excited that this is opening the door for us to work on VZV!!

viewer/src/constants/time.js
johnclary: Removing it is good 🙏

viewer/src/views/summary/Summary.jsx
johnclary: Hmm, yeah—i'll open a backlog issue to add a VZE-style env badge in the nav

viewer/.prettierrc
johnclary: i am kinda inclined to stick with not overriding the default (80?) since (I think) that would be consistent with our other apps. do these other formatting changes take inspiration from somewhere else?

viewer/_redirects
```diff
@@ -1,6 +0,0 @@
-#
-# React
-#
-/static/*       /static/:splat          200
-/public/*       /public/:splat          200
-/*              /index.html             200
```
johnclary, line 6: Not sure if this is related, but the routing in the netlify preview is not resolving after going to `/map` and refreshing the page:
https://deploy-preview-2048--vzv-staging.netlify.app/viewer/map

Image

johnclary: I'm also getting a little confused about the `/viewer` prefix—it's only included in the deploy preview route after clicking on the Map/Summary nav buttons.

viewer/package.json
```diff
@@ -96,6 +99,18 @@
...
+    "eslint": "^9.39.4",
+    "eslint-config-prettier": "^10.1.8",
+    "eslint-plugin-prettier": "^5.5.6",
+    "eslint-plugin-react": "^7.37.5",
+    "eslint-plugin-react-hooks": "^7.1.1",
+    "eslint-plugin-react-refresh": "^0.5.2",
+    "globals": "^17.6.0",
+    "prettier": "^3.8.3",
+    "react-axe": "^3.5.4",
+    "typescript-eslint": "^8.60.1",
```
mddilley, line 88: **Edit:** Now realizing that this will probably be a non-issue by the time this merges. Glad to see our cooldown is working though!

Chia helped me catch this, but I noticed that, when I try to install this `typescript-eslint` version, it is too new for our `.npmrc` `min-release-age` so i can't install it.

Might be worth trying 8.60.0 depending on when that released! we have `"typescript-eslint": "^8.59.0",` in Moped which I only landed on because the Moped Vite work was done before this newer version existed.

johnclary (commented):
Wowwww no more CRA! I was able to spin up locally and run all the tests easily. We have TS in VZV!

My questions are related to routing and netlify. I'm curious what you think moving to staging and prod deployments will look like as well. I'm quite fuzzy on the routing there.

viewer/package.json
mddilley: I'm pretty sure that this is leftover CRA ESLint config that can be removed since it is replaced by `eslint.config.mjs` but you might want to double-check me

mddilley: I removed this browserslist block from Moped since it tells CRA about what browsers to support. Since VZV is public-facing, I think that there could be a need to define this in vite.config.js using build.target (https://vite.dev/config/build-options#build-target)

cc @johnclary

viewer/package.json
```diff
@@ -27,6 +27,7 @@
     "chartjs-plugin-stacked100": "^0.7.1",
     "chroma-js": "2.4.2",
     "date-fns": "^2.30.0",
+    "events": "^3.3.0",
     "events-polyfill": "^2.1.2",
```
mddilley, line 31: **Edit: I think that I pieced it together in a later comment**

Did the `events` package fix anything specific? I'm wondering because I suspect there could be a way to handle in the Vite config but only because that worked for some things in Moped. This could be completely different and mostly want to learn more! It has been fun to see how the different codebases require different things.

viewer/package.json
mddilley: I also think `react-app-polyfill` is a CRA specific polyfill, but I also feel like, if removing it breaks something, it is probably worth keeping for now to avoid chasing down what the polyfill is doing on top of the rest of this work.

viewer/package.json
```diff
@@ -42,28 +43,30 @@
...
     "react-map-gl-geocoder": "^2.2.0",
-    "react-scripts": "5.0.1",
     "react-use": "^14.3.0",
     "react-with-styles-interface-aphrodite": "^6.0.1",
     "reactstrap": "^8.10.1",
     "reaviz": "^8.3.0",
     "serialize-javascript": "^7.0.5",
     "styled-components": "^4.4.1",
     "url": "^0.11.4",
+    "vite-plugin-svgr": "^5.2.0",
```
mddilley: I _think_ that this could be installed as a dev dependency only since I saw its install command was `npm install --save-dev vite-plugin-svgr` in the docs. I suspect because it does the svg transforms in the build and local dev steps only and isn't needed by the deployed app.

viewer/vite.config.js
mddilley: ok now i see where `events` plugs in which makes my earlier comment about this package irrelevant. 👍

viewer/vite.config.js
```diff
@@ -0,0 +1,30 @@
+import { defineConfig } from "vite";
+import react from "@vitejs/plugin-react";
+import svgr from "vite-plugin-svgr";
+
+export default defineConfig(() => {
+  return {
```
mddilley, line 6: @roseeichelmann I think that you may need to add the `base` key to this config since this app uses `/viewer/` in the routing. I had to add `base: "/moped/"` in Moped since it always includes `/moped/` in the route ([code here](https://github.com/cityofaustin/atd-moped/blob/b17c9caf14e8051c3bb7dc5bb68c4f28fadf03c7/moped-editor/vite.config.js#L9)).

mddilley (commented):
Hooray!! Awesome to see this in action!! 😎 The build is lightning fast. ⚡  I added some comments about some small things that I think can be removed or changed and one question about browser support config.

viewer/package.json
johnclary: ooo neat—i'll open a follow-up for this. @roseeichelmann you can ditch the `browserslist` entry here though

viewer/.prettierrc
roseeichelmann: i can set it to 80, the reason why im setting it at all here is bc it will make sure we all are using the same printWidth, anything we dont set here is up to the individual devs settings in their code editor. which is probably set default for most people, but if anyone has adjusted their own settings then this wall make sure we are all consistent. ill update it to 80

i got the other settings just from looking at my default settings and looking at recommended examples online

viewer/_redirects
roseeichelmann: phew that was hard to figure out but i got it -- you were right, deleting the _redirects caused this. i instead added a netlify.toml file like we do in [moped](https://github.com/cityofaustin/atd-moped/blob/b17c9caf14e8051c3bb7dc5bb68c4f28fadf03c7/moped-editor/netlify.toml)  and had to fiddle a lot with the settings in the file to finally get it to work, but the redirects and /viewer prefix are working as expected now

viewer/package.json
roseeichelmann: I removed it and it seems to not have broken anything, it was definitely some old cruft ty!

viewer/package.json
roseeichelmann: done!

roseeichelmann:
@mddilley @johnclary this is ready for re-review, thanks for the thorough testing yall!

viewer/.prettierrc
johnclary: oh i see—smart 🙏 I don't have a strongly held opinion about line length other than wanting to be consistent at least across the VZE as well.

viewer/netlify.toml
johnclary, line 1: 🚀

johnclary (approved):
The routing issues are all fixed. This looks ready to me!!

I will open the handful of follow-up issues we discussed momentarily. And @roseeichelmann lets work together to build of the Netlify and AWS todos for staging and ultimately the prod launch.

viewer/src/views/summary/Summary.jsx
johnclary: * https://github.com/cityofaustin/atd-data-tech/issues/28915

viewer/package.json
johnclary: I opened https://github.com/cityofaustin/atd-data-tech/issues/28916 and reached out to the digital accessibility office in case they can help.

mddilley (approved):
Looks great! Goodbye CRA and hello to the future! Thanks for taking this on, and it was fun to learn even more about Vite config!! 🙏🚢 🙌

mddilley:
@roseeichelmann @johnclary I'm not sure if this is my machine only, but I couldn't get the Mapbox tiles to load on my local dev server. I switched out with the VZE token, and it worked even though the two tokens have 100% the same settings when I looked in the Mapbox dashboard. Hoping this is just me but wanted to flag!

johnclary:
> I couldn't get the Mapbox tiles to load on my

Are you seeing an network errors along with that, Mike? Local is working normally for me 🤔

mateoclarke (approved):
Tested locally and in the staging branch. Great work getting this switched over! I'm stoked for more Vite! 

One question I have is should we update the 1PW file "Vision Zero Viewer (VZV) Environment File"? It still references `REACT_APP_MAPBOX_TOKEN` but maybe should be `VITE_MAPBOX_TOKEN`. I think the token is valid though. 

✨ 🚀

mddilley:
@johnclary i was seeing 403s but i just retested and it is working fine so it was my machine after all. 🙃  Sorry for the false alarm!

roseeichelmann:
@mateoclarke good idea! i just updated it :pray:

PR #2055, opened 2026-06-08
Optimize fatality details page for slide deck / screencap

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/28459

frb-layout-1

## Testing

**URL to test:** [Netlify](https://deploy-preview-2055--vze-staging.netlify.app/)

1. From the fatalities list, view the details for a temporary crash.

2. Confirm that you are now able to dismiss the temp crash banner.

3. Observe the changes. It will be helpful to look at prod side-by-side.

* Increased font size
* Updated the top row (map + units cards) column ratio to 5:7
* Update middle row (diagram + narrative + details cards) column ratio to 6:6 with narrative + details now stacked
Increase font size

4. I left the [dark mode theme refactor](https://austininnovation.slack.com/archives/CM9MK950S/p1781015424624429) out of scope

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/app/fatalities/[record_locator]/page.tsx
+++ b/editor/app/fatalities/[record_locator]/page.tsx
@@ -189,7 +189,7 @@ export default function FatalCrashDetailsPage({
           </Row>
         </Col>
       </Row>
-      <Row>
+      <Row style={{ fontSize: "18px" }}>
         <Col className="mb-3" sm={12} md={12} lg={6}>
           <CrashRecommendationCard
             crash_pk={crash.id}
--- a/editor/app/fatalities/[record_locator]/page.tsx
+++ b/editor/app/fatalities/[record_locator]/page.tsx
@@ -100,8 +100,8 @@ export default function FatalCrashDetailsPage({
           <CrashIsTemporaryBanner crash={crash} allowDelete={false} />
         )
       }
-      <Row>
-        <Col className="mb-3" sm={12} md={6}>
+      <Row style={{ fontSize: "18px" }}>
+        <Col className="mb-3" sm={12} md={5}>
           <Card className="h-100">
             <Card.Body>
               <Table>
@@ -156,31 +156,37 @@ export default function FatalCrashDetailsPage({
             </Card.Body>
           </Card>
         </Col>
-        <Col className="mb-3" sm={12} md={6}>
+        <Col className="mb-3" sm={12} md={7}>
           <FatalityUnitsCards crash={crash} onSaveCallback={onSaveCallback} />
         </Col>
       </Row>
-      <Row>
-        <Col className="mb-3" sm={12} md={6} lg={4}>
+      <Row style={{ fontSize: "18px" }}>
+        <Col className="mb-3" sm={12} md={6} style={{ minHeight: "625px" }}>
           <CrashDiagramCard crash={crash} crashRefetch={refetch} />
         </Col>
-        <Col className="mb-3" sm={12} md={6} lg={4}>
-          <CrashNarrativeEditableCard
-            crash={crash}
-            onSaveCallback={onSaveCallback}
-          />
-        </Col>
-        <Col className="mb-3">
-          <DataCard<Crash>
-            record={crash}
-            isValidating={false}
-            title="Details"
-            columns={otherCardColumns}
-            mutation={UPDATE_CRASH}
-            onSaveCallback={onSaveCallback}
-            shouldShowColumnVisibilityPicker={true}
-            localStorageKey="crashPageOther"
-          />
+        <Col className="mb-3" sm={12} md={6}>
+          <Row>
+            <Col className="mb-3">
+              <CrashNarrativeEditableCard
+                crash={crash}
+                onSaveCallback={onSaveCallback}
+              />
+            </Col>
+          </Row>
+          <Row>
+            <Col>
+              <DataCard<Crash>
+                record={crash}
+                isValidating={false}
+                title="Details"
+                columns={otherCardColumns}
+                mutation={UPDATE_CRASH}
+                onSaveCallback={onSaveCallback}
+                shouldShowColumnVisibilityPicker={true}
+                localStorageKey="crashPageOther"
+              />
+            </Col>
+          </Row>
         </Col>
       </Row>
       <Row>
--- a/editor/app/fatalities/[record_locator]/page.tsx
+++ b/editor/app/fatalities/[record_locator]/page.tsx
@@ -100,7 +100,7 @@ export default function FatalCrashDetailsPage({
           <CrashIsTemporaryBanner crash={crash} allowDelete={false} />
         )
       }
-      <Row style={{ fontSize: "16px" }}>
+      <Row className="fatality-details-row">
         <Col className="mb-3" sm={12} md={5}>
           <Card className="h-100">
             <Card.Body>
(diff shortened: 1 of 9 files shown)


editor/components/CrashIsTemporaryBanner.tsx
```diff
@@ -28,6 +28,7 @@ export default function CrashIsTemporaryBanner({
   return (
     <>
       <Alert
+        dismissible
```
johnclary: I think this may be handy in case Mary wants to exclude the banner from screenshots, thereby increasing screen real estate.

editor/utils/images.ts
```diff
@@ -76,7 +76,11 @@ export function useImage({
         setImageUrl(data.url);
         setIsLoading(false);
       } catch (err) {
-        console.error(`Error fetching ${recordType} image:`, err);
+        if (`${err}`.includes("404")) {
+          console.warn(err);
+        } else {
+          console.error(`Error fetching ${recordType} image:`, err);
+        }
```
johnclary, line 83: This is muting the annoying Nextjs error popup that happens locally. I wish the image API returned `{url: null}` instead of 404ing on `GET` requests 😞

editor/components/CrashIsTemporaryBanner.tsx
chiaberry: does she also screenshot the Crash Details page? I guess it doesnt hurt having it dismissible there as well...

editor/app/fatalities/[record_locator]/page.tsx
```diff
@@ -156,34 +156,40 @@ export default function FatalCrashDetailsPage({
...
           </Card>
         </Col>
-        <Col className="mb-3" sm={12} md={6}>
+        <Col className="mb-3" sm={12} md={7}>
           <FatalityUnitsCards crash={crash} onSaveCallback={onSaveCallback} />
         </Col>
       </Row>
-      <Row>
-        <Col className="mb-3" sm={12} md={6} lg={4}>
+      <Row style={{ fontSize: "18px" }}>
```
Comment (chiaberry) [inline editor/app/fatalities/[record_locator]/page.tsx]: The increased font feels jarring, navigating from the rest of the app to the fatalities page.

editor/app/fatalities/[record_locator]/page.tsx
```diff
@@ -156,34 +156,40 @@ export default function FatalCrashDetailsPage({
...
-            isValidating={false}
-            title="Details"
-            columns={otherCardColumns}
-            mutation={UPDATE_CRASH}
-            onSaveCallback={onSaveCallback}
-            shouldShowColumnVisibilityPicker={true}
-            localStorageKey="crashPageOther"
-          />
+        <Col className="mb-3" sm={12} md={6}>
+          <Row>
```
Comment (chiaberry) [inline editor/app/fatalities/[record_locator]/page.tsx]: what do you think about putting a min Height on this row, so even if the Narrative is empty, its got a little bit of height to the card?

editor/app/fatalities/[record_locator]/page.tsx
Comment (johnclary) [reply editor/app/fatalities/[record_locator]/page.tsx]: I'm open to ideas! We could add a transition so that the text slowly increases in size for ~30 seconds.

Comment (chiaberry) [reply editor/app/fatalities/[record_locator]/page.tsx]: I tried 16px and it didnt look as jarring to me. But what about the option to have a button that makes the font larger for screenshot purposes, so that way day to day usage of the VZE has consistent font sizes across the board. I am assuming the screenshots only happen right before powerpoint presentation creation?

Comment (roseeichelmann) [reply editor/app/fatalities/[record_locator]/page.tsx]: oh i was gonna say i love the increased font size and its gonna help mary so much with her screenshots bc as is the text was teensy tiny. i didnt really find it jarring when navigating around the app. future issue to have a button for it could be a cool idea though, but for now i think increasing the text size should be priority

roseeichelmann (approved):
i think this is a big improvement and will really help mary with visibility of everything when shes presenting 🙏

chiaberry (approved):
I am not the biggest fan of the font changes, but I also recognize that I am missing context that y'all have from your meetings with Mary. ✅

Charlie-Henry (approved):
I'm sure y'all considered it but maybe a way to "print" a fatality page might serve this purpose better than screenshotting? Looks good to me though

editor/app/fatalities/[record_locator]/page.tsx
Comment (johnclary) [reply editor/app/fatalities/[record_locator]/page.tsx]: FWIW, Chia, I'm not a fan of this solution either. It is a compromise I can live with because the page is purpose-built for the FRB and the alternatives are either also flawed or a heavy lift. 16px is the app-wide default text size, btw.

johnclary:
>  maybe a way to "print" a fatality page 

Yeah! You can set up CSS media queries specific to printing, and I think this is clearly The Way to handle this user need. I spent a decent amount of time trying to get that set up, but the CSS got really complex really quickly so I decided to abandon it.

editor/app/fatalities/[record_locator]/page.tsx
Comment (chiaberry) [reply editor/app/fatalities/[record_locator]/page.tsx]: hm interesting, when I inspected the text size on the crash details page for example, the inspector said the text was 14px in most places

editor/app/fatalities/[record_locator]/page.tsx
Comment (johnclary) [reply editor/app/fatalities/[record_locator]/page.tsx]: I added some flex classes so that narrative card is always going to at least fill it's half column and make for a consistent layout 🙏

johnclary:
OK, i dropped to 16px down from 18px and added flex utilities on the narrative card to keep its height from collapsing. i think this is in a good place!

Screenshot 2026-06-11 at 11 22 10 AM

chiaberry:
agreed!! i already approved but heres another little boat 🚢

editor/app/fatalities/[record_locator]/page.tsx
```diff
@@ -100,8 +100,8 @@ export default function FatalCrashDetailsPage({
           <CrashIsTemporaryBanner crash={crash} allowDelete={false} />
         )
       }
-      <Row>
-        <Col className="mb-3" sm={12} md={6}>
+      <Row style={{ fontSize: "16px" }}>
```
Comment (mateoclarke) [inline editor/app/fatalities/[record_locator]/page.tsx]: Since we're using the font sizing in multiple places (I count 3), I'd advocate for a single class to make this more maintainable, something like `.fatality-details-row`.

Alternatively, you could start to lean into react-bootstrap variant and size theming props like these examples: https://react-bootstrap.netlify.app/docs/getting-started/theming/

mateoclarke (approved):
Loosely help opinion about font size maintainability that would be a cheap fix, otherwise looks good! 

Going to point out two more minor visual things, one is out of scope but I noticed it in prod when comparing the differences.

1. dismiss `X` looks weird, not centered vertically when the delete button is on the crash details page.

Image

2. Why is the "Location" card header smaller than all the other headers? (out of scope of these changes, but indicator of some brittleness in our styling)

Image

editor/app/fatalities/[record_locator]/page.tsx
Comment (johnclary) [reply editor/app/fatalities/[record_locator]/page.tsx]: TIL about the special react-bootstrap theming 🙏  I went ahead and added a CSS class for this one

johnclary:
@mateoclarke 

1. I went ahead and made the dismissible prop specific to fatality details page, thereby bunting on trying to override styles of that banner close button
2. Intentional design decision the location card title doing double duty as the label for the hyperlinked ID and saving vertical space. Agree it could be cleaned up. It also irks me that the **Diagram** header is extra tall because it holds the keyboard shortcut

PR #2057, opened 2026-06-10
Add dry run flag to incident linker

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/27761

## Testing

**URL to test:** Locally

1. If you haven't tested any of the CAD ETLs before, follow the `./etl/cad_incidents` readme to get set up for local dev. 
 
2. Apply migrations and metadata.

3. Run the ETL using the `--dry-run` flag.

```
docker compose -f docker-compose.yml -f docker-compose.local.yml run import incident_linker.py --dry-run
```

Your script output should look like this:

```
INFO:root:Fetching unprocessed incidents that occurred before 2026-06-09 16:11:20.836905+00:00...
INFO:root:  Found 5,000 unprocessed incidents
```

Note that default limit of 5k records to process is taking effect, and that no records are actually processed ✅ 

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/etl/cad_incidents_import/incident_linker.py
+++ b/etl/cad_incidents_import/incident_linker.py
@@ -85,6 +85,7 @@ def main(args):
     logging.info(f"  Found {len(incidents):,} unprocessed incidents\n")
 
     if args.dry_run:
+        logging.info(f"Dry run: aborting further processing")
         return
 
     processed_ids = set()
--- a/etl/cad_incidents_import/incident_linker.py
+++ b/etl/cad_incidents_import/incident_linker.py
@@ -15,7 +15,7 @@ from utils.queries import (
 MIN_RECORD_AGE_HOURS = 24
 DISTANCE_THRESHOLD_M = 500
 TIME_THRESHOLD_MINUTES = 60
-MAX_RECORD_TO_PROCESS = 1000
+MAX_RECORD_TO_PROCESS = 5000
 
 
 def point_geojson(lon, lat):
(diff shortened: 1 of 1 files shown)


etl/cad_incidents_import/incident_linker.py
```diff
@@ -15,7 +15,7 @@
 MIN_RECORD_AGE_HOURS = 24
 DISTANCE_THRESHOLD_M = 500
 TIME_THRESHOLD_MINUTES = 60
-MAX_RECORD_TO_PROCESS = 1000
+MAX_RECORD_TO_PROCESS = 5000
```
johnclary, line 18: I increased this to make sure we have head-room to process a backlog of a week or so's worth of records.

mddilley (approved):
Tested with the new dry run flag, and I see that no incidents were processed. 🚢 🙏

frankhereford (approved):
This passes the eyeball test with the new argument and the straight forward action taken when it's true. Thanks for adding this John.

etl/cad_incidents_import/incident_linker.py
```diff
@@ -84,6 +84,9 @@ def main(args):
     incidents = data["cad_incidents"]
     logging.info(f"  Found {len(incidents):,} unprocessed incidents\n")
 
+    if args.dry_run:
+        return
```
chiaberry, line 89: What do you think about logging before the return something about how this a dry run?

chiaberry (approved):
Dry run flag is a success! I put a comment about logging, it would be nice to have some closure.

johnclary:
Y'all are the best—thanks for the quick review 🙌

PR #2062, opened 2026-06-15
Address eslint errors

roseeichelmann:
## Associated issues
Closes https://github.com/issues/assigned?issue=cityofaustin%7Catd-data-tech%7C28868

Fixes to ESLint errors using the auto `npm run lint --fix` command but most are manual fixes

## Testing

**URL to test:** 
Local

**Steps to test:**
1. Start up your local dev environment. From the `/viewer` folder run `nvm use` and `npm run lint`. You should get 0 errors :partying_face: 
2. Search the editor for `eslint-disable` and see I've ignored two errors to be addressed in a follow up issue -- they required a bit of a refactor that was starting to feel like a lot to cram into one PR and wasnt as easy as the others to wrap my head around
3. Test around the app & compared to production to make sure everything is working as expected. Test the different map filters including date, time, injury type. Test all the diff filters on the summary page cards as well.

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/viewer/netlify.toml
+++ b/viewer/netlify.toml
@@ -1,5 +1,11 @@
+# Default used by main/deploy previews
 [build]
   publish = "build"
+  command = "npm run lint && npm run build:staging"
+
+# Production specific build command
+[context.production]
+  command = "npm run lint && npm run build:production"
 
 # This rule handles the root path and redirects to /viewer/
 [[redirects]]
--- a/viewer/src/Components/Widgets/SummaryWidget.jsx
+++ b/viewer/src/Components/Widgets/SummaryWidget.jsx
@@ -86,7 +86,7 @@ const SummaryWidget = ({
         <Row>
           <Col>
             {/* Show spinner while waiting for data, add thousands separator to total */}
-            {!!totalsObject ? (
+            {totalsObject ? (
               <h2 className="h1 total">
                 {numberWithCommas(totalsObject[currentYear])}
               </h2>
--- a/viewer/src/constants/crashTypes.js
+++ b/viewer/src/constants/crashTypes.js
@@ -1,0 +1,20 @@
+export const CRASH_TYPES = {
+  fatalitiesAndSeriousInjuries: {
+    name: "fatalitiesAndSeriousInjuries",
+    textString: "Fatalities and Serious Injuries",
+    queryStringCrash: "(death_cnt > 0 OR sus_serious_injry_cnt > 0)",
+    queryStringPerson: "(prsn_injry_sev_id = 4 OR prsn_injry_sev_id = 1)",
+  },
+  fatalities: {
+    name: "fatalities",
+    textString: "Fatalities",
+    queryStringCrash: "(death_cnt > 0)",
+    queryStringPerson: "(prsn_injry_sev_id = 4)",
+  },
+  seriousInjuries: {
+    name: "seriousInjuries",
+    textString: "Serious Injuries",
+    queryStringCrash: "(sus_serious_injry_cnt > 0)",
+    queryStringPerson: "(prsn_injry_sev_id = 1)",
+  },
+};
--- a/viewer/src/constants/map.js
+++ b/viewer/src/constants/map.js
@@ -1,3 +1,14 @@
+// Build query string for crash type filter
+export const createModeFilterString = (isMapTypeSet, config) => {
+  if (isMapTypeSet.fatal && isMapTypeSet.injury) {
+    return `${config.fatalSyntax} ${config.operator} ${config.injurySyntax}`;
+  } else if (isMapTypeSet.fatal) {
+    return `${config.fatalSyntax}`;
+  } else if (isMapTypeSet.injury) {
+    return `${config.injurySyntax}`;
+  }
+};
+
 export const mapFilterReducer = (mapFilters, action) => {
   const { type, payload } = action;
 
--- a/viewer/src/constants/map.js
+++ b/viewer/src/constants/map.js
@@ -1,10 +1,21 @@
+// Build query string for crash type filter
+export const createModeFilterString = (isMapTypeSet, config) => {
+  if (isMapTypeSet.fatal && isMapTypeSet.injury) {
+    return `${config.fatalSyntax} ${config.operator} ${config.injurySyntax}`;
+  } else if (isMapTypeSet.fatal) {
+    return `${config.fatalSyntax}`;
+  } else if (isMapTypeSet.injury) {
+    return `${config.injurySyntax}`;
+  }
+};
+
 export const mapFilterReducer = (mapFilters, action) => {
   const { type, payload } = action;
 
   switch (type) {
     case "setInitialModeFilters":
       const initialFiltersArray = payload;
-      return initialFiltersArray;
+      return payload;
     case "updateModeSyntax":
       const isMapTypeSet = payload;
 
--- a/viewer/src/constants/map.js
+++ b/viewer/src/constants/map.js
@@ -13,10 +13,11 @@ export const mapFilterReducer = (mapFilters, action) => {
   const { type, payload } = action;
 
   switch (type) {
-    case "setInitialModeFilters":
+    case "setInitialModeFilters": {
       const initialFiltersArray = payload;
-      return payload;
-    case "updateModeSyntax":
+      return initialFiltersArray;
+    }
+    case "updateModeSyntax": {
       const isMapTypeSet = payload;
 
       const updatedModeFilters = mapFilters.map((filter) => ({
@@ -24,9 +25,11 @@ export const mapFilterReducer = (mapFilters, action) => {
         syntax: createModeFilterString(isMapTypeSet, filter),
       }));
       return updatedModeFilters;
-    case "updateModeFilters":
+    }
+    case "updateModeFilters": {
       const updatedFiltersArray = payload;
       return updatedFiltersArray;
+    }
     default:
       return null;
   }
--- a/viewer/src/constants/map.js
+++ b/viewer/src/constants/map.js
@@ -1,0 +1,22 @@
+export const mapFilterReducer = (mapFilters, action) => {
+  const { type, payload } = action;
+
+  switch (type) {
+    case "setInitialModeFilters":
+      const initialFiltersArray = payload;
+      return initialFiltersArray;
+    case "updateModeSyntax":
+      const isMapTypeSet = payload;
+
+      const updatedModeFilters = mapFilters.map((filter) => ({
+        ...filter,
+        syntax: createModeFilterString(isMapTypeSet, filter),
+      }));
+      return updatedModeFilters;
+    case "updateModeFilters":
+      const updatedFiltersArray = payload;
+      return updatedFiltersArray;
+    default:
+      return null;
+  }
+};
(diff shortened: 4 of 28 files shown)


viewer/src/Components/Widgets/SummaryWidget.jsx
```diff
@@ -86,7 +86,7 @@ const SummaryWidget = ({
         <Row>
           <Col>
             {/* Show spinner while waiting for data, add thousands separator to total */}
-            {!!totalsObject ? (
+            {totalsObject ? (
```
roseeichelmann, line 89: fixes https://eslint.org/docs/latest/rules/no-extra-boolean-cast

viewer/src/constants/crashTypes.js
roseeichelmann, line 1: Fixes `react-refresh/only-export-components` error which flags breaking of [hot module replacement](https://vite.dev/guide/features#hot-module-replacement). The fix is to create a separate file for code that needs to be exported instead of keeping them in the same file with exported components. Had to do this a couple times

viewer/src/constants/map.js
roseeichelmann, line 1: More fixes to `react-refresh/only-export-components`

viewer/src/utils/store.jsx
```diff
@@ -26,7 +25,10 @@ export default function StoreProvider({ children }) {
   // SideDrawer should never be open when not mobile
   const isTablet = useIsTablet();
   useEffect(() => {
-    !isTablet && setIsOpen(false);
+    if (!isTablet) {
+      // eslint-disable-next-line react-hooks/set-state-in-effect -- fix in issue #29044
```
roseeichelmann, line 29: i was having a hard time figuring out how to do this differently so i kicked the can to a sep issue

viewer/src/views/map/InfoBox/MapPolygonInfoBox.jsx
```diff
@@ -5,16 +5,19 @@ import { StyledPoylgonInfo } from "./infoBoxStyles";
 const MapPolygonInfoBox = ({ crashCounts, isMapTypeSet }) => {
   const createCrashContent = (crashCounts) => {
     const content = [];
-    isMapTypeSet.fatal &&
+    if (isMapTypeSet.fatal) {
```
roseeichelmann, line 8: out of the box eslint enforces the use of if statements instead of && 🤷 

i know we use this pattern a lot and are used to it so what are yalls thoughts?

viewer/src/views/map/Map.jsx
```diff
@@ -153,21 +158,25 @@ const Map = () => {
     isDragging ? "grab" : isHovering ? "pointer" : "default";
 
   // Set interactive layer IDs to allow cursor to change if isHovering
-  useEffect(() => {
-    const interactiveLayerIds = [
+  const interactiveLayerIds = useMemo(() => {
```
roseeichelmann, line 160: this is an example of how i fixed some of the `Error: Calling setState synchronously within an effect can trigger cascading renders.` errors. In this case and many of the others i was able to use `useMemo` instead and get rid of the state

viewer/src/views/nav/SideMapControl.jsx
```diff
@@ -97,29 +96,6 @@ const createModeFilterString = (isMapTypeSet, config) => {
   }
 };
 
-export const mapFilterReducer = (mapFilters, action) => {
```
roseeichelmann, line 100: this was moved into its own export file

viewer/src/views/nav/SideMapTimeOfDayChart.jsx
```diff
@@ -33,64 +31,68 @@ export const SideMapTimeOfDayChart = ({ filters }) => {
...
       axios.get(apiUrl).then((res) => {
         setChartData(res.data);
       });
+    }
   }, [dateRange, mapPolygon, mapFilters]);
 
-  useMemo(() => {
+  // Compute time window data from crash data
+  // Returns array of crash counts per time window
+  const timeWindowData = useMemo(() => {
```
roseeichelmann, line 45: i removed unnecessary setState calls in the following two useMemo hooks to fix `Error: Calling setState from useMemo may trigger an infinite loop`

viewer/src/views/summary/CrashesByYearAverage.jsx
```diff
@@ -1,16 +1,14 @@
...
 import { format, setMonth } from "date-fns";
 import { Bar } from "react-chartjs-2";
 
 import { colors } from "../../constants/colors";
 
 const CrashesByYearAverage = ({ avgData, currentYearData }) => {
-  const [chartData, setChartData] = useState({});
-
-  useEffect(() => {
+  const chartData = useMemo(() => {
```
roseeichelmann, line 8: was able to just memoize this and remove state

viewer/src/views/summary/CrashesByYearCumulative.jsx
```diff
@@ -1,24 +1,25 @@
...
+import React, { useMemo } from "react";
 import { format, setMonth } from "date-fns";
 import { Line } from "react-chartjs-2";
 import { colors } from "../../constants/colors";
 
 const CrashesByYearCumulative = ({ avgData, currentYearData }) => {
-  const [chartData, setChartData] = useState({});
-
-  useEffect(() => {
+  const chartData = useMemo(() => {
```
roseeichelmann, line 7: saame here, memoizing and removing state

viewer/src/views/summary/PeopleByDemographics.jsx
```diff
@@ -254,27 +255,37 @@ const PeopleByDemographics = () => {
...
   }, [chartType, demoData, crashType]);
 
-  // On crash or chart type change, replace chart data and colors from config
-  useEffect(() => {
-    if (demoData.status === "formatted") {
-      const [configName, config] = Object.entries(chartConfigs).find(
-        ([configName, config]) => config.label === chartType
-      );
+  // Derive chart data
+  const chartData = useMemo(() => {
```
roseeichelmann, line 259: once again, memoizing instead of using state

viewer/src/views/map/map-style.jsx
```diff
@@ -239,7 +239,7 @@ export const cityCouncilDataLayer = {
 };
 
 // Styles for MapPolygonFilter
-export function getEditHandleStyle({ feature, state }) {
+export function getEditHandleStyle({ state }) {
```
roseeichelmann, line 242: fixing unused vars error

viewer/src/views/map/InfoBox/MapPolygonInfoBox.jsx
chiaberry: I guess this is a side effect of them not liking `!!` boolean casting

I dont have a preference one way or another...

viewer/src/views/map/Map.jsx
```diff
@@ -99,16 +104,16 @@ const Map = () => {
       filters,
       dateRange,
       mapPolygon,
-      mapTimeWindow
+      mapTimeWindow,
     );
 
-    setCrashCounts(null); // Clear stale totals before fetch
-    !!apiUrl &&
+    // setCrashCounts(null); // Clear stale totals before fetch
```
chiaberry: does this need to stick around?

viewer/src/views/summary/CrashesByYearCumulative.jsx
```diff
@@ -1,24 +1,25 @@
...
           const num = parseFloat(data[valueKey]);
           const roundNum = (num) => Math.floor(num * 100) / 100;
 
-          i === 0 && acc.push(roundNum(num));
-          i !== 0 && acc.push(roundNum(acc[i - 1] + num));
+          if (i === 0) {
+            acc.push(roundNum(num));
+          } else {
+            acc.push(roundNum(acc[i - 1] + num));
+          }
```
chiaberry, line 22: i like this way better than the old way of doing it

chiaberry (approved):
👏🏼 looks good! i am clicking through the update trying to find anything broken, watching the graphs go up and down, filtered the map etc. everything looks the same as production

viewer/src/utils/store.jsx
johnclary: Thanks for opening [the issue](https://github.com/cityofaustin/atd-data-tech/issues/29044)!

viewer/src/views/map/InfoBox/MapPolygonInfoBox.jsx
johnclary: The VZE linter actually also guards against this: 

```
Expected an assignment or function call and instead saw an expression @typescript-eslint/no-unused-expressions 
```

It only comes up (I think) when we use `&&` plus some side effect function call, so I'm cool with refactoring as needed 👍

viewer/src/views/map/Map.jsx
johnclary: Good stuff!

johnclary (approved):
Wow, all clean! So many code fixes in here—thank you for taking all this on and getting our linter back up and running. This looks totally ready to got to me! 🙌 🚢

viewer/src/views/map/InfoBox/MapPolygonInfoBox.jsx
mddilley: i saw this a lot in Moped code too, and these days i'm finding the if blocks easier to read left to right to know there is some control flow happening. i support this change!

mddilley (approved):
Tested locally and read the code. Looks amazing, and thank you for cracking open some of this older code and improving it! This will be so great for future development in here. 🙌 😎 🙏

viewer/src/views/map/Map.jsx
roseeichelmann: no, good catch!

viewer/netlify.toml
```diff
@@ -1,5 +1,11 @@
+# Default used by main/deploy previews
 [build]
   publish = "build"
+  command = "npm run lint && npm run build:staging"
+
+# Production specific build command
+[context.production]
```
mddilley, line 7: I think having the context-specific build commands all set up is great, and I'm wondering if the project should have `.env.staging`. and `.env.production` waiting in the wings for future env-specific but not sensitive config like in the Moped project (if having blank .env files doesn't break anything). Like [this one](https://github.com/cityofaustin/atd-moped/blob/main/editor/.env.staging) in Moped.

We could always add later on too!

mddilley (approved):
Reviewed the latest commit, and I like that the Netlify build config defaults to staging and then handles production context. I left one question that doesn't block this at all.

viewer/netlify.toml
johnclary: interesting—so do these files allow you to simulate staging/prod locally? or are they consumed by netlify?

johnclary (approved):
Looks good to me!!

viewer/netlify.toml
mddilley: the way that i understand it is:
1. `npm run build:staging` runs telling Vite to use staging mode and consume any secrets in `.env.staging`. So, the file name really means `.env.[mode]`
2. Then, any secrets set in the Netlify environment like our Mapbox token are added in when building and deploying

That is how Moped knows which Cognito user pool to reference in which environment for example.

johnclary: ah, this is cool. keeping as much of the env as possible in version control is appealing. i'll follow up on this 🙏

PR #2063, opened 2026-06-16
Adds process to link crashes to Vision Zero incidents

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/28971

## Testing

**URL to test:** 🏠  Local

1. Replicate your DB if you haven't done so since June 15th. This ensures you have existing VZ incidents in your DB

4. Use your SQL client to enable debug logging, then execute the following tests cases to see how the trigger is working.

``` sql
-- enable debug logging
SET client_min_messages = 'DEBUG';
```

5. Creates a new VZ incident - No matching record found

```sql
INSERT INTO
    crashes_cris (cris_crash_id, crash_timestamp, latitude, longitude, cris_schema_version)
VALUES
    (1, '2000-01-06T00:00:00', 0, 0, '2026');
```

> DEBUG:  Not enough data to match incident. Created incident [integer]

6. Crash matches to incident based on case ID

```sql
INSERT INTO
    crashes_cris (cris_crash_id, crash_timestamp, latitude, longitude, cris_schema_version, case_id, investigat_agency_id)
VALUES
    (2, '2000-01-06T00:00:00', 0, 0, '2026', '200460268', 74);
```

> DEBUG:  Matched to incident 216883 based on case ID 200460268

7. Matched based on approximate time and location

```sql
INSERT INTO
    crashes_cris (cris_crash_id, crash_timestamp, latitude, longitude, cris_schema_version, case_id, investigat_agency_id)
VALUES
    (3, '2016-04-25 05:56:40+00', 30.443672, -97.743103, '2026', '123fake', 74);
```

> DEBUG:  Matched to incident 6 based on geo/temporal

8. Crash which matches multiple existing incidents

```sql
INSERT INTO
    crashes_cris (cris_crash_id, crash_timestamp, latitude, longitude, cris_schema_version, case_id, investigat_agency_id)
VALUES
    (4, '2026-02-07 03:41:00+00', 30.39341472, -97.67192309, '2026', '123fake456', 74);
```

> DEBUG:  Multiple incident matches found

9. Query the `vz_incident_records_view` to see `vz_incidents` associated with a crash record. The query should return 6 total rows referencing 3 different `vz_incident_id` values.

```sql
WITH
    start_ids AS ( SELECT vz_incident_id FROM crashes WHERE vz_incident_id IS NOT NULL
    )
    SELECT * FROM start_ids ids INNER JOIN vz_incident_records_view v ON ids.vz_incident_id = v.vz_incident_id;
```

11. Run the down migration 🧹 

---

#### Ship list

- [ ] Code reviewed


database/migrations/default/1781537795232_crashes_incident_link/up.sql
```diff
@@ -0,0 +1,159 @@
...
+
+COMMENT ON COLUMN public.crashes.vz_incident_id is 'The vz_incidents foreign key.';
+
+CREATE INDEX idx_crashes_vz_incident_id ON crashes(vz_incident_id);
+CREATE INDEX idx_vz_incident_match_status ON crashes(vz_incident_match_status);
+
+ALTER TABLE cad_incidents
+    ADD CONSTRAINT cad_incidents_agency_type_check CHECK (
+            agency_type IN ('AUSTIN PD', 'FIRE', 'AUSTIN-TRAVIS COUNTY  EMS')
+    ),
```
johnclary, line 25: I am adding strong typing to these `agency_type` strings that we get from CAD because we are going to be building a lot of business logic on them. We want to know if CAD decides to change these for some reason.

database/migrations/default/1781537795232_crashes_incident_link/up.sql
```diff
@@ -0,0 +1,159 @@
...
+    ADD COLUMN agency_type_short text generated always AS (
+        CASE
+            WHEN agency_type = 'AUSTIN PD' THEN 'apd'
+            WHEN agency_type = 'FIRE' THEN 'afd'
+            ELSE 'ems'
+        END
+    ) stored;
```
johnclary, line 32: I quickly regretted not creating this column when we initialized the CAD table. the original `agency_type` strings are long and ugly and annoying to type in a query.

database/migrations/default/1781537795232_crashes_incident_link/up.sql
```diff
@@ -0,0 +1,159 @@
...
+
+comment on column cad_incidents.agency_type_short is 'The abbreviated name of the responding agency';
+
+CREATE INDEX idx_cad_incidents_master_incident_number ON cad_incidents(master_incident_number);
+CREATE INDEX idx_cad_incidents_agency_type_short ON cad_incidents(agency_type_short);
+
+CREATE OR REPLACE VIEW public.vz_incident_records_view AS
```
johnclary, line 39: As you'll see in the trigger function that follows, this view gives us a uniform table to interact with for finding similar incidents that overlap in space/time. The idea being that we will add EMS and AFD incident records to it once we bring them into the **Vision Zero Incident** fold.

database/migrations/default/1781537795232_crashes_incident_link/up.sql
```diff
@@ -0,0 +1,159 @@
...
+ADD COLUMN vz_incident_matched_ids BIGINT[],
+ADD CONSTRAINT crashes_vz_incident_match_status_check CHECK (
+    vz_incident_match_status IN (
+        'unprocessed',
+        'created_by_automation',
+        'matched_by_automation_case_id',
+        'matched_by_automation_geo_temporal',
+        'multiple_matches_by_automation',
+        'matched_by_manual_qa')
+);
```
johnclary, line 13: the `unprocessed` status will linger only until we run a backfill.

database/migrations/default/1781537795232_crashes_incident_link/up.sql
```diff
@@ -0,0 +1,170 @@
...
+        ci.master_incident_number   AS record_incident_number,
+        ci.response_date            AS record_timestamp,
+        ci.address                  AS record_address,
+        ci.geom                     AS geom
+    FROM cad_incidents ci
+    WHERE ci.vz_incident_id IS NOT NULL;
+
+COMMENT ON VIEW public.vz_incident_records_view IS 'Unified view of records (crashes, cad_incidents) belonging to a vz_incident, exposed under a common schema for cross-type queries and geo-temporal matching.';
+
+CREATE OR REPLACE FUNCTION public.crashes_match_vz_incident()
```
johnclary, line 67: Trigger that handles the crash-vz_incident matching. The logic is:

  *  [APD crash records only] If the crash's case ID matches an APD CAD incident #, link to the same incident as the CAD record.
  * If the crash matches exactly one existing VZ incident based on time (+/- 30 minutes) and distance (500m) to any of the incident's member records, link to the incident
  * If the crash matches multiple VZ incidents, save the matched incident IDs in to the crash's match metadata. Don't create a new incident.
  * If the crash matches no other incidents, create a new incident

database/README.md
```diff
@@ -503,14 +530,18 @@ SELECT
                 jsonb_build_object(
```
johnclary, line 530: Update geojson query that now includes crashes ✨

database/migrations/default/1781537795232_crashes_incident_link/up.sql
chiaberry: Are there two spaces between COUNTY and EMS?

database/migrations/default/1781537795232_crashes_incident_link/up.sql
```diff
@@ -0,0 +1,170 @@
...
+        -- distinct container is counted once, regardless of member type or count.
+        SELECT count(*), array_agg(v.vz_incident_id)
+        INTO v_match_count, v_matched_ids
+        FROM (
+            SELECT v.vz_incident_id
+            FROM vz_incident_records_view v
+            WHERE v.geom IS NOT NULL
+              AND v.vz_incident_id IS NOT NULL
+              AND v.record_timestamp >= (NEW.crash_timestamp - time_threshold)
+              AND v.record_timestamp <= (NEW.crash_timestamp + time_threshold)
```
chiaberry, line 122: > existing VZ incident based on time (+/- 30 minutes)

time_threshold is 60 minutes, so this is incidents that happen within an 120 min window of the crash?

chiaberry (approved):
Ran through test steps, thanks for including examples of each scenario!

database/migrations/default/1781537795232_crashes_incident_link/up.sql
```diff
@@ -0,0 +1,170 @@
...
+    IF (NEW.investigat_agency_id = 74 AND NEW.case_id IS NOT NULL) THEN
+        -- First try: exact match APD crashes on case_id
+        raise debug 'Attempting to match case ID %', NEW.case_id;
+
+        SELECT vz_incident_id
+        INTO v_vz_incident_id
+        FROM cad_incidents
+        WHERE agency_type = 'AUSTIN PD'
+        AND master_incident_number = NEW.case_id
+        LIMIT 1;
```
mddilley, line 92: is there a possibility of multiple CAD incidents that match a crash case number? Just trying to understand if we are taking the first match here for a specific reason or not and also understand the data better.

database/migrations/default/1781537795232_crashes_incident_link/up.sql
mddilley: very cool to see these two joined together, and it looks nice and ready for future tables to enrich the story of incidents overall 🚀

mddilley (approved):
Ran through all the test queries and viewed the new output of the geojson sample query. Thanks for documenting how all the pieces fit together, and I'm happy to see the patterns used for the EMS incident-person matching coming into play again. Looks great to me! 🚢 🙌

database/migrations/default/1781537795232_crashes_incident_link/up.sql
johnclary: Yes! I wasn't sure where to put a [sic]. This is an artifact from the CAD system.

database/migrations/default/1781537795232_crashes_incident_link/up.sql
johnclary: There really shouldn't be. My understanding is that `master_incident_number` is unique per agency. I was hesitant to put a unique constraint on that column but perhaps I should go ahead and do that 🤔

database/migrations/default/1781537795232_crashes_incident_link/up.sql
roseeichelmann: good call!

database/migrations/default/1781537795232_crashes_incident_link/up.sql
roseeichelmann: i think if we expect them to be unique then it would be good to put a constraint so we can make sure we are proving that with our data

database/migrations/default/1781537795232_crashes_incident_link/up.sql
roseeichelmann: good catch chia, i cant tell if its intended to be a 60 min window or 120 min window

roseeichelmann (approved):
tested and query outputs were all as expected, this looks great! thanks for the documentation as well 🙏

johnclary:
Closing this in favor of https://github.com/cityofaustin/vision-zero/pull/2068

PR #2074, opened 2026-06-30
Create Vision Zero incidents list page

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/28753

This PR adds a basic VZ incident listing backed by a materialized view. The page is only visible to editors and admins.

## Testing

1. You'll want to replicate from prod if you haven't done so in the last week-ish to make sure you've got the latest VZ incidents.
2. Apply migrations and metadata: Clock how long it takes to apply this migration. We might want to consider manually applying this materialized view and using a placeholder statement in the migration itself.
3. Start your VZE
4. Use the side nav to navigate to the new `/incidents` list page. It should load normally. Test all the things:
* Sort by address column
* Show/hide columns using the settings menu (right side of table controls)
* Use the date filter presets, and also the custom data range
* Search by address, other columns
* Use the advanced filter toggles (filter by agency, has/does not have crash report, jurisdiction)
5. Use the toggle to switch to the map view, click around on some features. Use the hyperlinked incident ID to visit the incident details page
6. Confirm the incident details page has a breadcrumb and renders text like `Incident details page for ID #<some-number>`.
7. Log out, then log in with the credentials called **Vision Zero Editor (VZE) Test User: Read-only viewer** and confirm that the **Incident** page is hidden from the side nav.

---

#### Ship list

- [ ] Code reviewed


database/migrations/default/1786050676000_vz_incidents_view/up.sql
```diff
@@ -0,0 +1,136 @@
+DROP VIEW IF EXISTS vz_incidents_view;
+DROP VIEW public.vz_incident_records_view;
+
+--
+-- add additional columns to view in_austin_full_purpose, location_id, is_location_reviewed
```
johnclary, line 5: adding columns to the incident_records_view so that we can use it as the basis for the list view.

editor/app/ems/page.tsx
```diff
@@ -25,21 +25,7 @@ export default function EMS() {
   return (
     <UserEventsLogger eventName="ems_list_view">
       <div className="h-100 d-flex flex-column">
```
johnclary, line 24: Removes the `Beta` tag from the EMS list

editor/components/EMSCardHeader.tsx
```diff
@@ -3,22 +3,8 @@ import { LuInfo } from "react-icons/lu";
 
 export default function EMSCardHeader() {
   return (
-    <>
```
johnclary, line 6: Removes the `Beta` tag from the EMS card on the crash details page

editor/components/TableMap.tsx
```diff
@@ -96,6 +97,10 @@ export const TableMap = ({ mapRef, geojson, mapConfig }: TableMapProps) => {
 
   const PopupComponent = getPopupComponent(mapConfig.popupComponentName);
 
+  const centerFeature = selectedFeatures?.length
+    ? center(selectedFeatures[0])
+    : null;
```
johnclary, line 102: Makes it possible to have other selectable geojson types besides `Point`. I was initially messing around with `MultiPoint` map features. I scrapped that but left in this little code improvement.

editor/configs/routes.ts
```diff
@@ -36,6 +37,12 @@ export const routes: Route[] = [
     label: "EMS",
     icon: LuAmbulance,
   },
+  {
+    path: "incidents",
+    label: "Incidents",
+    icon: FaRoad,
```
johnclary, line 43: Not my favorite icon, but I couldn't find anything else that seemed appropriate.
https://react-icons.github.io/react-icons/search/

database/migrations/default/1783347751354_vz_incidents_view/up.sql
```diff
@@ -0,0 +1,138 @@
...
+        FALSE                            AS is_location_reviewed
+    FROM afd__incidents afd;
+
+
+COMMENT ON VIEW public.vz_incident_records_view IS
+    'Unified view of crash-related records (crashes, cad_incidents, ems__incidents, afd__incidents)'
+    'exposed under a common schema for cross-type queries and geo-temporal matching.';
+
+
+CREATE MATERIALIZED VIEW vz_incidents_view AS
```
johnclary: After many attempts at tuning this query, i opted for a materialized view that takes ~5s to create locally. 

This feels good enough while we're still in proof-of-concept. Long term, I think we'll want to consider possibly setting up a proper table with the triggers that keep it current, or maybe we can live with having this be on an hourly refresh ( a la `location_crashes_view`).

I'm still uncertain about what the finally shape of this view needs to be, so I think for now we can just create this materialized view once via migration and let it become stale while we further refine the future state.

database/migrations/default/1786050676000_vz_incidents_view/up.sql
```diff
@@ -0,0 +1,138 @@
...
+    in_austin_full_purpose
+FROM (
+    SELECT
+        v.vz_incident_id as id,
+        COUNT(*) AS record_count,
+        ARRAY_AGG(DISTINCT record_incident_number ORDER BY record_incident_number) as incident_numbers,
+        ARRAY_AGG(DISTINCT record_table_name ORDER BY record_table_name) as record_tables,
+        ARRAY_AGG(DISTINCT record_responding_agency ORDER BY record_responding_agency) AS responding_agencies,
+        (ARRAY_REMOVE(
+            ARRAY_AGG(v.record_address ORDER BY is_location_reviewed desc, v.record_timestamp, v.record_id), NULL
```
johnclary, line 128: So what we're doing here is picking a single member record to serve as the representative record in the list view. We select that based on:

- If `is_location_reviewed` is `true`, use that record: it is a crash report that has had it's position edited
- Otherwise take the record with the earliest response date, using the arbitrary `record_id` as a tiebreaker

Again, this is just for the proof of concept: we may find a more sophisticated/informed way of doing this later.

database/migrations/default/1786050676000_vz_incidents_view/up.sql
```diff
@@ -0,0 +1,138 @@
...
+
+CREATE MATERIALIZED VIEW vz_incidents_view AS
+SELECT
+    id,
+    record_count,
+    incident_numbers,
+    record_tables,
+    '$' || array_to_string(record_tables, '$,$') || '$' AS record_tables_str,
+    responding_agencies,
+    '$' || array_to_string(responding_agencies, '$,$') || '$' AS responding_agencies_str,
```
johnclary, line 100: We need these string-concatenated fields for search purposes. This is a work around for the fact that Hasura does not have a `_not_contains` operator on array types.

editor/components/VzMapPopupContent.tsx
```diff
@@ -0,0 +1,39 @@
+import Link from "next/link";
+import { formatArrayToString, formatIsoDateTime } from "@/utils/formatters";
+import { GeoJsonProperties } from "geojson";
+
+export interface VzMapPopupContentProps {
+  properties: GeoJsonProperties;
+}
+
+export default function VzMapPopupContent({
```
johnclary, line 9: Image

editor/configs/vzListViewTable.ts
```diff
@@ -0,0 +1,207 @@
...
+            id: "other",
+            column: "responding_agencies_str",
+            operator: "_nlike",
+            value: "%$ems$%",
+          },
+          {
+            id: "other",
+            column: "responding_agencies_str",
+            operator: "_nlike",
+            value: "%$apd$%",
```
johnclary, line 72: Here's that string-concatenated field coming into play: this is how we find non-apd crashes.

Image

editor/configs/vzListViewTable.ts
```diff
@@ -0,0 +1,207 @@
...
+        label: "Has crash report",
+        groupOperator: "_and",
+        enabled: false,
+        inverted: false,
+        filters: [
+          {
+            id: "has_crash_report",
+            column: "record_tables_str",
+            operator: "_ilike",
+            value: "%$crashes$%",
```
johnclary, line 100: And here's how we find incidents that don't have a `crashes` record—aka there is no crash report.

editor/types/tableMapConfig.ts
```diff
@@ -42,7 +46,7 @@ export interface TableMapConfig {
   /**
    * Name of the geojson transformer function to use when converting the input table to geojson
    */
-  geojsonTransformerName: "latLon";
+  geojsonTransformerName: "latLon" | "pointFeature";
```
johnclary, line 49: This lets us consume the geojson `Point` that comes out of the `vz_incident_records_view` rather than reconstructing it via lat/lon. Sorta feels like to the right way to have done this from the beginning.

editor/configs/routes.ts
```diff
@@ -36,6 +37,12 @@ export const routes: Route[] = [
...
+    allowedRoles: ["editor", "vz-admin"],
```
johnclary, line 44: Only editors and admins will be able to see this view while we're in this alpha stage of development

database/migrations/default/1783347751354_vz_incidents_view/up.sql
mddilley: 👍 and I clocked this migration at about 7 Mississippis to apply on my machine

database/migrations/default/1786050676000_vz_incidents_view/up.sql
```diff
@@ -0,0 +1,153 @@
...
+        NULLIF(
+            ARRAY_AGG(DISTINCT record_incident_number ORDER BY record_incident_number) 
+            FILTER (WHERE record_incident_number IS NOT NULL),
+            ARRAY[]::text[]
+        ) as incident_numbers,
+        NULLIF(
+            ARRAY_AGG(DISTINCT record_table_name ORDER BY record_table_name) 
+            FILTER (WHERE record_table_name IS NOT NULL),
+            ARRAY[]::text[]
+        ) as record_tables,
```
mddilley, line 121: are the incident numbers and record tables expected to have corresponding values in their indices? I'm trying to think through how this data is related, and if filtering out `null` would offset the index in an unexpected way. I might be misunderstanding how the two columns are related!

editor/configs/routes.ts
mddilley: When I was testing, I saw that the viewer cannot see the incidents page nav link, but they can navigate to it and to the upload non-CR3s page through a url. I can't remember if role-based routing permissions are implemented or relevant to VZE but wanted to check on that.

editor/configs/vzListViewTable.ts
mddilley: nice thanks for connecting the earlier comment to this one. it was a good reminder about why we do a similar thing for team members, etc. in the Moped advanced search 🙏

editor/configs/vzListViewTable.ts
```diff
@@ -0,0 +1,207 @@
...
+  sortColName: "record_timestamp",
+  sortAsc: false,
+  searchFilter: {
+    id: "search",
+    value: "",
+    column: "address",
+    operator: "_ilike",
+    wildcard: true,
+  },
+  searchFields: [{ label: "Address", value: "address" }],
```
mddilley: Not related to the code changed in this PR but one thing that was a small speed bump for me when testing the search was being able to know which fields that I could search in the list view. I might have missed something or it might be really obvious to users who are familiar with the app but wanted to mention.

mddilley (commented):
This looks good to ship to me! I asked a couple questions, and I can turn this into an approval real fast after that. Thanks John! Very nice to see what the shift to incidents looks like in the editor! 🙌

editor/configs/vzListViewTable.ts
roseeichelmann: An incident cant both have a crash report and not have one, but you can toggle both on as if thats possible. Would it be a big lift to make it so you can only toggle one at a time?

database/migrations/default/1783347751354_vz_incidents_view/up.sql
roseeichelmann: this migration took only like 4 ms for me!

editor/app/ems/page.tsx
roseeichelmann: wahoo i was just wondering when  this feature was going to graduate from "Beta"

editor/app/incidents/page.tsx
```diff
@@ -0,0 +1,45 @@
...
+          <div className="fs-3 fw-bold me-2">Incidents</div>
+          <div className="align-self-center">
+            <Badge bg="info">Beta</Badge>
+          </div>
+        </div>
+        <div className="fw-light text-secondary mb-2">
+          <AlignedLabel>
+            <LuInfo className="me-2" />
+            <span>
+              Incidents are currently in beta. Data may be inaccurate or change
```
roseeichelmann: can we make this line break on smaller viewports?

Image

editor/configs/routes.ts
roseeichelmann: i like it

editor/configs/routes.ts
roseeichelmann: oo i think thats a loophole we didnt consider and should fix, i was also able to replicate this.

question though for john -- why is the Incidents list page not allowed for read only?

roseeichelmann (approved):
Its so awesome to see the incidents coming together in the UI, ive gotta be honest its been kind of hard for me to wrap my heard around incidents and incident matching, but already getting to see it laid out in this list page and map view its helping my visual brain connect the dots.

i also have a few comments/questions as well and i do think the issue mike pointed out about read only still being able to navigate to the page should be addressed if its intended for them not to see this page, but i am ready to approve asap as well!

editor/configs/vzListViewTable.ts
chiaberry: Similarly to Mike, when I followed the test instructions "Search by address, other columns", I tried guessing what other columns to search on (like putting in "ems") and didnt get any results. I couldnt remember if we could search on incident number of not

editor/configs/vzListViewColumns.tsx
```diff
@@ -0,0 +1,70 @@
...
+    valueFormatter: formatArrayToString,
+  },
+  {
+    path: "record_count",
+    label: "Record count",
+    sortable: true,
+  },
+  {
+    path: "record_tables",
+    label: "record_tables",
```
chiaberry: Why is this label in snake case and not "Record tables"?

editor/components/VzMapPopupContent.tsx
chiaberry: I liked searching for Burnet and then toggling the map and seeing most of the incidents lining up with the street, except one or two. The future incident details page is gonna be great

database/migrations/default/1783347751354_vz_incidents_view/up.sql
chiaberry: it took 4 seconds for me

chiaberry (approved):
This is looking great. I am also curious about the permissions around the incident list page. Is that temporary until the Incident details pages are implemented? I'm thinking editors+ are more informed about things in progress...

database/metadata/databases/default/tables/public_cad_incidents.yaml
```diff
@@ -8,6 +8,15 @@ object_relationships:
   - name: vz_incident
     using:
       foreign_key_constraint_on: vz_incident_id
+  - name: vz_incidents_view
```
johnclary: I just realized that we should name this `vz_incidents_list_view` to be consistent with our other record views. i'm going to fix this later because I already have a branch going.

database/migrations/default/1786050676000_vz_incidents_view/up.sql
johnclary: each row will always have a `record_table`, but sometimes an incident number can be null. not a concern here, but you are picking up on potential weirdness with the way to way grab `address` and `point_feature` using the sorting logic below: it's possible that either one of these are null, in which case we could potentially grab the address from one member record and the point from a different record. it's an extreme edge case but definitely a detail to iron out with the VZ team.

editor/configs/routes.ts
johnclary: We decided to keep the incident page obscured from read-only users while it's under development in order to avoid confusing users. It should be harmless if they somehow manage to reach the page, right?

Good to have a gut check on non-CR3s. Read-only users don't have update permissions on the non-cr3 table, so if they did hack the routing the page is harmless. 

It's a bit sloppy that we don't redirect users away from restricted pages or show an error message. I'll open a backlog issue for that 🙏 

* https://github.com/cityofaustin/atd-data-tech/issues/29486

editor/configs/vzListViewTable.ts
johnclary: It would be a lift, yeah. FWIW this behavior is consistent with our column visibility toggles as well 🤷‍♂️

editor/configs/vzListViewTable.ts
johnclary: Sorry for that confusion: i forgot that i didn't enable searching on other columns. I added searching by **Incident number** as well—I think that will be valuable 🙏

editor/app/incidents/page.tsx
johnclary: Yeah! Thanks for catching that. the `AlignedLabel` helper uses `text-nowrap`, so i swapped that out for a regular div.

editor/configs/vzListViewColumns.tsx
johnclary: An oversight! Fixed 🙏

editor/configs/vzListViewTable.ts
roseeichelmann: okay thats fine then! i figured it would be a big lift

mddilley (approved):
Forgot to come back, read comments, and change this to an approval. Doing that now!  Thanks for the responses. 🙏🚢 ✅

mateoclarke (approved):
Tested locally and works as described. The migration took longer for me than other folks, but I wasn't counting mississippis. I'd say around 30-60 sec? Seems like a non-issue tho. 🚢

PR #2086, opened 2026-07-16
Creates DB view of crash risk factors

### LINKED ISSUE #29179: Create database view of crash risk factors
As part of https://github.com/cityofaustin/atd-data-tech/issues/29176, create a database view of crash risk factors based on our [business rules](https://docs.google.com/document/d/1YPVyAz72YUMo-kMvN65dMjN7DKba6OEqGJGQWMTwWAc/edit?usp=sharing). 

These business rules use a combination of unit-level contributing factors as well as other crash attributes. For example, the rules state that we should assign the "Impaired Driving" category when a `charges` value includes `"DWI"`.

The deliverable for this work is a view called `crash_risk_factors_view` that returns an array of crash risk factors per crash. The view structure should look like this:

`id` | `risk_factors`
-- | --
1813452 | `["Distracted driving", "Impaired driving"]`
1813440 | `["Distracted driving"]`
1813452 | `null`
1813452 | `["Visual obstruction", "Speeding", "Impaired driving"]`

Where `id` is the `crashes.id` value and `risk_factors` is the contributing factors category array computed from the business rules doc.

### Implementation

I used the below prompt with Claude Sonnet 5. Here is the full [chat](https://claude.ai/share/5fbdb262-c2e9-420c-bd87-a3f66c19ba63).
 

> I have a table called `crashes` with a related table `units`. Each `units` record holds five contributing factor ID fields (`contrib_factor_1_id`, `contribut_factor_2_id`, etc.). 
> 
> I want to be able to assign/generate contributing factors at the crash level based on these fields, plus some other crash attributes. I also want to bucket/simplify contributing factor names, because our contributing factor lookup table has 70+ values.
> 
> For example, if any `contrib_factor_x_id` field is 45 (“Had been drinking”) OR if any of the of the crash `charges` (another related table) contains “DWI”, then we can add “Impaired driving” to the list of crash-level contributing factors we generate.
> 
> I have a list of all these rules, but I’m not sure how to approach this. I think I might need to iterate through all of my custom contributing factor categories one by one to see if conditions are met. E.g. “Distracted Driving”, “Failure to Yield” etc.

The LLM suggested a strategy that makes a lot of sense to me, which is:
- Create a new lookup table in the DB that holds our `contrib_factr` mappings to each category (more on that below)
- In the view itself, union the units `contrib_factrr_1`, `contrib_factr_2` fields into a single column  in a CTE
- Use CTEs/querying to produce the results

If we want to go with the new lookup table that defines how CRIS contributing factors map to our custom contributing factor categories, this could be a `lookups.contrib_factor_categories` table like this:

`contrib_factor_id`  (references lookups.contrib_factr)| `risk_factor`
-- | --
20 (DRIVER INATTENTION) | Distracted driving
40 (FATIGUED OR ASLEEP) | Distracted driving
72 (CELL/MOBILE PHONE USE) | Distracted driving
4 (CHANGED LANE WHEN UNSAFE) | Distracted driving
17 (DISREGARD TURN MARKS AT INTERSECTION) | Failure to yield

Where the table primary key is the unique (`contrib_factor_id`, `contrib_factor_category`) combination.

mateoclarke:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/29179

## Testing

**URL to test:** 
Local

**Steps to test:**
- Spin up local VZE: `./vision-zero replicate-db`
- Make sure migrations are run; `hasura migrate apply` & `hasura metadata apply`
- Open Hasura console: `cd database` then `hasura console` _or poke around in your DB GUI of choice_
- Check for new lookup table `risk_factor_categories ` - http://localhost:9695/console/data/default/schema/lookups/tables/risk_factor_categories/browse
- Check for new view `crash_risk_factors_view` -  http://localhost:9695/console/data/default/schema/public/views/crash_risk_factors_view/browse
- Try these GraphQL queries...

**Preview the new view:**
```graphql
query {
  crash_risk_factors_view(
    where: { risk_factors: { _is_null: false } }
    limit: 20
  ) {
    id
    risk_factors
  }
}
```
 
**Preview crashes risk factors compared to source data on units:**
```graphql
query {
  crashes(
    where: {
      crash_risk_factors_view: {
        risk_factors: { _is_null: false }
      }
    }
    limit: 10
  ) {
    id
    record_locator
    crash_risk_factors_view {
      risk_factors
    }
    units {
      contrib_factr { id label }
      contrib_factr_2 { id label }
      contrib_factr_3 { id label }
      contrib_factr_p1 { id label }
      contrib_factr_p2 { id label }
    }
    charges_cris {
      charge
    }
  }
}
```

**Test a filtered view**
```graphql
query {
  crash_risk_factors_view(
    where: { risk_factors: { _contains: ["Impaired driving"] } }
    limit: 20
  ) {
    id
    risk_factors
  }
}
```

---

#### Ship list

- [ ] Code reviewed


Charlie-Henry (approved):
I see these two new tables in my DB viewer 💯 

Image

database/migrations/default/1784264703617_crash_risk_factors/up.sql
```diff
@@ -0,0 +1,248 @@
...
+-- Source: Crash contributing factors business rules: https://docs.google.com/document/d/1YPVyAz72YUMo-kMvN65dMjN7DKba6OEqGJGQWMTwWAc/edit?tab=t.0#heading=h.5wz64nv7k8b0
+-- Note: CRIS id 72 (CELL/MOBILE PHONE USE) was removed in CRIS v30.1; use 75–78.
+CREATE TABLE lookups.risk_factor_categories (
+    contrib_factor_id integer NOT NULL REFERENCES lookups.contrib_factr (id) ON UPDATE CASCADE ON DELETE RESTRICT,
+    risk_factor_category text NOT NULL,
+    PRIMARY KEY (contrib_factor_id, risk_factor_category)
+);
+
+-- Ensure referenced CRIS lookup rows exist (no-op when already loaded from a dump).
+INSERT INTO lookups.contrib_factr (id, label, source)
```
johnclary: We don't want to be inserting `cris`-attributed values in to the lookup table, because our [lookup table helper](https://github.com/cityofaustin/vision-zero/tree/main/toolbox/get_lookup_table_changes) is in charge of keeping the lookups in sync with the CRIS schema. Since this is fully a no-op I think you can just  remove this step.

database/migrations/default/1784264703617_crash_risk_factors/up.sql
```diff
@@ -0,0 +1,248 @@
+-- Mapping of CRIS contrib_factr IDs to Vision Zero crash-level risk factors.
+-- Source: Crash contributing factors business rules: https://docs.google.com/document/d/1YPVyAz72YUMo-kMvN65dMjN7DKba6OEqGJGQWMTwWAc/edit?tab=t.0#heading=h.5wz64nv7k8b0
+-- Note: CRIS id 72 (CELL/MOBILE PHONE USE) was removed in CRIS v30.1; use 75–78.
+CREATE TABLE lookups.risk_factor_categories (
+    contrib_factor_id integer NOT NULL REFERENCES lookups.contrib_factr (id) ON UPDATE CASCADE ON DELETE RESTRICT,
+    risk_factor_category text NOT NULL,
+    PRIMARY KEY (contrib_factor_id, risk_factor_category)
```
johnclary, line 8: I missed this in scoping: would you add a column called `source` that defaults to `'vz'`? It does not need to be exposed in graphql. This a convention we carry around in all the lookup tables, including our custom ones, e.g. `lookups.mode_category`. It's a breadcrumb that this is a custom lookup table and not a CRIS thing.

Image

database/migrations/default/1784264703617_crash_risk_factors/up.sql
```diff
@@ -0,0 +1,248 @@
...
+);
+
```
johnclary, line 10: Would you also add a table-level comment here that describes what this table is?

database/migrations/default/1784264703617_crash_risk_factors/up.sql
```diff
@@ -0,0 +1,248 @@
...
+                    (units.contrib_factr_3_id),
+                    (units.contrib_factr_p1_id),
+                    (units.contrib_factr_p2_id)
+            ) AS factor_ids (contrib_factor_id)
+            INNER JOIN lookups.risk_factor_categories AS categories
+                ON categories.contrib_factor_id = factor_ids.contrib_factor_id
+        WHERE
+            units.is_deleted = FALSE
+            AND factor_ids.contrib_factor_id IS NOT NULL
+            AND factor_ids.contrib_factor_id <> 0
```
johnclary, line 107: thanks for dealing with the weird `NONE` contrib factor 😐

database/migrations/default/1784264703617_crash_risk_factors/up.sql
```diff
@@ -0,0 +1,248 @@
...
+            people
+            INNER JOIN units ON units.id = people.unit_id
+        WHERE
+            people.is_deleted = FALSE
+            AND units.is_deleted = FALSE
+            AND (
+                people.prsn_alc_rslt_id = 1
+                OR people.prsn_drg_rslt_id = 1
+                OR (
+                    people.prsn_bac_test_rslt ~ '^[0-9]*\.?[0-9]+$'
```
johnclary, line 126: Maybe add an inline comment here with an example or two of what will be matched?

Tricky problem to solve here with this being a text field!

database/migrations/default/1784264703617_crash_risk_factors/up.sql
```diff
@@ -0,0 +1,248 @@
...
+            all_categories
+        GROUP BY
+            id
+    )
+SELECT
+    crashes.id,
+    NULLIF(aggregated.risk_factors, '{}'::text[]) AS risk_factors
+FROM
+    crashes
+    LEFT JOIN aggregated ON aggregated.id = crashes.id;
```
johnclary, line 204: This view looks great!

johnclary (commented):
This is looking great—thanks for getting all of these rules into place. Just a few minor changes needed 🚀

chiaberry (approved):
i too see the two tables, and nice extra fix on the failing build. I am not sure I completely understand why it failed on an empty db, but I get that it happens in our build/CI process. 

🚢

johnclary (approved):
Looks good! Thanks, Mateo! 🚢 🚢 🚢

roseeichelmann (approved):
this looks great!! 🚀

mddilley (approved):
Awesome! Very cool to see the Google doc categorization translated to a DB view. And love the comments in here. 🚢 🙌

PR #2073, opened 2026-06-29
Upgrade to react 18

roseeichelmann:
## Associated issues
https://github.com/issues/assigned?issue=cityofaustin%7Catd-data-tech%7C28869

This is going to become the collector branch for this work

**Changes made in this PR:**

- Upgrade React 16 -> 18

- @material-ui/core → @mui/material (v5)

- hookrouter → React Router 

- reactstrap V8  →  V9

-  bootstrap v4 →  v5.33

- upgraded chartjs-plugin-stacked100

- react-chartjs-2 v2 → v4. Should update to v5 eventually **(future issue?)**

- chart.js  v2 →  v3, in the future could update to v4 but thats not compatible with our chartjs-plugin-stacked100 so we can update in the future & get a new plugin but thats a **later issue**

- react-map-gl v7 → v8

- reaviz v8 → v16

- react-dates → react-datepicker

## Testing

**URL to test:** 
Local

**Steps to test:**
1. Check out this branch, delete your local node-module folder and package-lock files.
2. Run `npm install --legacy-peer-deps` the flag ignores dependency warnings for now when installing packages, as we are tackling all of this in phases (netlify preview isnt going to work for now until we address all these dependencies)
3. Run the viewer, compare it to staging. Keep in mind i wanted to come to a stopping point so some of the issues with the header and footer formatting i plan to address next PR. But see that the Viewer is running in React 18, with React Router, bootstrap v4, etc etc and is completely functional!! :sparkles: 

**Follow up PRs todo list:**

- Upgrade to React 19 and in doing so will have to update @fortawesome/react-fontawesome and react-chartjs-2 to version 5.3.1
- Fix all formatting issues that arise
- https://github.com/cityofaustin/atd-data-tech/issues/29449
- https://github.com/cityofaustin/atd-data-tech/issues/15051
- https://github.com/cityofaustin/atd-data-tech/issues/28870
- https://github.com/cityofaustin/atd-data-tech/issues/29454

---

#### Ship list

- [ ] Check migrations for any conflicts with latest migrations in `main` branch
- [ ] Confirm Hasura role permissions for necessary access
- [ ] Code reviewed
- [ ] Product manager approved

PR #2088, opened 2026-07-22
Replace date range date picker

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/15051

## Testing

Test locally -- It will be helpful if you have tested Rose's previous PRs that are in this same stack (https://github.com/cityofaustin/vision-zero/pull/2073). It might also be beneficial to edit the dates in prod to get a feel of how it works now. 

Use the right `npm` version and install the date-picker package

```bash
nvm use
npm install --legacy-peer-deps
```

then fire up your viewer, `npm run dev`

Navigate to the map and interact with the date pickers on the left. Click to select a start date and end date. Clicking the arrow icon will reset the date span. 

---

#### Ship list

- [ ] Code reviewed


viewer/src/views/nav/SideMapTimeOfDayChart.jsx
```diff
@@ -118,11 +118,6 @@ export const SideMapTimeOfDayChart = ({ filters }) => {
     }
   };
 
-  const createTooltipData = (tooltipItem) => {
-    const index = tooltipItem.index;
-    return `${timeWindowPercentages[index]}% (${timeWindowData[index]})`;
-  };
-
```
chiaberry, line 125: linting alerted me to this not being used

viewer/src/constants/time.js
```diff
@@ -44,9 +43,10 @@ export const currentYearString = summaryCurrentYearStartDate.slice(0, 4);
 export const prevYearString = summaryLastYearStartDate.slice(0, 4);
 
 // Map time data
-export const mapStartDate = moment(dataStartDate);
-
-export const mapEndDate = moment(dataEndDate);
+// export const mapStartDate = format(dataStartDate, "yyyy-MM-dd")
+// export const mapEndDate = format(dataEndDate, "yyyy-MM-dd");
+export const mapStartDate = dataStartDate;
+export const mapEndDate = dataEndDate;
```
chiaberry: I still need to go through the app and see where mapStartDate and mapEndDate are used before I remove and just use data*Date

viewer/src/views/map/helpers.js
```diff
@@ -39,16 +36,18 @@ export const createMapDataUrl = (
...
 
   // SideMapControlDateRange uses null to check if user set dates so
   // need to handle it and avoid unnecessary API calls
   if (dateRange.start === null || dateRange.end === null) return null;
-  const startDate = convertDateToSocrataFormat(dateRange.start, "T00:00:00");
-  const endDate = convertDateToSocrataFormat(dateRange.end, "T23:59:59");
+
+  // convert dates to socrata format
+  const startDate = format(dateRange.start, "yyyy-MM-dd") + "T00:00:00";
+  const endDate = format(dateRange.end, "yyyy-MM-dd") + "T23:59:59";
```
chiaberry, line 53: thinking of adding validation

viewer/src/views/nav/SideMapControlDateRange.jsx
```diff
@@ -121,157 +27,97 @@ const SideMapControlDateRange = ({ type }) => {
...
+        id={`map-start-date-${type}`}
+        selected={start}
+        startDate={start}
+        endDate={end}
+        onChange={handleStartDateChange}
+        dateFormat={"MM/dd/yyyy"}
+        selectsStart
+        minDate={dataStartDate}
+        maxDate={dataEndDate}
+        showMonthYearDropdown
```
johnclary: I don't see a practical use case for the month picker and am in favor of removing it

Image

viewer/src/views/nav/SideMapControlDateRange.jsx
```diff
@@ -121,157 +27,97 @@ const SideMapControlDateRange = ({ type }) => {
...
+      <StyledDatePicker
+        id={`map-end-date-${type}`}
+        selected={end}
+        startDate={start}
+        endDate={end}
+        onChange={handleEndDateChange}
+        dateFormat={"MM/dd/yyyy"}
+        selectsEnd
+        minDate={dataStartDate}
+        maxDate={dataEndDate}
```
johnclary: I think the max date should default to and not exceed the current date. I find it confusing that I can't set the end date to today. I want to know that I'm looking at the most recent data (even though the Socrata dataset does not have data less than two weeks old).

For the min date—I think defaulting to the data start date is fine, but I think users should be able to query all data in the ODP! Setting the min to 1/1/2014 would cover that.

viewer/src/views/nav/SideMapControlDateRange.jsx
```diff
@@ -121,157 +27,97 @@ const SideMapControlDateRange = ({ type }) => {
...
-        isDayBlocked={isOutsideDateLimits} // Grey out dates
-        numberOfMonths={!canTwoMonthsFit || (isTablet && !isMobile) ? 1 : 2}
+      <StyledDatePicker
+        id={`map-start-date-${type}`}
+        selected={start}
+        startDate={start}
+        endDate={end}
+        onChange={handleStartDateChange}
+        dateFormat={"MM/dd/yyyy"}
+        selectsStart
```
johnclary: TIL this component has support for showing ranges. It seems busted though—it does not highlight dates outside of the current month. I lean toward nixing this and just have the single date highlighted for both start and end pickers.

Image

viewer/src/views/nav/SideMapControlDateRange.jsx
```diff
@@ -121,157 +27,97 @@ const SideMapControlDateRange = ({ type }) => {
...
+        minDate={dataStartDate}
+        maxDate={dataEndDate}
+        showMonthYearDropdown
+        popperPlacement="bottom-start"
+        // withPortal // maybe for mobile
```
johnclary: It's looking promising as-is, though it would be nice to actually test on a mobile device. Future enhancement it would be nice to use native pickers on mobile since they are better optimized for the task and more familiar to users.

johnclary (commented):
Looking really good, Chia! I left some feedback that I hope is helpful. Did not take a close look at the code itself.

viewer/src/views/nav/SideMapControlDateRange.jsx
chiaberry: okay I will nix it -- I thought the range option was preventing me from putting end dates earlier than start dates, but i was just able to do just that directly inputting into the fields.

johnclary:
Thanks, Chia! I am adding the rest of the team for review as well.

viewer/src/views/nav/SideMapControlDateRange.jsx
```diff
@@ -4,115 +4,21 @@ import ThemedStyleSheet from "react-with-styles/lib/ThemedStyleSheet";
...
-      },
-      spacing: {
-        ...DefaultTheme.reactDates.spacing,
-        displayTextPaddingLeft_small: 4,
-        displayTextPaddingRight_small: 4,
-        displayTextPaddingBottom_small: 4,
-      },
-    },
-  };
+const minDate = new Date("2014-01-01");
```
roseeichelmann: the min default date for me is showing 01/01/2022, should we make 01/01/2014 the default?

roseeichelmann (approved):
I think this looks great and overall is an improvement!

 when clicking thru the months the position of the calendar popup would move from above to below. Is there a way to prevent that? See below video:

https://github.com/user-attachments/assets/df197d28-793a-4bdc-8050-64e3fa73f398

viewer/src/views/map/helpers.js
johnclary: I think this works fine without validation! An invalid date input (e.g. `"12/xy/2015"`) is rejected cleanly. Did you have something else in mind?

viewer/src/views/nav/SideDrawer.jsx
```diff
@@ -96,10 +96,9 @@ const SideDrawer = () => {
           >
             <SideDrawerContent type="temporary" />
           </DrawerStyled>
-
           <DrawerStyled
             id="permanent-drawer"
-            variant="permanent"
+            variant="permanent" // this is always mounted, but when on mobile display is none
```
johnclary, line 99: 🙏

viewer/src/views/nav/SideMapControlDateRange.jsx
johnclary: My thinking was that the default date start should match the summary page, which is Jan 1st <5 years ago. Users now have the option to go back to the beginning of our dataset, which is not a thing in prod.

Does that seem like a decent compromise?

viewer/src/views/nav/SideMapControlDateRange.jsx
```diff
@@ -121,164 +27,94 @@ const SideMapControlDateRange = ({ type }) => {
...
-        orientation={isMobile ? "vertical" : "horizontal"} // More mobile friendly than horizontal
-        isOutsideRange={() => false} // Enable past dates
-        isDayBlocked={isOutsideDateLimits} // Grey out dates
-        numberOfMonths={!canTwoMonthsFit || (isTablet && !isMobile) ? 1 : 2}
+      <StyledDatePicker
+        id={`map-start-date-${type}`}
+        selected={start}
+        onChange={handleStartDateChange}
+        dateFormat={"MM/dd/yyyy"}
+        minDate={minDate}
```
johnclary: A minor thing that feels off with this picker is that if you input a date before the `minDate`, the picker reverts to whatever date was perviously saved. E.g., if i refresh the page and set the start to `1/1/2000`, it reverts to `1/1/2022`.

It would be nice if inputting `1/1/2000` autocorrected to `1/1/2014` — that would be a clear signal to users that 2014 is the data cutoff. 

Would you mind checking to see if this component exposes some way for us to do that?

viewer/src/views/nav/SideMapControlDateRange.jsx
```diff
@@ -4,115 +4,21 @@ import ThemedStyleSheet from "react-with-styles/lib/ThemedStyleSheet";
 import aphroditeInterface from "react-with-styles-interface-aphrodite";
 import DefaultTheme from "react-dates/lib/theme/DefaultTheme";
```
johnclary: It looks like we might be able to remove this and the dusty dependency on `react-dates` 👀

johnclary (commented):
Wow, this is looking really great. Much improved UX, no WSOD, and a lot of nice code cleanup as well. This is super close—I had a few comments.

viewer/src/views/nav/SideMapControlDateRange.jsx
chiaberry: I cant find an exposed way to handle this, I have been experimenting with reading [the raw input](https://reactdatepicker.com/#example-read-raw-input-value-on-change) with the hopes to trap the update and replace the wrong date with 2014, but it is happening separately from the raw input handler

johnclary: I PR'd a minor fix for this and requested review from you and Rose
https://github.com/cityofaustin/vision-zero/pull/2117

chiaberry: I will wrap up my current review and test 2117, but looking at it (and after having experimented with that handler), it looks great

johnclary: I have merged that work in here. Noting that Chia separately found that the date pickers are not working on mobile. We will address that via https://github.com/cityofaustin/atd-data-tech/issues/29925 and the scope of this PR can remain concerned with desktop use only.

frankhereford (approved):
I tested this locally, and it works great. Code changes all make sense. Thank you Chia! 

The only thing I can think of, and I believe you have already mentioned it, is that you can set the end-date to prior to the start-date. The app does do the right thing -- show no crashes -- so I wouldn't let that block shipping.

johnclary (approved):
Thanks, Chia!

This is ready to go 🚢

Noting that https://github.com/cityofaustin/atd-data-tech/issues/29925 is going to block us from releasing this React 18 upgrade work.

PR #2089, opened 2026-07-22
Address layout and style issues

roseeichelmann:
## Associated issues
Closes https://github.com/issues/assigned?issue=cityofaustin%7Catd-data-tech%7C29449

## Testing

**URL to test:** 
Local

**Steps to test:**

1. Will be helpful if you first follow test steps in the [previous PR ](https://github.com/cityofaustin/vision-zero/pull/2073) to get setup locally
2. Really the test is to compare with production and see if there are any differences we dont like, mostly i was fixing the formatting of the Header, Footer, and map view side toggle
3. Make sure you test different view port widths including mobile!
4. The bootstrap upgrade came with responsive font size based on screen size (bigger or smaller) which i actually like, it makes the banner font bigger on my screen which i think makes it more accessible but lmk what yall think
5. There are tiny tiny diffs in other spacing that i think are fine, i dont really want to start pixel pushing unless something looks bad so lmk what yall think!

---

#### Ship list

- [ ] Code reviewed


viewer/src/views/nav/Footer.jsx
roseeichelmann, line 1: I had to adjust the footer logo seal and  i adjusted where it sits bc i think it looks better like this

Now:

Image

Before:

Image

viewer/src/views/nav/SideMapFooter.jsx
roseeichelmann, line 1: i refactored this component to use styled components like some of our other components, i realized theres a lot of inconsistency in how we are styling across the app

viewer/src/views/summary/CrashesByMode.jsx
```diff
@@ -369,8 +369,10 @@ const CrashesByMode = () => {
                         stacked: true,
                       },
                     },
-                    legend: {
-                      display: false,
+                    plugins: {
+                      legend: {
+                        display: false,
```
roseeichelmann, line 374: @johnclary pointed out in the previous PR that this legend was showing when it wasnt before, its bc this option needs to be moved under the plugins umbrella

viewer/src/views/summary/Summary.jsx
```diff
@@ -63,8 +63,8 @@ const Summary = () => {
                     places to live, but too many of our fellow Austinites are
                     killed or seriously injured in traffic crashes each year. To
                     learn more about the City's transportation safety
-                    initiatives, visit Austin Transportation's Vision Zero
-                    Program{" "}
+                    initiatives, visit Austin's Transportation and Public Works
```
roseeichelmann, line 66: thanks to @frankhereford for this rec in the last PR!

viewer/src/views/nav/Footer.jsx
johnclary: Much better, thanks!

viewer/src/views/nav/Footer.jsx
```diff
@@ -39,25 +39,30 @@ const Footer = () => {
...
+      }
     }
 
     /* Prevent links from overlapping CoA logo */
     @media only screen and (max-width: ${responsive.bootstrapLarge}px) {
       .coa-logo {
         position: relative;
         left: 40px;
-        top: 45px;
+        top: 70px;
```
johnclary, line 65: Would be nice to remove the pixel pushing and use flexbox, but I am fine leaving that for another day which may not ever come.

viewer/src/views/nav/Header.jsx
```diff
@@ -108,8 +108,8 @@ const Header = () => {
             // In Summary view, match padding and margins of Summary content below
             className={`${
               isSummaryView
-                ? "px-xs-0 mx-xs-0 ps-md-2 pe-md-1 px-lg-3 mx-lg-4"
-                : "px-0"
+                ? "px-xs-0 mx-xs-0 ps-md-2 pe-md-1 px-lg-3 mx-lg-4 d-flex justify-content-between"
+                : "px-0 d-flex justify-content-between"
```
johnclary, line 112: 🙏

viewer/src/views/nav/SideMapFooter.jsx
johnclary: Looks great—thank you for working through it!

johnclary (commented):
Rose, this looks great—all the major issues are resolved.

I'm noticing one lingering change to this version of the viewer that I'd like to address, which is that font-sizes are scaled up a bit from prod.

##### Prod 👇 

Image

#### This version 👇 

Image

For example, the widget card footer text is `0.875rem` in prod, and that's doing a better job of keeping the footer text from wrapping. These images also highlight the slightly adjusted `secondary` text color and decreased padding compared to prod, but I don't think those changes need to be reverted.

Would you mind tinkering with the font sizes in the summary widgets to get them to match prod?

This is just an aside after I went down a rabbit hole on this: after inspecting the computed styles  on the page, it kinda seems like the Material `CssBaseline` is overriding styles in prod that it's not overriding in the upgraded viewer. It appears that in prod we were unintentionally import the MUI `CssBaseline` as a global stylesheet via `SideDrawer.jsx`. I don't think that import is actually needed at all anymore since I can't tell it making a difference when i remove it from your branch 🤷

viewer/src/views/nav/Footer.jsx
```diff
@@ -128,6 +126,12 @@ const Footer = () => {
               </Col>
             ))}
           </Row>
+          <img
+            alt="City of Austin seal"
```
frankhereford: Do you think this alt text should say "City of Austin logo" since they replaced the seal with the A-letter mark?

viewer/src/views/nav/SideMapFooter.jsx
```diff
@@ -1,56 +1,62 @@
...
+`;
 
+const SideMapFooter = () => {
   return (
-    <div className="side-map-footer" style={footerStyle}>
+    <FooterContainer>
       {/* Logo */}
-      <img style={logoStyle}
+      <Logo
         alt="City of Austin seal"
```
frankhereford: Here's another one of the seal / logo changes if you change it.

frankhereford (approved):
This looks awesome Rose. I can't tell the difference between local and production, except in a few places where you have improved things. 🚢🚢🚢

I don't think this is related to this work at all, but the map is not responding to scroll-wheel input to zoom in and out. It works in staging though, and I think I remember this being addressed recently, but I wanted to mention it if that's not what you're expecting.

chiaberry (approved):
I think the increased font size for the banner is good, but the widgets font size feels too big. This isnt related to these style issue fixes, but the animation when filtering on the "travel by mode" chart feels off, I believe John may have mentioned this previously but I couldnt find where. 

going to approve, figure you will either address the font size here or in the parent branch

viewer/src/views/nav/Footer.jsx
roseeichelmann: I updated, thank you!

roseeichelmann:
Updated to fix the discrepancy in widget font size @johnclary !

johnclary (approved):
Thanks, @roseeichelmann!


Probably as a result of removing the MUI css baseline, the font-family appears to be coming through for me as apple-system font. We lost Helvetica.

Like I mentioned [here](https://austininnovation.slack.com/archives/CM9MK950S/p1788207179440739) we are declaring font-family in a few different places and its enough of a mess to warrant its own issue. So i'll make that one. This can merge IMO 🙏  🚢

PR #2093, opened 2026-07-23
Adds basic incident details page

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/29381

This PR adds a basic incident details page which can be reached from the [incident list view](https://github.com/cityofaustin/vision-zero/pull/2074). The goal here is just to provide basic visibility to Xavier and Joel about which records have been grouped into an incident. Further scoping and research needed to design a version of the page that will be useful and accessible to all users.

## Testing

**URL to test:** Local

⚠️  You will need to replicate your DB again if you haven't done so since testing https://github.com/cityofaustin/vision-zero/pull/2074 ⚠️

1. Start your local stack and apply migrations and metadata.
2. Navigate to the Incidents list view (`http://localhost:3002/editor/incidents`). You can use the filter toggles to find incidents with lots of different record types (hint: enable the AFD, EMS and "Has crash report" toggles)
3. Navigate to the incident details page from the records in list view. Verify these things:

* The page title shown your browser tab is **Incident <incident-id>**
* The map loads with a single blue pin in the center. (Although an incident may not have a location if it is comprised of a single crash report with no location - [example](http://localhost:3002/editor/incidents/651876))
* Each record table is populated accordingly, and each table's column visibility picker is functional

---

#### Ship list

- [ ] Code reviewed


database/metadata/databases/default/tables/public_vz_incidents_list_view.yaml
```diff
@@ -1,6 +1,52 @@
 table:
-  name: vz_incidents_view
+  name: vz_incidents_list_view
```
johnclary, line 2: Renaming this to be consistent with our other list views 😇

editor/app/incidents/[id]/page.tsx
```diff
@@ -1,5 +1,53 @@
...
-import { use } from "react";
+import { use, useEffect } from "react";
+import { notFound } from "next/navigation";
+import Col from "react-bootstrap/Col";
+import Row from "react-bootstrap/Row";
+import {
+  LuAmbulance,
+  LuClipboardPen,
+  LuFlame,
+  LuHeadset,
```
Comment (johnclary) [inline editor/app/incidents/[id]/page.tsx:10]: * I don't think I like the flame icon
 * We should probably switch the clipboard icon to the car burst icon we use for crashes in the side nav. We may also want to rename **Crashes** to **Crash reports** elsewhere in the app 🤔 
 
Wherever we land, these icons are going to play an important role in helping users distinguish what kind of record they're looking at. They will also be used in map markers.

One idea I have been tinkering with is having chip icons that show at a glance what kind of records are attached the incident. Sorta like this:

Image

editor/components/LocationMap.tsx
```diff
@@ -1,4 +1,4 @@
-import { useMemo, MutableRefObject } from "react";
+import { useMemo, RefObject } from "react";
```
johnclary, line 1: `MutableRefObject` was deprecated

editor/components/IncidentMapCard.tsx
johnclary, line 1: Just rendering a single pin on the map for now. I have yet another branch that takes the map a step further by showing all member record locations.

editor/app/incidents/[id]/page.tsx
Comment (frankhereford) [reply editor/app/incidents/[id]/page.tsx:10]: The flame doesn't bother me, I think it's OK. Especially, if I recall correctly, I think you had changed the color to more of a yellow in the work you demoed today in the planning meeting. 

I say this in humor -- The one that gets me is the `FaRoad` in the page title. I think it's fine, and I'm not advocating for changing the picture of the road -- it's just that my eyes always see a pair of lungs, and I can't unsee it. 

As for the chips at the top -- that's an awesome idea. Maybe you could add a number in them that shows the number of that type of records shown in the table below? Maybe order them largest to smallest? They could be links to anchors at the tables they refer to maybe too? Just some random ideas

frankhereford (approved):
This looks great John. I clicked all over the page and verified the items you listed in your PR comment. 🚢 🚢 🚢 This is an awesome launching off point for the incidents details page.

One thing you might add or spin off is the 404 page support similar to [how it works on the crash detail page](http://localhost:3002/crashes/not-an-id) when you feed it a bogus ID in the URL.

Thanks again for this awesome work!

database/views/materialized/vz_incidents_list_view.sql
mddilley, line 1: i had some issue with my local app caching the old name of the view, and, in debugging, I saw that the old view definition stuck around in a .sql file that could be cleaned up `vz_incidents_view.sql` 🧹

editor/app/incidents/[id]/page.tsx
Comment (mddilley) [reply editor/app/incidents/[id]/page.tsx:10]: same i think the flame instantly makes me think of AFD. And +1 about the chips. I could imagine being able to quickly see how many different data sources contribute to an incident being very helpful!

editor/components/RecordTypeBadge.tsx
```diff
@@ -0,0 +1,52 @@
...
+export function RecordTypeBadge({
+  icon: Icon,
+  label,
+  iconStyle,
+}: RecordTypeBadgeProps) {
+  return (
+    <div className="d-flex">
+      <AlignedLabel>
+        <Icon className={`fs-4 me-2`} style={iconStyle && { ...iconStyle }} />
+        <span className="fw-light fs-6 text-dark me-2 my-auto">{label}</span>
```
mddilley: Small thing but I noticed that the labels on these badges are very close to the background color in dark mode. 🌚  Seems like a small tweak and looks great otherwise!

Image

editor/types/vzIncidentRecord.ts
mddilley, line 1: @johnclary I saw the renaming of this type earlier in the code, and I see that there is still a `VzIncident` type in `vzincident.ts`. is that a separate type still needed or something for cleanup? Just checking because my memory is failing me right now! 😅

mddilley (commented):
Everything looks great on this new page! Awesome to see the data and relationships that we've been seeing in the database start to be represented in the UI. 🤩 I left a couple comments about small things, and I really have only one question and then I'll approve afterward. Thanks!

editor/app/incidents/[id]/page.tsx
Comment (mateoclarke) [reply editor/app/incidents/[id]/page.tsx:10]: I like the chip idea too

database/views/materialized/vz_incidents_list_view.sql
mateoclarke: yup. I had the same problem. I had to rename the view manually. Something was weird with the migration, but because I had run it already locally testing the other branch.

mateoclarke (approved):
Its super cool to see the app on the verge of a major level up. 🍄 

VZ team is going to 🤯!!!

editor/components/RecordTypeBadge.tsx
johnclary: Thanks for catching that! I removed the `text-dark` that was causing the problem.

database/views/materialized/vz_incidents_list_view.sql
johnclary: Sorry about that—it didn't occur to me that you would need to fully replicate again in order for this branch to work after testing the parent branch. I'll note that in the test steps.

mddilley (approved):
Came back to follow up on my last comments, and this looks so good! 🚢 🙌 😎

PR #2101, opened 2026-07-28
Fix React Complier/Nextjs v16 Linter errors

### LINKED ISSUE #25372: Upgrade VZE to NextJS v16
* Take a look at NextJS v16
* See what the upgrade path looks like for the VZE
* Ship it if it looks good!

mateoclarke:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/25372

Here are the 5 main linter errors we tripped and a little bit I learned about each:

**`react-hooks/set-state-in-effect`**: Calling `setState` inside `useEffect` (ex: syncing props to state) causes an extra render after paint. We should prefer deriving state, adjusting during render, or event handlers. ([more info](https://react.dev/learn/you-might-not-need-an-effect))

**`react-hooks/immutability`**: Mutating props, state, or other values that React treats as snapshots (ex: `obj.field = x` then `setState(obj)`). Updates should create new objects/arrays. ([more info](https://react.dev/reference/rules/components-and-hooks-must-be-pure))

**`react-hooks/purity`**: Impure work during render (`Date.now()`, `Math.random()`, etc.). Render should be a pure function of props/state so Strict Mode / Compiler can safely re-run it. ([more info](https://react.dev/reference/rules/components-and-hooks-must-be-pure))

**`react-hooks/preserve-manual-memoization`**: Your `useMemo` / `useCallback` dependency list doesn’t match what the Compiler infers (ex: dependency array list `user?.email` but the callback closes over `user`). Manual memo must stay consistent so the Compiler can preserve it. ([more info](https://react.dev/reference/eslint-plugin-react-hooks/lints/preserve-manual-memoization))

**`react-hooks/incompatible-library`**: APIs that break Compiler memoization, in our case, React Hook Form’s `watch()`. Prefer `useWatch`, which is designed for render-phase subscription. ([more info](https://react-hook-form.com/docs/usewatch))

## Testing

**URL to test:** 
https://deploy-preview-2101--vze-staging.netlify.app/editor/crashes

**Steps to test:**

I tried to identify the key functionality of the components and methods that were touched in this React Compiler lint refactor. Here's my checklist...

### Crashes list (`/crashes`)

- [ ] Table loads rows on first visit  
- [ ] Hard refresh, filters should restore and table still loads  
- [ ] Test the TableSearch input. Typing should feel snappy (no keystroke lag). The submit button should toggle from disable to active with keystrokes. Clicking submit should render results. Clearing input should reset query. Switching “Search by” column should work
- [ ] Date presets YTD / 1Y / 5Y / All refresh results. Custom date range + Apply works
- [ ] Advanced filters should work and resetting filters restores defaults and clears search. Reset hides when form is clean 
- [ ] Pagination / page size still work  
- [ ] Download icon pops modal, CSV downloads properly, modal closes. After, you should be able to reopen and download still works  
- [ ] Map view: Test the multi-feature popup next/prev. Click another group/stack of pins and the counter should reset to 1 of X  

### Navbar search

- [ ] Change search type (Case / Crash / EMS / Location). The selection should stick after refresh  
- [ ] Search a known Crash ID. It should navigate, then the input clears
- [ ] Try an Invalid ID. It should give an error. 

### Crash details

- [ ] Map marker shows at crash location. Edit/drag location. It should keep marker on top  
- [ ] Diagram: rotate / pan / zoom, the Save button should toggle from disabled to active when dirty. After save, it should persist after refresh  
- [ ] Diagram or victim photo upload: file preview appears. Cancel/close clears. Upload should refresh  

### Temp crash delete (editor/admin)

_If you don't have one, create a temp crash with something (like a Note) to transfer._

- [ ] From the banner, click Delete. In the modal, there is a typeahead search, select a target crash. Then the list shows what will transfer. Test transferring. 
- [ ] Cancel without deleting should work.
---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/components/CrashSearchTypeahead.tsx
+++ b/editor/components/CrashSearchTypeahead.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useEffect, useMemo, useState, useCallback } from "react";
+import { useMemo, useState, useCallback } from "react";
 import type { KeyboardEvent } from "react";
 import Form from "react-bootstrap/Form";
 import Spinner from "react-bootstrap/Spinner";
@@ -32,10 +32,15 @@ export default function CrashSearchTypeahead({
   const [searchInput, setSearchInput] = useState("");
   const [showDropdown, setShowDropdown] = useState(false);
   const [highlightedIndex, setHighlightedIndex] = useState(0);
+  const [prevSelected, setPrevSelected] = useState(selected);
 
-  useEffect(() => {
-    if (!selected) setSearchInput("");
-  }, [selected]);
+  // Reset the input when the parent clears the selection (no effect needed)
+  if (selected !== prevSelected) {
+    setPrevSelected(selected);
+    if (!selected) {
+      setSearchInput("");
+    }
+  }
 
   // Constrain address search by time to reduce result count
   const minCrashTimestamp = useMemo(() => {
--- a/editor/components/DeleteTemporaryCrashModal.tsx
+++ b/editor/components/DeleteTemporaryCrashModal.tsx
@@ -163,11 +163,15 @@ export default function DeleteTemporaryCrashModal({
   const tempFatalityId =
     tempFatalities.length > 0 ? tempFatalities[0].id : null;
 
-  // Detect photo existence via the API
-  const [hasPhotoToTransfer, setHasPhotoToTransfer] = useState(false);
+  // Detect photo existence via the API. Derive "no photo" when the modal is
+  // closed or there is no fatality id, so we don't need a sync setState reset.
+  const [photoCheck, setPhotoCheck] = useState<{
+    personId: number | null;
+    hasPhoto: boolean;
+  }>({ personId: null, hasPhoto: false });
+
   useEffect(() => {
     if (!showModal || !tempFatalityId) {
-      setHasPhotoToTransfer(false);
       return;
     }
     let cancelled = false;
@@ -178,9 +182,13 @@ export default function DeleteTemporaryCrashModal({
           `${process.env.NEXT_PUBLIC_CR3_API_DOMAIN}/images/person/${tempFatalityId}`,
           { headers: { Authorization: `Bearer ${token}` } }
         );
-        if (!cancelled) setHasPhotoToTransfer(res.ok);
+        if (!cancelled) {
+          setPhotoCheck({ personId: tempFatalityId, hasPhoto: res.ok });
+        }
       } catch {
-        if (!cancelled) setHasPhotoToTransfer(false);
+        if (!cancelled) {
+          setPhotoCheck({ personId: tempFatalityId, hasPhoto: false });
+        }
       }
     };
     checkPhoto();
@@ -188,6 +196,12 @@ export default function DeleteTemporaryCrashModal({
       cancelled = true;
     };
   }, [showModal, tempFatalityId, getToken]);
+
+  const hasPhotoToTransfer =
+    showModal &&
+    tempFatalityId !== null &&
+    photoCheck.personId === tempFatalityId &&
+    photoCheck.hasPhoto;
 
   const hasVictimPhotoToTransfer =
     tempFatalities.length > 0 && hasPhotoToTransfer;
@@ -288,7 +302,7 @@ export default function DeleteTemporaryCrashModal({
     editedCardFields,
     selectedTarget,
     targetRecommendation,
-    user?.email,
+    user,
     shouldTransferPhoto,
     tempFatalityId,
     targetFatalityPersonId,
--- a/editor/components/ImageUploadModal.tsx
+++ b/editor/components/ImageUploadModal.tsx
@@ -8,8 +8,8 @@ import {
   Spinner,
   CloseButton,
 } from "react-bootstrap";
-import { Dispatch, SetStateAction, useState, useEffect } from "react";
-import { useForm } from "react-hook-form";
+import { Dispatch, SetStateAction, useState, useEffect, useMemo } from "react";
+import { useForm, useWatch } from "react-hook-form";
 import { useGetToken } from "@/utils/auth";
 import AlignedLabel from "@/components/AlignedLabel";
 import { LuTrash } from "react-icons/lu";
@@ -47,7 +47,6 @@ export default function ImageUploadModal({
   recordId,
 }: ImageUploadModalProps) {
   const [isSubmitting, setIsSubmitting] = useState(false);
-  const [previewUrl, setPreviewUrl] = useState<string | null>(null);
   const [error, setError] = useState<string | null>(null);
   const [isDeleting, setIsDeleting] = useState(false);
 
@@ -58,7 +57,7 @@ export default function ImageUploadModal({
     handleSubmit,
     reset,
     formState: { errors },
-    watch,
+    control,
   } = useForm<FormData>({
     mode: "onChange",
     defaultValues: {
@@ -67,25 +66,32 @@ export default function ImageUploadModal({
     },
   });
 
-  const file = watch("file");
+  const file = useWatch({ control, name: "file" });
 
   const url = `${process.env.NEXT_PUBLIC_CR3_API_DOMAIN}/images/${imageType}/${recordId}`;
 
-  // Keeps track of file updates and errors to update preview URL
-  useEffect(() => {
-    setError(null);
+  // Create a preview object URL when a valid file is selected; revoke on change
+  const previewUrl = useMemo(() => {
     if (file && file.length > 0 && !errors.file) {
-      const url = URL.createObjectURL(file[0]);
-      setPreviewUrl(url);
-
-      // Return cleanup function that revokes this specific URL
-      return () => {
-        URL.revokeObjectURL(url);
-      };
-    } else {
-      setPreviewUrl(null);
+      return URL.createObjectURL(file[0]);
     }
+    return null;
   }, [file, errors.file]);
+
+  useEffect(() => {
+    return () => {
+      if (previewUrl) {
+        URL.revokeObjectURL(previewUrl);
+      }
+    };
+  }, [previewUrl]);
+
+  // Clear upload errors when the selected file changes
+  const [prevFile, setPrevFile] = useState(file);
+  if (file !== prevFile) {
+    setPrevFile(file);
+    setError(null);
+  }
 
   /**  Uploads the image to the API */
   const onSubmit = async (data: FormData) => {
@@ -171,7 +177,6 @@ export default function ImageUploadModal({
       size="lg"
       onHide={handleClose}
       onExited={() => {
-        setPreviewUrl(null);
         reset();
       }}
     >
--- a/editor/components/NavBarSearch.tsx
+++ b/editor/components/NavBarSearch.tsx
@@ -1,5 +1,5 @@
 import { useRouter } from "next/navigation";
-import { useEffect, useState } from "react";
+import { useEffect, useState, useSyncExternalStore } from "react";
 import Spinner from "react-bootstrap/Spinner";
 import Button from "react-bootstrap/Button";
 import Dropdown from "react-bootstrap/Dropdown";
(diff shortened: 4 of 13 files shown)


editor/components/CrashSearchTypeahead.tsx
```diff
@@ -32,10 +32,15 @@ export default function CrashSearchTypeahead({
...
-  useEffect(() => {
-    if (!selected) setSearchInput("");
-  }, [selected]);
+  // Reset the input when the parent clears the selection (no effect needed)
+  if (selected !== prevSelected) {
+    setPrevSelected(selected);
+    if (!selected) {
+      setSearchInput("");
+    }
+  }
```
johnclary, line 43: Yeah, this is great. It is weird to me to see a bare a `setState()` call inside a component render, but it's obviously fine because it's protected by the case statements. Will take some getting used to to break this habit in my React code.

editor/components/DeleteTemporaryCrashModal.tsx
```diff
@@ -163,11 +163,15 @@ export default function DeleteTemporaryCrashModal({
   const tempFatalityId =
     tempFatalities.length > 0 ? tempFatalities[0].id : null;
 
-  // Detect photo existence via the API
-  const [hasPhotoToTransfer, setHasPhotoToTransfer] = useState(false);
+  // Detect photo existence via the API. Derive "no photo" when the modal is
+  // closed or there is no fatality id, so we don't need a sync setState reset.
+  const [photoCheck, setPhotoCheck] = useState<{
```
johnclary, line 168: 👍

editor/components/ImageUploadModal.tsx
```diff
@@ -67,26 +66,33 @@ export default function ImageUploadModal({
...
+      }
+    };
+  }, [previewUrl]);
+
+  // Clear upload errors when the selected file changes
+  const [prevFile, setPrevFile] = useState(file);
+  if (file !== prevFile) {
+    setPrevFile(file);
+    setError(null);
+  }
```
johnclary, line 94: Nodding along with the changes—pairing current and prev state vars is a clean pattern ✨

editor/components/NavBarSearch.tsx
```diff
@@ -87,16 +87,31 @@ const getValidSearchField = (key: string | null): AnySearchField => {
   return foundSearchField || SEARCH_FIELDS[0];
 };
 
+const subscribeToNavSearchStorage = (onStoreChange: () => void) => {
+  window.addEventListener("storage", onStoreChange);
+  return () => window.removeEventListener("storage", onStoreChange);
```
johnclary, line 92: Big TIL for me about the `storage` event and that it listens to storage changes that happen in another tab/window. I tested changing the search field with two VZE tabs open, and they stay in sync 😵

editor/components/NavBarSearch.tsx
johnclary, line 1: Really interesting refactor here 🍿 

This feature has been a bit of a thorn in my side since its inception. The state management has always felt off. This is such a huge improvement 🚀

editor/components/TableExportModal.tsx
```diff
@@ -113,25 +112,21 @@ export default function TableExportModal<T extends Record<string, unknown>>({
...
+    });
+    const blob = new Blob([csvContent], { type: "text/csv;charset=utf-8;" });
+    const url = URL.createObjectURL(blob);
+    const anchor = document.createElement("a");
+    anchor.href = url;
+    anchor.download = formatFileName(exportFilename);
+    anchor.click();
+    URL.revokeObjectURL(url);
+    onClose();
+  };
```
johnclary, line 129: ✨

editor/components/TableSearch.tsx
```diff
@@ -19,14 +19,30 @@ export interface SearchSettings {
 }
 
 /**
- * Record search component that plugs into the query builder
+ * Record search component that plugs into the query builder.
+ *
+ * The input value is kept in local state so keystrokes do not re-render
```
johnclary, line 24: When rendering lots of rows, the search input in prod IS the tiniest bit laggy—and it bummed me out when i noticed it. The solution is so clear now 🥹

editor/components/TableWrapper.tsx
```diff
@@ -74,22 +161,86 @@ export default function TableWrapper<T extends Record<string, unknown>>({
...
+  /**
+   * Client gate only — do NOT also wait on column-visibility localStorage here.
+   * TablePaginationControls must mount to load column visibility; blocking on
+   * that flag prevents the query from ever firing (blank /crashes page).
+   */
+  const isClient = useSyncExternalStore(
+    () => () => {},
+    () => true,
+    () => false
+  );
```
johnclary, line 173: a TIL, obviously, to use this hook to gate client-only stuff 🍿

editor/components/TableWrapper.tsx
johnclary, line 1: This is a chonky refactor—and really highlights how much trouble I've had wiring up the local storage in react. Super cool to see this majorly upgraded 🙌

editor/eslint.config.mjs
```diff
@@ -13,13 +13,12 @@ const eslintConfig = defineConfig([
...
-      "react-hooks/preserve-manual-memoization": "warn",
-      "react-hooks/purity": "warn",
-      "react-hooks/immutability": "warn",
-      "react-hooks/incompatible-library": "warn",
+      // React Compiler rules from eslint-config-next 16
+      "react-hooks/set-state-in-effect": "error",
+      "react-hooks/preserve-manual-memoization": "error",
+      "react-hooks/purity": "error",
+      "react-hooks/immutability": "error",
+      "react-hooks/incompatible-library": "error",
```
johnclary, line 22: 🤩

editor/components/TableWrapper.tsx
```diff
@@ -74,22 +161,86 @@ export default function TableWrapper<T extends Record<string, unknown>>({
...
+
+  const rawStoredConfig = useSyncExternalStore(
+    (onStoreChange) => {
+      window.addEventListener("storage", onStoreChange);
+      return () => window.removeEventListener("storage", onStoreChange);
+    },
+    () => localStorage.getItem(localStorageKey),
+    () => null
+  );
```
johnclary, line 182: Similar to the top nav behavior, this causes search settings (page size, filter toggles, map view—everything) to sync across tabs which are looking at the same list view—but only if those tabs are open to the default view and have not been interacted with at all. It's pretty neat, and I think totally harmless.

johnclary (approved):
Dang, what a huge upgrade to our codebase! So many TILs in here for me. Thanks for laying down the patterns we can reach for as we continue dev'ing in this new and exciting React world. 

These rules are a bit daunting, but it's clear from the performance improvements you're bringing into this PR that the rules are going to guide us to write better code. I've tested the heck out of this and feel that this is ready to bring forward 🚀 🚢 🚢 🚢 🚢

editor/components/CrashSearchTypeahead.tsx
mddilley: +1 def feels very new to me! Love that there is still more to learn about useEffect usage! 🙏 

side note, i noticed that a note about [React Compiler](https://react.dev/learn/react-compiler) is added to the ["You Might Not Need an Effect" page](https://react.dev/learn/you-might-not-need-an-effect) which is another new thing that i've yet to look into.

editor/eslint.config.mjs
```diff
@@ -13,13 +13,12 @@ const eslintConfig = defineConfig([
...
       "no-unused-vars": "off",
       "@typescript-eslint/no-unused-vars": "error",
-      // React Compiler rules newly enabled by eslint-config-next 16.
-      // Warn for now; fix and promote to error in a follow-up.
-      "react-hooks/set-state-in-effect": "warn",
-      "react-hooks/preserve-manual-memoization": "warn",
-      "react-hooks/purity": "warn",
-      "react-hooks/immutability": "warn",
-      "react-hooks/incompatible-library": "warn",
+      // React Compiler rules from eslint-config-next 16
```
mddilley: i'm just realizing that, when I reviewed the last PR, I was conflating [React Compiler](https://react.dev/learn/react-compiler) (the new auto-memo React library) with the ordinary stuff that React does under the hood. I wanted to check if I'm understanding that the `react-hooks/preserve-manual-memoization` and `react-hooks/incompatible-library` rules are for when/if we [switch that on in the Next.js config](https://nextjs.org/docs/app/api-reference/config/next-config-js/reactCompiler).

if so, it could be nice to group those rules and add a comment for the future. either way, just another way that Next.js is making me feel stuck in my ways with them pushing new and exciting stuff. 😄

mddilley (approved):
I ran through all the test steps, and I couldn't find any regressions. There is a ton to learn in this PR! Thanks for taking on these refactors, and it was fun taking the time to digest the changes. I'll definitely be thinking about these patterns the next time that I reach for a useEffect. 🙏 🙌 🚢

roseeichelmann:
Is anyone else dealing with being redirected to the login page after every manual refresh?

roseeichelmann:
Im getting a whole bunch of these tile-loading NextJS errors in the console about how Mapbox GL JS is interacting with Next.js's server-side rendering. [Here is my convo](https://chat.deepseek.com/share/ai31yi9d28n26qslwo) with the bot about it. But it seems like NextJS v16 has changed how it handles SSR and now we need to add 'use client' to the top of our map component file

roseeichelmann (commented):
I combed through the app and couldnt find anything broken, you did so many thorough refactors in this PR and it looks so great!! Also your test steps have me reflecting on how much the app has leveled up in the past couple of years and the really cool features we have implemented

Im curious if adding the "use client" directive at the top of the parent map file will fix all the tile errors im getting in the console on the crash details/fatality details page, and curious if others are seeing these errors too.

editor/eslint.config.mjs
mateoclarke: Great question Mike. I hadn't connected the dots that we'd need to add `reactComplier: true,` to the [next config](https://github.com/cityofaustin/vision-zero/blob/mateo%2F25372_nextjs_v16_lint_fixes/editor/next.config.mjs#L9). Should we enable that?

mateoclarke:
> Is anyone else dealing with being redirected to the login page after every manual refresh?

@roseeichelmann I was able to log out and back in with no issues. Should I try something else to reproduce?

mateoclarke:
> Im getting a whole bunch of these tile-loading NextJS errors in the console about how Mapbox GL JS is interacting with Next.js's server-side rendering. [Here is my convo](https://chat.deepseek.com/share/ai31yi9d28n26qslwo) with the bot about it. But it seems like NextJS v16 has changed how it handles SSR and now we need to add 'use client' to the top of our map component file

@roseeichelmann I'm not seeing these errors in local but I am seeing something similar in staging. In my console, I see 401 Unauthorized errors related to nearmap tiles. Seem like our env vars in staging are out of sync with 3rd party pws? See also "Network" tab. Maybe we check these credentials during launch party tomorrow? 🤷

roseeichelmann:
@mateoclarke That makes sense if its our staging env vars being out of sync. I think checking in during our launch today is a good idea. 

The issue with being redirected to the login page was happening to me when i was refreshing my page on the crashes list, and i was able to get it to happen 4x. But now i cant reproduce it at all so :shrug:

roseeichelmann (approved):
Sorry for the noise, this looks great to ship and we can just check on those env vars during the launch party!

johnclary:
> That makes sense if its our staging env vars being out of sync. I think checking in during our launch today is a good idea.

Confirming that it is expected behavior that nearmap tiles won't load in deploy previews. It would be a nice enhancement to disable the layer entirely in previews so that we don't have those busted API calls screaming at us. 

> The issue with being redirected to the login page was happening to me when i was refreshing my page on the crashes list, and i was able to get it to happen 4x. But now i cant reproduce it at all so 🤷

Please do let me know if/when you encounter this. I am always concerned that our auth refresh token flow is somehow misconfigured 😅

PR #2104, opened 2026-08-03
Add ARIA labels to chart elements

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/29454

## Testing

**URL to test:** 

**Steps to test:**
Testing is just to check out the aria-labels i added to the chart elements :+1: 

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/viewer/src/views/nav/SideMapTimeOfDayChart.jsx
+++ b/viewer/src/views/nav/SideMapTimeOfDayChart.jsx
@@ -163,7 +163,7 @@ export const SideMapTimeOfDayChart = ({ filters }) => {
           ref={(ref) => (chartRef.current = ref)}
           data={data}
           height={250}
-          aria-label="TODO"
+          aria-label="Horizontal bar chart showing crash distribution by time of day. Click any bar to filter the map to that time range"
           onClick={handleBarClick}
           options={{
             indexAxis: "y",
--- a/viewer/src/views/nav/SideMapTimeOfDayChart.jsx
+++ b/viewer/src/views/nav/SideMapTimeOfDayChart.jsx
@@ -163,7 +163,7 @@ export const SideMapTimeOfDayChart = ({ filters }) => {
           ref={(ref) => (chartRef.current = ref)}
           data={data}
           height={250}
-          aria-label="Horizontal bar chart showing crash distribution by time of day. Click any bar to filter the map to that time range"
+          aria-label="Horizontal bar chart showing crash distribution by time of day"
           onClick={handleBarClick}
           options={{
             indexAxis: "y",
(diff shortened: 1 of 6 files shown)


viewer/src/views/nav/SideMapTimeOfDayChart.jsx
```diff
@@ -163,7 +163,7 @@ export const SideMapTimeOfDayChart = ({ filters }) => {
           ref={(ref) => (chartRef.current = ref)}
           data={data}
           height={250}
-          aria-label="TODO"
+          aria-label="Horizontal bar chart showing crash distribution by time of day. Click any bar to filter the map to that time range"
```
mddilley: Maybe the instructions push this beyond a label? I'm not well-versed in best practices so I looked at MDN and saw this blurb on the [aria-description](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-description) page linked from the [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label) page that could help guide us:

> The aria-description attribute is similar to [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label) in that both provide a text string to associate with the element, but a label should be short and concise, while the description can be longer as it is intended to provide more context and information.

mddilley (approved):
Looks good! Thanks for making this improvement! 🙏 🚀

frankhereford (approved):
I think that Mike is on to something with his comment about `aria-description` vs `aria-label`, but I also think this is a clear improvement and is 100% shippable. Thank you! 🚢🚢🚢

viewer/src/views/nav/SideMapTimeOfDayChart.jsx
mateoclarke: Love this a11y in the weeds feedback. Good to be precise with this stuff, especially on public-facing sites.

mateoclarke (approved):
Love when `TODO`s are red in the diff 🚀

viewer/src/views/nav/SideMapTimeOfDayChart.jsx
johnclary: TIL aria-description. 

I think "Horizontal bar chart showing crash distribution by time of day" is a great label for this chart. `Click any bar to filter the map to that time range` can go in `aria-description`, or just remove it since we're not putting that kind of instruction on our other charts 🤷

johnclary (approved):
Thanks, Rose!

PR #2108, opened 2026-08-05
Shows all incident member records on details map

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/29677

This PR updates the incident details map to show all of the member records with their agency icon markers. This ended up being quite a lot of code to abstract how we compute the initial map bounds. It leaves us ready to handle other arbitrary feature collections on maps, and I can already think of a few use cases (historical crashes on the fatality details page, showing the location polygon on the crash details page...).

## Testing

**URL to test:** Local

1. Start your stack and apply migrations and metadata.
2. Just as you did in https://github.com/cityofaustin/vision-zero/pull/2093, you'll want to browse some incidents that have multiple records by different agencies.
3. On the incident details page, confirm that once the map appears on the screen, it immediately updates to fit all map markers into view. 
4. Observe that each of the incident's records are represented on the map with a marker, with the caveat that it's common for markers to be exactly overlapping and thus obscured. (We have https://github.com/cityofaustin/atd-data-tech/issues/29683 to help with this problem).
6. Drag/zoom the map away from the default view and use the navigation control in the top-right corner of the map to reset the map extend to inlcude all features. (I think we should change the icon of this control button to a house/home)
7. Nice! Now head to a crash details page and confirm the map loads normally with the map centered on the crash location (indicated by a blue map pin). Edit the crash location, save, and confirm that map extent updates to center the updated crash location.
8. Head to an EMS incident details page, and confirm the map loads normally and is centered on the location of the EMS incident (indicated by a red ambulance marker)

---

#### Ship list

- [ ] Code reviewed


editor/declarations.d.ts
johnclary, line 1: Fixes the typescript linter being unhappy with css imports.

editor/components/CrashMapCard.tsx
```diff
@@ -52,7 +52,7 @@ export default function CrashMapCard({
   });
 
   const [isEditing, setIsEditing] = useState(false);
-  const [mapLatLon, setMapLatLon] = useState<LatLon>({
+  const [draftLatLon, setDraftLatLon] = useState<LatLon>({
```
johnclary, line 55: renamed this for clarity

editor/components/IncidentMapCard.tsx
```diff
@@ -22,14 +61,19 @@ export default function IncidentMapCard({ incident }: IncidentMapCardProps) {
...
+
   return (
     <Card className="h-100">
       <Card.Body className="p-1 crash-header-card-body" ref={mapContainerRef}>
         <PointMap
-          savedLatitude={incident?.latitude || 0}
-          savedLongitude={incident?.longitude || 0}
+          autoFitBounds
+          savedLatitude={null}
+          savedLongitude={null}
```
johnclary: Unlike all of our other details map, this one does not take a primary saved lat/lon that is used as the focal point of the map. Instead we will auto-fit the map bounds based on all features we render.

editor/components/IncidentMapMarker.tsx
```diff
@@ -0,0 +1,51 @@
...
+ */
+export default function IncidentMapMarker({
+  id,
+  name,
+  latitude,
+  longitude,
+  ...props
+}: IncidentMapMarkerProps) {
+  // register each marker in the map context to enable autoFitBounds
+  useRegisterMapFeature(id, { latitude, longitude });
```
johnclary, line 24: This new fancy context helper enables to keep track of features that we add to the map, so that we can interact with them later (specifically when we want to auto-fit the map bounds).

editor/components/PointMap.tsx
```diff
@@ -107,48 +119,43 @@ export const PointMap = ({
...
   savedLongitude,
   isEditing,
-  mapLatLon,
-  setMapLatLon,
+  draftLatLon,
+  setDraftLatLon,
   CustomMarker,
   children,
   customLayerToggles,
+  autoFitBounds = true,
```
johnclary, line 125: All of this refactoring is in service of being able to fit the map bounds to multiple arbitrary features, rather than a single point. It's backwards compatible.

Without this context-based approach, we'd have to rig up some state that we pass as a prop to this component which contains an array of all of the geojson features we want to include in our bounds calculations. That would be redundant and pretty gnarly because this component is abstracted to accept the map features as `{children}` — and the children are react nodes, not raw geojson data.

Now that we have this in place, we should separately be able to:

- Enhance the EMS details map to use the new `useRegisterMapFeature()` hook to make use of auto-fitting there. That map currently displays nearby CR3 and Non-CR3 crashes that often fall outside the map's initial viewport
- Deprecate the `LocationMap` component, which is quite redundant to this component and exists mostly because it needed to show a polygon instead of a point as its central feature 😅

editor/components/PointMap.tsx
```diff
@@ -162,6 +169,91 @@ export const PointMap = ({
...
+  /**
+   * This callback will update registeredFeatures whenever
+   * a new feature is registered in map registry context
+   */
+  const onFeaturesChange = useCallback(
+    (features: Map<string, Geometry | Geometry[]>) => {
+      setRegisteredFeatures(features);
+    },
+    []
+  );
```
johnclary, line 195: ☝️ These two declarations are everything we need to enable our feature registry context to do it's thing. Any child component that registers itself will land in our `registeredFeatures` object .

editor/components/PointMap.tsx
```diff
@@ -162,6 +169,91 @@ export const PointMap = ({
...
+   * Construct a FeatureCollection from the feature registry
+   */
+  const allFeatures = useMemo(() => {
+    const feats: Feature<Geometry>[] = [];
+
+    if (savedLatitude != null && savedLongitude != null) {
+      feats.push(point([savedLongitude, savedLatitude]));
+    }
+
+    registeredFeatures.forEach((geom) => {
```
johnclary, line 209: `registeredFeatures` is a [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map), which looks like an `Object` but has some neat features, like `forEach` ✨

editor/components/PointMap.tsx
```diff
@@ -162,6 +169,91 @@ export const PointMap = ({
...
+      if (Array.isArray(geom)) {
+        geom.forEach((g) => feats.push(feature(g)));
+      } else {
+        feats.push(feature(geom));
+      }
+    });
+
+    return featureCollection(feats);
```
johnclary, line 217: these `turf` helpers (`featureCollection`, `point`, `feature`) are nice shortcuts to having to write geojson by hand.

editor/package.json
```diff
@@ -18,6 +18,7 @@
     "@mapbox/mapbox-gl-geocoder": "^5.1.2",
     "@turf/bbox": "^7.2.0",
     "@turf/center": "^7.3.5",
+    "@turf/helpers": "^7.3.5",
```
johnclary, line 21: These were already bundled in our app as a dependency for the other turf packages we use—there's no bundle hit afaik 😇

editor/contexts/MapFeatureRegistry.tsx
johnclary, line 1: Here's the context provider and helpers. This code is pretty challenging to follow, especially with the typescript markup. Hopefully stepping through the `PointMap` and the `IncidentMapMarker` components helps.

editor/utils/constants.ts
```diff
@@ -7,6 +7,7 @@ export const COLORS = {
   primary: "#1276d1",
   secondary: "#6b7676",
   warning: "#ffd22f",
+  afd_orange: "#fdae61",
```
johnclary, line 10: yeah—the map marker colors need some work, and i don't feel great about stashing this color here 🙈

editor/tsconfig.json
```diff
@@ -30,6 +30,7 @@
   },
   "include": [
     "next-env.d.ts",
+    "declarations.d.ts",
```
johnclary, line 33: Fixes the typescript linter being unhappy with css imports.

editor/components/PointMap.tsx
```diff
@@ -162,6 +167,91 @@ export const PointMap = ({
...
+      map.fitBounds(totalBounds, {
+        padding: 60,
+        maxZoom: DEFAULT_MAP_PAN_ZOOM.zoom,
+        duration: 0,
+      });
+      hasFitRef.current = true;
+    };
+
+    if (map.isStyleLoaded()) fit();
+    else map.once("data", fit);
```
johnclary, line 259: the "data" load event seems to be working well in preventing the map from initializing (over the capitol building) and then jumping to incident location a few milliseconds after. I'm not seeing any jumping on map load. Are you??

editor/configs/crashesColumns.tsx
```diff
@@ -391,4 +392,4 @@ export const vzIncidentDetailscrashColumns: ColDataCardDef<Crash>[] = [
   crashesColumns.private_dr_fl,
   crashesColumns.in_austin_full_purpose,
   crashesColumns.location_id,
-];
+]);
```
johnclary, line 413: I added this `makeReadOnly` helper to deal with the awkward dance of wanting to reuse column definition with some minor changes (ie disabling editing).

editor/components/IncidentMapCard.tsx
```diff
@@ -4,6 +4,45 @@ import { MapRef } from "react-map-gl";
...
+/**
+ * Hook which returns a map marker for each record that the incident
+ * is comprised of
+ */
+const useIncidentMarkers = (incident: VzIncidentListRow) => {
+  const markers = incident?.vz_incident_records_view
+    ?.filter((record) => record.geom)
+    .map((record, i) => {
+      /**
+       * Defaut to the "crashes" marker and use the appropriate
```
frankhereford, line 19: typo in 'default'

editor/components/IncidentMapCard.tsx
```diff
@@ -4,6 +4,45 @@ import { MapRef } from "react-map-gl";
...
+import IncidentMapMarker from "@/components/IncidentMapMarker";
+import { RECORD_TYPE_BADGES } from "@/components/RecordTypeBadge";
+
+/**
+ * Hook which returns a map marker for each record that the incident
+ * is comprised of
+ */
+const useIncidentMarkers = (incident: VzIncidentListRow) => {
+  const markers = incident?.vz_incident_records_view
+    ?.filter((record) => record.geom)
```
frankhereford, line 16: Nice `filter` here -- I was about to leave a comment about null island at 0°, 0° below, but this catches the missing geometry case.

editor/components/PointMap.tsx
frankhereford: Nope - no capitol building for me. Looks great - loading behavior never crossed my mind.

editor/components/PointMap.tsx
```diff
@@ -32,6 +34,9 @@ import { COLORS } from "@/utils/constants";
 import { z, ZodFormattedError } from "zod";
 import "mapbox-gl/dist/mapbox-gl.css";
 import { MapAerialSourceAndLayer } from "@/components/MapAerialSourceAndLayer";
+import { MapFeatureRegistryProvider } from "@/contexts/MapFeatureRegistry";
```
frankhereford, line 37: This mechanism is ✨ -- what an elegant solution!

editor/contexts/MapFeatureRegistry.tsx
```diff
@@ -0,0 +1,165 @@
...
+ * time a child registers, unregisters, or changes its geometry,
+ * `onFeaturesChange` fires with a fresh snapshot `Map` of everything
+ * currently registered (id -> geometry).
+ *
+ * Registrations are tracked in a ref (not state) internally, so the
+ * provider itself never re-renders on registration changes — only the
+ * `onFeaturesChange` callback fires, letting the parent decide how to
+ * respond (e.g. store it in its own state to trigger a bounds refit).
+ *
+ * `onFeaturesChange` shold be a function with a stable reference that
```
frankhereford, line 60: typo in 'should'

editor/contexts/MapFeatureRegistry.tsx
```diff
@@ -0,0 +1,165 @@
...
+      type: "Point",
+      coordinates: [latLon.longitude, latLon.latitude],
+    });
+    return () => ctx.unregister(id);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [ctx, id, latLon?.latitude, latLon?.longitude]);
+}
+
+/**
+ * Register a set of GeoJSON features. Pass a stable id uniqe to this
```
frankhereford, line 148: typo in 'unique'

editor/contexts/MapFeatureRegistry.tsx
frankhereford: Your comments / typescript documentation help a ton. I think your explanations and example are a great example of what makes typescript hum and they explain what's going on here really well.

frankhereford (approved):
This is really top-shelf work. It works just like one would expect and it was fun to follow along in your code. My spell-checker picked up a few typos, but in terms of impactful comments -- I'm coming up empty on what to suggest for improvement. I think this is eminently shippable. 

To your suggestion about making the 'zoom to bounding box' button into a 🏠 icon -- I think you're probably right, although my personal preference is the dot inside the square. It's currently showing just so exactly what it does, but I get that that might be a very GIS specific observation.

editor/components/MapFitBoundsControl.tsx
mddilley, line 1: totally unrelated to these changes, but it jumped out to me this time around that this control is a different height and width than the other Mapbox controls. 

this might be intentional and i don't have a strong opinion on this at all but could be something to bundle with the house/home icon change that you mentioned in the test steps which i think sounds like a good change to me!

editor/components/PointMap.tsx
mddilley: +1 i didn't see any jumping on map load or flash of the capitol building. This part is always so tricky, and I love seeing mapbox instance methods come in handy! 🙌

editor/contexts/MapFeatureRegistry.tsx
```diff
@@ -0,0 +1,165 @@
...
+) {
+  const ctx = useContext(MapFeatureContext);
+
+  useEffect(() => {
+    if (!ctx || !latLon) return;
+
+    ctx.register(id, {
+      type: "Point",
+      coordinates: [latLon.longitude, latLon.latitude],
+    });
```
mddilley, line 141: made me think about the `point` turf helper used in PointMap - could that be used here? either way, those helpers are TILs for me and definitely made me think about possible usage in Moped where we do this stuff by hand. 🍄

editor/contexts/MapFeatureRegistry.tsx
mddilley: i agree! the comments help a lot, and it was a fun way to think through navigating state vs. refs, object references, and unintended re-renders that can be caused by contexts. good stuff here!

mddilley (approved):
I followed all the test steps, and I couldn't find any regressions in the modified fit bounds control. Very cool to see the next part of the incidents UI fall into place, and I'm excited to see the next steps too. Some fun React code in here too with the context/ref! 🚢 🙌

editor/components/IncidentMapCard.tsx
johnclary: TIL "null island" - just a perfect term for this

editor/components/MapFitBoundsControl.tsx
johnclary: Good call. Yes!

editor/contexts/MapFeatureRegistry.tsx
johnclary: Had the same thought and then realized that the [`point`](https://turfjs.org/docs/api/point) helper constructs a `Feature` but what we're looking for here is a `Geometry`. 

There is a tension in this new registry and the pointmap between `Feature` and `Geometry` that does give me pause. It might be cleaner to work with `Feature` objects across the board since we will almost always be rendering and interacting with Features. Will think about this some more.

Update: see https://github.com/cityofaustin/vision-zero/pull/2116/changes#r3812790817.

editor/components/IncidentMapCard.tsx
johnclary: If the typos disappear you'll know I've been take over by a bot 😅

johnclary:
Thanks for your reviews @frankhereford and @mddilley. I have withdrawn other review requests and migrated the context-based feature registration over to https://github.com/cityofaustin/vision-zero/pull/2115. 

This PR stack is going to close without merge in the near future - [slack discussion](https://austininnovation.slack.com/archives/CM9MK950S/p1786993270013719?thread_ts=1786984599.687119&cid=CM9MK950S).

PR #2115, opened 2026-08-18
Adds location polygon to crash details map

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/28468

This PR adds the crash location polygon to the crash details map. 

This work was motivated by my desire to port over the `autoFitBounds` map improvements that I implemented in https://github.com/cityofaustin/vision-zero/pull/2108. I copied the code and comments over from that PR where relevant (you do not need to review that PR).

Screenshot 2026-08-18 at 11 11 16 AM

Screenshot 2026-08-18 at 11 11 21 AM

Screenshot 2026-08-18 at 11 11 30 AM

## Testing

**URL to test:** [Netlify](https://deploy-preview-2115--vze-staging.netlify.app/)

1. If local, you'll need to run `nvm use && npm install` before staring the VZE
2. Login as an editor or admin. Visit a few crash details pages, and you should see a location polygon on the map whenever the crash has a polygon associated with it.
4. Repeatedly edit a crash's location and save to test out moving a point into and out of a polygon. The polygon should update on the map whenever the crash's associated location ID changes.
5. Try refreshing the page a few times and confirm that the map renders with with the crash marker and location polygon fully within the map extent. 
6. Check out dark mode/light mode using the **Streets** basemap and confirm that the polygon outline contrast is ok
7. Visit the location details page for any location and confirm that the location loads normally and the map looks fine. Note that the polygon will not auto-fit to the map bounds on this page. [Separate work](https://github.com/cityofaustin/atd-data-tech/issues/29852) is required for that.

Other enhancements that would be nice to have (issues incoming):

- A legend or label on the polygon that indicates what it is
- Ability to see nearby polygons while editing the crash location
- Ability to see the **Location group ID** in the map card header: this info hints at whether the crash wants to intersect with level 5 or non-level 5 polygons
- Other things?

---

#### Ship list

- [ ] Code reviewed


editor/components/CrashMapCard.tsx
```diff
@@ -52,7 +56,7 @@ export default function CrashMapCard({
   });
 
   const [isEditing, setIsEditing] = useState(false);
-  const [mapLatLon, setMapLatLon] = useState<LatLon>({
+  const [draftLatLon, setDraftLatLon] = useState<LatLon>({
```
johnclary, line 59: renamed this for clarity

editor/components/LocationPolygonLayer.tsx
```diff
@@ -0,0 +1,77 @@
...
+ */
+export default function LocationPolygonLayer({
+  location,
+}: LocationPolygonLayerProps) {
+  const polygonFeature = useLocationFeature(location);
+
+  useRegisterMapFeatures(
+    "locationPolygon",
+    polygonFeature ? [polygonFeature] : null
+  );
```
johnclary: This new fancy context helper enables us to keep track of features that we add to the map, so that we can interact with them later (specifically when we want to auto-fit features into the map bounds).

editor/components/PointMap.tsx
```diff
@@ -96,6 +101,11 @@ interface PointMapProps {
    * Configs for adding custom layer toggles to the basemap control
    */
   customLayerToggles?: CustomLayerToggle[];
+  /**
+   * If true, fit the map to every registered feature (saved point +
+   * anything children report via useRegisterMapFeature). Default: true
+   */
+  autoFitBounds?: boolean;
```
johnclary, line 108: All of this refactoring is in service of being able to fit the map bounds to multiple arbitrary features, rather than a single point. It's backwards compatible.

Without this context-based approach, we'd have to rig up some state that we pass as a prop to this component which contains an array of all of the geojson features we want to include in our bounds calculations. That would be redundant and pretty gnarly because this component is abstracted to accept the map features as `{children}` — and the children are react nodes, not raw geojson data.

Now that we have this in place, we should separately be able to:

- Enhance the EMS details map to use the new `useRegisterMapFeature()` hook to make use of auto-fitting there. That map currently displays nearby CR3 and Non-CR3 crashes that often fall outside the map's initial viewport
- Enhance the crash details map to show CAD or EMS incidents

editor/components/LocationMap.tsx
johnclary, line 1: The location details map does not auto-fit the location polygon, because it needs to be updated with the same behavior as `PointMap`.

But we should actually be able to fully replace this component with `PointMap`. I left that out of scope and just moved helper utils out of here in favor of the new `LocationPolygonLayer` component.

editor/components/PointMap.tsx
```diff
@@ -169,6 +174,91 @@ export const PointMap = ({
...
+  /**
+   * This callback will update registeredFeatures whenever
+   * a new feature is registered in map registry context
+   */
+  const onFeaturesChange = useCallback(
+    (features: Map<string, Geometry | Geometry[]>) => {
+      setRegisteredFeatures(features);
+    },
+    []
+  );
```
johnclary, line 195: ☝️ These two declarations are everything we need to enable our feature registry context to do it's thing. Any child component that registers itself will land in our `registeredFeatures` object .

editor/components/PointMap.tsx
```diff
@@ -169,6 +174,91 @@ export const PointMap = ({
...
+   * Construct a FeatureCollection from the feature registry
+   */
+  const allFeatures = useMemo(() => {
+    const feats: Feature<Geometry>[] = [];
+
+    if (savedLatitude != null && savedLongitude != null) {
+      feats.push(point([savedLongitude, savedLatitude]));
+    }
+
+    registeredFeatures.forEach((geom) => {
```
johnclary: `registeredFeatures` is a [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map), which looks like an `Object` but has some neat features, like `forEach` ✨

editor/components/PointMap.tsx
```diff
@@ -169,6 +174,91 @@ export const PointMap = ({
...
+        padding: 60,
+        maxZoom: DEFAULT_MAP_PAN_ZOOM.zoom,
+        duration: 0,
+      });
+      hasFitRef.current = true;
+    };
+
+    if (map.isStyleLoaded()) fit();
+    else map.once("data", fit);
+  }, [autoFitBounds, mapRef, allFeatures, totalBounds]);
```
johnclary, line 260: the "data" load event seems to be working well in preventing the map from initializing (over the capitol building) and then jumping to incident location a few milliseconds after. I'm not seeing any jumping on map load. Are you??

[Other comments](https://github.com/cityofaustin/vision-zero/pull/2108/changes#r3725343537)

editor/contexts/MapFeatureRegistry.tsx
johnclary, line 1: https://github.com/cityofaustin/vision-zero/pull/2108/changes#r3724473114

editor/contexts/MapFeatureRegistry.tsx
```diff
@@ -0,0 +1,166 @@
...
+) {
+  const ctx = useContext(MapFeatureContext);
+
+  useEffect(() => {
+    if (!ctx || !latLon) return;
+
+    ctx.register(id, {
+      type: "Point",
+      coordinates: [latLon.longitude, latLon.latitude],
+    });
```
johnclary: https://github.com/cityofaustin/vision-zero/pull/2108/changes#r3754021747

editor/components/LocationPolygonLayer.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+const polygonLayerFill: FillLayerSpecification = {
+  id: "location-polygon-fill",
+  source: "location-polygon",
+  type: "fill",
+  paint: {
+    "fill-color": "orange",
+    "fill-opacity": 0.1,
+  },
+};
```
johnclary, line 42: I added a faint black outline underneath the orange line — it helps quite a bit with contrast. See comparison below.

I also added a faint transparent fill which I'm on the fence about. I think it helps convey that the location is a surface that intersects with crash points 🤷 

**This PR**

Image

**Prod**

Image

editor/components/PointMap.tsx
chiaberry: no jumping here!

editor/components/LocationPolygonLayer.tsx
chiaberry: i like the faint transparent fill

chiaberry (approved):
Tested and reviewed the code (and some of the old comments on the other PR) and its great!  🚢

editor/components/LocationPolygonLayer.tsx
roseeichelmann: i like the design a lot!

roseeichelmann: i like the transparent fill too

editor/components/PointMap.tsx
roseeichelmann: no jumping here either

roseeichelmann (approved):
tested, this looks great! im super excited about this feature. i think the design looks great. so cool to update the location point and see the new polygon get outlined.

editor/components/LocationPolygonLayer.tsx
mddilley: +1 i reviewed PRs in the wrong order, and i was wondering if my eyes were playing tricks on me! I agree that the fill ties it all together ✨

editor/components/LocationPolygonLayer.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { FillLayerSpecification, LineLayerSpecification } from "mapbox-gl";
+import { Feature, GeoJsonProperties, MultiPolygon } from "geojson";
+import { feature } from "@turf/helpers";
+import { COLORS } from "@/utils/constants";
+
+interface LocationPolygonLayerProps {
+  location: Location;
+}
+
+const polygonLayerLine: LineLayerSpecification = {
```
mddilley, line 14: it really is so nice to have the mapbox types as a reference for these style specs because i always find their shapes to be a bit tricky 🚀

mddilley (approved):
Looking great! I reviewed the new changes in here since the last PR, and I'm really glad to see the feature registry code still make it into the app. The locations are a great addition to the map on the crash details page. 🚢 🚀

editor/components/LocationPolygonLayer.tsx
johnclary: Cool, thanks y'all! Xavier also said he likes the fill.

PR #2118, opened 2026-08-19
Restrict crash report PDF access to editor and admin users  

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/29773

Screenshot 2026-08-19 at 12 47 58 PM

## Testing

**URL to test:** Local

1. In the `./api` directory, you might want to update your `AWS_S3_BUCKET_ENV` var to `"staging"`. Alternatively, you'll need to test with older crashes (+3 years) because we should have crash reports for those in the dev bucket.
2. Start your local DB + Hasura + Flask API: `./vision-zero local-stack-up` will take care of that for you. See the `./api` readme if you've never run the Flask API locally.
3.  Run the API tests by following the instructions in `/api/tests/readme.md`. If you've previously set up your env for testing, it is as easy as running this command from the `./api` directory:

```shell
# from ./api
docker compose -f docker-compose-tests.yaml run --rm cr3-user-api-test
```

4. Head to the `./editor` directory and make sure your `.env.local` file has you correctly pointed to the local flask API:

```shell
# double-check this env var
NEXT_PUBLIC_CR3_API_DOMAIN=http://localhost:8085
```

5.. Start your VZE, and login as using the read only creds in 1pass: **Vision Zero Editor (VZE) Test User: Read-only viewer**

6. Visit a few crash details pages for crashes that are more than 2 months old and confirm you cannot open any crash report PDFs. The **Open crash report button** should be disabled, and hovering over the button reveals a tooltip with a helpful message.

7. Log out and back in to the VZE with editor or admin credentials. Confirm that you are able to open crash reports normally.

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/components/CrashNarrativeCard.tsx
+++ b/editor/components/CrashNarrativeCard.tsx
@@ -3,7 +3,11 @@ import Button from "react-bootstrap/Button";
 import { FaFilePdf } from "react-icons/fa6";
 import AlignedLabel from "@/components/AlignedLabel";
 import { Crash } from "@/types/crashes";
-import { useGetToken } from "@/utils/auth";
+import { hasRole, useGetToken } from "@/utils/auth";
+import { useAuth0 } from "@auth0/auth0-react";
+import { OverlayTrigger, Tooltip } from "react-bootstrap";
+
+const allowedCrashReportDownloadRoles = ["editor", "admin"];
 
 // Downloads pdf from the CR3 API and opens it in a new tab
 export const onDownloadCR3 = async ({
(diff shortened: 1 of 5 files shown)


editor/components/CrashNarrativeCard.tsx
```diff
@@ -3,10 +3,14 @@ import Button from "react-bootstrap/Button";
...
 import AlignedLabel from "@/components/AlignedLabel";
 import { Crash } from "@/types/crashes";
-import { useGetToken } from "@/utils/auth";
+import { hasRole, useGetToken } from "@/utils/auth";
+import { useAuth0 } from "@auth0/auth0-react";
+import { OverlayTrigger, Tooltip } from "react-bootstrap";
 
-// Downloads pdf from the CR3 API and opens it in a new tab
-export const onDownloadCR3 = async ({
+const allowedCrashReportDownloadRoles = ["editor", "admin"];
```
mddilley: @johnclary I think this needs to be changed to `vz-admin` since I was seeing a disabled button when testing the admin role like Rose was too

mddilley: I'm also noticing that I can't download crash reports using `AWS_S3_BUCKET_ENV="dev"` in my local api environment (using `"prod"` does work and downloads the matching report by crash id). This might be the way it has always been so ignore if so!

editor/components/CrashNarrativeCard.tsx
```diff
@@ -44,22 +48,39 @@ export const onDownloadCR3 = async ({
...
         <Card.Title>Narrative</Card.Title>
-        <Button
-          size="sm"
-          onClick={() => onDownloadCR3({ crash, getToken })}
-          disabled={!isCr3Stored}
+        <OverlayTrigger
+          placement="top"
+          overlay={
+            <Tooltip id="copy-address-tooltip" hidden={canDownloadCrashReport}>
+              Vision Zero team members can access crash reports - contact them for assistance
```
chiaberry: This may stray into being too pedantic but I really want this to say "Only Vision Zero team members can access crash reports - contact them for assistance"

editor/components/CrashNarrativeCard.tsx
johnclary: The issue with the `dev` bucket is that we don't have recent crash reports. So you're experiencing a de-sync of what your db replica thinks exists in S3 versus what is actually there.

Setting this env var to `"staging"` would be better, because you do run the risk of mutating prod otherwise. I'll update the test steps 🙏

editor/components/CrashNarrativeCard.tsx
roseeichelmann: I like chias wording here too

editor/components/CrashNarrativeCard.tsx
mddilley: ahhh ok. i just updated my env file to use staging for next time around. thank you!

chiaberry (approved):
tested again! got the 42 passing tests in the api, logged in as a read only user and couldnt click the crash report button on the crash page nor the fatality page. Signed in again as moi and was able to click the buttons. 🚢

roseeichelmann (approved):
im not sure what was going on before, but i am now able to click the open crash pdf button on my personal account as well as with the vz admin and editor test accounts! and read only is disabled as expected 🚢

mddilley (approved):
Works great! Couldn't download as a read-only, and the tests still passed as well. 🚢 🙏

editor/components/CrashNarrativeCard.tsx
johnclary: Small details are very welcome — I made this change 👍

PR #2122, opened 2026-08-20
Redirects users who try to view restricted pages

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/29486

This PR adds a hook that re-directs unauthorized users away from restricted routes. Prior to this change, if a user reached a restricted page it would simply not fully load due to failing API requests.

Screenshot 2026-08-20 at 2 30 18 PM

## Testing

**URL to test:** [Netlify](https://deploy-preview-2122--vze-staging.netlify.app/editor/crashes)

1. Login in using the **Test User: Read-only viewer** account

2. Try visiting the following pages and confirm you are redirected to the `/unauthorized` page.
- [EMS list page](https://deploy-preview-2122--vze-staging.netlify.app/editor/ems)
- [EMS details page](https://deploy-preview-2122--vze-staging.netlify.app/editor/ems/26196-0215)
- [Upload non-CR3 page](https://deploy-preview-2122--vze-staging.netlify.app/editor/upload-non-cr3)

3. On the `/unauthorized` page, use the **Ask for help** link and confirms it takes you to the service desk request form (I should update that url to pre-populate the app ID!)
3. Repeat steps with the editor and admin users (**Test User: Editor** and **Test User: VZ Admin**) and confirm you can reach all pages without issue.
 
---

#### Ship list

- [ ] Check migrations for any conflicts with latest migrations in `main` branch
- [ ] Confirm Hasura role permissions for necessary access
- [ ] Code reviewed
- [ ] Product manager approved

PR #2125, opened 2026-08-20
Clean up permissions and roles type checks

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/29939

See also https://github.com/cityofaustin/vision-zero/pull/2122.

## Testing

**URL to test:** Local, because branch of branch

Literally test the entire app with all user roles 😭 

Editor, Admin users:

- Crash details: Edit crash location
- Crash details: EMS card is visible
- Crash details: transfer temp crash data
- Crash details: edit field on the **Summary**, **Flags**, or **Other** 
- Crash details: edit field on **Units** or **People**
- Crash details: edit crash notes
- Crash details: edit crash recommendations
- Crash details: edit crash diagram orientation
- Fatality details: Edit charges and contrib factors of unit
- Fatality details: Edit person image
- Fatality details: Edit narrative summary
- EMS details: link record to person

Read-only user:

- EMS list: hidden from nav for read-only users
- Upload non-CR3 page: hidden from nav for read-only users
- Crash details: EMS card is invisible
- Crash details: no edit buttons/functionality present
- Fatality details: no edit buttons/functionality present

Admin user only:

- ⚠️  You don't have to make these edits — if you see the edit/save buttons, you're good! 
- Add user
- Edit user
- Delete user

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/app/fatalities/[record_locator]/page.tsx
+++ b/editor/app/fatalities/[record_locator]/page.tsx
@@ -10,11 +10,9 @@ import FatalityUnitsCards from "@/components/FatalityUnitsCards";
 import CrashNarrativeEditableCard from "@/components/CrashNarrativeEditableCard";
 import { GET_CRASH, UPDATE_CRASH } from "@/queries/crash";
 import { Crash } from "@/types/crashes";
-import { canViewEms } from "@/utils/auth";
 import { useDocumentTitle } from "@/utils/documentTitle";
 import { formatIsoDateTimeWithDay, formatYear } from "@/utils/formatters";
 import { useQuery } from "@/utils/graphql";
-import { useAuth0 } from "@auth0/auth0-react";
 import CrashDiagramCard from "@/components/CrashDiagramCard";
 import DataCard from "@/components/DataCard";
 import { crashesColumns } from "@/configs/crashesColumns";
@@ -23,6 +21,8 @@ import NotesCard from "@/components/NotesCard";
 import UserEventsLogger from "@/components/UserEventsLogger";
 import { INSERT_CRASH_NOTE, UPDATE_CRASH_NOTE } from "@/queries/crashNotes";
 import CrashIsTemporaryBanner from "@/components/CrashIsTemporaryBanner";
+import { ADMIN_EDIT_ROLES, hasRole } from "@/utils/auth";
+import { useAuth0 } from "@auth0/auth0-react";
 
 const otherCardColumns = [
   crashesColumns.case_id,
@@ -43,12 +43,13 @@ export default function FatalCrashDetailsPage({
 
   const { record_locator: recordLocator } = use(params);
   const { user } = useAuth0();
+  const includeEms = hasRole(ADMIN_EDIT_ROLES, user)
 
   const typename = "crashes";
 
   const { data, error, refetch } = useQuery<Crash>({
     query: recordLocator ? GET_CRASH : null,
-    variables: { recordLocator, includeEms: canViewEms(user) },
+    variables: { recordLocator, includeEms },
     typename,
   });
 
--- a/editor/app/fatalities/[record_locator]/page.tsx
+++ b/editor/app/fatalities/[record_locator]/page.tsx
@@ -43,7 +43,7 @@ export default function FatalCrashDetailsPage({
 
   const { record_locator: recordLocator } = use(params);
   const { user } = useAuth0();
-  const includeEms = hasRole(ADMIN_EDIT_ROLES, user)
+  const includeEms = hasRole(ADMIN_EDIT_ROLES, user);
 
   const typename = "crashes";
 
@@ -102,10 +102,7 @@ export default function FatalCrashDetailsPage({
       {
         // show alert if crash is a temp record, hide delete button on fatalities pages
         crash.is_temp_record && (
-          <CrashIsTemporaryBanner
-            crash={crash}
-            dismissible
-          />
+          <CrashIsTemporaryBanner crash={crash} dismissible />
         )
       }
       <Row className="fatality-details-row">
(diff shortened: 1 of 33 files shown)


editor/components/CrashDiagramCard.tsx
```diff
@@ -150,11 +151,11 @@ export default function CrashDiagramCard({
 
   const { user } = useAuth0();
 
-  const isReadOnlyUser = user && hasRole(["readonly"], user);
```
johnclary, line 153: I decided to invert this `isReadOnlyUser` logic after realizing that it would resolve to `false` if `user` was `undefined`.  It's not actually a valid code path since non-users get redirected to the login page, but it seemed just conceptually wrong to fall back to edit permissions instead of falling back to readonly permissions.

editor/utils/permissions.ts
johnclary, line 1: @mateoclarke I'm curious what you think about this file— I initially thought it would make sense to declare all of our page/component-specific permissions here as exports, but in pretty much all of the code except the `/users` pages we can simply use `ADMIN_EDIT_ROLES`.

mateoclarke: On one hand, I like small files. But there is already so much role-based/permission-based logic in the auth file that I see your dilemma. We don't have super complex roles and gates (and that probably won't change) so having page/table/component-specific permissions is pretty rare. As your code here already points out, a better name for the constant I made is probably just `ADMIN_EDIT_ROLES`.

editor/utils/auth.ts
```diff
@@ -2,55 +2,74 @@ import { useCallback, useEffect } from "react";
...
+    user?.["https://hasura.io/jwt/claims"]?.["x-hasura-allowed-roles"];
+
+  if (!role) {
+    console.warn(
+      "User has malformed/missing Hasura JWT claims. No allowed role found."
+    );
+    return ["readonly"];
+  }
+  return role;
+};
```
johnclary, line 41: So we're now falling back to `"readonly"` if the User has a malformed/missing Hasura claims object in their auth0 profile. Before this change, I _think_ we were effectively falling back to admin/edit permissions because the user's role name would be `""`.  (Not that they actually had edit  permissions — JWT validation would fail).

I briefly tried to take this a step further by adding a type guard/coercion to this function, and using a fancy  custom `useTypedAuth0()` hook that uses our `CustomUser` as a generic. But it started to seem like too much overhead when we simply:

- Trust the Auth0 API contract to return the correct custom claims
- Ensure that type safety so that, if a user does have malformed claims, they experience no crashing of the app but they won't be able to do anything in the app.

Charlie-Henry (commented):
I clearly don't know what I'm doing sorry

I tired on admin account:

- [x] Crash details: Edit crash location
- [x] Crash details: EMS card is visible
- [ ] Crash details: transfer temp crash data
- [ ] Crash details: edit field on data card
  -  I'm not sure what this means 😅 
- [ ] Crash details: edit field on related record table
  -  I don't see a related records section?
- [x] Crash details: edit crash notes
- [x] Crash details: edit cash recommendations
- [ ] Crash details: edit crash diagram orientation
  -  none of my crash diagrams were loading
- [ ] Fatality details: Edit charges and contrib factors of unit
  -  I couldn't find the charges section, I can't edit the contributing factors?
- [x] Fatality details: Edit person image
- [x] Fatality details: Edit narrative summary
- [ ] EMS details: link record to person
  - Idk how to do this 😅 

Read-only user:

- [x] EMS list: hidden from nav for read-only users
- [x] Upload non-CR3 page: hidden from nav for read-only users
- [x] Crash details: EMS card is invisible
- [x] Crash details: no edit buttons/functionality present
- [x] Fatality details: no edit buttons/functionality present

Admin user only:

Image

- [ ] Add user
- [ ] Edit user
- [ ] Delete user

editor/app/fatalities/[record_locator]/page.tsx
```diff
@@ -43,12 +43,13 @@ export default function FatalCrashDetailsPage({
 
   const { record_locator: recordLocator } = use(params);
   const { user } = useAuth0();
+  const includeEms = hasRole(ADMIN_EDIT_ROLES, user)
```
Comment (frankhereford) [inline editor/app/fatalities/[record_locator]/page.tsx]: Missing `;` here

editor/app/users/[user_id]/page.tsx
```diff
@@ -11,12 +11,12 @@ import { FaUserEdit, FaUserAltSlash } from "react-icons/fa";
...
 import UserModal from "@/components/UserModal";
 import PermissionsRequired from "@/components/PermissionsRequired";
-import { formatRoleName, useGetToken } from "@/utils/auth";
+import { ADMIN_ROLE, formatRoleName, useGetToken } from "@/utils/auth";
 import { useUser } from "@/utils/users";
 import { User } from "@/types/users";
 import { formatIsoDateTime } from "@/utils/formatters";
 
-const allowedUserEditRoles = ["vz-admin"];
+const allowedUserEditRoles = [ADMIN_ROLE]
```
Comment (frankhereford) [inline editor/app/users/[user_id]/page.tsx]: Missing `;` here

editor/utils/auth.ts
```diff
@@ -2,55 +2,74 @@ import { useCallback, useEffect } from "react";
...
-    // check this user's raw JSON in the Auth0 user database
-    return "";
-  } else {
-    return roles[0] || "";
+export const getHasuraRoleName = (
+  roles: HasuraUserRoleName[]
+): HasuraUserRoleName => {
+  {
+    return roles[0];
   }
```
frankhereford: I think you can drop this inner set of curly braces left over from the if/else.

frankhereford (approved):
John, I think this is working really well. I was able to work through the list both as a read-only user and then as an editor, and observed the expected app behavior. I found a few tiny things in the code, but nothing that wouldn't have been picked up by the linter. Thanks for this big clean-up!

chiaberry (approved):
I think I passed all the test steps, I was only able to edit the charges on a temp fatality record, which is think is the correct behavior. 

Echoing Charlie's confusion on "Crash details: edit field on data card & Crash details: edit field on related record table", I took those to mean editing some of the fields on the summary, flags and details cards and also being able to edit some of the fields on the units and people cards. 

I like Frank's suggestion of removing the EMS incident option from the nav search. Or if not removing it, including a message that says EMS incident search is unavailable. Right now if you put in anything it just fails silently.

roseeichelmann (approved):
tested every role with every action, this is all working perfectly. charges can only be edited on the fatalities page for temp crash fatalities, which is how its supposed to work. this all looks great!

PR #2113, opened 2026-08-14
Fix map issues in React 18 upgrade

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/28870

## Testing

**URL to test:** 

**Steps to test:**

---

#### Ship list

- [ ] Check migrations for any conflicts with latest migrations in `main` branch
- [ ] Confirm Hasura role permissions for necessary access
- [ ] Code reviewed
- [ ] Product manager approved

PR #2126, opened 2026-08-20
Address some map upgrade issues

chiaberry:
This PR addresses some of the issues that have arisen from upgrading react-map-gl to v7
I branched off of Rose's upgrade work (https://github.com/cityofaustin/vision-zero/pull/2113/)

## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/29941

## Testing

Test locally. From the viewer directory, 
```
npm install --legacy-peer-deps
```

and then `npm run dev` to get it running. 

**Steps to test:**

Test the Overlays
1. ASMP layers. Toggle on and then test showing only certain levels
2. High Injury Network, toggle on and off
3. Toggle on the city council overlay.  Toggle off. Click in a council district. there is a popup. 

Test the Infobox popup by clicking any of the crash markers. Click the clipboard icon to copy the crash number. You may see a warning in your console, it's a [known reactstrap issue](https://github.com/reactstrap/reactstrap/issues/2819)

Test drawing a polygon. 

---

#### Ship list

- [ ] Code reviewed


viewer/src/views/map/InfoBox/MapInfoBox.jsx
```diff
@@ -29,7 +29,7 @@ const MapInfoBox = React.memo(
     const buildSeriousInjuriesOrFatalitiesConfig = (info) => [
       {
         title: "Date/Time",
-        content: format(new Date(info.crash_timestamp_ct), "MM/dd/yyyy H:m a"),
+        content: format(new Date(info.crash_timestamp_ct), "MM/dd/yyyy H:mm a"),
```
chiaberry: Sneaking this in here after someone pointed it out in dev sync, this changes the time from `19: 7 PM` to `19:07 PM`

I think the inclusion of 24 hour clock and PM is redundant, but not all americans are used to the 24 hour time format.

viewer/src/views/map/MapComponent.jsx
```diff
@@ -507,6 +487,7 @@ const MapComponent = () => {
       {...viewState}
       onMove={onMove}
       mapboxAccessToken={import.meta.env.VITE_MAPBOX_TOKEN}
+      mapStyle="mapbox://styles/mapbox/light-v9"
```
johnclary: Would you mind bumping us to`light-v11`?

viewer/src/views/map/MapComponent.jsx
```diff
@@ -516,13 +497,15 @@ const MapComponent = () => {
       preserveDrawingBuffer={false}
     >
       {baseSourceAndLayer}
+      <Source {...asmpSourceConfig}>
+        {buildAsmpLayers(asmpConfig, overlay)}
```
johnclary, line 501: Not a new problem, and not causing issues on my machine, but these layer-building functions are running 100+ times per map interaction. We should memoize them. Should I drop that into a separate issue?

viewer/src/views/map/MapPolygonFilter.jsx
```diff
@@ -21,7 +21,7 @@ const MapPolygonFilter = ({ setMapPolygon }) => {
 
     .mapboxgl-ctrl-top-right {
       position: absolute;
-      top: 107px;
+      top: 116px;
```
johnclary: Thanks for fixing this. I got curious and set this to `200px` and now i see an extra set of draw controls. Any idea why we have the duplicated control buttons?

Image

I also noticed that while drawing:

1. The polygon line does not appear until you add a second point. In production, once you add the first point of the polygon, a dotted line appears and follows your cursor.
2. In prod, the polygon outline for the currently-being-edited-edges is dotted and has color differentiation. In this branch we have a solid blue line at all times, and the the polygon vertices are smaller + lacking a white outline.

\#1 feels like a regression we should fix; I am less concerned \#2. Can you spot what might have changed between this branch and prod that's causing this?

Image

Image

viewer/src/views/map/MapComponent.jsx
```diff
@@ -516,13 +497,15 @@ const MapComponent = () => {
...
+      </Source>
       {!!mapData && renderCrashDataLayers()}
       {selectedFeature && renderSelectedLayer()}
-      {buildAsmpLayers(asmpConfig, overlay)}
       {buildHighInjuryLayer(overlay)}
       {!!cityCouncilOverlay && overlay.name === "cityCouncil" && (
         <Source type="geojson" data={cityCouncilOverlay}>
-          <Layer {...cityCouncilDataLayer} />
+          <Layer beforeId="road-street" {...cityCouncilDataLayer} />
```
johnclary: When I click on the council district layer I hit a WSOD 😭 

In prod we have a little popup with an animated circle. We don't need to restore the animated circle marker, but we should bring the pop-up back to life. Happy to make a separate issue for this as well.

Image

johnclary (commented):
Chia, thanks for working through these tricky upgrades. I ran into a few problems, but overall this is looking great 🙏 

Happy to split off issues where you think it makes sense.

viewer/src/views/map/InfoBox/MapInfoBox.jsx
johnclary: Oof. Thanks for catching this. I am quite surprised to see 24hr time here. Would you mind switching this to 12-hour time: `"h:mm a"` — I wonder if that was the original intent 🤔

viewer/src/views/map/MapPolygonFilter.jsx
frankhereford: I'm not sure if this is helpful or not, but I also noticed that as you're adding vertices to the polygon, only the 0th vertex and the most recently clicked one get a marker.

viewer/src/views/map/map-style.jsx
```diff
@@ -142,12 +142,7 @@ export const buildHighInjuryLayer = (overlay) => {
   const highInjuryNetworkLayerConfig = {
     id: " highInjuryNetwork",
```
frankhereford: There is a leading space here in this ID. When I fix it, I don't notice any difference in the app, but it may be worth fixing here unless there is a reason that I'm missing.

frankhereford (approved):
Thanks for this work Chia, I am not seeing any additional issues beyond what John has pointed out. One thing I am seeing, for sure, is a responsive map. I'm able to scroll in and out freely, which was the major bug IIRC. This is a big improvement!

viewer/src/views/map/MapPolygonFilter.jsx
```diff
@@ -99,13 +93,41 @@ const MapPolygonFilter = ({ setMapPolygon }) => {
...
+          },
           {
             id: "gl-draw-polygon-midpoint",
             type: "circle",
             filter: ["all", ["==", "$type", "Point"]],
             paint: {
-              "circle-radius": 3,
-              "circle-color": "#fbb03b",
+              "circle-radius": 4,
+              "circle-color": "#3bb2d0",
```
chiaberry: i find that the orange vertex color is too close to the serious injury marker color, what do you think about making it blue instead of orange?

roseeichelmann (approved):
Chia i tested thoroughly and im not seeing any issues! the map is really responsive, i can add and take away all of the layers, polygon draw is working fine for me, i think this is at a good stopping point!

viewer/src/views/map/map-style.jsx
chiaberry: thank you!

viewer/src/views/nav/SideMapControlDateRange.jsx
```diff
@@ -1,6 +1,5 @@
 import React, { useState, useEffect } from "react";
 import { StoreContext } from "src/constants/context";
-import aphroditeInterface from "react-with-styles-interface-aphrodite";
```
chiaberry, line 3: missed removing this in the date picker PR

viewer/src/views/map/MapComponent.jsx
```diff
@@ -302,8 +303,8 @@ const MapComponent = () => {
         ...selectedFeature,
         properties: {
           ...selectedFeature.properties,
-          latitude: event.lngLat[1],
-          longitude: event.lngLat[0],
+          latitude: event.lngLat.lat,
+          longitude: event.lngLat.lng,
```
chiaberry, line 307: this fixes the white screen crashing when clicking the city council layer, the popup is back (but the animated icon is still missing)

viewer/src/views/map/MapPolygonFilter.jsx
chiaberry: I removed the extra buttons that were being added and restored the styles, addressing the regressions. Let me know what else you find

viewer/package.json
```diff
@@ -46,7 +45,6 @@
     "react-map-gl": "^7.1.7",
     "react-router-dom": "^6.30.4",
     "react-use": "^17.6.1",
-    "react-with-styles-interface-aphrodite": "^6.0.1",
```
johnclary, line 49: 🧹

viewer/src/views/map/InfoBox/InfoCard.jsx
```diff
@@ -42,22 +43,24 @@ const InfoCard = ({ content }) => {
                       textAlign: "center",
                       fontSize: "12px",
                       lineHeight: 1.42857,
+                      backgroundColor: colors.info,
```
johnclary, line 46: Adding this background color remove the btn hover effect that paints the button slightly darker. Is it possible to bring the hover effect back?

viewer/src/views/map/helpers.js
```diff
@@ -86,3 +86,111 @@ export function useMapEventHandler(eventName, callback, mapRef) {
...
+  {
+    id: "gl-draw-midpoint",
+    type: "circle",
+    filter: ["all", ["==", "meta", "midpoint"]],
+    paint: {
+      "circle-radius": 3,
+      "circle-color": orange,
+    },
+  },
+];
```
johnclary, line 196: looks great!

viewer/src/views/map/MapPolygonFilter.jsx
```diff
@@ -120,24 +79,26 @@ const MapPolygonFilter = ({ setMapPolygon }) => {
...
+            (editType === "draw.create" || editType === "draw.update") &&
+            event.features.length > 0
+          ) {
+            const feature = event.features[0];
             if (
               feature &&
               feature.geometry &&
               feature.geometry.type === "Polygon"
             ) {
+              console.log(feature)
```
johnclary, line 93: Remove this? 🧹

viewer/src/views/map/MapComponent.jsx
johnclary: interesting that the shape of this data changed! thanks for fixing!

viewer/src/views/map/InfoBox/InfoCard.jsx
johnclary, line 1: I think this change was introduced by removing the MUI css baselline: The text in the info box is `12px` now, but `14px` in prod. Can you make the text size 14px in this info box and also the council district popup?

johnclary (commented):
@chiaberry This looks so good! I Just the tiniest changes needed. 

Very cool that you can now drag a drawn shape around. It's also easier to close the drawn polygon—really nice improvements.

One other small thing: in order to install this i had to comment out this line in `.npmrc`:

```
# min-release-age=7
```

Otherwise I got a package resolution error:

```
❯ npm install --legacy-peer-deps

npm warn deprecated @types/mapbox-gl@3.5.0: This is a stub types definition. mapbox-gl provides its own type definitions, so you do not need this installed.
npm error code ETARGET
npm error notarget No matching version found for zod@4.5.4 with a date before 8/28/2026, 2:42:48 PM.
npm error notarget In most cases you or one of your dependencies are requesting a package version that doesn't exist.
```

The other workaround is to use a pre-v24 version of node when installing. This should resolve itself in <7 days.

viewer/src/views/map/MapPolygonFilter.jsx
mddilley: the draw UI looks so good! i noticed a couple of differences between prod and this branch, but i think they are really small edge cases:
1. if you cross your lines when selecting, the correct points are selected but you get different styling (see below)
2. if you click away from the selected polygon in this branch, the polygon goes out of focus and the trash can icon doesn't remove the selection when clicked. you need to click the polygon again to make it active and then delete.

#### this branch
Screenshot 2026-09-09 at 3 19 12 PM

#### prod
Screenshot 2026-09-09 at 3 19 52 PM

mddilley (approved):
I tested this out every way that I could, and I couldn't break it or find anything no one else had already mentioned or has been fixed already. Thanks for updating and making this map usable!! 🙏 🙌

viewer/src/views/map/InfoBox/InfoCard.jsx
chiaberry: I am going to address this hover in a future PR...

PR #2127, opened 2026-08-24
Create crash geolocation_provider lookup

### LINKED ISSUE #29767: Create crash `geolocation_provider`  lookup
We currently identify the source of a crash location using read-time querying: if the lat/lon values in the `crashes` table are different from those in the `crashes_cris` table, we assume the crash location was provided via manual Q/A.

This has limitations:
 
- It adds query overhead and confounds the query planner, leading to long query execution for custom reports, such as those that Xavier uses in [his QA/QC dashboard](https://github.com/cityofaustin/atd-data-tech/issues/29843). It's also the reason why we don't show this information on the crashes list in the VZE!
- We are limited to a simple binary of whether or not a crash location was edited, but we have aspirations to add [at least one additional provider](https://github.com/cityofaustin/atd-data-tech/issues/29877): CAD (`cad_incidents`).

#### In scope

- Add a new lookup table to the database — `lookups.geolocation_providers` with the following lookup values:

`id` | `label` | `source`
-- | -- | --
1 | `"cris"` | `"vz"`
2 | `"apd_cad"` | `"vz"`
3 | `"manual_qa"` | `"vz"`

- Add a new, non-nullable column to the `crashes` table called `geolocation_provider_id` which references the new lookup table. Default value is `1`.

- On the crash details page of the VZE
  - Use the new `geolocation_provider` to populate the location **Provider** displayed in the map card header. 
  - When a crash location is edited, set the `geolocation_provider_id` provide value to `"manual_qa"` ID

- Complete a one-time backfill of this column for all crashes. If the lat/lon values in the `crashes` table is different from those in the `crashes_cris` table, set the value to the `"manual_qa"` ID

### Out of scope

* https://github.com/cityofaustin/atd-data-tech/issues/29877

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/29767

## Testing

**URL to test:** 
Local bc migrations

**Steps to test:**
2. Go to a crash details page, see that the Provider on the map card says TxDOT
3. Edit the coordinates of the crash and see that the Provider automatically updates to Manual Q/A
4. Now to test the backfill, run the following statement in your db manager:
```
UPDATE crashes
SET geolocation_provider_id = 3,
updated_by = 'system'
WHERE EXISTS (
    SELECT 1
    FROM crashes_cris
    WHERE crashes_cris.id = crashes.id
    AND (crashes.latitude IS DISTINCT FROM crashes_cris.latitude
         OR crashes.longitude IS DISTINCT FROM crashes_cris.longitude)
);
);
```
5. Now when you open up crashes, if it has coordinates that are different from the crashes_cris table, it will have Manual Q/A as the geocoder provider.
---

#### Ship list

- [ ] Code reviewed


editor/components/CrashMapCard.tsx
```diff
@@ -22,14 +22,20 @@ import LocationPolygonLayer from "@/components/LocationPolygonLayer";
 
 const allowedMapEditRoles = ["vz-admin", "editor"];
 
+const formattedGeoProvider = {
+  cris: "TxDOT CRIS",
+  apd_cad: "APD CAD",
```
roseeichelmann, line 29: Is this how we wanna format the APD CAD label?

chiaberry (approved):
🚢 ! I see the expected geocoder provider after moving the location around and after running the backfill locally. (took about 5 minutes for the backfill to finish)

database/migrations/default/1787588257933_create_table_lookups_geolocation_provider/up.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE "lookups"."geolocation_provider" ("id" serial NOT NULL, "label" text NOT NULL, "source" text NOT NULL DEFAULT 'vz', PRIMARY KEY ("id") , UNIQUE ("id"));
+
+INSERT INTO "lookups"."geolocation_provider"("id", "label", "source") VALUES (1, E'cris', E'vz');
+
+INSERT INTO "lookups"."geolocation_provider"("id", "label", "source") VALUES (2, E'apd_cad', E'vz');
+
+INSERT INTO "lookups"."geolocation_provider"("id", "label", "source") VALUES (3, E'manual_qa', E'vz');
+
+alter table "public"."crashes" add column "geolocation_provider_id" integer
+ not null default '1';
```
johnclary, line 10: Would you add a foreign key constraint between this column and the lookup table?

database/metadata/databases/default/tables/public_crashes.yaml
```diff
@@ -44,6 +44,15 @@ object_relationships:
...
           schema: public
+  - name: geolocation_provider
+    using:
+      manual_configuration:
+        column_mapping:
+          geolocation_provider_id: id
+        insertion_order: null
+        remote_table:
+          name: geolocation_provider
+          schema: lookups
```
johnclary: I don't _think_ there is any functional tradeoff between using a manual relationship config vs the built-in `foreign_key_constraint_on` method, but I guess I prefer the latter for consistency.

editor/components/CrashMapCard.tsx
johnclary: I'm afraid so, yes. 

I don't like the acronyms, but "APD Computer-Aided Dispatch" is too long. I considered "APD Dispatch", but I think it's worth using "CAD" specifically because that's how we will label CAD records when we add more CAD features (like showing CAD records on maps). 

But I could be convinced otherwise!

editor/components/CrashMapCard.tsx
```diff
@@ -86,11 +92,10 @@ export default function CrashMapCard({
...
           <span className="fw-bold me-2">Provider </span>
           <span>
-            {" "}
-            {hasCoordinates
-              ? isManualGeocode
-                ? "Manual Q/A"
-                : "TxDOT CRIS"
+            {hasCoordinates && geolocationProvider
+              ? formattedGeoProvider[
+                  geolocationProvider as keyof typeof formattedGeoProvider
```
johnclary: This type assertion is hinting to me that we should add strong typing for the geolocation provider rather than the generic `LookupTableOption`. That will let you define the exact allowed values on the `Crash` type, this component's props, and the `formattedGeoProvider` lookup. Something like this: 

```typescript
type GeolocationProviderCode = "cris" | "apd_cad" | "manual_qa";

const formattedGeoProvider = {
  cris: "TxDOT CRIS",
  apd_cad: "APD CAD",
  manual_qa: "Manual Q/A",
} satisfies Record<GeolocationProviderCode, string>;

// plug this into the Crash type
type GeolocationProvider = {
  id: number;
  label: GeolocationProviderCode;
};
```

johnclary (commented):
Looking good, Rose! My feedback is pretty minor. The tests worked perfectly.

editor/components/CrashMapCard.tsx
roseeichelmann: I think its probably the best option

editor/components/CrashMapCard.tsx
roseeichelmann: Okay this was a good idea -- i put all this code in the `CrashMapCard` component and exported the `GeolocationProvider` type into the `crashes.ts` file, is that what you were envisioning?

editor/app/crashes/[record_locator]/page.tsx
```diff
@@ -129,7 +129,7 @@ export default function CrashDetailsPage({
             mutation={UPDATE_CRASH}
             locationId={crash.location_id}
             location={crash.atd_txdot_location}
-            isManualGeocode={crash.crashes_list_view.is_manual_geocode}
```
Comment (johnclary) [inline editor/app/crashes/[record_locator]/page.tsx:132]: Just noting that we will remove `is_manual_geocode` from this view via https://github.com/cityofaustin/atd-data-tech/issues/29844.

johnclary (approved):
Looks great! Thanks for tightening up those types 🚢 🚢 🚢 🚢 🚢

PR #2128, opened 2026-08-26
Salvage various updates from the incident branches

johnclary:
## Associated issues

- In service of https://github.com/cityofaustin/atd-data-tech/issues/30060
- See also: https://github.com/cityofaustin/vision-zero/pull/2132

## Testing

**URL to test:** [Netlify](https://deploy-preview-2128--vze-staging.netlify.app/)

**Steps to test:**

1. Log in as an editor or admin and visit the EMS list view (`/ems`) and confirm the **Beta** badge and info message have been from the title (see [prod](https://visionzero.austin.gov/editor/ems) for comparison)
2. Visit the crash details page and confirm the **Beta** badge and info message have been from the card header of the **EMS Patient Care** card
3. Let me know what you think about keeping the CAD type and column definitions in the code base even though we're not actively using them.

---

#### Ship list

- [ ] Code reviewed


editor/configs/crashesColumns.tsx
```diff
@@ -70,6 +73,7 @@ export const crashesColumns = {
   in_austin_full_purpose: {
     path: "in_austin_full_purpose",
     label: "In Austin Full Purpose Jurisdiction",
+    valueFormatter: formatYesNoString,
```
johnclary, line 76: We're not rendering this anywhere at the moment, but it will be properly formatted if we do

editor/configs/cadColumns.tsx
johnclary, line 1: We don't need these columns at the moment, but it seemed like a waste to through them away since we should at some point want to show CAD records to users 🤷

editor/types/cadIncident.ts
```diff
@@ -0,0 +1,33 @@
+import { Point } from "geojson";
+
+export type CadIncident = {
```
johnclary, line 3: Worth keeping this type def around, right?

editor/app/ems/page.tsx
johnclary, line 1: Removes the **Beta** label from the EMS list page

editor/components/EMSCardHeader.tsx
johnclary, line 1: Removes the **Beta** label from the EMS card header on the crash details page

editor/types/cadIncident.ts
chiaberry: I think yes. I have hoarding tendencies, but I think this is still something useful to keep. We can reassess later if it turns crufty.

chiaberry (approved):
the badges are gone! 🚢 

I am trying to remember what else in your incident branches would be good to save. I have the vaguest memory of one of your incident branches adding code to be able to adjust column visibility/edit ability depending on what card the columns were on. I am going to look to see if my memory is right and if it made it into the codebase. I havent looked at the teardown PR yet, so maybe things will be clearer then.

editor/configs/cadColumns.tsx
roseeichelmann: I also think yes its fine to keep around if we are going to be circling back on this soon

roseeichelmann (approved):
all the badges are gone as expected 👍

frankhereford (approved):
These changes make sense to me, and I'm fine with keeping the columns and their associated metadata around if you think it's potentially useful. Thanks John for this.

PR #2129, opened 2026-08-27
Update crash with CAD incident lat/longitude if not provided by CRIS

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/29877

## Testing

**URL to test:** 
Local

**Steps to test:**
1. Apply the migration
2. In order to test the trigger we have to reproduce what would happen when there is an insertion of a new crash into `crashes_cris` that has a matching `case_id`  with a record in `cad_incidents` that is within 2 days of the crash timestamp. In order to do this I found a crash that is already matched to a `cad_incident` that follows those criteria and cloned that row. Repeat my steps here to test:
   - In your Hasura console, filter the `crashes_cris` table by `case_id` = 180751950. Click on the button that says "Clone this row"
   - Now edit the latitude and longitude to be null in the new row you are creating and change the `cris_crash_id` so that it is unique, and then click Save
   - The trigger should run on the new row created in the `crashes` table and it should now have a latitude and longitude derived from the `cad_incident` table and the `geolocation provider_id` should be 2, which maps out to APD CAD.
3. Find this crash details page in the editor and see that it says "APD CAD" for the Provider label on the map card

---

#### Ship list

- [ ] Code reviewed


database/migrations/default/1787861098253_crashes_fill_cad_coordinates_trigger/up.sql
```diff
@@ -0,0 +1,40 @@
...
+        IF v_lat IS NOT NULL AND v_lon IS NOT NULL THEN
+            NEW.latitude := v_lat;
+            NEW.longitude := v_lon;
+            NEW.geolocation_provider_id := 2; --"apd_cad"
+        END IF;
+    END IF;
+    
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
```
johnclary, line 34: Would you add a `COMMENT` that describes what this function does?

database/migrations/default/1787861098253_crashes_fill_cad_coordinates_trigger/up.sql
```diff
@@ -0,0 +1,40 @@
...
+
+-- This trigger must fire before crashes_set_spatial_attributes_on_insert trigger, so it must come first alphabetically
+CREATE TRIGGER crashes_fill_cad_coordinates_before_insert
```
johnclary: Looking at the triggers on `crashes`, I think we want to make this absolute earliest trigger to fire, so that any lat/lon it populates get picked up by `crash_insert_ems_match_trigger`.

What do you think about prefixing this with trigger name with `00_`? And if y'all agree I'll open a tech debt issue to adopt this convention across all our tables:

```sql
00_crashes_fill_cad_coordinates_before_insert 
10_crash_insert_ems_match_trigger
20_crashes_set_spatial_attributes_on_insert
-- etc
````

File: database/migrations/default/1787861098253_crashes_fill_cad_coordinates_trigger/up.sql [hunk 3]
```diff
@@ -0,0 +1,40 @@
...
+       AND NEW.case_id IS NOT NULL THEN
+        
+        -- Search for matching CAD incident
+        SELECT ci.latitude, ci.longitude
+        INTO v_lat, v_lon
+        FROM cad_incidents ci
+        WHERE ci.agency_type_short = 'apd'
+          AND ci.master_incident_number = NEW.case_id
+          AND NEW.crash_timestamp BETWEEN ci.response_date - INTERVAL '2 days' 
+                                      AND ci.response_date + INTERVAL '2 days'
```
johnclary, line 21: I used `explain analyze` on this query and we are hitting our incident number index cleanly. This is a very cheap query.

```sql
Index Scan using idx_cad_incidents_master_incident_number on cad_incidents ci  (cost=0.42..2.66 rows=1 width=16) (actual time=0.163..0.174 rows=1.00 loops=1)
"  Index Cond: (master_incident_number = '262321438'::text)"
"  Filter: ((agency_type_short = 'apd'::text) AND ('2026-08-21 02:19:00+00'::timestamp with time zone >= (response_date - '2 days'::interval)) AND ('2026-08-21 02:19:00+00'::timestamp with time zone <= (response_date + '2 days'::interval)))"
  Index Searches: 1
  Buffers: shared hit=6
Planning Time: 0.460 ms
Execution Time: 0.227 ms
```

johnclary (commented):
@roseeichelmann this looks really good. I had an older snapshot lying around, so i downloaded a few CAD and CRIS CSVs and tested this with a week of imports. 

In my testing, we had 11 of 171 APD crash reports come in with missing lat/lon, and all of them were geolocated with a CAD record. Super cool!

I left minor feedback. Thanks for shipping this! 🚀

database/migrations/default/1787861098253_crashes_fill_cad_coordinates_trigger/up.sql
roseeichelmann: Okay so triggers cant start with a number, but we could prefix like a_, b_, c_?

johnclary: I think the reason that incrementing by `00_`, `10_`, is conventional is because you can squeeze in additional triggers later without renaming the whole lot (`02_`, `04_`...). so actually maybe starting at `10_` is superior because you have room to add an earlier trigger...

chiaberry (approved):
Thanks for the test steps! 

Image

If you can't begin trigger names with numbers, does prefixing with an underscore work?

johnclary (approved):
thanks, Rose! 🚢

PR #2139, opened 2026-09-04
The crash map marker is red when not editing

### LINKED ISSUE #29934: The crash map marker is red when not editing
The map marker on the crash details page should be blue when not editing, and red when editing. If you edit a location by moving the marker outside of a location polygon, it persists as red after saving.

Image

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/29934

## Testing

**URL to test:** 
[Netlify](https://deploy-preview-2139--vze-staging.netlify.app/editor/crashes)

**Steps to test:**
Go to a crash details page, the location marker should be blue. Edit the coordinate to be outside of a polygon (not on a major roadway, so try putting it over a house). On edit the marker should turn red, on save the marker should turn blue.

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/editor/components/PointMap.tsx
+++ b/editor/components/PointMap.tsx
@@ -300,7 +300,7 @@ export const PointMap = ({
       {/* editable + not editable point layers */}
       {savedLatitude && savedLongitude && !isEditing && (
         <Marker
-          key={dynamicMarkerKey}
+          key={`marker-view-${dynamicMarkerKey}`}
           latitude={savedLatitude}
           longitude={savedLongitude}
           color={COLORS.primary}
@@ -308,10 +308,10 @@ export const PointMap = ({
       )}
       {isEditing && draftLatLon && (
         <Marker
-          key={dynamicMarkerKey}
+          key={`marker-edit-${dynamicMarkerKey}`}
           latitude={draftLatLon.latitude}
           longitude={draftLatLon.longitude}
-          color={isEditing ? COLORS.danger : undefined}
+          color={COLORS.danger}
         />
       )}
     </MapGL>


editor/components/PointMap.tsx
```diff
@@ -300,18 +300,18 @@ export const PointMap = ({
       {/* editable + not editable point layers */}
       {savedLatitude && savedLongitude && !isEditing && (
         <Marker
-          key={dynamicMarkerKey}
+          key={`marker-view-${dynamicMarkerKey}`}
```
roseeichelmann, line 303: makes sure we are mounting a new marker when the `isEditing` state changes, instead of reusing the stale marker with the wrong color

johnclary (approved):
Tricky! Thanks for the fix!! 🚢 🚢 🚢

editor/components/PointMap.tsx
mddilley: took me a second but then i saw the `view` and `edit` parts of the keys. cool! and a good reminder about how important distinct keys can be to React when deciding to reuse something or not. 😎

mddilley (approved):
Works just as expected. Nice fix to a sneaky little bug! 🐛 👋 🚢

chiaberry (approved):
nice fix! 🚢

frankhereford (changes requested):
This works great showing when you're edit mode by turning the marker red, but on reading the issue, I think the marker is supposed to stay red if it's saved on a location outside of a VZ polygon. If this was refined further and I missed the updated requirements, please let me know, and I'll toggle this to an approval. Thanks!

frankhereford (approved):

PR #2140, opened 2026-09-09
Fixes mobile date picker and cleans up filter UI/UX

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/29925

Screenshot 2026-09-10 at 1 07 48 PM

## Testing

**URL to test:** Local, because branch of branch of branch

1. Test out the date filters. Try to break them! Shrink your display down and make sure the native date pickers appear whenever you are using the collapsed side-drawer. It would be extra helpful if you test with your mobile device as well. You can do this with by starting the viewer with `npm run dev -- --host`: vite will display the server address on your local network.

I have tested on Chrome on macos, and safari on macos and iOS. More browsers/platforms welcome.

2. Test out the other filters: Injury type, travel mode, and crash time

---

#### Ship list

- [ ] Code reviewed


viewer/src/constants/responsive.js
```diff
@@ -12,7 +12,7 @@ export const responsive = {
   headerButtonHeight: 50,
   headerButtonWidth: 200,
   headerLogoOffset: 34,
-  drawerWidth: 300,
+  drawerWidth: 350,
```
johnclary, line 15: The date input text was wrapping on my iphone at 300px. It'd be nice to go 100% on extra-small displays, but I left that out of scope.

viewer/src/views/nav/SideMapControlDateRange.jsx
```diff
@@ -26,6 +26,8 @@ const clampToMinDate = (raw) => {
 };
 
 const SideMapControlDateRange = ({ type }) => {
+  const isMobile = type === "temporary";
```
johnclary: The app renders the temporary/mobile side drawer at the bootstrap medium breakpoint, so that's when we're falling back to native date pickers. 

We might want to fallback to native date pickers on smaller displays....

viewer/src/views/nav/SideMapControlDateRange.jsx
```diff
@@ -7,159 +7,209 @@ import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
...
+  height: 16px;
+  margin: 2px;
+  right: 1px;
+`;
+
+const StyledCalendarIcon = styled(FontAwesomeIcon)`
+  ${calendarInputIconStyles}
+`;
+
+const StyledActionRow = styled.div`
```
johnclary, line 68: it'd be good to move these declarations outside of component render throughout the app — we do it a lot.

viewer/src/views/nav/SideMapControl.jsx
johnclary, line 1: The changes to this file add the border and title to the injury type and travel mode filters.

viewer/src/views/nav/SideMapControlDateRange.jsx
johnclary, line 1: Summary of changes:
- Adds an "Apply date filter" ("Apply dates", "Apply" - ?) button. No more parsing + applying date inputs instantly `onChange`
- Use native date pickers on <=tablet-sized displays
- Use an undo instead of redo icon

viewer/src/views/nav/SideMapControlDateRange.jsx
```diff
@@ -7,159 +7,209 @@ import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
...
+        </StyledActionRow>
       </StyledButtonContainer>
-      <div className="form-text">
-        <span>
-          <FontAwesomeIcon
-            icon={faInfoCircle}
-           className="me-1"
-          />
-        </span>
-        <span>Data starts in 2014</span>
```
johnclary, line 161: I removed this text to clean up visual clutter. I don't mind bringing it back, but it feels more obvious  now that when you input a date before 2014 it is coerced to 1/1/2014 because that's the earliest date available 🤷 LMK what you think 🙏

viewer/src/views/nav/SideMapControlDateRange.jsx
```diff
@@ -7,159 +7,209 @@ import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
...
+          )}
+          {!isMobile && (
+            <StyledCalendarIcon
+              title="Date range"
+              icon={faCalendar}
+              color={colors.dark}
+            />
+          )}
+        </StyledDateRow>
+        <StyledActionRow>
```
johnclary, line 194: I went back and forth on always showing a button below the date inputs — with it being disabled when there are no changes to apply. I really don't like the "disabled" style of our dark button: it's not obvious that the button is disabled. That's the main reason I opted for hiding these buttons unless there is an action available. 

If I were re-designing the side drawer styles, I would nix the black entirely and use primary/non-dark styles. I intend to mock that up!

viewer/src/views/nav/SideMapTimeOfDayChart.jsx
johnclary, line 1: Summary of changes:
- Add the outline border and title to this filter section
- Add helper text about clicking a time range to filter on it
- Rename  "All Times" button to "Reset" and only show it if there's something to reset

viewer/src/views/nav/SideMapTimeOfDayChart.jsx
```diff
@@ -152,7 +162,9 @@ export const SideMapTimeOfDayChart = ({ filters }) => {
   };
 
   return (
-    <Container className="px-0 mt-3">
+    <StyledButtonContainer className="mt-3">
+      <h6>Crash time</h6>
+      <span className="form-text">Click a time range to filter</span>
```
johnclary, line 167: Is there a better way to phrase this?

viewer/package.json
```diff
@@ -86,6 +86,7 @@
     "globals": "^17.6.0",
     "prettier": "^3.8.4",
     "react-axe": "^3.5.4",
+    "typescript": "^5.7.0",
```
johnclary, line 89: `npm lint` was busted because of this dependency. We have some lint errors that need to be cleaned up. Will open a separate issue.

viewer/src/views/nav/SideMapControlDateRange.jsx
```diff
@@ -7,159 +7,209 @@ import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
...
-            color={colors.dark}
-            onClick={() => null}
-          />
-        )}
+      <StyledButtonContainer className="picker-outline">
+        <h6>Crash date</h6>
+        <StyledDateRow>
+          {isMobile ? (
+            <>
+              <StyledNativeDateInput
```
johnclary, line 149: I initially introduced native date pickers because the DatePicker component was fully busted on Android. The new "Apply" button probably fixes the underlying issue with DatePicker though. I personally prefer native date pickers on IOS, so I left them in. 

The problem, however, is that we are keying on tablet display size when we decide to render the native pickers (this is flowing down from the conditional rendering of the "temporary" (aka collapsible)  side drawer). So, if you're using desktop and shrink your window down you'll get native date pickers. They work fine, so I think this is ok? We could probably improve this by specifically trying to detect touch devices, which seems to be a bit tricky FWIW.

viewer/src/views/nav/SideMapControlDateRange.jsx
```diff
@@ -7,159 +7,209 @@ import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
...
   }, [start, end, setMapDate]);
 
+  const handleApply = () => {
+    const clampedStart = pendingStart < minDate ? minDate : pendingStart;
+    const clampedEnd = pendingEnd < minDate ? minDate : pendingEnd;
+    setStart(clampedStart);
+    setEnd(clampedEnd);
+    setPendingStart(clampedStart);
+    setPendingEnd(clampedEnd);
+  };
```
johnclary, line 112: Now that we are handling the date conversions on button click, we could add error feedback if the user enters an end date that is earlier than the start date. I didn't bother to do this in this PR.

viewer/src/views/nav/SideMapControlDateRange.jsx
chiaberry: I agree with removing it

chiaberry (commented):
I am going to keep testing next week and test on my phone, but I wanted to put this screenshot in before I forget

The calendar is being cropped on my chrome browser. I see you removed popper-placement, I remember messing with the popper props in order to get the calendar to not be pushed off to the side, but they were finicky. 

Image

viewer/src/views/nav/SideMapControlDateRange.jsx
```diff
@@ -7,159 +7,209 @@ import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import {
   faCalendar,
   faInfoCircle,
```
chiaberry, line 9: this icon isnt being used anymore

viewer/src/views/nav/SideMapControlDateRange.jsx
chiaberry: I was experimenting with using `navigator.userAgentData` to see if we could reliably see if the browser was mobile, and it doesnt seem to work on my phone. `navigator.userAgent` does return something, but unclear if it would work on other devices, so agreed, bit tricky.

viewer/src/views/nav/SideMapControlDateRange.jsx
chiaberry: since you removed it, I dont think we need the fragment around StyledButtonContainer anymore

viewer/src/views/nav/SideMapTimeOfDayChart.jsx
chiaberry: I think this works

chiaberry (commented):
Works great on android! Holding off on approving until the calendar popup placement is fixed, but the changes are great improvements. I also dont like the active buttons being the dark color, i was thrown off by the buttons on mobile and didnt know what was selected and what was active at first. 

Image
Image

viewer/src/views/nav/SideMapControlDateRange.jsx
roseeichelmann: I think i would like the Reset button more if it was a different color than the Apply Filter button. Like if it was outlined only or something, but its confusing to me to apply a filter and the button stays the same but says Reset instead

johnclary: Thanks for that suggestion. Both reset buttons (date picker and time chart) now use an outline style for the reset action.

viewer/src/views/nav/SideMapControlDateRange.jsx
```diff
@@ -4,163 +4,209 @@ import styled from "styled-components";
...
   }, [start, end, setMapDate]);
 
+  const handleApply = () => {
+    const clampedStart = pendingStart < minDate ? minDate : pendingStart;
+    const clampedEnd = pendingEnd < minDate ? minDate : pendingEnd;
+    setStart(clampedStart);
+    setEnd(clampedEnd);
+    setPendingStart(clampedStart);
+    setPendingEnd(clampedEnd);
+  };
```
frankhereford, line 108: One thing that caught my eye as I was messing with the viewport during testing -- if you enter in a date in one more (Desktop vs Mobile) and then switch to the other mode, the previous or default value shows up in the related place. I don't think it's worth trying to sync them up as a user is entering a date, but if it's possible to set the other mode's value on-Apply, that might be worth it. 

I also think this is something that only really comes up in browser testing, and I wouldn't expect it to be an issue with real users, so that makes me think the added complexity may not be worth it. I could go either way, but I'd thought I'd share the observation.

frankhereford (approved):
I think this works really well. I tested on IE on windows and Arc on mac -- both chromium based and they had similar if not identical behavior to my eyes. I think the code makes sense - this looks good to me, thanks John!

chiaberry (approved):
popper is back in view, retested on android just in case and it looks great. 🚢

roseeichelmann (approved):
I tested on firefox on my iphone and this is all working great!! awesome improvements 🚀 

i think if anything, there could be more of a visual cue on mobile that the Crash date is editable.

viewer/src/views/nav/SideMapControlDateRange.jsx
johnclary: thanks for flagging this. we have made the creative decision to have the app mount both the mobile side drawer and desktop sidedrawer at the same time, using `display: none` to control visibility. this leaves us with split state across both. i'll open a separate issue to address this 👍

johnclary:
Thanks, y'all. Just want to mention that Rose an I discussed the firefox native date pickers via Slack. It seems that Firefox's dev tools ("Responsive design mode") is buggy/not reliable when it comes to the date input popups. I was able to reproduce the issue Rose saw via browserstack on Firefox desktop. Thanks for confirming it's a non-issue, Rose.

PR #2147, opened 2026-09-14
Update comprehensive cost values in crash_injury_metrics_view

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/29987

## Testing

run locally. 

```sql
select
	id,
	cris_crash_id,
	est_comp_cost_crash_based,
	est_total_person_comp_cost
from
	crash_injury_metrics_view
where
	cris_crash_id = 13635921
```
Use this query to see the comp cost. Before migrations the cost is 20,000 after applying the migration the cost is 25,000. 

Test down migration. 

---

#### Ship list

- [ ] Code reviewed


mddilley (approved):
Tested locally, and I see the updated fallback comprehensive cost value after running the up migration and then was able to restore the previous value with the down migration. Thanks Chia!! 🚢

johnclary (approved):
thanks, Chia!

frankhereford (approved):
Works great - thank you for updating this calculation Chia!

PR #2149, opened 2026-09-15
Upgrade to react-map-gl v8

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/30254

## Testing

**URL to test:** Local

1. Install and start the VZV

```
nvm use && npm install --legacy-peer-deps
```

2. Confirm the VZV summary page loads normally, then focus on the map tab and confirm

- Map tiles and markers are loaded
- Map interactions are normal
- Map pop-ups are normal
- Draw tool works normally
- Map layer toggles work normally
- Council district map popup works normally
- Sidebar filters work normally

Bonus points if you test on mobile or windows! I have tested via BrowserStack and we're looking good.

---

#### Ship list

- [ ] Code reviewed


viewer/src/views/map/MapComponent.jsx
```diff
@@ -391,15 +389,7 @@ const MapComponent = () => {
     };
 
     return (
-      <Source id="selectedCrash" type="geojson" data={selectedFeature}>
-        {/* <AnimatedIcon
```
chiaberry, line 395: are you going to want to bring something like this back eventually?

chiaberry (approved):
map is working great!

viewer/src/views/map/MapComponent.jsx
mddilley, line 1: While I was stacking different map features, I noticed that the infobox that shows counts in the polygon selection and the council district popup can't exist at the same time. The map will only show one or the other. I can't remember if that was intentional or not but might be worth a spin-off issue. 🤷 

Image
Image

mddilley (approved):
Nice! I tested on port 3000, and the tiles loaded immediately. Thanks for fixing since it gets tricky when debugging port and allow list. The rest of the map features work just like they did before too. 🙏 🚢 ✨

viewer/src/views/map/MapComponent.jsx
chiaberry: I am able to replicate this in prod, but I think it might be worth a spin off issue...

johnclary: Thanks for flagging: https://github.com/cityofaustin/atd-data-tech/issues/30259

viewer/src/views/map/MapComponent.jsx
johnclary: I don't particularly miss the animated marker but if there's popular demand...

PR #2153, opened 2026-09-16
Ability to filter crashes list based on crashes which need manual QA review

### LINKED ISSUE #29844: Ability to filter crashes list based on crashes which need manual QA review
Vision Zero team users need the ability to filter the crashes list based on which crashes have not gone through the team's quality assurance (QA) review checks, so that they can easily find and prioritize crashes for review.

A crash is considered to have its manual review complete if it meets the following criteria:

- `crashes.geolocation_provider_id` = `3` (`manual_qa`)
- The `movement_id` value of all units in the crash is not `null`
- The `veh_trvl_dir_id` value of all units in the crash is any of: `(1, 3, 5, 7, 9, 11)` (NORTH, EAST, SOUTH, WEST, NOT APPLICABLE, UNKNOWN)

### In scope
 
- Add a boolean column to the `crashes_list_view` called `is_manual_review_qa_complete` that resolves as `TRUE` when it meets the criteria described above
- Remove the `is_manual_geocode` column from the `crashes_list_view`. It should no longer be needed or referenced anywhere in the VZE
- Evaluate whether any additional indexes are needed on the `crashes` or `units` tables. E.g., should we index `crashes.geolocation_provider_id`? 
- On the crashes list page in the VZE, add a filter toggle to the **Internal** filter card called **Needs manual review/QA**. When toggled, the crashes list should only display crashes where `is_manual_review_qa_complete` is `false`

### Questions

- Should we spit this logic out into a separate database view to isolate the business logic? For the sake of the PG query optimizer, John is leaning toward writing this directly into the existing `crashes_list_view`, since its already rolling up units.

Image

roseeichelmann:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/29844

## Testing

**URL to test:** 
Local

**Steps to test:**
1. Check out this branch locally run the latest production replication (**important**) , run the migrations and metadata
2. On the Crashes list page use the Needs manual review/QA filter. For the past 1y, i had ~ 7,900 results
3. Click on one of the crashes that needs manual review/qa. Edit the crash coordinates, edit all the units to make sure they each have a Movement value, and that the Direction is one of North, South, West, East, etc.
4. Refresh the crashes list view with the Needs manual review/QA filter, that crash should no longer appear bc you successfully QA'd it!
5. Test the down migration

---

#### Ship list

- [ ] Code reviewed


editor/configs/crashesListViewTable.ts
```diff
@@ -274,6 +274,20 @@ const crashesListViewfilterCards: FilterGroup[] = [
           },
         ],
       },
+      {
```
johnclary, line 277: I had to reset my table filters in order to see this. I think we ought to bump the config version (can't remember exactly where that is set) to force the refresh. Downside being all users will have their stored filters blasted away. That has been our standard practice of late though.

johnclary (commented):
[sending in this feedback while i am distracted with other things for a bit]

editor/configs/crashesListViewTable.ts
roseeichelmann: okay i think i bumped the config version -- lemme test it

roseeichelmann: Okay the bumped config version fixed this -- i tested in my chrome browser instead of firefox and didnt have to click the Reset filters button

chiaberry (approved):
i edited a crash location and the units and saw it disappear from the manual QA list! I have 8105 records that need manual QA from the past 1Y and i just replicated prod.

johnclary (approved):
Tested this a bunch — it works great! Way to go Xavier and having incredible manual qa coverage. 

Thanks for shipping this in time for the release, Rose!

🚢 🚢 🚢 🚢 🚢

mddilley (approved):
Tested locally, and I was able to filter down to "needs manual review/QA", click into a crash record, QA and update it, and come back to the list to see that it was processed and no longer included in the filter. Feels like a nice workflow to me! 🚢 🙌

PR #2155, opened 2026-09-17
Fixes annoying council district popup when drawing

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/30262

## Testing

**URL to test:** Local

**Steps to test:**

1. Start  the app: `nvm use && npm install && npm run dev`
2. Head to the map tab and turn on the council districts layer. Click around on the map and verify the council district popup opens normally and displays the name of the district in which you clicked.
3. With the council district layer still enabled, use the polygon draw tool to draw a polygon. Note that the council district popup does not annoyingly pop up every time you click the map to draw a node in the polygon. You can compare this to the parent branch to see the bug.
4. Turn off the council district layer and draw another polygon — confirm that all works normally.

---

#### Ship list

- [ ] Code reviewed


frankhereford (approved):
Thanks for this John. I was able to go through the testing steps without any trouble. 🚢 

One thing I noticed -- and I'm not sure if it's related to this PR or not -- is that I'm unable to click on a crash inside a polygon when it's being used as a filter. Maybe something to look at down the road. Thanks again!

roseeichelmann (approved):
bug is gone!! nice squeezing this in ✨

mddilley (approved):
Bug gone! I also noticed what Frank described and agree that it could be important to be able to inspect the crash details through the popups inside a selection. 🚀

johnclary:
Thanks @frankhereford for flagging that one. Haven't confirmed yet if this is a regression introduced in this PR or earlier work, but it is definitely a regression from prod. I'll see if I can restore that here or will open a follow-up issue 🙏

chiaberry (approved):
annoying popup gone! it does seem like a regression in this branch that you cant select a crash in the polygon, my hunch is its because when you are drawing a polygon the interactive layer ids are an empty array _but_ i can sometimes get the council district popup to appear with a polygon drawn so maybe my hunch is wrong

sometimes I had to click in and out of the polygon and then back in to be able delete, but i think thats the way it has been since the upgrades

johnclary:
Thanks for that additional info — will hold off on merge and ship as a patch 🙏

johnclary:
Folks, this is overlapping with the broader fix for broken mobile pop-ups. I'm closing this in favor of https://github.com/cityofaustin/vision-zero/pull/2160, which will hopefully fix all the things.

PR #2159, opened 2026-09-17
Fixes VZV filter styles bug

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/30324

## Testing

* [Netlify deploy preview](https://deploy-preview-2159--vzv-staging.netlify.app/viewer/)

Take a look at and test out the toggle buttons in these locations 👇 

### Summary tab

* **By Year & Month** chart
* **Crashes by mode** chart
  * I also updated the styles of the Monthly / Cumulative buttons to match the color and border radius of the other toggles.
* **By time of day** chart

### Map tab

* Injuries filter buttons at the top of the side bar

Please also confirm the rest of the app is looking normal.

---

#### Ship list

- [ ] Code reviewed


viewer/package.json
```diff
@@ -45,7 +45,7 @@
     "reactstrap": "^9.2.3",
     "reaviz": "^16.2.2",
     "serialize-javascript": "^7.1.1",
-    "styled-components": "^4.4.1",
+    "styled-components": "^6.5.3",
```
johnclary, line 48: I had a lot of trouble reproducing this bug locally yesterday. I wasn't able to do so until i served up the static site build. I narrowed that weirdness down to our old version of styled-component, but now I can reproduce the bug locally on any version of styled components 🙃 

Anyway, this styled-components bump seems to have been a clean upgrade.

viewer/src/views/summary/Components/CrashTypeSelector.jsx
```diff
@@ -53,13 +39,14 @@ const CrashTypeSelector = ({ setCrashType, componentName }) => {
   }, [setCrashType, activeTab]);
 
   return (
-    <StyledButton>
+    <>
       <Button
```
johnclary, line 43: I was really confused for a bit while working on these buttons because i didn't realize the default button variant in reactstrap is `secondary`. It is `primary` in `react-bootstrap` 😵‍💫 

I think this restores us to nearly the same style we had before — in which we used a light button style but the secondary-dark active style.

Charlie-Henry (approved):
The text in buttons on the map tab are looking a little too close to spilling out especially as compared to the homepage. Other than that looks good 💯 

Image

Image

chiaberry (commented):
I *still* dont understand why you had to upgrade the package to get the bug to surface locally, but im glad you figured that out and fixed it!

I like that you also updated the style of the Monthly/Cumulutive buttons (and the Race/Ethnicity, Age and Sex buttons) to be more pill shaped and match the other buttons. And I think the Time of Day buttons should stay more rounded square, but they still have the bug where none of them are styling as selected.

johnclary:
Oops — I overlooked the **By time of day** chart and also realized the **By demographics** chart was affected by my other updates. I just updated the test steps. Please take a look — thanks @chiaberry.

viewer/src/App.css
```diff
@@ -25,7 +25,23 @@ div.Tooltip-module_tooltip__omIxd {
...
 }
 
-.mapbox-gl-draw_polygon.disabled{
+.mapbox-gl-draw_polygon.disabled {
   filter: opacity(0.3) drop-shadow(0 0 0 #fff);
   pointer-events: none;
 }
+
+.btn.chart-toggle-button {
+  font-size: 14px;
```
chiaberry, line 34: In prod the non injury severity toggle buttons all have a font size of 16px. I didn't notice the font change until I looked at the year toggles on the time of day card. Maybe also because they became pill shaped, but they feel too small to me at 14px

johnclary: I did not notice that! Hmm, I don't think it make sense for those toggles to have larger text. Are they more important?

viewer/src/constants/responsive.js
```diff
@@ -12,7 +12,7 @@ export const responsive = {
   headerButtonHeight: 50,
   headerButtonWidth: 200,
   headerLogoOffset: 34,
-  drawerWidth: 350,
+  drawerWidth: 365,
```
johnclary, line 15: I had to increase the drawer width even more to accommodate the padding between the filter toggles on the map. The mobile experience is regressing now, because once the side drawer opens you have to find space to tap on the right side of the screen to cause the side nav to collapse. I think we need a close/hide button at the top of the side drawer on mobile 😕 

Edit: See also https://github.com/cityofaustin/vision-zero/pull/2160/changes#r4049826488

viewer/src/App.css
chiaberry: You are right, they arent more important. I think aesthetically I liked them being bigger but I dont have a good reason to say why they should be bigger

viewer/package.json
johnclary: I just noticed that we are getting bombed with console warnings about styled-components re-rendering constantly. I opened https://github.com/cityofaustin/vision-zero/pull/2161 to address this.

chiaberry (approved):
🚢

frankhereford (approved):
I clicked around on the sets of radio-like buttons you mention in your testing instructions, and they all look ship-shape to me. Thanks John!

PR #2160, opened 2026-09-18
Fixes mobile map pop-ups

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/30325
* https://github.com/cityofaustin/atd-data-tech/issues/30262

This PR fixes the problem of map pop-ups not rendering at all on mobile. The source of the issue, as far as I could reckon, was conflicting interaction listeners with the draw tool. I worked with Claude to write up fresh draw handlers and clean up the map state. This seems to have done the trick. We can draw on mobile and we can look at popups on mobile.

I also put in the fix for the council district popup firing off while drawing.

I also updated the the mobile nav drawer so that it immediately closes when you use a nav button. This is to work around the problem of not really being able to close the side drawer given the wider width of it introduced via https://github.com/cityofaustin/vision-zero/pull/2159#discussion_r4049164197.

This PR does revert that ability to drag a drawn polygon around on the map. This was functionality we _added_ in the React 18 upgrade that went live yesterday — it was not previously possible to do this. 

## Testing

**URL to test:** [Netlify](https://deploy-preview-2160--vzv-staging.netlify.app/viewer/map)

Please test using your phone and your laptop. [Browser stack](https://live.browserstack.com/dashboard?source=bstacklocal-3.7.8) is also available. 

1. Head to the map tab and click on some crash points. Confirm the popup looks/works normally.
2. Test the drawing UX. Create and delete a polygon. Try deleting a polygon while in the midst of drawing.
3. Turn on the council district layer and repeat your tests. Confirm that the council district popup is not being [annoying](https://github.com/cityofaustin/atd-data-tech/issues/30262).
4. Use the navigation to nav back and forth between map and summary. Note the new behavior in which the mobile drawer closes automatically when you navigate. 

I have tested android (chrome + firefox) and ios (safari).

---

#### Ship list

- [ ] Code reviewed


viewer/src/views/nav/SideDrawerMobileNav.jsx
johnclary, line 1: This change causes the mobile side drawer to immediately close when you use a nav button. 

Upside: it's possible to close the side drawer if you're using mobile.

Downside: you don't see the map filter options unless you open the menu again.

Ideal states?
- separate nav/menu toggles for filtering vs switching back to the summary view?
- Make the side drawer narrow again by stacking the crash the type filters

chiaberry: The downside doesnt feel like too much of a downside to me, I think this works fine

chiaberry (approved):
Tested a lot on my phone and the pop ups worked great. 

The code refactoring looks good. when I worked on this feature the code still had defined buttons but also the built in mapbox control buttons turned on and I am kicking myself a little for not sticking with our manual buttons. The draw polygon button that turns into a trashcan is a nice touch, makes it very obvious that only one polygon can be drawn at a time.

Charlie-Henry (approved):
Tested on ios + safari and it worked great 💯

frankhereford (approved):
I tested this in my browser and on iOS - both work great. Thanks for this clean up, and especially for the great inline comments you've added. Super helpful ✨

johnclary:
>  kicking myself a little for not sticking with our manual buttons

@chiaberry I was happy to see them go at the time 🙃

